PR #187, opened 2014-11-22
Fixed runner to use system SHELL env in ':shell' command

amerlyq:
I use zsh with extended globbing, etc, and miss it in ranger -- so there it is.
Variant of spawning subshell for one command a time is too slow, comparing to ':'


amerlyq:
@hut, @Vifon, any of you can delete already that lingering branch. 
And any other similar stale branches.
Because each new ranger fork uselessly copies them as well.
I suppose, tidying main repository from branches merged by pull requests is what that github button about.
https://github.com/blog/1335-tidying-up-after-pull-requests

Issue #185, opened 2014-11-17
w3m image preview damaged when file list scrolled down


tex:
I'm using urxvt. There might be some connection to #86.
My problem is that when file list is scrolled down and it is image and it is to be previewed by w3m, the image is shown damaged.
The preview is ok if file list is not scrolled.
It looks like there might be problem in logic in ranger: the image is previewed before file list is scrolled.

nfnty:
Are you still having this issue?

tex:
Yes. I did update ranger and try image preview and image previews are still getting corrupted.

nfnty:
Have you tried using `preview_images_method urxvt`?

nfnty:
Continue discussion in #86.

PR #192, opened 2014-11-29
make ImageDisplayer display images without blocking

tex:
fix to #185

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -189,8 +189,6 @@ class W3MImageDisplayer(ImageDisplayer, FileManagerAware):
         self.process.stdin.write(input_gen)
         self.process.stdin.flush()
         self.process.stdout.readline()
-        self.quit()
-        self.is_initialized = False
 
     def clear(self, start_x, start_y, width, height):
         if not self.is_initialized or self.process.poll() is not None:
@@ -557,6 +555,7 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
         self.backend = None
         self.stream = None
         self.pix_row, self.pix_col = (0, 0)
+        self.prev_image_size = (0, 0)
 
     def _late_init(self):
         # tmux
@@ -611,13 +610,6 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
         self.needs_late_init = False
 
     def draw(self, path, start_x, start_y, width, height):
-        self.image_id += 1
-        # dictionary to store the command arguments for kitty
-        # a is the display command, with T going for immediate output
-        # i is the id entifier for the image
-        cmds = {'a': 'T', 'i': self.image_id}
-        # sys.stderr.write('{}-{}@{}x{}\t'.format(start_x, start_y, width, height))
-
         # finish initialization if it is the first call
         if self.needs_late_init:
             self._late_init()
@@ -631,6 +623,10 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
             #     raise ImageDisplayError(str(warn[-1].message))
         box = (width * self.pix_row, height * self.pix_col)
 
+        if (self.prev_image_size != (image.width, image.height)):
+            self.clear(start_x, start_y, width, height)
+            self.prev_image_size = (image.width, image.height)
+
         if image.width > box[0] or image.height > box[1]:
             scale = min(box[0] / image.width, box[1] / image.height)
             image = image.resize((int(scale * image.width), int(scale * image.height)),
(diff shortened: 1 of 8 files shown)


nfnty:
Rebase this PR and make sure that all tests pass. See the updated CONTRIBUTING and HACKING documents for more info. Thanks!

soredake:
@tex ?

tex:
Well, I'm still alive, but buried with work that helps me pay my bills, and also spending my whole free time with my children. Not having that much free time as I would like to have (actually,  almost no free time). Sorry.

aldiyarablyazov:
Is this still being worked on? Previewing images still results in laggy navigation with iTerm2. Will we be seeing a merge anytime soon?

tex:
Hi, I still have no time for this. But I wonder whether it is really needed. I tried ranger over remote sshfs and it works almost ok. When I scroll down over a directory full of images by holding down arrow ranger does not tries to show images so it scrolls down fast, as expected. When I stop it tries to show image and yes, here it blocks until the image is shown. I would call it just a slight inconvienence, not a big problem.
Feel free to close this merge request or tell me to close it.

toonn:
No one else is actively working on this either. Consider it a far-future plan. The ueberzug preview method does side-step the issue because all ranger needs to do is tell ueberzug to display something, so it doesn't block much, same story for mpv.

tex:
I ported it to master, for all backends, tested with w3m and kitty and it works nice in my NixOS. Will do some clean up and push it soon.

tex:
Try it. Works for me (tm).

seebye:
If it's not too much trouble it would be great if the delay could be disabled for ueberzug by default as frames will already be dropped if ueberzug can't keep up with the drawing requests.  
(Also in my opinion it's not so nice not to show images during scrolling by default.
(At least that's how I understood this: An images will be shown if there isn't another draw request within the delay time))

tex:
@seebye It is fine that ueberzug can handle it without this patch, but I believe it doesn't harm it. You still can set preview_image_delay to 1.
Yes, it works as you described, I do not see a problem. I want to see image when I stop scrolling. Not in some arbitrary times in between that just interrupts (even a for while) my scrolling.

seebye:
> that just interrupts (even a for while) my scrolling.

I never said anything against the concurrent execution.

> I believe it doesn't harm it.  

Well, going from dropping frames to not seeing anything at all isn't really an improvement.  
Just as I already said in my opinion delaying the preview is a bad default setting.  
As the preview gives you an idea about the content of the folder during scrolling.  
Also especially for images and videos the preview is a lot more valuable to find a desired file.

ranger/ext/img_display.py
```diff
@@ -686,8 +688,6 @@ def clear(self, start_x, start_y, width, height):
         # kitty doesn't seem to reply on deletes, checking like we do in draw()
         # will slows down scrolling with timeouts from select
         self.image_id -= 1
-        self.fm.ui.win.redrawwin()
-        self.fm.ui.win.refresh()
```
tex: I see sometimes artifacts in UI, that was probably "fixed" by calling redrawin and refresh here, but I needed to remove it from here because concurrent assess to fm.ui.win caused heavy UI damage.

Where in ranger is the "main" loop? Something that awaits key-presses? Perhaps it would be better to modify it there.

tex:
My final attempt. No threads, works perfectly for me.

GermainZ:
I've tested this branch (I'll call it `tex`) for a few terminals/preview methods.  

I'm testing with Python 2.7.17 and 3.8.2.

| terminal | preview method | py2 | py3 | notes                                                                                             |
|----------|----------------|-----|-----|---------------------------------------------------------------------------------------------------|
| kitty    | kitty          | ✔   | ✔   |                                                                                                   |
| kitty    | w3m            | ❌   | ❌   | Image is previewed for a split second, then disappears                                            |
| kitty    | ueberzug       | ✔   | ✔   |                                                                                                   |
| urxvt    | urxvt          | ✔   | ✔   |                                                                                                   |
| urxvt    | w3m            | ✔   | ✔   |                                                                                                   |
| urxvt    | ueberzug       | ✔   | ✔   |                                                                                                   |
| xterm    | w3m            | ❌   | ❌   | Image is sometimes previewed for a split second, then disappears. Sometimes doesn't appear at all |
| xterm    | ueberzug       | ✔   | ✔   |                                                                                                   |
| termite  | w3m            | ❌   | ❌   | Image is previewed for a split second, then disappears                                            |
| termite  | ueberzug       | ✔   | ✔   |                                                                                                   |

(`master` works in all tested cases.)

Notes:
- With `preview_images_delay = 0`, the performance is the same as `master` since no threads are used.

tex:
@GermainZ Thanks for testing, please try with update. Fixed w3m behaviour. Also try with pictures of different size please.

tex:
BTW this version does not utilize threads anymore and well, that makes it better fit to current ranger architecture...

GermainZ:
Hey! I don't have time to do all the tests again right now, but it seems to work for kitty+w3m. However, for kitty+kitty, the image is not cleared when going backwards in a directory (i.e. press `h`).

tex:
@GermainZ Thanks! Replicated and fixed!

PR #203, opened 2014-12-15
Add a 'cycle' option to direction

PotatoesMaster:
This allows to loop on a file list, going at the other side of the list when trying to go beyond the boundaries (e.g. pressing down when the last file is selected). 

Test command: `move up=1 cycle=true`

Note: in `minimum + pos % (maximum + offset - minimum)`, I am not sure if the use of 'offset' is appropriate (i.e. not sure whether it is broken or not).

### DIFF
--- a/ranger/ext/direction.py
+++ b/ranger/ext/direction.py
@@ -81,6 +81,9 @@ class Direction(dict):
     def percentage(self):
         return 'percentage' in self and self['percentage']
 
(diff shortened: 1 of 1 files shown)


hut:
Nice patch, thanks! Except for one thing:
```move up=1 cycle=false``` will still cycle

hut:
Fixed & merged.

PR #205, opened 2014-12-16
Allow sxiv_workaround_hook to Work if Command Contains "sxiv "

angelic-sedition:
Currently the command in the rifle.conf must start with "sxiv " for all images in the directory to be opened. My sxiv command for opening images, however, does not start with sxiv and calls another script first. Is there a reason that the command should start with "sxiv " (or "feh ") as opposed to just containing it?

### DIFF
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -126,14 +126,14 @@ class FM(Actions, SignalDispatcher):
         # The requirements to use it are:
         # 1. set open_all_images to true
         # 2. ensure no files are marked
-        # 3. call rifle with a command that starts with "sxiv " or "feh "
+        # 3. call rifle with a command that contains "sxiv " or "feh "
         def sxiv_workaround_hook(command):
             import re
             from ranger.ext.shell_escape import shell_quote
 
             if self.settings.open_all_images and \
                     len(self.thisdir.marked_items) == 0 and \
-                    re.match(r'^(feh|sxiv) ', command):
+                    re.match(r'^.*(feh|sxiv) ', command):
 
                 images = [f.basename for f in self.thisdir.files if f.image]
                 escaped_filenames = " ".join(shell_quote(f) \
@@ -143,12 +143,12 @@ class FM(Actions, SignalDispatcher):
                         "$@" in command:
                     new_command = None
 
-                    if command[0:5] == 'sxiv ':
+                    if 'sxiv ' in command:
                         number = images.index(self.thisfile.basename) + 1
                         new_command = command.replace("sxiv ",
                                 "sxiv -n %d " % number, 1)
 
-                    if command[0:4] == 'feh ':
+                    if 'feh ' in command:
                         new_command = command.replace("feh ",
                             "feh --start-at %s " % \
                             shell_quote(self.thisfile.basename), 1)


angelic-sedition:
@hut
Okay, I tried overriding it in the Rifle class like in the plugin_new_macro.py example, but I get an error when trying to open anything (`FM.move: [], {'right': 1}`):
```
import ranger.ext.rifle
old_preprocessing_hook = ranger.ext.rifle.Rifle.hook_command_preprocessing

def sxiv_workaround_hook(command):
    (modified)

ranger.ext.rifle.Rifle.hook_command_preprocessing = sxiv_workaround_hook
```
I'm guessing that I'm doing something horribly wrong, but I'm not sure what it is.

hut:
What's the output when you run ranger with the `--debug` option?

angelic-sedition:
I named the new function "sxiv_workaround_hook_modified".

```
ranger version: 1.6.1, executed with python 3.4.2
Locale: en_US.UTF-8
Current file: /home/angelic_sedition/Move/2014-07-06-133534_895x394_scrot.png
  File "/usr/lib/python3.4/site-packages/ranger/core/main.py", line 139, in main
    fm.loop()
  File "/usr/lib/python3.4/site-packages/ranger/core/fm.py", line 326, in loop
    ui.handle_input()
  File "/usr/lib/python3.4/site-packages/ranger/gui/ui.py", line 211, in handle_input
    self.handle_key(key)
  File "/usr/lib/python3.4/site-packages/ranger/gui/ui.py", line 147, in handle_key
    self.press(key)
  File "/usr/lib/python3.4/site-packages/ranger/gui/ui.py", line 162, in press
    quantifier=keybuffer.quantifier)
  File "/usr/lib/python3.4/site-packages/ranger/core/actions.py", line 230, in execute_console
    cmd_class(string, quantifier=quantifier).execute()
  File "/usr/lib/python3.4/site-packages/ranger/api/commands.py", line 376, in execute
    return self._based_function(*args, **keywords)
  File "/usr/lib/python3.4/site-packages/ranger/core/actions.py", line 432, in move
    result = self.execute_file(selection, mode=mode)
  File "/usr/lib/python3.4/site-packages/ranger/core/actions.py", line 389, in execute_file
  File "/usr/lib/python3.4/site-packages/ranger/ext/rifle.py", line 311, in execute
    cmd = self.hook_command_preprocessing(cmd)
  File "/usr/lib/python3.4/site-packages/ranger/core/fm.py", line 160, in sxiv_workaround_hook
    return old_preprocessing_hook(command)
TypeError: sxiv_workaround_hook_modified() takes 1 positional argument but 2 were given
```

hut:
Adding `self` as the first argument in the function definition should fix it

angelic-sedition:
I get `AttributeError: 'Rifle' object has no attribute 'settings'` then. I'm using the same workaround_hook function as in the pull request. Do I need to modify it further?

hut:
Oh, right, the default hook references the `fm` object as `self`. You need to get `fm` somewhere. Try this:

```python
import ranger.api

old_hook_ready = ranger.api.hook_ready

def hook_ready(self):
    # "self" is now the FileManager instance
    old_preprocessing_hook = self.rifle.hook_command_preprocessing

    def sxiv_workaround_hook(command):
        # your code

    self.rifle.hook_command_preprocessing = sxiv_workaround_hook

ranger.api.hook_ready = hook_ready
```

Note that I didn't use an extra `self` argument in the function definition here. This is because now we override the method of the rifle object, not the rifle class.

angelic-sedition:
Ah that makes more sense and works. Thanks a lot!

PR #211, opened 2015-01-01
renameConsole -> rename_append + special cases for hidden files and directories

NoSuck:
Because the file extension for ".python_history" is not "python_history".


knight-96:
There is still a case that isn't included. If I want to `rename_append` a file, such as '.python.history', the cursor will go to the end, which is not what I expect.

I change the line to the following:
```
if cf.basename.rfind('.') > 0 and not cf.is_directory:
```

Hope it helps :)

PR #212, opened 2015-01-01
follow_link + documentation

NoSuck:
I use this one fairly often.  Not guaranteed to work with BSD's readlink implementation (though it shouldn't crash, either).

### DIFF
--- a/doc/ranger.pod
+++ b/doc/ranger.pod
@@ -833,6 +833,7 @@ ranger.  For your convenience, this is a list of the "public" commands including
  filter_inode_type [dfl]
  find pattern
  flat level
+ follow_link
  grep pattern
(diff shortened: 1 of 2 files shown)


hut:
Check out the key bindings "gl" and "gL", I think they are what you're looking for.

NoSuck:
Clever bindings.  I'd been neglecting those.
However, I see one disadvantage:  "gL" moves to the parent directory of a link target but fails to select the target itself.  I'll do a little more digging and see if something elegant presents itself.

hut:
We could simply add the feature to automatically select the target file when calling gL on a file. Note that the behaviour of calling gL on a directory is slightly different: it makes ranger cd *into* the target of the link.

PR #215, opened 2015-01-13
Modified cd command to select original file, when following links.

NoSuck:
Good suggestion.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -107,7 +107,8 @@ class cd(Command):
             self.shift()
             destination = os.path.realpath(self.rest(1))
             if os.path.isfile(destination):
-                destination = os.path.dirname(destination)
+                self.fm.select_file(destination)
+                return
         else:
             destination = self.rest(1)
 


hut:
Strangely, this has no effect for me. Can you describe the behaviors of a relevant command without and with your patch?

NoSuck:
Strange.  I tested with both ranger-stable 1.6.1 and the current git master.  I have "cd -r %f" mapped to "O" and will substitute the default config of "gL" for these examples.
```bash
■ pwd
/tmp/test
■ ls -RgG --time-style=+""
.:
合計 0
drwxr-xr-x 2  80  links
drwxr-xr-x 2 100  orig

./links:
合計 0
lrwxrwxrwx 1 27  hotcakes_abs.txt -> /tmp/test/orig/hotcakes.txt
lrwxrwxrwx 1 20  hotcakes_rel.txt -> ../orig/hotcakes.txt

./orig:
合計 0
-rw-r--r-- 1 0  flapjacks.txt
-rw-r--r-- 1 0  hotcakes.txt
-rw-r--r-- 1 0  pancakes.txt
```
With the patch, gL on either `/tmp/test/links/hotcakes_abs.txt` or `/tmp/test/links/hotcakes_rel.txt` changes the directory of the current tab to `/tmp/test/orig/` and selects `hotcakes.txt`.  Without the patch, the directory is changed but no file is selected (defaulting to `flapjacks.txt`, if the current sort is alphabetical and no previous selections had been made in `orig`).  Does something else happen on your end?

hut:
Okay, I tested it again and it worked as you described. No idea what went wrong the last time, and I'm sorry for the inconvenience of you having to describe this simple patch so exhaustively.

PR #217, opened 2015-01-17
Tab switch action

NoSuck:
A speedy companion to cd.  Works well when you're working in a lot of directories but want the convenience of quickly switching to one of them.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -229,13 +229,13 @@ map S shell $SHELL
 
 map :  console
 map ;  console
-map !  console shell 
+map !  console shell
 map @  console -p6 shell  %%s
-map #  console shell -p 
-map s  console shell 
-map r  chain draw_possible_programs; console open_with 
-map f  console find 
-map cd console cd 
+map #  console shell -p
+map s  console shell
+map r  chain draw_possible_programs; console open_with
+map f  console find
+map cd console cd
 
 # Change the line mode
 map Mf linemode filename
@@ -258,7 +258,7 @@ map <F3> display_file
 map <F4> edit
 map <F5> copy
 map <F6> cut
-map <F7> console mkdir 
+map <F7> console mkdir
 map <F8> console delete
 map <F10> exit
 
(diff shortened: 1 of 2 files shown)


hut:
Thanks for contributing! But I'm not yet convinced that this is useful. Can you remind me of your use case?

NoSuck:
Sure.  Basically, I feel it is important for ranger to open tabs intelligently, as there are almost no situations that involve having two tabs opened to the same directory.

The simple usage case:  I use `tab_switch` with system directories.  Before, I would first scan the open tabs to see if `/etc` was already open.  If it wasn't open, I would press "gn" to open a new tab then "ge" to move to `/etc`.

Now, I just press blindly press "Ge", and a new tab is opened (or switched to) intelligently.

I also have a shell script that accepts arguments to `find`, symlinking results to `/tmp/find` (or whatever).  With a `tab_switch` action, I can do something like press "xi" to symlink into `/tmp/find` all images found recursively in %s while simultaneously opening `/tmp/find` in a tab (if it's not already open, in which case its open tab is activated).  I can then view the non-pr0n images at a leisurely pace, close `/tmp/find`'s tab, and resurrect `/tmp/find` instantaneously at a later point, with "xv" for videos, "xa" for audio and so forth.

It may sound a bit complicated, but it's really nice to be able to press two keys (or even one, depending on the config) and immediately have the results of a pre-programmed `find` call linked into a single directory that appears magically without wasting a new tab... It does sound complicated, though, which is why I hope to prepare a screencast to demonstrate how awesome it is.

hut:
Okay, I kind of see the point. I'll merge it once I read & tested the code. I don't believe the extra bindings are justified though, because you bind them to G which is already the "move to the end" key (though this is obviously fixable) but also because more keys = more confusion, generally.

So unless more people ask me to add key bindings, I'll skip that commit.

hut:
I merged all the commits not dealing with key bindings now. :) Thanks.
And while wondering how to document this, I noticed that there's no documentation for tab_new, tab_open, and all the other implicit commands. =S I think I'll just open a bug report on this issue.

NoSuck:
Great!  I debated for a while about documentation... Perhaps the following should be removed from the man page?  I think it'd be just fine for tab_new and friends to be mentioned in the docs.

> There are additional commands which are directly translated to python functions, one for every method in the ranger.core.actions.Actions class. They are not documented here, since they are mostly for key bindings, not to be typed in by a user.  Read the source if you are interested in them.

PR #221, opened 2015-01-17
:meta may be used on multiple selected files

Vifon:


hut:
Thank you for your contributions :)

PR #222, opened 2015-01-18
Priority change

HerrNieschnell:
I think if someone has sxiv, feh or mirage installed, it should have a higher priority than eog or eom, because they are normally automatically installed along with mate or gnome.

it fits the philosophy of browser or documents default priority.


hut:
Yup, thanks :)

HerrNieschnell:
you are welcome, glad I could help :blush:

PR #234, opened 2015-01-28
config.commands: added ability to create multilevel directories

punzik:
As "mkdir -p", e.g. :mkdir dir1/dir2/dir3


hut:
Good idea! :)

PR #236, opened 2015-01-29
Quickjump (Vim-easymotion) navigation mode

Klinkenstecker:
I worked on an additional navigation mode, that is inspired by Vim-easymotion (or AceJump in emacs) and in line with the keyboard-navigation in browers like dwb (see also the ##1## part in http://lists.nongnu.org/archive/html/ranger-users/2014-02/msg00000.html). I prefer it most time over the filter solution, because the number of keystrokes needed to select a file  doesn't depend on the similarity of the filenames in the directory.  

This navigation mode comes with an own keymap. E.g. this allows to copy files without leaving the terminal/scout command. 

I used this alot in the past weeks, and hope that I have found all bugs and quirks, so I think it's time to create this pull request and hope that you like this feature.


nfnty:
Very interesting! Going to check this out soon.

hut:
Cool feature^^
Why is every commit included twice?

Klinkenstecker:
Mmm, when I look into my local history, it doesn't show all the double commits, only the original once. I think that happend yesterday while I rebased my branch to your master, which caused a lot of conflicts and manual merging. But I'm not a git-expert (for sure there is an option that creates a better history).

If you think that this is a problem, I could try to rewrite my changes (before the rebase) in the quick_jump branch to a single commit and then create a pull request from this point, but I think that this will cause conflicts when you try to merge the pull request.

Or does someone else has an idea how to fix this issue?

hut:
It seems like you both rebased *and* merged. It looks like the history line up to commit 19e254782148c930e2edf9836375d61e92dabe0a is the right one. Check out the branch "quickjump" that I made and see if this is alright: https://github.com/hut/ranger/tree/quickjump

Klinkenstecker:
Yeah, the branch you made includes all the needed commits.

hut:
Wonderful. I'll need some time to read through everything. Sadly I'm under time pressure now, so it might take a while. ;_;

hut:
Can you give me a use case for the pause function? I don't think I understand the one you described about copying files without leaving the console.

Klinkenstecker:
The pause function has nothing todo with the possibility to configure the Quick-Jump feature in a way, that allow to trigger commands while you are in the console.

Use case pause function: 
The pause function e.g. allows to combine the "traditional" filter function with the quick jump mode. My filter alias is scout -efgtiQ. When the filter still show some different files, and I want select one of those, i just press <tab> to activate the Quick-Jump and select one of this files by pressing the corresponding key. Or when i travel using with -q and enter a directory with alot of files, and I want e.g. want to filter .rc files, I press <tab>rc<tab> and have know the reduced list of files quick selectable.

Non leaving the console:
While the Quick-Jump mode is activate, you can't use the console in the normal way, so I have added an own keymap for the Quick-Jump mode (but when you pause the Quick-Jump mode, the console keymapping is used). Additionally you can define, which characters are used for the Quick-Jump tags. So if you remove a character from this list, you can use them in the Quick-Jump keymap. 

This is an excerpt of my quick jump related rc.conf:
===
set quick_jump_letters fdsartbvcwxiuhlz
qmap <tab>       eval fm.ui.quick_jump.smart_tab()
cmap <tab>       eval fm.ui.quick_jump.smart_tab()

qmap q eval fm.ui.console.close()

qmap <Space> mark_files toggle=True
qmap e shell -fs emacsclient %f

qmap <DELETE>   console delete

qmap yp paste
qmap yo paste overwrite=True
qmap yl paste_symlink relative=False
qmap yL paste_symlink relative=True
qmap yY cut
qmap yy copy

qmap '<any>  enter_bookmark %any
qmap '<bg>   draw_bookmarks
qmap m<any>  set_bookmark %any
qmap m<bg>   draw_bookmarks
=== 

So as you can see, I can even use the bookmarks while traveling.

hut:
Yes, I can see how that's useful =)

Vifon:
I've been testing Quickjump for a few days. I've found the following issues:

1. If I use `scout` with `-afQ` and select a directory using Quickjump, the filter becomes permanent even though I didn't use `-p` option. A wild guess: maybe the filter is being undone after performing `chdir`?

2. `travel` now lacks a trailing space after the first change of the directory.

Should I open the individual tickets on your fork, @Klinkenstecker?

Klinkenstecker:
@Vlfon: Thanks for testing and reporting the issues, I will try to reproduce and fix them this weekend. And yes, please open individual tickets on my fork.

Vifon:
@Klinkenstecker: It seems you have issue tracker disabled, so I can't.

Klinkenstecker:
@Vlfon: Upps, sorry. The tracker is enabled now.

amerlyq:
Will we have this feature some point in time or not?

kohane27:
Is this feature available? Any updates on this please? I've been trying exactly to achieve this. Currently I keep spamming `jjjj` to move to files. Filtering with `/` seems excessive so I didn't want to use it.

Looking at [the plugin list](https://github.com/ranger/ranger/wiki/Plugins), we only have plugins to jump between dir ([ranger-autojump](https://github.com/fdw/ranger-autojump), [ranger-zjumper](https://github.com/ask1234560/ranger-zjumper) and [ranger-zoxide](https://github.com/jchook/ranger-zoxide)) but not between files.

PR #239, opened 2015-02-01
core.actions: tab_switch uses realpath

Vifon:
This way the different representation of the same path are treated in the same way.


hut:
Are you sure you want realpath and not abspath?

Vifon:
A good question. I think it's a matter of personal preference. In my opinion `realpath` makes sense here but `abspath` makes sense too. If you feel `abspath` would be better, I'm ok with this.

hut:
I'll just merge this until someone complains. It really doesn't make much of a difference anyway.

PR #242, opened 2015-02-01
config.commands: bulkrename works with `:flat -1`

### LINKED ISSUE #241: `:bulkrename` doesn't work with `:flat -1`
When flattening subdirectories with `:flat`, the filenames in `:bulkrename` are only basenames rather than having their relative path prepended. That way, `mv` can't find the files and the command fails.

Vifon:
Fixes #241.


hut:
Thanks <3

PR #245, opened 2015-02-04
doc/ranger.1: added some information about the config file loading process

Vifon:
I've added the manpage sections discussed on the IRC channel recently. Feel free to remove some of it as there is some redundancy.


hut:
Thank you. In 8545ec6101ba09cc27ef6183d1f7b3c7c9643a63 I specified which files exactly are affected by this.

PR #246, opened 2015-02-06
add qpdfview to rifle.conf

bastorran:
Hello,

there is a PDF viewer called qpdfview. I added this one to rifle.conf

Greetings
bastorran


hut:
Thanks =)

PR #253, opened 2015-02-12
Fix renaming with insufficient privileges

### LINKED ISSUE #97: renaming breaks with insufficient permissions
when renaming a directory with :rename and you don't have sufficient permissions, the preview disappears.

UnrealQuester:
Closes #97
When renaming a file, the internal file pointer will be set to the new
file name without checking if the file actually got renamed. This would
cause the file pointer to be set to an invalid file if renaming failed
because of bad privileges. With no valid file, the previev window will
disappear. This patch adds a return value to the core renaming method so
it can be checking by the caller. The rename command now checks if the
rename succeeded and only then sets the new file pointer.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -715,10 +715,10 @@ class rename(Command):
         if access(new_name, os.F_OK):
             return self.fm.notify("Can't rename: file already exists!", bad=True)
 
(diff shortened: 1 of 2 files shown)


hut:
Thank you for the fix. :)

PR #254, opened 2015-02-12
Deleting a file with a cached image preview no longer crashes ranger

Vifon:
If user deletes a file with an associated cached image preview (via scope.sh), ranger used to check this now non-existent file's modification time and crash in the process. I've added a check whether a file still exists.


hut:
Thank you

PR #257, opened 2015-02-18
Fixed the file macros with :flat enabled

Vifon:
The `%f`, `%s` and `%t` file macros used to return only the file basename, even if it was not located in the current directory. It would break the `%d/%f` idiom and many others when `:flat` was enabled. Now they return a path relative to the current directory (which is the same as the basename when `:flat` is disabled).

Additionally, I've renamed the fsobject.drawn_basename to fsobject.relative_path according to the discussion on the #ranger IRC channel and used it wherever possible.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -793,14 +793,12 @@ class bulkrename(Command):
     def execute(self):
         import sys
         import tempfile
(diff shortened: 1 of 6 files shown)


hut:
Thanks and sorry for the wait :)

PR #264, opened 2015-02-22
Linemode API added to make the custom linemodes possible

Vifon:
I have moved most of the linemode logic into a separate module and added a simple API to allow creating the custom linemodes. The abstract base class was used to catch as many errors as possible as soon as possible -- most of the implementations will be supplied by the users.


hut:
Thank you, good job \o/

PR #265, opened 2015-02-28
restructed quick-jump/scout execution

Klinkenstecker:
Vifon reported some issues two wees ago, and while I studied the cause, I found that the problems are caused by "reimplement" stuff in the quick_jump class instead of scout.execute. I think this pull request is a better integration of the quick-jump mode into scout.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1089,7 +1089,9 @@ class scout(Command):
 
         if self.MARK in flags or self.UNMARK in flags:
             value = flags.find(self.MARK) > flags.find(self.UNMARK)
-            if self.FILTER in flags:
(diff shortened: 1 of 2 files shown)


nfnty:
Has this been tested? Ready for merging?

Klinkenstecker:
I use it for a long time now without problems, but of course it's possible that there are use cases that I didn't checked. And also this is a pull-request for a branch that is based on Ranger 1.61 (and I didn't updated Ranger since then), so it's possible that merging this branch into the current version is maybe not that easy.

nfnty:
Yeah, I know.

Alright, I'll merge this. `quickjump` needs to be rebased and updated for it to be merged. Open a new issue if you're interested.

PR #268, opened 2015-03-08
Add iTerm2 image format support to image previews

### LINKED ISSUE #183: Compatibility with iTerm2's imgcat
The nightly iTerm2 build features a new script called "imgcat" that allows you to view images inside of the terminal. Instead of using w3m, this would be a great alternative since for the life of me, I can't get it to work on Mac.

Here's more info regarding the matter;

http://iterm2.com/images.html

I've tried adding it to the scope.sh file, but I can't seem to get it to work. Is there anyway it could be made to work with ranger?

kattrali:
Adds support to `ImageDisplayer for the iTerm2 image format. Screenshot:

Super meta screenshot

## Changes

* Separated `preview_images` from w3m by adding configuration options `use_w3m_image_preview` and `use_iterm2_image_preview`. All default to `false` and are documented in `rc.conf`.
* Split `ImageDisplayer` into a generic stub and implementations for w3m and iTerm2 support

## Discussion

The changes should make it easy to add support for other image preview types as available. I also considered adding in a displayer for `img2txt` (and still can), but first wanted to make sure this approach makes sense as a part of the hut codebase. The major motivation was that the [iTerm2 format](http://iterm2.com/images.html) does not support being parsed through a pager, and thus could not be used in `scope.sh`.

Fixes #183.

Thoughts?


mdarby:
:+1:

hut:
Helpful patch, thank you!

Just one issue: I think it would be better to keep the `preview_images` setting for toggling this feature, and add another setting for chosing the backend. Maybe name it `preview_images_method` (for tab-completion relatedness with the `preview_images` setting) and accept values like "w3m" and "iterm2". That way we can also extend it in the future.

kattrali:
@hut Gotcha. That makes sense! Updating the patch.

kattrali:
Updated and set `w3m` as the default `preview_images_method`

hut:
Everything looks fine. :) Merged it. Thank you again.

kattrali:
Yay! Thank you. :beers:

Issue #260, opened 2015-02-19
confirm_on_delete inconsistency


leebrooks0:
According to the comments at https://github.com/hut/ranger/blob/master/ranger/config/rc.conf#L34 the value is always by default, but it is set to multiple?

randynobx:
Comment changed to reflect the actual default, multiple.

PR #269, opened 2015-03-09
updated comment to reflect current default value for confirm_on_delete

randynobx:
Fixed comment concerning default value of confirm_on_delete in rc.conf.

Issue #260


hut:
Thank you very much

PR #270, opened 2015-03-09
:linemode and :default_linemode tab completion

Vifon:
The linemode commands may be tab-completed now. I've also cleaned up a bit the :default_linemode changes from the previous pull request.


hut:
I merged it, but left out the "minor :default_linemode cleanup" commit. "fm.thisfile" should not be used in this case, as it is not initialized yet during the time of loading rc.conf. This is a deal breaker because that command is primarily intended to be run in rc.conf.

PR #271, opened 2015-03-10
added solarized colorscheme to list of available-by-default colorschemes...

randynobx:
... in rc.conf


hut:
Oh, thanks, I missed that.

PR #274, opened 2015-03-12
os module does not have "mkdirs" function 

anekos:
# error message

```
'module' object has no attribute 'mkdirs'
```

# refs

https://docs.python.org/2/library/os.html?highlight=makedirs#os.makedirs
https://docs.python.org/3/library/os.html?highlight=makedirs#os.makedirs

ranger is best file manager! Thanks.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1244,7 +1244,7 @@ class Actions(FileManagerAware, EnvironmentAware, SettingsAware):
 
     def mkdir(self, name):
         try:
-            os.mkdirs(os.path.join(self.thisdir.path, name))
+            os.makedirs(os.path.join(self.thisdir.path, name))
         except OSError as err:
             self.notify(err)
 


hut:
Hi, what do you do to produce this error message? As far as I can see, the code is never called. Instead, when you type `:mkdir ...`, the code in ranger.config.commands.mkdir is used instead (see https://github.com/hut/ranger/commit/eb2c650f)

anekos:
My defined function in .config/ranger/commands.py uses "self.fm.mkdir"

I should not use this function?

The part of my commands.py
```
class mkseldir(Command):
  """
  :mkseldir <dirname>

  Creates a directory with the name <dirname>, and select it.
  """

  def execute(self):
    dirname = join(self.fm.thisdir.path, expanduser(self.rest(1)))
    if not lexists(dirname):
      self.fm.mkdir(dirname)
      if lexists(dirname):
        self.fm.thisdir.load_content(schedule=False)
        self.fm.select_file(dirname)
      else:
        self.fm.notify("Failed to mkdir!", bad=True)
    else:
      self.fm.notify("file/directory exists!", bad=True)
```

anekos:
I checked on 81b65ab18a805b333fc08315d28695c8732ad314 with python3

hut:
Thank you. The method is not used by ranger since 301b488b987b629add14c8a4831d237387ac5d28, but I guess it makes sense to keep it as an API of sorts, for use cases like yours.

PR #275, opened 2015-03-12
core.tab: added the after_chdir_hook

Vifon:
The added `after_chdir_hook` is called after changing the current ranger directory. The details are described in the docstring of the `api.chdir_hook` decorator.

I've got a feeling that I might have used the ranger signals as the `Tab.enter_dir` function calls `self.fm.signal_emit` but I've noticed it only after I've already finished. Additionally the signals seem like they are not meant to be used by the user, unlike the new solution.


hut:
Yep, the signal "cd" of the "fm" object should be used for this. See https://github.com/hut/ranger/wiki/Signals-and-Hooks . I'll add a little tutorial on how to use signals properly some time, until then you can just look at how it's done in the source code by searching for lines with "signal_bind" and "signal_emit".

PR #276, opened 2015-03-12
the register_linemode function reimplemented as a decorator

Vifon:
While I was working on #275, it occurred to me that the `register_linemode` function could be replaced with a bit more elegant decorator. In my opinion using it would be much more natural from the user's point of view.


hut:
Thanks =)

PR #280, opened 2015-03-16
Fix a typo s/self\.notify/self\.fm\.notify/

anekos:
s/self\.notify/self\.fm\.notify/


hut:
Thank you, didn't notice that ;_;

PR #282, opened 2015-03-21
ranger.py: fixed issues with $tempfile in embedded shell

eiiches:
Previously, $tempfile (namely, /tmp/chosendir) was not cleaned up correctly when the ranger process quit in `pwd` without moving to other directory.  This causes permission errors in multi-user environments trying to overwrite $tempfile created by a different user.

This commit solves the problem in two ways:
 - Correctly clean up temporary files
 - Avoid writing to the same temporary by using mktemp(1)

### DIFF
--- a/ranger.py
+++ b/ranger.py
@@ -9,7 +9,7 @@
 # default is simply "ranger". (Not this file itself!)
 # The other arguments are passed to ranger.
 """":
-tempfile='/tmp/chosendir'
+tempfile="$(mktemp)"
 ranger="${1:-ranger}"
 test -z "$1" || shift
 "$ranger" --choosedir="$tempfile" "${@:-$(pwd)}"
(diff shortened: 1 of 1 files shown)


eiiches:
I realized `--suffix` option of mktemp(1) isn't very portable. Removing all the options of mktemp(1) might be good, although I'm not so sure about this (I do not own Mac or BSDs to test with).

randynobx:
I can test with Mac (OS X Lion) and FreeBSD. What exactly are you looking to test with mktemp?

hut:
Thank you for the patch, but BSD's mktemp indeed has no --suffix option.

eiiches:
Thank you for your responses, and I'm sorry I hadn't have time to reply earlier.

Yes, I wanted know if mktemp on BSDs have --suffix option.
As it turned out it doesn't (thanks @hut), I'm updating the patch, to call mktemp without the option.

I tested this patch on Ubuntu 14.04 LTS and CentOS 6.5, and also confirmed that mktemp is included in the default installation of Cygwin (latest version).

hut:
Thank you for the update =)

jbyler:
Note that this change doesn't work on Mac OS X.  See my PR #330 with a fix.

PR #283, opened 2015-03-22
added tag management for rename, delete, bulkrename, and moving of files

### LINKED ISSUE #118: Tag management when moving/deleting files
Tags should be removed from the tags file when the tagged files are deleted. When moving or renaming files, they should keep their tag.

randynobx:
fixed typo in rename

added recursive checking to move, rename, and delete; refactored previos code

bugfix: moved recursive tag checking for delete to commands.py

Fixes #118

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -523,10 +523,16 @@ class delete(Command):
                 self._question_callback, ('n', 'N', 'y', 'Y'))
         else:
             # no need for a confirmation, just delete
+            for f in self.fm.tags.tags:
+                if str(f).startswith(self.fm.thisfile.path):
+                    self.fm.tags.remove(f)
             self.fm.delete()
 
     def _question_callback(self, answer):
         if answer == 'y' or answer == 'Y':
+            for f in self.fm.tags.tags:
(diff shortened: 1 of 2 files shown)


hut:
The approach for `:bulkrename` is unreliable, since the command consists of 2 steps: 1. edit the file file names, 2. review the commands. In step 2, the user can still change his mind and edit those commands. Your patch ignores step 2.

I don't know a good way to fix it. I don't think parsing the whole shellscript file that's edited in step 2 is an option. But we could at least check if the user changed the file in step 2 and only update the tags if he didn't.

hut:
Thanks :) I made a follow-up commit ( af17ede9ff216d09495672e00a36ba60fbf8bee3 ) fixing some issues. I don't remember what the exact problem was, but under some case, tags were not updated with `:bulkrename`. I removed the use of sha1 because it adds overhead with no gain (AFAIK comparing 2 long strings is faster in python than generating a checksum of them and comparing the checksum)

Also, please pay a bit more attention to coding style next time. For example, instead of `hash1= ...` write `hash1 = ...`, instead of `for old,new` write `for old, new`. See also https://www.python.org/dev/peps/pep-0008/

PR #290, opened 2015-03-29
Fix: djvu files are being opened with image viewers

arwer13:
This fixes, for instance, my problem with djvu files opened by image viewers. It also might be welcome for other possible "image-documents" types.  But I'm not sure, maybe there is a better way to cope with it.


hut:
Nice solution. Thank you :)

PR #292, opened 2015-03-30
:flat statusbar indicator added

Vifon:
Additionally I've removed the leading space from the filter indicator and added it to the front of the right side of the statusbar as a separate element.


hut:
Thank you and sorry for the delay :3

Issue #146, opened 2014-08-17
Using <Space> in map commands


notalex:
It would be great if one could use `<Space>` in map commands. Using a literal space does not play well with scripts that remove trailing whitespaces.

Also, `console find ` does not clearly indicate that there is a space at the end, while `console find<Space>` is much more expressive.

hut:
Fixed in fabab01bda21868be0d47e053c65b48a9102b820, see #293

PR #293, opened 2015-03-31
bugfix: prevent deletion of important extra spaces

andya9:
This fixes a bug in those editors which automatically delete extra spaces at the end of lines; without the fix when (after editing rc.conf) i.e. pressing r to open files we need to manually add a space between "open_with" and the command

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -241,13 +241,13 @@ map S shell $SHELL
 
 map :  console
 map ;  console
-map !  console shell 
+map !  eval fm.open_console('shell ')
 map @  console -p6 shell  %%s
-map #  console shell -p 
-map s  console shell 
-map r  chain draw_possible_programs; console open_with 
-map f  console find 
-map cd console cd 
+map #  eval fm.open_console('shell -p ')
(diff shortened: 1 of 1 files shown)


hut:
I realize this is an issue, but IMO this is not the ideal way to deal with this. Some extra symbol should be added that internally translates to a space.

This is related to #146.

andya9:
Thank’you for answering; do you mean we should create a variable space = " " and then join it to the command strings? I’m a beginner with python, but if this is what you mean I’ll do it

hut:
I just noticed that all commands except `alias`, `eval` and `map` allow the use of macros. So we can solve this by adding a macro named `%space` and expand it to `" "`. I implemented that in fabab01bda21868be0d47e053c65b48a9102b820.

andya9:
Thank’you very much @hut !
I’m not able to understand macros yet, but I’m glad it works  :smile: 
When I’ll be better with programming, I hope I can help more; I really like ranger :+1:

hut:
Np.

Macros are really easy, they're like abbreviations. For example. when you write %f, it's replaced with the current file name. Now, when you write %space, it's replaced with a space. :)

PR #302, opened 2015-04-18
Merge command

BarryPH:
I added the merge command for pasting files as was requested here: https://github.com/hut/ranger/issues/142. If used with the cut command files that can be merged will be, then the source directory of the cut will be deleted.

I did at one point set the merge command, when cutting, to merge all possible files and then only delete files from the source directory that were merged but I felt that was redundant as it is very similar to just copying and merging so I reverted it.

I still have all the code if you would prefer the second method for cutting and merging.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -343,6 +343,7 @@ map A  eval fm.open_console('rename ' + fm.thisfile.basename)
 map I  eval fm.open_console('rename ' + fm.thisfile.basename, position=7)
 
 map pp paste
+map pm paste merge=True
 map po paste overwrite=True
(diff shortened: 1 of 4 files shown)


hut:
Wouldn't it lead to data loss if a file already existed but was empty?

BarryPH:
Yes it would. That is why I was contemplating the other implementation. I can change the code to leave already existing files untouched if you think that would be best. It would be safer.

nfnty:

dufferzafar:
@hut I can see that the PR hasn't been merged, but is this feature available in ranger now?

toonn:
Don't think it's possible yet. You can overwrite but not merge.

PR #306, opened 2015-04-20
[STUPID+FIXME] Show modification date in main column. Terribly defunct code.

eMPee584:
Not meant for merging.. can this be done better?

### DIFF
--- a/ranger/container/directory.py
+++ b/ranger/container/directory.py
@@ -8,7 +8,7 @@ import re
 
 from os import stat as os_stat, lstat as os_lstat
 from collections import deque
-from time import time
+from time import time, strftime, localtime
 
 from ranger.container.fsobject import BAD_INFO, FileSystemObject
 from ranger.core.loader import Loadable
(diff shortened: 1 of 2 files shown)


hut:
This sounds like a job for a so-called linemode, see https://github.com/hut/ranger/blob/master/ranger/core/linemode.py

And the example plugin: https://github.com/hut/ranger/blob/master/examples/plugin_linemode.py

eMPee584:
Yes definitely.. I cnoded this before I tripped over the new linemode
implementation. Still didn't quite grok it, the rofl_23 example is a bit
basic :smirk:

hut:
You can add an "infostring" method to your linemode class, like in FileInfoLinemode ( https://github.com/hut/ranger/blob/master/ranger/core/linemode.py#L96 ). There, you can simply return the string you generated with strftime.

hut:
And https://github.com/hut/ranger/wiki/Custom-linemodes

PR #318, opened 2015-05-15
Added cmap for forward word deletion.

NoSuck:
The `<C-delete>` shortcut probably won't work with most terminal configs.


hut:
Is there a terminal that it does work on?

NoSuck:
I don't know.  Is there a way to map [3;5~ in ranger?  Vim can handle it, so I know it's possible...

hut:
ranger can't deal with that mapping. If you want to bind it, you need to extend the keybinding parser and the handle_input method in ranger.gui.ui. Not worth the effort IMO but up to you.

NoSuck:
Eh, I took that part out.  Please accept my puny patch!  It's not the size that counts!

hut:
Sure. Could you squash those commits into 1? I'd do it myself but you may want your name as the author/commiter :p

NoSuck:
Hey, I learned something today.  SQUASHED.

hut:
Thanks. And I didn't even notice that it was your first time ;)

PR #324, opened 2015-06-01
allow ranger to be sourced (". ranger") without being installed in PATH

Vifon:
It makes a difference if the user wants to clone the ranger repository
and start ranger using the full path to "ranger.py" without installing
it anywhere.

Previously it was possible to pass the path to ranger in the next
argument but it is unnecessary since we already have the path of the
currently executed script (unless the user wants to override it, then it
is still possible).

### DIFF
--- a/ranger.py
+++ b/ranger.py
@@ -10,7 +10,7 @@
 # The other arguments are passed to ranger.
 """":
 tempfile="$(mktemp)"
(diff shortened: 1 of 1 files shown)


hut:
That is a neat trick. Thanks =)

hut:
Except that it doesn't work in bash. D=

PR #330, opened 2015-06-10
Fix mktemp invocation to work on Mac OS X and Linux

jbyler:
mktemp on Mac OS X (and probably all BSDs) just returns an error when
invoked without any arguments.  The -t option used in this change is
interpreted differently on Mac OS X and Linux, and is deprecated on
Linux, but this invocation works as expected on both.

See discussion at https://unix.stackexchange.com/questions/30091
Another alternative would be to use Python's tempfile module:
https://docs.python.org/2/library/tempfile.html

### DIFF
--- a/examples/bash_automatic_cd.sh
+++ b/examples/bash_automatic_cd.sh
@@ -8,7 +8,7 @@
 # original directory.
 
 function ranger-cd {
-    tempfile="$(mktemp)"
+    tempfile="$(mktemp -t tmp.XXXXXX)"
     /usr/bin/ranger --choosedir="$tempfile" "${@:-$(pwd)}"
     test -f "$tempfile" &&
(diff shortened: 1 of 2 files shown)


jbyler:
Discussion: PR #282 broke my shell wrapper on Mac OS X.  This fixes again and avoids breaking for Linux.

Tested on Mac OS X 10.10.3 "Yosemite" and RHEL 6.1.

eiiches:
I'm the author of the PR #282 and am sorry for the inconvenience.
+1. This works perfectly on my Ubuntu 14.04.

hut:
I'm not very comfortable with using a deprecated feature but if it's the only portable option, I'll go with it.

jbyler:
Thanks @hut!  There are other options that don't use a deprecated feature, actually.  Here's one that's also works fine on Mac OS X and linux:

    tempfile="$(mktemp 2>/dev/null || mktemp -t tmp)"

PR #334, opened 2015-06-17
DEFAULT_PAGER/EDITOR

s5unty:
I think "sensible-pager" and "sensible-editor" are better :)


hut:
Heyho, I'll gladly take the TMPDIR patch but setting sensible-something as a default is a no-go because it's specific to debian.
Thank you^^

hut:
Btw, thank you very much for maintaining the debian ranger package and I really appreciate your desire to submit your patches to upstream =) @s5unty

PR #335, opened 2015-06-21
put the previewed image in right place and clean image(only w3mimgdisplay)

ssh0:
This small changes make the previewed image to put in right place and properly clean the image properly.

Before:

(with border)
before_with_border

(without border)
before_without_border

After:

(with border)
after_with_border

(without border)
after_without_border


hut:
Thank you :D

PR #336, opened 2015-06-21
added examples/plugin_ipc.py

Vifon:
I have added a new example plugin which allows to control ranger from the external scripts.


hut:
Looks good, thanks man

Issue #332, opened 2015-06-12
Broken filename shortening


hut:
> I've found what looks like a bug in ranger, although it seems quite harmless: when filenames are abbreviated because they're too long, the last line has an "1" instead of the tilde or ellipsis; this does not always happen (it seems to depend on the window's height).
>
>See the attached screenshot as an example; it shows ranger without config (so the problem doesn't come from any option setting).
>
> My version of ranger is "ranger-stable 1.7.1", running on Arch.

![screenshot](https://cloud.githubusercontent.com/assets/90548/8129690/e36e87dc-110b-11e5-8783-74d847e05a43.png)

hut:
fixed in c831d7fdd8abe37d6f46332f8e4cf8f272f16660

PR #337, opened 2015-06-21

ajtluser:
shortened.


hut:
You closed this, does it mean the patch doesn't work?

ajtluser:
Sorry. Still kinda new to this. Didn't know of a better way to change my pull request.

PR #338, opened 2015-06-21

ajtluser:
shortened.


hut:
It seems wrong. `_fix_surrogates(args)` only ever accepts 1 argument. When you write `_fix_surrogates(*args)` after a `len(args) > 1` check, it's guaranteed to pass >1 arguments to it, resulting in an exception.

Am I missing something?

ajtluser:
Yep, you're right. As far as I can tell now, the issue is that the function addstr can sometimes raise an exception even after having printed some characters. So it would print some, raise the exception, and the program would start printing the string again from the beginning but with the cursor now at the wrong coordinates.
At least that's what I think is happening.
I tried to fix the fix by storing the initial x, y in temporary variables, and then in case of an exception moving the cursor back to those coordinates. I think it works but it seems I made a mess when I tried to update this pull request. My apologies.

PR #339, opened 2015-06-25
Use setuptools if present

dufferzafar:
The main reason for this is the presence of `develop` command in setuptools.

So by running `sudo python setup.py develop` - ranger gets installed, but you can easily make changes to the code and won't have to reinstall it.

You can read more about it here:
http://pythonhosted.org//setuptools/setuptools.html#development-mode

It basically links an egg, so I had to add an appropriate line to .gitignore as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *~
+*.egg-info
 *.pyc
 *.pyo
 stuff/*
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,12 @@
 # This file is part of ranger, the console file manager.
 # License: GNU GPL version 3, see the file "AUTHORS" for details.
 
-import distutils.core
+# Setuptools is superior but not available in the stdlib
+try:
+    from setuptools import setup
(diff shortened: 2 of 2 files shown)


hut:
There's no downside as far as I can see, so why not. There you go.

Vifon:
It seems to break the main ranger executable when using its embedded sh-based launcher (`. ranger`) by replacing `bin/ranger` with a wrapper. Any way to fix it while keeping setuptools?

hut:
Yep, we need the original ranger.py because it's a specially designed polyglot file which can be interpreted as a bash script too, to activate an extra feature. (To change directory in the original shell after closing ranger.)

Then there is another issue: When using --prefix to install ranger to a directory outside of PYTHONPATH, it fails with an error:

2015-07-16-222600_628x427_scrot

hut:
I opened an issue about this at the setuptools repo:
https://bitbucket.org/pypa/setuptools/issues/444/ability-to-copy-scripts-unmodified

Issue #267, opened 2015-03-08
my bookmarks stopped being saved


tkossak:
I created some bookmarks with ```m``` key, and they work as they should, but now when i create new bookmark and restart ranger the bookmark is not saved: when i press apostrophe i don't see it on the list and it is not present in the file ```~/.config/ranger/bookmarks``` (i have rw permissions to the file).

Some options in ```~/.config/ranger/rc.conf```:
>    set show_hidden_bookmarks true
>    set autosave_bookmarks true

Please help, thank you.

hut:
does it happen with 1 particular directory, or everywhere?

tkossak:
ok i fixed it - it was my fault, the owner of the ```~./config/ranger/bookmarks``` was root instead of me. Don't know how it happened but I changing the owner to me and it works again.

Issue #59, opened 2013-01-20
bookmarks file owner


mortbauer:
every now and then i start ranger as "su" which works just fine, but after closing ranger it changes the owner and the group of the bookmarks file of my normal user to root, which isn't very nice.
 It isn't a very big deal but maybe it would be easy to change that? I don't know, but ranger is the best i wanna say. Thanks

PR #340, opened 2015-06-26
preserve permissions when updating bookmarks. Fix for #59

igoyak:
Proposed fix for the (non-)bug #59, where the owner of the bookmarks file is changed to root when running ranger as root but with a user's environment. This could be called user error, but this patch will hopefully prevent further confusion such as #267. 

The try-except statement catches the (weird) case where the file is already owned by another user. 

Should this also be applied to other files managed by ranger?


hut:
Thank you very much

PR #353, opened 2015-08-04
Add ability to use ranger as file-chooser in gvim

ngoonee:
Previously RangeChooser() would just fail silently in gvim. With this change, it will open an xterm and open ranger there. Everything else works as before.


hut:
Thank you very much

ngoonee:
You're welcome, thank YOU for such a useful piece of software

PR #358, opened 2015-08-17
Implemented Subversion VCS support.

rdburns:
Implemented initial Subversion VCS support for Ranger. Based it off of the Hg support. Seems to work well, feedback welcome.

### DIFF
--- a/ranger/container/fsobject.py
+++ b/ranger/container/fsobject.py
@@ -281,6 +281,8 @@ class FileSystemObject(FileManagerAware, SettingsAware):
                 backend_state = self.settings.vcs_backend_hg
             elif self.vcs.vcsname == 'bzr':
                 backend_state = self.settings.vcs_backend_bzr
+            elif self.vcs.vcsname == 'svn':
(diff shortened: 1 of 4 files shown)


hut:
Looks good so far. I cleaned it up a little, I hope it all still works as intended:

* removed whitespace in some places
* removed unused import
* added a "r" in front of strings using backslashes like "\s" in regular expressions

I couldn't really successfully test it though, all it changed was add a "(svn)" in the status line. This happens before and after the clean up.

Also there are 2 problems in the ranger/ext/vcs/svn.py for which I need your input:

Line 202:
```python
    def get_branch(self):
        """Returns the current named branch, if this makes sense for the backend. None otherwise"""
        return None
        branch = self._svn(self.path, ['branch'], catchout=True)
        return branch or None
```

and line 154:
```python
    def checkout(self, rev):
        """Checks out a branch or revision"""
        raise NotImplementedError
        self._svn(self.path, ['update', rev])
```

Both functions contain code which is never executed. Please either remove the line that prematurely terminates the function or remove the code after that line.

PR #359, opened 2015-08-17
make mv/cp generators yield progress

notEvil:
a small patch to `CopyLoader` which would allow more flexible plugins (e.g. do mv/cp in elevated subprocess using some custom Loader for progress update)


hut:
Looks good, thanks :)

notEvil:
Thx :)
just realized that `import time` in ranger/ext/shutil_generatorized.py is not necessary. My mistake

hut:
I removed it, thanks ^^

hut:
Actually, this breaks the progress bar when copying multiple files at once. Can you please take a look at the problem?

notEvil:
you are right. a few lines were missing. with the new commit it should be fixed. should I create a new pull request?

hut:
Nope, it's fine :) thanks!

Issue #341, opened 2015-06-27
fish shell and :shell command


iamevn:
I'm actually still getting #300 or something like it on 1.7.1
```
fish: Variables may not be used as commands. Instead, define a function like 'function SHELL; /usr/bin/fish $argv; end' or use the eval builtin instead, like 'eval $SHELL'. See the help section for the function command by typing 'help function'.
Standard input: $SHELL
                ^
evan@archpad ~> ranger --version
ranger-master 1.7.1

Python 3.4.3 (default, Mar 25 2015, 17:13:50) 
[GCC 4.9.2 20150304 (prerelease)]
```

jihygk:
I also have the same issue.
```
jaykim@SVN082-Servian-2 ~/ranger> ranger --version
ranger-stable 1.7.1

Python 2.7.6 (default, Sep  9 2014, 15:04:36)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)]
jaykim@SVN082-Servian-2 ~/ranger>
```

tannhuber:
Same problem here.
@jihygk Your workaround also doesn't work: `Command not found`
So I use:
```
map S shell zsh
```
with the following `$HOME/.zshrc`:
```
#!/bin/zsh
/usr/local/bin/fish
exit
```
I use the same workaround for invoking fish from vim.

PR #368, opened 2015-09-08
really fix #300 and #341

### LINKED ISSUE #300: Starting fish shell with `:shell $SHELL`
After updating from v1.6.1 to v 1.7.0, I can no longer start a shell from within ranger by pressing S.
Fish tries to start and immediately returns with the error message:

    fish: Variables may not be used as commands. Instead, define a function like 'function SHELL; /usr/bin/fish $argv; end' or use the eval builtin instead, like 'eval $SHELL'. See the help section for the function command by typing 'help function'.
    Standard input: $SHELL

iamevn:
os.environ['SHELL'] returns '/usr/bin/fish' which != 'fish'
this should fix fish compatibility for #300 and #341.

### DIFF
--- a/ranger/core/runner.py
+++ b/ranger/core/runner.py
@@ -146,7 +146,7 @@ class Runner(object):
         # Set default shell for Popen
         if popen_kws['shell']:
             # This doesn't work with fish, see #300
-            if os.environ['SHELL'] != 'fish':
+            if not 'fish' in os.environ['SHELL']:
                 popen_kws['executable'] = os.environ['SHELL']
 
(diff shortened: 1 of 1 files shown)


hut:
I think the `"fish" in os.environ["SHELL"]` condition is too broad. I don't know another shell that contains "fish" in its name, but maybe one day there will be one...

On the other hand, it's probably more likely that fish sets yet another $SHELL string like `/usr/bin/fish2` in the future...?

Anyway, this is just a workaround and needs to be maintained in the future either way. So I'll just go with your version.

PR #369, opened 2015-09-12
Get rid of garbage `-E` character in preview display on mac

### LINKED ISSUE #344: scope.sh -E causes issues on OS X
The `-E` flag that follows `/bin/echo` on lines 31 and 39 in the default **scope.sh** causes:

* `-E`to be output at the start of every previewed text file
* Previews of PDFs, archives and HTML pages to not appear

Removing the `-E` prevents the two above issues (but not sure if that has knock-on effects?) - it seems like it considers the `-E` as part of the echo input rather than a flag for it.

Platform: OS X Yosemite (10.10.4)
Terminal: Terminal.app and iTerm2
Shell: zsh

ptzz:
* OSX does not support -E with /bin/echo

* On Linux the -E option is set by default

Fixes #344

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -28,7 +28,7 @@ maxln=200    # Stop after $maxln lines.  Can be used like ls | head -n $maxln
 
 # Find out something about the file:
 mimetype=$(file --mime-type -Lb "$path")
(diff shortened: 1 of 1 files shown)


hut:
Makes sense, thank you

Issue #371, opened 2015-09-14
counldnt' preview image without extension name


royaso:
ranger 1.6 
ubuntu package

if 'picture' is a image but without the '.jpg' extension name, preview dont work

collector1871:
Try to look for a file called `scope.sh`. 
In my Debian it is located in:
`/etc/ranger/data/scope.sh` - this is symlink `/usr/share/pyshared/ranger/data/`.

hut:
fixed in #372.

royaso:
I used rxvt-unicode , w3mimgdisplay  to display images. It was fine , only problem is noe suffix name no image preview.

 After the fix , only img2txt no matter whether or not suffix name

hut:
You need to update your ~/.config/ranger/scope.sh

royaso:
@hut You are right! Thanks! IT works!!

royaso:
@hut WHOA! I dont know why, but It gets better than I expected!!

I just try it in terminator which doesnt support images preview natively even though w3m-img is installed, But Now IT works with it too!! AMAZING!

Issue #263, opened 2015-02-21
Custom image previews ignoring the "preview_images" option


Vifon:
The custom images previews (these using the `$cached` env variable) are completely ignoring the `preview_images` option. In my opinion they should be affected by it just like the regular image previews.

**Solution 1**: Just don't display the generated image when `preview_images` is disabled. **Problem**: `scope.sh` would not be able to fall back to a textual preview.

**Solution 2**: Set `$cached` only when `preview_images` is `true` and then `scope.sh` would branch on `[ -n "$cached" ]`. **Problem**: the backward compatibility. The existing `scope.sh`s using this feature would need to consider the possibility of `$cached` being empty. It would break the existing configs.

**Solution 3**: Add an another variable to indicate whether an image should be generated and ignore the image if it still is created. It would work like the previous solution but would not break the existing configs.

hut:
fixed in #372

PR #372, opened 2015-09-14
Use scope.sh for all the previews

GermainZ:
… instead of calling w3mimgdisplay for (potentially non-supported, e.g.
SVG files) images, which also has the side effect of making custom
previews for images impossible.

This patch also handles #263 using solution 3, in order not to break
backward compatibility:

> **Solution 3:** Add an another variable to indicate whether an image
> should be generated and ignore the image if it still is created. It would
> work like the previous solution but would not break the existing configs.

scope.sh has also been slightly refactored to take into account this
change.

Should also take care of #371.

---

````
One Scope to rule them all, One Scope to find them,
One Scope to bring them all and in the terminal preview them
In the Land of Ranger where the Data lies.
````

### DIFF
--- a/ranger/container/file.py
+++ b/ranger/container/file.py
@@ -76,8 +76,6 @@ class File(FileSystemObject):
         if self.fm.settings.preview_script and \
                 self.fm.settings.use_preview_script:
             return True
-        if self.image and self.fm.settings.preview_images:
-            return True
         if self.container:
             return False
         if PREVIEW_WHITELIST.search(self.basename):
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -799,14 +799,11 @@ class Actions(FileManagerAware, SettingsAware):
             return
 
         pager = self.ui.open_pager()
-        if self.settings.preview_images and self.thisfile.image:
-            pager.set_image(self.thisfile.realpath)
+        f = self.thisfile.get_preview_source(pager.wid, pager.hei)
+        if self.thisfile.is_image_preview():
+            pager.set_image(f)
         else:
-            f = self.thisfile.get_preview_source(pager.wid, pager.hei)
-            if self.thisfile.is_image_preview():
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by hut** [comment]
Excellent, thank you. Sorry for the wait -- because it's *you*, I had to take special care to verify that there was no malicious code embedded in the patch. :3

**Comment by GermainZ** [comment]
One day, I'll overwhelm you with patches and sneakily slip some code to open
https://rms.sexy on ranger startup.

<3

On Sat, Sep 19, 2015 at 10:31 PM, hut < notifications@github.com [notifications@github.com] > wrote:
Excellent, thank you. Sorry for the wait -- because it's you , I had to take special care to verify that there was no malicious code
embedded in the patch. :3

—
Reply to this email directly or view it on GitHub
[https://github.com/hut/ranger/pull/372#issuecomment-141700705] .[https://github.com/notifications/beacon/ACxRlVwBemNw7cvz54wUDCxm28_iS-63ks5oza-IgaJpZM4F8-73.gif]

**Comment by Vifon** [comment]
Should we consider #263 resolved?

**Comment by hut** [comment]
I believe so.

**Comment by hut** [comment]
@kattrali could you do a quick test if your iterm2 previews still work in the current git version? I'm not sure if this patch left your feature intact.

**Comment by kattrali** [comment]
@hut Sure, I'll check back in a bit after doing a few tests.

**Comment by kattrali** [comment]
@hut I checked running latest `ranger.py` against my existing [configuration](https://gist.github.com/kattrali/4152f05f774c5f6c6c4d#file-rc-conf-L62,L75):

```
set use_preview_script false
set preview_images true
set preview_images_method iterm2
```

and #268 was broken by 40bd7bc. I might have to get a bit clever to find a fix, since using a pager breaks the `iterm2` format.

hut:
Okay, thank you. I'll open an issue about this and try to find a fix tomorrow

PR #388, opened 2015-10-06
core/actions.py: Move some code inside the try block

Vifon:
The previous version seemed to be a mistake.


hut:
try/except/pass blocks, especially large ones, are bad practice in python. Is there any problem that's avoided with this patch?

Vifon:
The previous try block just didn't make any sense. It assigned a property to the local variable and then the variable was used outside of that block. Maybe we should remove this try altogether?

hut:
Right, thanks^^
Fixed in 91d784a677604c4bb82db3007c5606a258971acc

PR #389, opened 2015-10-06
Fix the doc extraction for "dump_commands" and add some new documentation

Vifon:
I've fixed the handling of the methods of the `Actions` class in the `dump_commands` function. Previously their docstrings were completely ignored. The new behavior is described in the commit message of 45ee436 and is applied to all the commands, not only the `Actions` class methods.

Additionally I've added the documentation to some of these methods.


hut:
Awesome, thanks :D

PR #390, opened 2015-10-06
Add an alternative highlight function

Vifon:
I've moved the highlight-specific arguments inside our wrapper and added an alternative one using pygments.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -44,7 +44,10 @@ dump() { /bin/echo "$output"; }
 trim() { head -n "$maxln"; }
 
 # wraps highlight to treat exit code 141 (killed by SIGPIPE) as success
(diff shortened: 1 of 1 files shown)


hut:
How about generalizing the highlight function to a generic "protect any command from sigpipes"-function? We could then use it for both highlight and pygmentize, and use the usual mechanism with the try/dump functions to determine which one should be started.

Vifon:
Like this?

hut:
Like 9f87b7d3a7374f369f0c04bd3758cc9da7a2a67b

hut:
My last attempt was incorrect, as you explained on IRC. Does 82ae8743adf352387dd1f3996af1dbe5a25a02a3 sound fine?

Vifon:
Looks good to me.

hut:
Merged

PR #391, opened 2015-10-08
scope.sh: Support unicode extensions

Ambrevar:
Most (all?) `tr` implementations do not support unicode.
Use `awk` instead for turning extensions to lowercase.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -30,7 +30,7 @@ maxln=200    # Stop after $maxln lines.  Can be used like ls | head -n $maxln
 
 # Find out something about the file:
 mimetype=$(file --mime-type -Lb "$path")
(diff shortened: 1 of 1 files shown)


hut:
BSD/MacOS/Linux have different default awk implementations, I believe. Do you have access to a BSD machine to test it on? Works fine on my linux.

Ambrevar:
No worries, AWK is POSIX, and 'tolower()' is shared across all known implementations.
See awk(1p).
'tr' implementation is also different across *BSD, GNU and OSX by the way.
Since you are an Arch Linux user, you can try 'nawk'. It is the reference implementation by Kernighan and It is used under FreeBSD at least.

hut:
Alright, thanks :)

PR #396, opened 2015-10-12
Allow the execution of the executable binary files via rifle

Vifon:
It shouldn't be a security issue because the user still needs to enable the executable bit of the binary.

### DIFF
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -103,6 +103,7 @@ ext rb  = ruby -- "$1"
 ext js  = node -- "$1"
 ext sh  = sh -- "$1"
 ext php = php -- "$1"
+mime application/x-executable = "$1"
 
 #--------------------------------------------
 # Audio without X


arwer13:
+1
I wonder why it isn't in default config.

hut:
I don't like the idea of accidently executing programs when opening them. Accidental opening of images or PDFs is no problem, but people may have some executables on their machine that they don't trust and that may contain malicious code.

As a compromise, I'll put this entry last, below the editor/pager entries, where it will never trigger by default.

Vifon:
> I don't like the idea of accidently executing programs when opening them.

Doesn't it already happen with the scripts? I see no major difference between the executable scripts executed by another process and an executable binary file. Even worse: in the case of the scripts the missing executable bit seems to be ignored. Maybe we should consider disabling it too. Either way, the current behavior seems inconsistent.

Vifon:
Ok, it seems the "code" section of rifle.conf seems to be ignored because these are text files. Why is it even there?

PR #398, opened 2015-10-15

ajtluser:
shortened.

### DIFF
--- a/ranger/gui/curses_shortcuts.py
+++ b/ranger/gui/curses_shortcuts.py
@@ -25,20 +25,28 @@ class CursesShortcuts(SettingsAware):
     """
 
     def addstr(self, *args):
+        y, x = self.win.getyx()
+
         try:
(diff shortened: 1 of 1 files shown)


ajtluser:
A couple months ago I tried to fix this one, but messed up my commit (sorry about that).

Below I copy-pasted a comment I made back then that should explain my attempt to fix this.

>Yep, you're right. As far as I can tell now, the issue is that the function addstr can sometimes raise an exception even after having printed some characters. So it would print some, raise the exception, and the program would start printing the string again from the beginning but with the cursor now at the wrong coordinates.
At least that's what I think is happening.
I tried to fix the fix by storing the initial x, y in temporary variables, and then in case of an exception moving the cursor back to those coordinates. I think it works but it seems I made a mess when I tried to update this pull request. My apologies.

I hope I got it right this time.

hut:
Yep, that was missing. Thank you.

PR #399, opened 2015-10-16
commands.py: Pass number to tab()

nfnty:
shift-tab functionality for scout etc.

### DIFF
--- a/ranger/api/commands.py
+++ b/ranger/api/commands.py
@@ -6,6 +6,7 @@
 import os
 import ranger
 import re
+import inspect
 from collections import deque
 from ranger.api import *
 from ranger.core.shared import FileManagerAware
@@ -106,7 +107,7 @@ class Command(FileManagerAware):
     def execute(self):
         """Override this"""
 
-    def tab(self):
+    def tab(self, tabnum):
         """Override this"""
 
     def quick(self):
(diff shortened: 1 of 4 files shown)


hut:
This would break every custom commands.py for a feature nobody asked for, even though I guess it makes sense to have this. Not sure what to do, maybe I can think of a cleaner way to add this fuctionality.

hut:
Some options:
* add a function called `tab_shift`
* add an attribute called `self.shift_pressed` (probably a bad idea because of unnecessary statefulness)

nfnty:
Fixed it by inspecting `tab()` arguments before calling it.

hut:
That's an elegant solution, but I wonder if it might confuse the users. In standard python, it never happens that you can override methods with different numbers of arguments, unless you use *args or **kwargs. At least I haven't ever seen it. Either way, I'll merge it. Most likely, nobody will ever perceive this as an issue. :D

hut:
Wait, IMO tabnum is kind of counterintuitive. How about using tab(self, shift=True/False) instead?

nfnty:
With tabnum you can bind shift to any size. I.e. `cmap <C-TAB> eval fm.ui.console.tab(-5)`

nfnty:
Rebased to master.

PR #404, opened 2015-10-20
colorscheme: check for compiled python files

lovebug356:
Some systems only provide *.pyc files


hut:
Didn't think of that one, thanks!

PR #409, opened 2015-10-24
Added command for jumping to first file/directory.

NoSuck:
Having selfishly hogged such functionality for months, I cleaned up the code in response to a user request.


hut:
How about another name for that command? Maybe "skip". I think the name "jump" could be useful for another command later on. E.g. if we integrate autojump into ranger or something.

nfnty:

PR #411, opened 2015-10-24
Rifle's ext condition now handles fringe cases.

NoSuck:
Previously, rifle would recognize certain non-extensions as file extensions.  Three examples:

/etc/modules.d directory opening as a D source code file
./jpg directory opening as JPEG file
~/.mkv text file opening as an MKV movie file

These modifications only evaluate a file extension if it is part of a file and has a file name before a period.


hut:
Thanks :3

PR #421, opened 2015-11-13
open_all_images support for imv

autrimpo:
[imv](https://github.com/eXeC64/imv) now supports sxiv-like 'start-at' option.
This adds imv support for the **open_all_images** option.


hut:
Thank you!

PR #422, opened 2015-11-14
core.fm: Save the version of the copied configs

Vifon:
This will allow to later check against the stored version upon upgrade to inform the user they may need to update their config. Like the recent issues with `scope.sh`. Even if we'll not automate it, it will be possible to ask the user for this info to help resolve the issues.

I've decided to use a plain-text format instead of using `repr` because the added complexity is negligible while the format being much more user friendly.

### DIFF
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -218,12 +218,31 @@ class FM(Actions, SignalDispatcher):
             self.input_blocked = False
         return self.input_blocked
 
+    def get_base_config_versions(self, version_file_path):
+        """get the versions of the configs the user's configs are based on"""
+        versions = {}
+        try:
+            with open(version_file_path, 'r') as version_file:
+                for line in version_file:
+                    config, _, version = line.strip().partition(": ")
+                    versions[config] = version
+        except IOError:
+            pass
+        return versions
+
     def copy_config_files(self, which):
         if ranger.arg.clean:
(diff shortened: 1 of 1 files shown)


hut:
I don't think this is the right approach because the `base_version` file will get out of sync as soon as the user starts fiddling with the configs. The best alternative that comes to mind are magic comments in every config file, what do you think of that? As in, looking for /.*version=([0-9.])+/i in the first 5 lines, or something.

Vifon:
I've considered the approach with magic comments. If you think it's manageable, then maybe it is, but it would require you to update all of them with each release. And it still may get out of sync if the user just copies some parts.

My idea is not to get a 100% correct information but at least some hint. Right now to cannot reason about the config age at all. If the user merges their config manually, they will probably notice the `base_version` file. We could document it somehow (a comment on top?) to tell the user to update it after a merge.

dquinton:
Hi, do you keep a "major changes" file somewhere?  with a list of added/changes values/code made to rc.conf, scope.sh (the files user is likely to fiddle) - then this file is copied to user config

This will help a user in case they want to use the default clean config files, if there's a problem, but if they just copy all their own mods to them w/o knowing any vital changes made, they will still get problems.  Then, rc.conf doesn't really need reverting to original version (I find), but scope might.

Vifon:
I'm working on an alternative superior solution, closing this one.

PR #428, opened 2015-11-16
Add the version compatibility checker (again)

Vifon:
This checker scans the first 5 lines of each config for a version specification and then compares it with a list of the oldest compatible versions.

I've initialized it with `1.7.9999` but feel free to change it.

By default ranger will refuse to run if there are some outdated configs. It can be overridden with `--force-outdated` (a new option) or `--clean`. In such case, it will print the information which config(s) in which version is/are to blame.

The unit tests and docstrings in the `ranger.ext.version_check` module should more or less explain how it works.

Currently a manual update of the config versions is needed. Updating them only makes sense when there is a known compatibility break. After some thinking I've concluded that automatic appending of the date of the latest Git commit wouldn't help much. Adding it would not hurt but it can be done later if needed. In my opinion the only benefit would be a narrowed down commit range containing a compatibility break when a user reports some issues along with a config version.

Currently there are no checks for a too new config version (compared to the running ranger).

Regarding the version convention for the config files for the devel version, my proposal would be to use `1.7.9999-git-2015-11-16` and so on. To clarify: there is no need to change the version convention of the whole ranger, though using the same leading part (before "-git...") would be preferable.


nfnty:
Could you list all reasons why this is a good idea? What are you specifically trying to solve?

Vifon:
Closing for now. I'll need to rewrite it sometime in the future.

Vifon:
Let me elaborate a bit.

Back then we used to receive multiple reports weekly about broken previews due to `scope.sh` being significantly changed. It would allow us to notify the user in a simple and orderly manner that their config needs to be updated, and which config specifically. It could be a simple one-time notification of course but it doesn't have as much flexibility and is less future-proof, for example when a user skips multiple versions. It would allow us to more freely break the compatibility when necessary.

PR #430, opened 2015-11-18
core.main: Log the fact of disabling the file previews for root

Vifon:
It seems like a good idea to document somewhere the fact that the previews are disabled for root. This patch appends an information to the <kbd>W</kbd> log.


hut:
Thanks, tho I don't think it will help much because most people don't look there ;_;

Vifon:
It will at least help with support: It was one of the first questions I've asked on IRC and I got nothing.

More importantly, I've grepped the whole repository for "root" and it's not documented anywhere.

PR #448, opened 2016-01-10
xdg: Add support for XDG_DATA_HOME

### LINKED ISSUE #303: bookmarks, tags and history should be in XDG_DATA_DIR or ~/.local/share/ranger

genkimarshall:
I think this is all. It didn't seem like any docs needed to be updated (?).
Also, not sure about the `_relpath` factoring out.
Resolves #303.

### DIFF
--- a/ranger/__init__.py
+++ b/ranger/__init__.py
@@ -33,6 +33,8 @@ VERSION = 'ranger-master %s\n\nPython %s' % (__version__, sys.version)
 # If the environment variable XDG_CONFIG_HOME is non-empty, CONFDIR is ignored
 # and the configuration directory will be $XDG_CONFIG_HOME/ranger instead.
 CONFDIR = '~/.config/ranger'
+# Similarily for XDG_DATA_HOME:
+DATADIR = '~/.local/share/ranger'
 
 # Debugging functions.  These will be activated when run with --debug.
 # Example usage in the code:
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -99,7 +99,7 @@ class FM(Actions, SignalDispatcher):
         self.image_displayer = self._get_image_displayer()
 
         if not ranger.arg.clean and self.tags is None:
-            self.tags = Tags(self.confpath('tagged'))
+            self.tags = Tags(self.datapath('tagged'))
         elif ranger.arg.clean:
             self.tags = TagsDummy("")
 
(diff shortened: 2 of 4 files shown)


hut:
Sounds like the right thing to do, but what about the 10k confused console cowboys who will wonder where their important bookmarks and tags went?

genkimarshall:
I was wondering about that. Is there any precedence of this sort of
breaking-change for ranger?

Here are some not-ideal, but possible thoughts at least for discussion...
- Wait until hypothetical 2.0.0 and document it as something you must
  handle.
- Detect: If history/tags/bookmarks are in the old location, print to the
  console saying this is deprecated. Then it can perhaps continue to
  use those until the user moves them over to XDG_CONFIG_HOME location.
- Or instead, it could exit with error, explaining that you must move them.
- It could perhaps be another command-line / env-var option, but...

Perhaps this sort of change is simply not worth it. In that case this PR
and the issue can simply be closed.

hut:
Using ~/.config/ranger/{tagged,bookmarks,history} if they exist, and ~/.local/share/ranger/* if they don't, sound like the most sensible way to handle this. In time, most users will have their data in ~/.local/share/ranger.

hut:
Oh, don't forget to place `~/.config/ranger/copy_buffer` file in `$XDG_DATA_HOME` too, in case it isn't yet.  It's used in the `:save_copy_buffer` and `:load_copy_buffer` commands which transfer yanked files to different ranger instances.

genkimarshall:
Quoting hut (2016-01-31 14:35:11)
> Oh, don't forget to place `~/.config/ranger/copy_buffer` file in
> `$XDG_DATA_HOME` too, in case it isn't yet.  It's used in the
> `:save_copy_buffer` and `:load_copy_buffer` commands which transfer
> yanked files to different ranger instances.

Thanks for the help/feedback. Unfortunately, school has just started and
because I'm currently sitting on a fresh install and not using ranger ATM,
I'm not finding enough motivation to finish this off right away. So if
anyone reading this wants to, please go ahead. Otherwise, I will get to
this sometime in the future.

soredake:
any updates?

nfnty:

genkimarshall:
@nfnty I think @hut makes good points in his last two comments. But, this PR doesn't implement those items yet.

I don't think I'm going to make progress on this in the near future. So to anyone: Please feel free to use this work in a new PR.

PR #466, opened 2016-01-29
update Iterm2 ImageDisplayer to work within tmux

a-morales:
On OSX when using ranger inside tmux, images won't display if the setting is enabled. This is a workaround similar to Iterm2's [`imgcat`](https://github.com/gnachman/iTerm2/blob/7fb596f04b8e8c7e730efe1a033317f121a2f943/tests/imgcat)

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -199,11 +199,18 @@ class ITerm2ImageDisplayer(ImageDisplayer, FileManagerAware):
         image_width = self._fit_width(
             image_width, image_height, max_cols, max_rows)
         content = self._encode_image_content(path)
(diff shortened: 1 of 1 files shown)


hut:
Looks good, can anybody else with iterm2 confirm that this actually works?

hut:
I'll just trust you with this and merge it. Thanks for your contribution.

jottr:
I can not confirm this to be working on OSX in iTerm2 when running from within tmux. 
It works without problems within a normal iTerm2 session. As soon as I enter a tmux session, image preview within ranger stops working. 

Tested with **ranger 1.7.2** installed from homebrew, and ranger installed from `HEAD`, via `brew install ranger --HEAD`.

`iTerm2: Build 2.9.20160426
`

Issue #447, opened 2016-01-09
Bookmarks aren't updated on renaming


evgorchakov:
It appears that if a directory is bookmarked and then renamed (via `:rename` or `:rename_append`), the bookmarks file isn't updated, which leads to that particular bookmark becoming inaccessible since that particular directory ceases to exist. Is this by design? If not, could there be an option to update the bookmarks file on renaming? I'm willing to work on it if necessary.

charignon:
I sent #469

hut:
Fixed in #469, thanks @charignon

hut:
Though it doesn't work yet if you rename the parent directory...

charignon:
That's right, I was planning to send another patch for that.

PR #469, opened 2016-01-31
bookmarks: fix issue #447 Bookmarks aren't updated on renaming

charignon:
I checked that it was working by:
- Creating a bookmark on a folder
- Renaming the folder
- Accessing the bookmark and checking that it got me to the right place


hut:
Thanks^^

PR #470, opened 2016-02-01
bookmarks: better fix for issue #447

charignon:
This is a better fix for issue #447, it works even if we change a parent of a bookmark.


hut:
This will cause a false-positive when e.g. the current file is `/home/hut` and the bookmark being checked is `/var/backups/home/hut/.config`.

PR #471, opened 2016-02-01
Add py.test tests structure and tests for bookmarks and history

charignon:
- Adds py.test tests for bookmarks and history. This will allow to easily switch to `collections.deque` for history without regression.
- Also do some minor renaming / removing unused methods
- Add the tests to the makefile (now requires py.test to run)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 stuff/*
 doc/ranger.1.html
 build
+pytestdebug.log
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,7 @@ test:
 		echo "Testing $$FILE..."; \
 		RANGER_DOCTEST=1 PYTHONPATH=".:"$$PYTHONPATH ${PYTHON} $$FILE; \
 	done
+	py.test tests
 
 man:
 	pod2man --stderr --center='ranger manual' --date='$(NAME)-$(VERSION)' \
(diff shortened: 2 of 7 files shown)


hut:
Sorry for the late reply.

Ranger currently uses doctests for testing, see for example https://github.com/hut/ranger/blob/master/ranger/ext/signals.py , you can run them with "`make test`". Why choose another framework?

charignon:
I sent test because the code is untested and has bugs (found by the tests). It is possible but hard to test anything beyond a single function with doctest. If you want to add the same amount of coverage for bookmarks for example your source code will be unlegible :)

hut:
Okay, I'm willing to give py.test a try, even if it breaks consistency, because it looks rather awesome from what I've seen. Thanks :)

About your History.rebase() fix in f54fe5ac282254f6ecd6de70d21b20889fae5b43,... Actually this was intended behavior to allow this use case:

1. You move to directory A and want to open another tab in the same directory.
2. Press `^N` to open a new tab in $HOME
3. Press `H` (history back) to instantly get back to directory A
4. Press `L` (history forward) to go forward to $HOME

I'm guessing that you noticed unexpected behavior when opening a new tab with `alt+number` instead of `^N`, because your patch fixes that while breaking the use case I described.

The proper fix would test if the top (or the current?) entries of both histories are the same and apply the offset only when they are. I'll tackle this later.

hut:
My bad, everything after "I'm guessing" turned out to be wrong after checking it. Now I don't see anything that your patch fixes, other than your unit test. Maybe we should define the behavior of History.rebase() more clearly and fix the unit test accordingly.

hut:
Fixed, the `index_offset` was simply calculated wrong

PR #490, opened 2016-03-08
Make natural_sort's behaviour better defined

lverweijen:
The old version relied on comparing integers to strings which is not
supported in Python 3 anymore and not a good practice in general anyway.
It was also the case that 'hello2' was ordered before 'hello' instead of
after, which I found to be non-intuitive.


hut:
Thank you, although I wonder how you could run those tests without a syntax error. Didn't work for me without e13b9212537e3cdcf586e946aff731aaf051b21a.

lverweijen:
It seems that's because the syntax changed in python 3.

PR #491, opened 2016-03-08
Support 256 colours if available

lverweijen:
For pygmentize and highlight.


hut:
Cool, didn't know this existed. Thanks^^

PR #503, opened 2016-03-20
open_all_images support for pqiv

noctuid:
This adds [pqiv](https://github.com/phillipberndt/pqiv) support for ranger's `open_all_images` option.

It might also be nice to put pqiv somewhere in the default `rifle.conf`. I don't know how high priority it would have, but I'd say it's one of the more rarely installed image viewers.


hut:
Thanks. And yes, let's add pqiv. Please either tell me how to invoke it or send me a patch. About the order: It has 37 stars on github, so I guess we can put it at the top. :P

PR #507, opened 2016-03-21
set_bookmark for directories other than the current

cirosantilli:
Sample usage:

    :set_bookmark a /home/me/a

Motivation: make it easier to set bookmarks from environment variables
on rc.conf.


hut:
Thank you!

PR #512, opened 2016-03-21
Ignore generated man pages and remove them from repo

cirosantilli:
Those files are genereated on make man, it's better to not git track output
files.


nfnty:
This breaks `setup.py`

hut:
I generally don't include generated files, but I make an exception for the man pages, because I think it's more important to have documentation than to save space. Also it would add a dependency to perl. No way. :)

PR #521, opened 2016-04-05
Add ModificationTimeLinemode

### LINKED ISSUE #520: Displaying mtime in linemode requires subclassing?
Hi,

Forgive me if this is a dupe, but I couldn't find any.

Am I correct in understanding that to display the mtime in the file list (`ls -l`-style) requires subclassing the LinemodeBase class?

If so, may I suggest that this be done and added as a built-in option?  It's good that we can sort by mtime, but sometimes it's necessary to see the actual mtimes.  Like right now, I am clearing out some old temp files, and I need to see the mtimes so I know which ones are old.

Thanks.

Vifon:
Closes #520.


hut:
Thanks, I merged alphapapa's improved version.

PR #523, opened 2016-04-06
Add MtimeLinemode and SizeMtimeLinemode

### LINKED ISSUE #520: Displaying mtime in linemode requires subclassing?

alphapapa:
Thanks to @Vifon for his help.  Fixes #520.  Conflicts with #521.

It might be worth considering making the `sizemtime` linemode the default, because it most closely resembles the output of `ls -l`.

### DIFF
--- a/ranger/container/fsobject.py
+++ b/ranger/container/fsobject.py
@@ -79,7 +79,8 @@ class FileSystemObject(FileManagerAware, SettingsAware):
     _linemode = DEFAULT_LINEMODE
     linemode_dict = dict(
         (linemode.name, linemode()) for linemode in
-        [DefaultLinemode, TitleLinemode, PermissionsLinemode, FileInfoLinemode]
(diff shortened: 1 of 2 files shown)


hut:
Merged, thanks. I also added some key bindings, `Mm` and `Ms`. I object against making it the default though, because it takes up too much space.

alphapapa:
Ah, good call on the keybindings!  And yeah, I understand that it could use a lot of space with ranger's other panes and such.  Thanks for your work on ranger.

Issue #483, opened 2016-02-29
copy mode = remove <yr> isn't working after file has been pasted


jarhat:
When a currently selected file is yanked with yy, the copy mode can be removed again with yr, that's alright..
But after having it pasted with pp, the copy mode cannot be removed with for some reason.
The file stays grey (in default colorscheme) and can still be pasted. To sort of clear the 'yank buffer', I could only yank another file and remove the copy mode here.

I'm having this with ranger 1.7.2.

hut:
Fixed by #526

PR #526, opened 2016-04-08
Add __eq__, __neq__, and __hash__ to class File

ajtluser:
Adding or removing a file or directory in the current directory made ranger update itself, by creating new File objects. 

Lack of custom __eq__, __neq__, and __hash__ methods meant those new File objects would not be equal or even have the same hash as preexisting File objects with the same path. 

In the file actions.py, class Actions, method copy, the functions difference_update and symmetric_difference_update would give bad results because some of their parameters would contain preexisting File objects and some would contain new File objects.

These new methods test for equality, and perform hashing on the File object's path field, so preexisting and new File objects are equal/have the same hash as long as they have the same path.

Fix for issue #483.

### DIFF
--- a/ranger/container/file.py
+++ b/ranger/container/file.py
@@ -96,3 +96,12 @@ class File(FileSystemObject):
             return self.fm.previews[self.realpath]['imagepreview']
         except KeyError:
             return False
+
+    def __eq__(self, other):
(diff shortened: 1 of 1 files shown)


hut:
Looks like something that may have unintended side effects, but I can't think of anything. So let's try it out. Thank you :)

PR #527, opened 2016-04-11
Program indices now line up nicely.

NoSuck:
... Am I the only one crazy enough to have rifle entries in the double digits?


hut:
Ty!

hut:
FYI I had to patch your patch: 523689c114f0688e8efb3831e13e5214a2a858f3

PR #528, opened 2016-04-11
use % formatter

EdwardBetts:


hut:
When using python2.7, your patch will always show 0% (unless of course it shows no number at all, like `All` or `Top`)

EdwardBetts:
Thanks, I fixed the bug.

hut:
Thanks, merged

PR #544, opened 2016-05-11
`draw_progress_bar_in_statusbar' does not exist

mrogalski:
A correction of a simple typo in one of the options described in the man page.


hut:
Thank you, next time please fix the man page in the doc/ranger.pod source file, from which the man page is compiled with `make man`

Issue #489, opened 2016-03-07
line numbers for the main column


v1kn:
A `set number` toggle for the main pane would improve precise movement and selection across the file system.  
Thank You

Delapouite:
The `set line_numbers` option has been added in the PR above and will be released in the next version of ranger. I guess this issue can be closed now.

v1kn:
great news.
when the new version is planned to land?

nfnty:
In a couple of days.

v1kn:
Looking forward to it.
Thank You.

PR #546, opened 2016-05-16
add line numbers in main column

mrogalski:
As requested in #489, add line numbers in main column.  Supports normal and relative variants.

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -133,7 +133,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.7.2" "04/15/2016" "ranger manual"
+.TH RANGER 1 "ranger-1.7.2" "16/05/16" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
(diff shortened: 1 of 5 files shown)


hut:
Thanks, this has been long awaited by many people!

I changed your patch in e01bf0062ad29f356878985a813b47ec86966803 so it only uses 1 option, `line_numbers={false,absolute,relative}` rather than two.

mrogalski:
No worries, I really missed this from vim.  
A'propos changing this file, could You please look at these two lines?
https://github.com/hut/ranger/blob/30cf3e0d640e7ea35c1268cfa871004dacd9e856/ranger/gui/widgets/browsercolumn.py#L261-L262

Are they really necessary?  I mean, the loop is `for line in range(self.hei):`,  so `line` should always be smaller than `self.hei`... unless I am missing something.
You have introduced this in 5417dda5fcb4950cb8979e311ae29d285615a633.

hut:
Yep, those lines seem to be redundant. Don't know why I added them. Want to send a PR to get the credit, or should I patch this?

mrogalski:
Made the PR just to get a pointless little green badge to feel like achieving something: hut/ranger#556 :)

PR #550, opened 2016-05-18
2 small changes to loading plugins

dflupu:
1. Use importlib instead of __import__ because the later messes up the global namespace within the plugin.
2. Try to load commands from the plugin, because it is a huge pain to inject new commands from within the plugin itself. I am aware of the commands file, but sharing a command by telling someone to paste some code at the end of a file isn't particularly fun.

### DIFF
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -6,6 +6,7 @@
 import os.path
 import sys
 import tempfile
+import importlib
 
 def main():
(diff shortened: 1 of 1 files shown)


hut:
Good ideas. I added documentation on 2. in commit fe56c9c8494f961505c6f63d9814f961ec070ae3 FYI

Issue #558, opened 2016-05-29
Starting ranger in just removed directory throws an exception


hahiserw:
Nothing serious I think. But ranger throws an exception instead of handling it somehow.

To reporduce:

Console 1:
mkdir test

Console 2:
cd test

Console 1:
rmdir test

Console 2:
ranger

mrogalski:
Seems straightforward to fix.

Thanks to You I had found this semi-related problem:
```sh
mkdir -p /tmp/a/b/c
cd /tmp/a/b/c
ranger
```

Now in a second console, do `rm -rf /tmp/a/`.  This doesn't cause an exception, but instead it silently prohibits the user from moving up one dir by `h`.  Ranger still works and I can still `cd`, but some kind of a message would be nice -- in such a situation mc creates a big red sign "Error: cannot change directory".

PR #560, opened 2016-05-29
Raise error if started in a deleted directory

mrogalski:
Fix for #558

### DIFF
--- a/ranger/core/tab.py
+++ b/ranger/core/tab.py
@@ -16,7 +16,12 @@ class Tab(FileManagerAware, SettingsAware):
         self.history = History(self.settings.max_history_size, unique=False)
         self.last_search = None
         self.pointer = 0
-        self.path = abspath(expanduser(path))
+        try:
+            self.path = abspath(expanduser(path))
+        except FileNotFoundError:
+            sys.stderr.write("Error: can't open `{}'\n".format(path))
+            raise SystemExit(1)
+
         self.pathway = ()
         # NOTE: in the line below, weak=True works only in python3.  In python2,
         # weak references are not equal to the original object when tested with


hut:
Isn't it a bad idea to have `raise SystemExit` in a function that can be called any time during runtime? Even though I couldnt' get it to shutdown ranger after a couple of minutes of trying, it seems like the wrong place for it. Let's maybe put it in the main function instead?

mrogalski:
Yes, I agree it is a bad idea, and a very unhealthy hack..  I apologize, but I can't think of a better way to do this.
The problem I had is that I wanted a nice and clean message with a reason for failure -- and that should include the filename.
You might try running `os.access(".", os.F_OK)` in a directory that doesn't exist -- it returns `True`, so this condition checks out fine, even though the directory certainly doesn't exist: https://github.com/ranger/ranger/blob/904d3df13a96c2ded55919ff2a3ed873b8da3c9e/ranger/core/main.py#L69

The thing that fails is the `abspath(expandpath(path))`, which I enveloped in the try-except block.  `abspath` on a path that doesn't exist throws a `FileNotFoundError`.  It is trivial to catch it, and it can be easily done in main, but in this situation (in a directory that doesn't exist) curiously `FileNotFoundError` is created with an `filename` equal to `None` -- not the path that does not exist.
So, three ways to solve this:

1. this unclean thing in this pull request,

2. handle a `FileNotFoundError` in main but give a nondescriptive error message, something like "some directories do not exist" (consider running it with several arguments, one of which does not exist), 

3. hybridize 1&2: catch `FileNotFoundError` in tab, but then rethrow it with a proper filename, and then catch it in main -- but in my view works pretty much the same as the pull request.  An it looks a bit silly:  an exception is raised, caught only to be rethrown just for the sake of catching it in main to be SystemExit'ed.

I opted for 1 which has a certain simplicity to it, although I am definitely not an expert on python programming.  You have much more experience.  If You want 2 or 3 (or maybe You have something better in mind?), I'd be happy to provide an implementation.

hut:
It may be a little clumsy, but we could just add another `abspath(expandpath(path))` for each target in targets before `fm = FM(paths=targets)` in main.py, and if it blows up, we know the file name and can display a proper error message, before FM was even initialized.

mrogalski:
Yeah, I thought about that as well.  The only problem I have with it is what would happen if the directory was deleted after we do something like:  
```
for t in targets:
   try:
       abspath(expandpath(t))
   except FileNotFoundError as e:
       print("no such directory: `{}'".format(t)")
```
but before the FM initialization.

nfnty:

PR #562, opened 2016-05-31
Fix wrong continued indentation and whitespaces.

stepshal:
From test results by pylint.


hut:
Some of the changes are really good ideas, but the continued line indentations as proposed by pylint don't make sense IMO.

It leads to code like
```python
self.my_long_function_name(self.some_other_function(foo,
                                                    bar
                                                    zar)
```
which leaves little room for function parameters, so we don't do it in ranger.

I'll fix the 7-space-indentation in that plugin file at least, though.

stepshal:
> Some of the changes are really good ideas
Like what? i make separate PR.

PR #572, opened 2016-06-14
ranger.py: place imports at the top of the module

stepshal:
From pylint.


hut:
Thanks, but actually, the location of the imports is there on purpose in this file. The embedded bash script (see comments) needs to be at the top to work.

PR #583, opened 2016-06-19
 ranger/config/commands.py: Make exactly one space before keyword

stepshal:


hut:
lining up things helps readability, so it's fine

PR #586, opened 2016-06-19
Block comment should start with '# '

stepshal:


hut:
This is (mostly) outcommented code. IMO outcommented code should not have anything added to it except the `#`. The two cases which don't concern outcommented code are fine, I'll fix them myself quickly

stepshal:
@hut Maybe just delete them? With modern revision control available, there is no reason to save commented-out code to your repository.

PR #590, opened 2016-06-19
 ranger/core/actions.py: do not assign a lambda expression, use a def

stepshal:


hut:
IMO lambdas are more natural in this case, also `def` one-liners are ugly. I'll close it, but if anybody has strong opinions about this, throw it in.

PR #591, opened 2016-06-19
Change ambiguous variable name 'l' to more meaningful 'line'

stepshal:
In some fonts, character 'l' are indistinguishable from the numeral one. So better use 'L' instead.

### DIFF
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -211,10 +211,10 @@ class CommandLoader(Loadable, SignalDispatcher, FileManagerAware):
                 except select.error:
                     sleep(0.03)
             if not self.silent:
-                for l in process.stderr.readlines():
+                for line in process.stderr.readlines():
                     if py3:
-                        l = safeDecode(l)
-                    self.fm.notify(l, bad=True)
+                        line = safeDecode(line)
+                    self.fm.notify(line, bad=True)
             if self.read:
                 read = process.stdout.read()
                 if py3:


lamarpavel:
Why would you use a font for coding where this is a problem? It seems rather dull to change the code because there is a chance that on some setup (that is arguably unfit for work) things might be hard to read.

Vifon:
Additionally this fix breaks the next line which should use the same variable regardless of the `if` branch taken.

stepshal:
@Vifon I changed all of them now.

stepshal:
@Vifon Sorry, fixed.

Vifon:
Now it's broken for the opposite case. Look, the lowercase `L` (`l`) is created in line 214, if we're running on Python 3 then `safeDecode` is applied to it and regardless of this step it is then passed to `self.fm.notify`. After your patch `safeDecode`'s result is stored in a separate variable so it won't get passed properly in both Python 2 and Python 3. Your first version worked for Python 2 while this one works for Python 3 but neither works for both.

Additionally, like @lamarpavel said, is that really such a great idea in the first place?

stepshal:
And how to make it work for both? That do it?
``` Python
                 for L in process.stderr.readlines():
                      if py3:
                         L = safeDecode(L)
                    self.fm.notify(L, bad=True)
```

Vifon:
Technically, yes. But if you're already renaming it, why not call it something more meaningful than `L`?

By the way, @hut, would it be okay to remove the `.readlines()` part? `process.stderr` should be an iterable anyway.

stepshal:
@Vifon 
>  But if you're already renaming it, why not call it something more meaningful than L?
Like what? I change it to `i`, which is similar to `l`.

Vifon:
For example `line` would do just fine, because that's what this variable holds.

stepshal:
@Vifon Okay, thanks.

hut:
Thanks, people ^^
And @Vifon: yes, I'll remove the redundant .readlines()

PR #592, opened 2016-06-19
Remove unexpected spaces around keyword / parameter equals

stepshal:


hut:
Thanks for cleaning up ranger :)

PR #593, opened 2016-06-19
Make exactly one space before operator

stepshal:


hut:
I think lining up assignments is fine, so let's leave it like this. Pylint's category for this is `bad-whitespace`, but I'm guessing we can't just ignore `bad-whitespace` because it includes different whitespace style issues as well... Hm.

PR #595, opened 2016-06-21
Unnecessary list literal - rewrite as a set literal.

stepshal:
It's unnecessary to use a list literal within a call to `set` since there is literal syntax for these types.


hut:
Set literals are implemented since python2.7, but ranger should run on python2.6

PR #596, opened 2016-06-21
Use single-quotes instead of doable-quotes

stepshal:
Tell me if it's okay with you, and I'll add the rest.


hut:
No, this is just bikeshedding.

PR #600, opened 2016-06-22
Add missing trailing commas

stepshal:


hut:
Not necessary

Vifon:
The first two would make sense in my opinion. Especially the one in `settings.py` as some dicts there already have trailing commas. The others are not likely to be extended or are not line-oriented (the second one in `git.py`), so I would leave them as they are.

PR #602, opened 2016-06-23
Do not enable line_numbers when it is not set to any value

Vifon:
Currently if the option `line_numbers` is not set to any value whatsoever in the config, it gets enabled by default. It's due to checking if it's `!= 'false'` which is true if it's an empty string. It seems `settings.ALLOWED_VALUES` is not enforced anywhere so I've added a helper function that checks for it and added it to these value comparisons. I do not consider it a final solution (manually calling `valid_value()` is ugly and still prone to bugs) but it's a good start.

### DIFF
--- a/ranger/container/settings.py
+++ b/ranger/container/settings.py
@@ -246,6 +246,12 @@ class Settings(SignalDispatcher, FileManagerAware):
     def _raw_set_with_signal(self, signal):
         self._raw_set(signal.setting, signal.value, signal.path, signal.tags)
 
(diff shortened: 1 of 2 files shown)


hut:
How about #608?

Vifon:
Looks good to me. In the future it would be good to actually enforce `ALLOWED_VALUES` and emit an error on value change, so that setting `line_numbers` to `HELL_NO!` would not enable it (in a hard to predict way).

nfnty:
Fixed by #608.

PR #603, opened 2016-06-23
Remove unused imports

stepshal:
I don't think some of them unused, so see for yourself.

### DIFF
--- a/ranger/api/__init__.py
+++ b/ranger/api/__init__.py
@@ -32,8 +32,6 @@ def hook_ready(fm):
     NOT print anything to stdout anymore from here on.  Use fm.notify instead.
     """
 
-from ranger.core.linemode import LinemodeBase
-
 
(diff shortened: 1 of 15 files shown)


ranger/api/__init__.py
```diff
@@ -32,8 +32,6 @@ def hook_ready(fm):
     NOT print anything to stdout anymore from here on.  Use fm.notify instead.
     """
 
-from ranger.core.linemode import LinemodeBase
```
Vifon: This import is used by the modules importing this module.

ranger/api/commands.py
```diff
@@ -7,8 +7,6 @@
 import ranger
 import re
 import inspect
-from collections import deque
-from ranger.api import *
```
Vifon: The same with this one, and probably many others.

PR #608, opened 2016-06-26
Sane setting fallback values

hut:
Every setting will be initialized with the first allowed value. Alternative to #602.

### DIFF
--- a/ranger/container/settings.py
+++ b/ranger/container/settings.py
@@ -83,13 +83,13 @@ ALLOWED_SETTINGS = {
 }
 
 ALLOWED_VALUES = {
-    'confirm_on_delete': ['always', 'multiple', 'never'],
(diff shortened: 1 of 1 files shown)


hut:
Although 'vcs_backend_git' is enabled in the rc.conf by default. Perhaps the ALLOWED_VALUES should reflect that.

ranger/container/settings.py
```diff
@@ -100,6 +100,10 @@ def __init__(self):
                     self._sanitize, priority=1.0)
             self.signal_bind('setopt.' + name,
                     self._raw_set_with_signal, priority=0.2)
+        for name, values in ALLOWED_VALUES.items():
+            assert len(values) > 0
+            assert name in ALLOWED_SETTINGS
```
Vifon: Wait, does it even make sense to check for that?

PR #610, opened 2016-06-26
Add github issue and pull request templates

stepshal:
CONTRIBUTING.md - doesn't do anything, just taking place in source code, so this making  them part of the repository configuration instead. Should work this time.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,15 +1,5 @@
-# Contributing to Ranger
-
-## Tips for bug reports
-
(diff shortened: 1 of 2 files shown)


hut:
Having CONTRIBUTING.md is better IMO because 1. it shows up even for people who don't use GitHub, and 2. it adds a notification to the "New Issue" action, saying "Please review the guidelines for contributing to this repository." which is (debatably) better than dumping the text into the text area which the user has to remove first.

We could have both, but that would result in duplication and redundancy, so I'm going leave it how it is for now.

PR #614, opened 2016-06-29
Refactor and improve the TERMCMD handling

### LINKED ISSUE #613: Open terminal with additional options
I'd like to open new terminal from ranger (for example with command like ```:terminal```) but not in a separate window as it happens now but in a new tab (let's assume I use terminal that supports tabs like gnome-terminal, roxterm etc).

For example in case of roxterm it's possible to execute it like ```roxterm --tab``` and new tab will appear instead of new window.

I tried to set TERMCMD to ```roxterm --tab```  but then ```xterm``` gets executed at all (I think that's because ```roxterm --tab``` could not be found in PATH by ```get_executables()```).

Any thoughts on how that might be achieved?

Vifon:
Now only the first token of TERMCMD needs to exist in PATH. Previously
the whole value was checked, which could contain additional parameters.

The check and term guessing were performed in two places. Now it's done
by the get_term() function to reduce the code duplication. In that
function the second check was indented one level deeper (relative to the
original code). Keeping both checks on the same level, while visually
pleasing, was conveying the wrong message. One never wants to perform
the second check if the first one fails. Nesting shouldn't be a problem:
if the nesting will need to be any deeper than this, this code will need
to be refactored anyway.

Fixes #613.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -493,14 +493,8 @@ class terminal(Command):
     Spawns an "x-terminal-emulator" starting in the current directory.
     """
     def execute(self):
-        import os
-        from ranger.ext.get_executables import get_executables
-        command = os.environ.get('TERMCMD', os.environ.get('TERM'))
-        if command not in get_executables():
-            command = 'x-terminal-emulator'
-        if command not in get_executables():
-            command = 'xterm'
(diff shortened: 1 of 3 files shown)


hut:
Makes a lot of sense, thanks!

PR #644, opened 2016-08-04
Fix map of open_with command

ssh0:
When I hit the `r` key, there was no white space after the prompt `open_with`...

It can be solved by simply adding one more `%` before `%space` to escape.

related to: #293


hut:
Thank you.

The bug is due to ab6643976a91e2517af224e6c2ab99f631660cd7, btw.

PR #645, opened 2016-08-04
Fix full folder size being shown incorrectly

Enzime:
The full folder size was previously being shown when all files currently displayed were being shown, however in combination with filtering and hidden files this could make the size totally wrong.


hut:
Thank you

> when all files currently displayed were being shown

s/shown/selected/ ;)

PR #647, opened 2016-08-05
get_cumulative_size now also calculates the cumulated items count of …

repnzscasb:
New feature: when calculating the cumulative size of selected directories, calculate the cumulative items count (files + directories) too.

Items count are displayed like `ncdu`, so that we have a fixed-width column and everything is kept aligned.

Added sorting option `oi` and `oI`.
Slightly adapted the statusbar layout.

I could rename all the methods to be `get_cumulative_size_count` but in this case I should also change config options like `autoupdate_cumulative_size`. This would break backward compatibility and I'm not sure whether it makes sense to do it now.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -428,6 +428,7 @@ map <a-9>     tab_open 9
 map or set sort_reverse!
 map oz set sort=random
 map os chain set sort=size;      set sort_reverse=False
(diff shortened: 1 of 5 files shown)


nfnty:

randynobx:
@repnzscasb Are you able to rebase and update the PR so it can be reviewed and merged? I'd love to see this implemented.

Issue #574, opened 2016-06-15
Implement binding of Ctrl+Arrow Keys


hut:

nfnty:
We should be looking into using something like [libtickit](http://www.leonerd.org.uk/code/libtickit/).

dquinton:
I need <a-bracketleft> <a-bracketright> but it won't work.
Also <c-1> etc
Alt- F keys ?

Q3CPMA:
Hello, as a hack, wouldn't the output of print_keys.py work (i.e. `<27; 91; 49; 59; 53; 67>`)?

Issue #266, opened 2015-03-07
Unable to map modifier keys


zaxebo1:
as suggested somewhere, I installed trash-cli and then in rc.conf of ranger, put the following line
map <DELETE> shell -s trash-put %s

so, now pressing "**delete**" button inside ranger, sends it to trash (just like what happens in nautilus)
# 

QUESTION 1:
Now, my question is that, (just like in nautilus) I want to define that pressing "**shift+delete**" should not send it to trash, but rather directly delete it (preferably showing warning before going to delete) . HOW to do it?
I tried all of the below lines in rc.conf:
 map <s-DELETE> shell -s rm -f %s 
map <s-DELETE>  console delete
but none of them worked. And additionally i want it to ask confirmation(y/n) before actually deleting it
# 

QUESTION2:
In ranger, How to select  multiple files(consecutively listed) 
( for example: In nautilus, we press Shift+Uparrow (or Shift+Downarrow) to select multiple consecutive files  )
actually if i have ten consecutive files - which need to be selected, then pressing "space" for each of them is very cumbersome. In nautilus, we have shift+Uparrow and shift+downarrow to select multiple consecutively listed files. How can i emulate it in ranger

hut:
Q1: `map <S-DELETE> delete` works for me. If you have an older ranger version where the key `<S-DELETE>` was not implemented yet, try `:map <383> delete`.
To ask for confirmation, add `set confirm_on_delete always` to your ~/.config/ranger/rc.conf.

Q2: Excerpt from the man page:

       Space         Mark a file.
       v             Toggle the mark-status of all files
       V             Starts the visual mode, which selects all
                     files between the starting point and the
                     cursor until you press ESC.  To unselect
                     files in the same way, use "uV".

zaxebo1:
Answer1) thanks for reply . I am using ranger version 1.6.1 on ubuntu 12.04 , in my PC this Answer1 is still not working. But now as you have given me the hint, i will look into it . Thanks :-)   

Answer2) I have reworded my question . Probably i was not able to convey my question clearly earlier. Now read my question 2 as :
**'(like in nautilus) can i somehow map my shift+uparrow and shift+down arrow keybindings also to the functionality of what "V and ESC" are currently providing in ranger. That is, can i additionally map keys such that till Shift+Downarrow is pressed that files are keep being marked for selection, until i lift the shift+Downarrow' . actually, I want to emulate nautilius keybindings as much close as possible.

eMPee584:
[...]

PR #648, opened 2016-08-05
[WIP] Add support for modifier combos + arrow keys

Enzime:
Currently works in termite, but not working in urxvt.
#574


Enzime:
Now supports urxvt :)

nfnty:

nfnty:
Potentially also solves #266.

nfnty:
This is very hackish and will break easily. We should be looking into using something like [libtickit](http://www.leonerd.org.uk/code/libtickit/) instead.

nfnty:
This will need a lot of work to implement in a general way. Continue discussion in #574.

PR #649, opened 2016-08-06
add hash as an escaped character in shell commands

nightsense:
This is required for shell commands to succeed with filenames that
start with the hash character. The 'mv' command invoked by 'bulkrename',
for instance, will fail with such files unless hashes are escaped.


hut:
Thanks!

PR #652, opened 2016-08-20
Fix preview column displayed unintentionally

Enzime:
The incorrect part of the conditional checks whether a folder is currently focused but does not check if the setting `preview_directories` is `True`. This condition is already covered in the `self.columns[-1].has_preview()` function thus can be removed from the conditional.


hut:
Thank you

PR #653, opened 2016-08-21
Change how column widths are calculated

Enzime:
By changing it, this means the columns will use the entire width if the geometry allows for it.

Instead of using a hard offset of one for the width of the column, this just changes the next position of the column to be one further than the column width.


nfnty:
Thanks!

PR #654, opened 2016-08-21
Fix cursor being lost when changing viewmodes

### LINKED ISSUE #639: Cursor disappears on changing viewmode (with key ~ or ":set viewmode!")
The title sums it up.

Enzime:
This results from finalize being called in order for the widgets. Previously when ranger changed viewmodes, it would delete then readd the widget object which meant that the finalize for the viewmode was called last instead of before console, which meant the finalize of the viewmode was setting
and overriding the cursor position.

Fixes #639.

### DIFF
--- a/ranger/gui/displayable.py
+++ b/ranger/gui/displayable.py
@@ -211,6 +211,7 @@ class DisplayableContainer(Displayable):
     New methods:
 
     add_child(object) -- add the object to the container.
+    replace_child(old_obj, new_obj) -- replaces old object with new object.
     remove_child(object) -- remove the object from the container.
 
     New attributes:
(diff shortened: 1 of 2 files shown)


hut:
Awesome^^ Thanks. I pushed a follow-up (8132885c2fbddc9e686f2285b307e28a1f29f2a7) to avoid crashes in DisplayableContainer.replace_child if old_obj isn't in the list.

Enzime:
Personally, I made it unsafe because adding it to the end is not an alternative to replacing it in the list, as the order of the `self.container` can have a great effect sometimes, so making it unsafe made it more likely to find other bugs.

PR #658, opened 2016-08-26
Added preview option for OpenDocument Text files using odt2txt

entropy-dev:
Added a line to scope.sh for OpenDocument Text files (odt, ods, odp, sxw). I've used it for quiet some time and the preview seems to work. However I'm not sure about the exit code(s).

The preview is generated by [odt2txt](https://github.com/dstosberg/odt2txt). It is contained in the Arch Linux community repository as well as in the repositories of Ubuntu, Debian, Fedora, OpenSuse and probably more distributions.

Tell me what you think.


hut:
Great, thank you very much

PR #661, opened 2016-09-06
Add correct path to ranger on OS X

arshiamufti:
Spent a while trying to figure out why the example script in `bash_automatic_cd.sh` didn't work on my Mac (OS X 10.11.6). Running `which ranger` on my machine produces `/usr/local/bin/ranger`, not `/usr/bin/ranger` (which I'm guessing applies to linux). Versions I'm using:
- ranger-stable 1.7.2
- Python 2.7.10


hut:
Thanks!

Issue #667, opened 2016-09-15
--choose* flags only work when selecting files within $PWD


cstrahan:
I'm doing something like the following:

```
echo >! paths.txt; ranger --choosefiles paths.txt; cat paths.txt
```

As long as I stay within the current working directory, can select files with `<space>` and they'll be printed out when I hit `<enter>`.

If I type `h` to go to the parent directory and select files, they _won't_ be written to `paths.txt`.

This is on ranger 1.7.2 with python 2.7.11.

mrogalski:
Check the pull request #668 if it suits Your needs and for possible bugs. 
Is the order of paths an issue?  Because right now You will get them in an unsorted way.

cstrahan:
For my needs, the order isn't important. I would suggest documenting that the order is currently undefined, allowing one to revisit the code and preserve the original selection order without fear of breaking people's expectations.

mrogalski:
You mean documenting in code and/or manpage?

cstrahan:
Correct -- just so no one somehow depends on the undefined order (granted, an unlikely scenario).

cstrahan:
@nfnty Thanks for merging! Would it be possible to drop a new point release relatively soonish? :wink:

nfnty:
A new release may take some time due to major changes to the codebase. Needs to be tested thoroughly.

PR #668, opened 2016-09-16
Proposed implementation of feature request #667

mrogalski:
Possible issue:  using set() to get unique paths, which changes order of chosen files.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -405,8 +405,15 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 open(ranger.args.choosefile, 'w').write(self.fm.thisfile.path)
 
             if ranger.args.choosefiles:
-                open(ranger.args.choosefiles, 'w').write("".join(
-                    fobj.path + "\n" for fobj in self.fm.thistab.get_selection()))
+                paths = []
+                for hist in self.fm.thistab.history:
+                    for fobj in hist.files:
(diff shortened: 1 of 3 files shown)


cstrahan:
Awesome! Would it be feasible to preserve the order?

mrogalski:
Yes and no.
First of all:  when I currently choose files/dirs in 1.7.2, I don't get them in the order of choice, but in the order of listing -- it even changes depending on how the directory is sorted.  And I can't figure out why -- so before I can figure that out, I have to say no.

But let's see the possibilities for yes: 
The simplest way to dispense with the nondeterminism is just wrap the `set()` in a `sorted()` function -- You get a predictable, lexicographic order, even files grouped by directories. 
To get the selected things in other dirs, the code I wrote traverses the history.  The question is: when the history is changed, what should be the order?  That is, I select files `a`, `b`, `c` in some dir, then I go down one dir, select files `d` and `e`, then I go up, select files `x`, `y`, `z`.  What should the order be? `a, b, c, d, e, x, y, z`?  `z, y, x, c, b, a, e, d`? Technically, the simplest solution to that is to store both file objects and their selection timestamp, and then sort them by timestamps, which hopefully should get You the `a, b, c, d, e, x, y, z` sort.  The second kind of sort should not be that difficult as well, though of course it requires more complicated code -- this patch is just the simplest way of achieving a solution.

nfnty:

PR #676, opened 2016-09-28
PAGER: split on " "

Profpatsch:
$PAGER could be something like `"less -R"`, and `-R` should be given as
argument, not as executable.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -4,6 +4,7 @@
 import codecs
 import os
 import re
+import shlex
 import shutil
 import string
 import tempfile
@@ -1186,8 +1187,7 @@ class Actions(FileManagerAware, SettingsAware):
             write("\n")
 
         temporary_file.flush()
-        pager = os.environ.get('PAGER', ranger.DEFAULT_PAGER)
-        self.run([pager, temporary_file.name])
+        self._call_pager(temporary_file)
 
     def dump_commands(self):
         temporary_file = tempfile.NamedTemporaryFile()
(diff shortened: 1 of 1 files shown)


nfnty:
Use `shlex.split()` instead.

ranger/core/actions.py
```diff
@@ -1225,8 +1223,13 @@ def write(string):
             write("%30s = %s\n" % (setting, getattr(self.settings, setting)))
 
         temporary_file.flush()
+        self._call_pager(temporary_file)
+
+    def _call_pager(self, file_):
         pager = os.environ.get('PAGER', ranger.DEFAULT_PAGER)
-        self.run([pager, temporary_file.name])
+        # PAGER might contain arguments, like "less -R"
+        pager = pager.split(" ")
```
nfnty: Use `shlex.split()`

Profpatsch: Done.

ranger/core/actions.py
```diff
@@ -1225,8 +1224,13 @@ def write(string):
...
 
         temporary_file.flush()
+        self._call_pager(temporary_file)
+
+    def _call_pager(self, file_):
         pager = os.environ.get('PAGER', ranger.DEFAULT_PAGER)
-        self.run([pager, temporary_file.name])
+        # PAGER might contain arguments, like "less -R"
+        pager = shlex.split(" ")
+        self.run(pager + [file_.name])
```
nfnty: Should be `shlex.split(pager)`. Refactor into:
```
def run_pager(self, *paths):
    self.run(shlex.split(os.environ.get('PAGER', ranger.DEFAULT_PAGER)) + list(paths))
```

Profpatsch: Right. Not sure what you mean by the refactor, though. I kind of like the split in three lines, this way the `split` can be documented.

Vifon: +1, looks good to me in three lines. I'd be somewhat partial towards using a new variable name after the split but do what you think is right.

nfnty: The refactor is for fixing the argument name, adding the ability to specify multiple paths and making it public for use.

This function is so simple that one line is more than enough, but either way works.

nfnty:
Merged with changes in 79deadc7a9662807c075e3c61f3eb9362e29683c. Thanks.

PR #681, opened 2016-10-10
Update the image preview method upon setting it in runtime

Vifon:
Previously the displayer was not changed after the initialization, so only the value read from rc.conf was used.


hut:
Thanks :)

PR #693, opened 2016-10-23
fixed 7z crash

valoq:
When 7zip archives created with -mhe=on (also known as "encrypt file list") are selected by ranger, the application crashes, because als (or rather 7z) promps for a password to list the files in the preview.

To prevent this, the 7z command can use the -p option with an empty string, preventing the preview for 7z archives with encrypted file lists.
The same applies when such archives are opened, causing an error about a wrong password instead. Archives without the -mhe=on option are not affected by this.

Alternatively "als -O-p" can be used to pass the option to 7z instead of using it directly.

### DIFF
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -182,7 +182,8 @@ ext xcf,                    X, flag f = gimp -- "$@"
 # Archives
 #-------------------------------------------
 # This requires atool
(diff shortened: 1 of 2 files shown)


hut:
Thanks! A short comment in the code would've been great since it's not so obvious. I'll add it for you.

PR #711, opened 2016-11-07
Do not suppress custom commands import exceptions

ael-code:
Original problem
----------------
All the import exceptions raised while trying to load custom commands
file were suppressed. This make difficult to understand why the custom commands
file was not loaded.

Fix
---
In the case of import exceptions , an explanatory message is printed
on stderr and the original exception is raised.

Bonus
-----
The path of custom commands file that caused the exception is printed
in the explanatory message.

### DIFF
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -265,14 +265,17 @@ def load_settings(fm, clean):
         allow_access_to_confdir(ranger.arg.confdir, True)
 
         # Load custom commands
-        if os.path.exists(fm.confpath('commands.py')):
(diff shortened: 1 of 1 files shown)


hut:
sure that it's a good idea to abort loading of ranger if command loading fails? the user could have a broken commands.py and still a working ranger as default commands.py is still loaded. How about using your new logging mechanism for this instead?

ael-code:
Actually this was fixed in 77f6c349f56ecf27c91cb394156c222e8525032e, it use the new logging approach to log the error without exiting

PR #715, opened 2016-11-09
Remove sxiv_workaround_hook function

AcidBear:
I'm getting an error while I'm trying to open image with _sxiv_ in a folder with thousands image files. The error is: 
> Argument list is too long

After getting this one I'm not able to quit from ranger by pressing _q_ key.
Also, if I close the terminal with ranger after this error, ranger's process uses a 100% of one core of the CPU in the background.

Without this function (exactly as in this commit) everything works well.

Log with ```--debug``` option:
```
  File "ranger/ranger/core/main.py", line 145, in main
    fm.loop()
  File "ranger/ranger/core/fm.py", line 368, in loop
    ui.handle_input()
  File "ranger/ranger/gui/ui.py", line 230, in handle_input
    self.handle_key(key)
  File "ranger/ranger/gui/ui.py", line 166, in handle_key
    self.press(key)
  File "ranger/ranger/gui/ui.py", line 181, in press
    quantifier=keybuffer.quantifier)
  File "ranger/ranger/core/actions.py", line 221, in execute_console
    cmd_class(string, quantifier=quantifier).execute()
  File "ranger/ranger/api/commands.py", line 416, in execute
    return self._based_function(*args, **keywords)
  File "ranger/ranger/core/actions.py", line 444, in move
    result = self.execute_file(selection, mode=mode)
  File "ranger/ranger/core/actions.py", line 401, in execute_file
  File "ranger/ranger/ext/rifle.py", line 371, in execute
    Popen_forked(cmd, env=self.hook_environment(os.environ))
  File "ranger/ranger/ext/popen_forked.py", line 21, in Popen_forked
    subprocess.Popen(*args, **kwargs)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 7] Argument list too long
```
Version of Python: 2.7.

Please remove this function or do something with it.
Thanks.

### DIFF
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -138,6 +138,7 @@ class FM(Actions, SignalDispatcher):
         # 1. set open_all_images to true
         # 2. ensure no files are marked
         # 3. call rifle with a command that starts with "sxiv " or "feh "
+        """
         def sxiv_workaround_hook(command):
             import re
             from ranger.ext.shell_escape import shell_quote
@@ -181,6 +182,7 @@ class FM(Actions, SignalDispatcher):
             return old_preprocessing_hook(command)
 
         self.rifle.hook_command_preprocessing = sxiv_workaround_hook
+        """
 
         def mylogfunc(text):
             self.notify(text, bad=True)


AcidBear:
P.S. I'm a little bit in frustration about my need to be involved in ranger's source code for it's normal work. It's not OK, I think.

ael-code:
> P.S. I'm a little bit in frustration about my need to be involved in ranger's source code for it's normal work. It is not OK, I think.

The idea of the FOSS software is to encourage collaboration in order to build better software. Don't be frustrated, be collaborative :)

AcidBear:
@ael-code Especially if it about removing existing functionality :)

Vifon:
Did you try to add `set open_all_images false` to your rc.conf? Are you using the newest ranger version from Git? Because I cannot reproduce your exact error.

AcidBear:
@Vifon Oh, I didn't know about this option. And if I disable it, everything works well... So, problem has been solved. But I think this feature is too buggy to be enabled by default. 
But thanks for the help anyway.
> I cannot reproduce your exact error.

With enabled `open_all_images` option or not? In git version the error appears in a directory with 25000 image files, if `open_all_images` is set to `true`.

Vifon:
> But I think this feature is too buggy to be enabled by default.

In 99% of cases there shouldn't be any problem and personally I find this feature really handy. As for the 1% I think we could catch this exception to fall back to the behavior used when `open_all_images` is set to `false`. That's why I'm trying to reproduce it myself.

> With enabled `open_all_images` option or not? In git version the error appears in a directory with 25000 image files, if `open_all_images` is set to `true`

Enabled of course. I asked because I got a different error and wanted to know whether it's some other related but independent bug or rather it is how it manifests in the new version. It seems back then I created too many files and ranger was crashing on memory allocation itself. Now I managed to reproduce it.

Vifon:
Hm, it seems to be a bit tougher than I expected. I hoped that I can run one command and if it fails, easily fall back to the command without the extra arguments. But it seems the extra arguments are added quite a while before the execution itself. I'll try to get around to it later as I'm pretty busy right now, unless @hut will do it before me.

For the record:

The sxiv extra arguments are added [here][1] and after some more processing, the command is executed [there][2]. The further processing is the biggest issue in my opinion.

[1]: https://github.com/ranger/ranger/blob/055247a002757ab439e68f8497b3935f28a839c4/ranger/ext/rifle.py#L316
[2]: https://github.com/ranger/ranger/blob/055247a002757ab439e68f8497b3935f28a839c4/ranger/ext/rifle.py#L371

Xender:
A possible fix unrelated to the bug itself: use sxiv's `-i` option:
`-i     Read names of files to open from standard input.`

This should have no problems with too much filenames.

Vifon:
@Xender Yes, it seems `-i` could be a viable workaround. It seems feh supports a similar `-f -` flag. pqiv has `-r` and imv seems to work like this [by default](https://github.com/eXeC64/imv#opening-images-via-stdin). Sounds pretty promising to me!

nfnty:
I see no reason to add a fix for this. If your OS has a too low limit for `ARG_MAX` (modern OS should have close to or more than `2097152`) then either increase the limit or disable `open_all_images`.

Open a new issue if you feel that there should be a better fix for it.

PR #720, opened 2016-11-15
Fixed trailing space bug in config parsing.

vlad7777:
Fixed the trailing space in config bug.
Basically, trailing spaces were not stripped, and if one unknowingly leaves a trailing space in config file, the line is ignored. Fixed by adding space to rstrip in actons.py.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -346,7 +346,7 @@ class Actions(FileManagerAware, SettingsAware):
         filename = os.path.expanduser(filename)
         with open(filename, 'r') as f:
             for line in f:
-                line = line.lstrip().rstrip(" \r\n")
+                line = line.strip(" \r\n")
                 if line.startswith("#") or not line.strip():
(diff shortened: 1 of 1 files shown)


Vifon:
+1, since we have `%space` to express a literal space now.

For the record, the config line wasn't ignored, it was just incorrect in a given context. For example there is no option called "preview_images " with a trailing space. If it was for example a "map ! console shell " keybind, it would make perfect sense since you want a trailing space in this interactively read line, although `%space` should be used instead.

I'd consider calling `.rstrip()` without arguments or even `.strip()` instead of `.lstrip().rstrip()`. Unless we want to keep the trailing tabs but I don't think so.

vlad7777:
Oh, I didn't know about %space.
Decided to simply go with strip(" \r\n").

hut:
Yeah, I guess it's finally time for this, as %space has been supported since ~1.7.0

Issue #713, opened 2016-11-08
Proposal: migration to standard logging library


ael-code:
I've started reading the code of ranger few days ago because I was not able to debug the custom commands file.

Going through the lines I've found out that logging practice it is not used a lot in ranger and it is handled in very unpractical way, logging function handler are passed from module to module and finally stored in central objects like FM. We need to keep in mind that since ranger has mainly a curses interface manage logging is not as easy as printing to stdin/stderr. I mean everything is running smoothly but we can improve it :) 

I've found some old functions (e.g. `ranger/__init__.py:log`) in the code intended to be used in debugging mode in order to write to a file but it seems that they are not used anywhere at the moment.

For all the above reasons I'm suggesting to use the standard python library to manage logging, this would bring a lot of benefits such as:
 - straightforward implementation of the multi log handler: you can see log from the curses interface, printing to a file while send it to the standard error.
 - thread safe logging without effort
 - possibility to use log levels (DEBUG, WARNING...) and to filter them at runtime according to the `--debug` flag
 - submodules, library, custom commands and plugins don't need to be aware of where the log object it is stored and how to use it, they just need to use the standard library.

I'm already trying to change the code in order to implement it but in the meanwhile I would like to know what do you think about this. Do you think that some major problem will arise trying to implement this?

ael-code:
I saw that some works on improving logging have been done recently by @nfnty on #701. But it was only to print the log in reverse order

ael-code:
I've a PR ready to fix this,  but it depends on #717.

Along with the new code I would like to write somewhere the guidelines for logging and outputting in order to inform new devs and standardize some common procedures.

Where do you think it will be the best place to put this kind of stuff? `ranger/HACKING.md` ?

PR #725, opened 2016-11-25
Migration to standard logging library

ael-code:
# Migration to standard logging library

## Introduction
You can read about the reason behind this PR on this issue:
https://github.com/ranger/ranger/issues/713

## Old logs and output managment

Four different ways of outputting stuff were used:
 - printing out on stdout:
	can be used only when the curses interface is not running,
    is usually used on critical error before exit ranger.
 - printing out on stderr:
	can be used only when the curses interface is not running,
	is usually used for warnings.
 - appending string to a queue in the FM class
	(these are the logs shown by the curses log viewer)
 - background processes and task usually use the `FM.notify` function
   to show a message on the bottom bar of ranger. Moreover this fuction,
   if the debug mode is active, will print the stack trace and exit the program.

### Problems
First of all it is difficult for new dev to understand which method of output
should be used for that particular case.
Moreover the logs (collected on the FM.log object) can only be seen through the
curses log viewer at runtime and it isn't possible to export them, for instance on a file.
For these reasons logs are almost not used at all in ranger. 

## New logging approach
The goal is to provide an easy api to log stuff and a straigthforward way of inspect them. This has been achieved using the standard logging library. The default behaviour is pretty similar to the old one, in the sense that all the the produced logs will be collected in a queue that can be inspected with the curses log viewer (`display_log` command). Moreover the `--logfile` cli option has been added and it can be used to specify a destination file for all the logs in such a way that the same log can be viewed at runtime as well as inspected after a program crash.

The verbosity and the format of the log message is controlled by the already existent `--debug` command line flag:

 - Normal mode:
	A concise log format will be used and only important message will be logged (log level > INFO)
	
	Example:
	```
	[INFO] Ranger version 1.7.2
	[INFO] Running on Python 3.5.2 (default, Jun 28 2016, 08:46:01) [GCC 6.1.1 20160602]
	[INFO] Process ID is 1497

	```

 - Debug mode:
	An extended log format will be used and all the message will be logged.
	
	Example:
	```
	23:17:43,719 [ranger.core.main] |INFO| Ranger version 1.7.2
	23:17:43,719 [ranger.core.main] |INFO| Running on Python 3.5.2 (default, Jun 28 2016, 08:46:01) [GCC 6.1.1 20160602]
	23:17:43,719 [ranger.core.main] |INFO| Process ID is 1515
	23:17:43,720 [ranger.core.main] |DEBUG| config dir: '/home/groucho/.config/ranger'
	23:17:43,720 [ranger.core.main] |DEBUG| cache dir: '/home/groucho/.cache/ranger'
	23:17:43,738 [ranger.core.actions] |DEBUG| Sourcing config file '/path/to/ranger/config/rc.conf'
	```

### Advantages
The main advantages of the new logging approach:

 - clean and standard API that can be used internally as well as in the custom command and plugins
 - support for [multilevel logging](https://docs.python.org/3/library/logging.html#logging-levels)
 - the infrastructure of loggers, handlers and filters allow to easily implement custom behaviour,
	that will be necessary in the feature.

### Logging in commands
All the subclass of the Command class can now use the
the class object `Command.log` for logging pourposes.
This is a standard `logging.Logger` instance, thus
the standard logging api must be used.

Example:

```
class dummy_cmd(Command):
    def execute(self):
        self.log.info("the dummy command has been invoked")
```

### Curses log viewer
You can open it with the `:display_log` command

Now shows the logs in descending order such that the last entry will correspond to the most recent log entry.
On every new open the pager will be positioned on the end of the log showing the most recents log entries.

### DIFF
--- a/ranger/__init__.py
+++ b/ranger/__init__.py
@@ -29,44 +29,9 @@ CACHEDIR = os.path.expanduser("~/.cache/ranger")
 USAGE = '%prog [options] [path]'
 VERSION = 'ranger-master %s\n\nPython %s' % (__version__, sys.version)
 
-try:
-    ExceptionClass = FileNotFoundError
-except NameError:
-    ExceptionClass = IOError
-try:
-    LOGFILE = tempfile.gettempdir() + '/ranger_errorlog'
-except ExceptionClass:
-    LOGFILE = '/dev/null'
-del ExceptionClass
 
 # If the environment variable XDG_CONFIG_HOME is non-empty, CONFDIR is ignored
 # and the configuration directory will be $XDG_CONFIG_HOME/ranger instead.
 CONFDIR = '~/.config/ranger'
-
-# Debugging functions.  These will be activated when run with --debug.
-# Example usage in the code:
-# import ranger; ranger.log("hello world")
-
-
-def log(*objects, **keywords):
-    """Writes objects to a logfile (for the purpose of debugging only.)
-    Has the same arguments as print() in python3.
-    """
-    from ranger import arg
-    if LOGFILE is None or not arg.debug or arg.clean:
-        return
-    start = keywords.get('start', 'ranger:')
-    sep   = keywords.get('sep', ' ')
-    end   = keywords.get('end', '\n')
-    _file = keywords['file'] if 'file' in keywords else open(LOGFILE, 'a')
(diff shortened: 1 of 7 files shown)


hut:
thanks! :kissing_smiling_eyes:

PR #740, opened 2016-12-18
ext.vcs: Improve logging, Lint using pylint and flake8

nfnty:


ael-code:
I didn't get why you renamed the `log` object into `LOG.` Now this file is not any more coherent with the rest of the code.

Moreover I think that logging improvement and lint introduction are completely different changes that should be proposed in different commits and even different PR. It is just that makes working with the commit history very hard.

In any case I'm happy that the logging system is being used... thanks :)

nfnty:
Because `LOG` is a constant. `ext.vcs` has always been linted with both `pylint` and `flake8` and your changes introduced linting errors. This is just a precursor to linting the whole project in the near future and integrating it with Travis CI, so that every new commit and PR is checked.

ael-code:
> @nfnty wrote
> Because LOG is a constant.

Well... it's not easy to define something constant in an object oriented programming language. Wikipedia tries to gives some definition of that [1].
In our specific case, since the object is an instance of `Logger` class, and since there are a lot of functions that make side effect on it (like `Logger.addHandler(hdlr)`,  `Logger.setLevel(lvl)`) I definitely wouldn't consider it a constant.

[1] https://en.wikipedia.org/wiki/Constant_(computer_programming)#Object-oriented_constants

nfnty:
Yes and Python is extremely dynamic in this regard. It's a constant in the sense that the variable is guaranteed to never point to a different object. All variables declared at module level SHOULD be using uppercase, marking them as special and to be treated as such. It's a convention for readability and linters will produce errors if you try to declare a variable at local scope with the same name.

PR #743, opened 2016-12-24
Linting with pylint and flake8

nfnty:
Everything is now fully linted and passing with minimally disabled/ignored warnings.

Pylint only works properly on `2.7`, `3.4` and `3.5`.

Refactoring for the future:
* Remove as many `ansible disable=` as possible.
* Enable `broad-except`: Will require rethinking how exceptions are handled.
* Remove `ranger.core.actions.Actions` from ignored classes: Could be done by making all methods raise `NotImplementedError` and implementing them in subclasses.
* Enable `cyclic-import`: Will require structural changes to the dependency tree.


nfnty:
Would be great to have this PR reviewed as soon as possible to reduce friction for new PRs.

hut:
I am personally against the rule to indent arguments of functions with the height of the parenthesis, like
```python
some_super_long_object_name.verbose_method_name(argument1,
                                                argument2,
                                                argument3)
```

Especially since it would break with long arguments. I would instead vote for an arrangement like this:
```python
some_super_long_object_name.verbose_method_name(
        argument1, argument2, argument3)
```

But maybe I'm just too old-fashioned... Any opinions?

nfnty:
I'm impartial, whatever makes sense and is the most readable for the specific situation.

Add comments to the lines that you see where reformatting is needed and I'll fix it.

Vifon:
+1 for a newline before the first argument. In some cases consider also this:

```
some_super_long_object_name.verbose_method_name(
        argument1,
        argument2,
        argument3)
```

nfnty:
I strongly disagree on having the last line end with `)` for that example. Readability suffers and makes refactoring harder.

These are the ones I think are acceptable, in order of usage:
```
some_super_long_object_name.verbose_method_name(argument0, argument1)
```
```
some_super_long_object_name.verbose_method_name(
    argument0, argument1, argument2, argument3)
```
```
some_super_long_object_name.verbose_method_name(argument0, argument1,
                                                argument2, argument3)
```
```
some_super_long_object_name.verbose_method_name(
    argument0,
    argument1,
    argument2,
    argument3,
)
```
very rarely:
```
some_super_long_object_name.verbose_method_name(argument0,
                                                argument1,
                                                argument2,
                                                argument3)
```

Vifon:
> I strongly disagree on having the last line end with `)` for that example.

Fair enough. Just my Lisp habits. I'm ok with both. Though now that we mention it, I feel that it fits Python's design even more than Lisp's since Python prides itself in being brace-less.

nfnty:
Python 3.6 scout filter error fixed by c38c7f2a11eeaa882e2253a7fcabd62b3fadb8fc was much harder to debug due to `broad-except` at [config/commands.py L1258-L1262](https://github.com/ranger/ranger/blob/c38c7f2a11eeaa882e2253a7fcabd62b3fadb8fc/ranger/config/commands.py#L1258-L1262). Would be great to have this merged soon so that more important work can be started.

nfnty:
Is this ready to be merged?

hut:
1. can we throw out the pylint errors "bad-continuation" and "bad-whitespace" and the associated changes in commit 8c0575caacf66022c06e61711aecf99545c3c9fc? That would avoid forcing us to do things like
```python
programs = [program for program in self.rifle.list_commands([target.path],
                                                            None)]
```
and would allow us to line up assignments like
```python
AUTO_OPEN       = 'a'
OPEN_ON_ENTER   = 'e'
FILTER          = 'f'
SM_GLOB         = 'g'
```
which I think is worth keeping.

2. why add the `from __future__ import (absolute_import, print_function)` line to every file? print is barely used (and even if, it can be used in a way that is compatible with python2 and 3 without importing `print_function` from `__future__`). I don't think the effect of importing `absolute_import` ever actually applies. I would rather skip that patch.

3. Why throw out python 2.6, 3.2 & 3.3 versions from `.travis.yml`?

4. Thanks for the good work, mate!

nfnty:
1. These problems were mostly caused by `autopep8` not formatting properly. Should be fixed with 51ec08da2f3c3c5466a60304dd94db9da45d99d5. Tell me if you see any other strange formatting.
2. This is mostly a precursor to mimic Python 3 behavior as closely as possible. I will add `division` and `unicode_literals` in the future.
3. Pylint doesn't support `<2.7` and `<3.4`.

hut:
This distorts metadata which could've been used for git-blame, but whatever. Go with it.

PR #745, opened 2017-01-01
allow spaces again at the end of lines in rc.conf

dflupu:
If you have lines such as:
```
map cw console rename
map /  console search_inc
```

then being able to have a trailing space that isn't stripped is pretty useful. Otherwise, you have to press space after pressing the bound key, before typing the arguments in the console.


nfnty:
Use `%space` instead, as discussed in #720

PR #746, opened 2017-01-02
Rifle - better support for archives

andya9:
This adds patool as an alternative to atool for dealing with archives. Also makes aunpack work nicely with multiple archives as input.


nfnty:
The `atool` improvement has been merged. Thanks.

`patool` will only be merged if there's more interest in it being included by default.

nfnty:
I'll close this for now, doesn't seem like there's much interest in `patool`. Request reopening if more interest shows up.

PR #762, opened 2017-01-11
Don't use bold on selected items for better readability

felf:
When inverting a selected item's colour and setting the attributes to bold, the text becomes gray on a coloured background. This makes it quite hard to read even with a big font. I propose to just use black instead.

### DIFF
--- a/ranger/colorschemes/default.py
+++ b/ranger/colorschemes/default.py
@@ -63,11 +63,11 @@ class Default(ColorScheme):
                 fg = black
                 attr |= bold
             if context.main_column:
-                if context.selected:
-                    attr |= bold
                 if context.marked:
                     attr |= bold
                     fg = yellow
+            if context.selected:
+                attr &= ~bold
             if context.badinfo:
                 if attr & reverse:
                     bg = magenta
--- a/ranger/colorschemes/snow.py
+++ b/ranger/colorschemes/snow.py
@@ -18,7 +18,7 @@ class Snow(ColorScheme):
         elif context.in_browser:
             if context.selected:
                 attr = reverse
-            if context.directory:
+            elif context.directory:
                 attr |= bold
 
         elif context.highlight:
--- a/ranger/colorschemes/solarized.py
+++ b/ranger/colorschemes/solarized.py
@@ -76,11 +76,11 @@ class Solarized(ColorScheme):
                 fg = 234
                 attr |= bold
             if context.main_column:
-                if context.selected:
-                    attr |= bold
                 if context.marked:
                     attr |= bold
                     bg = 237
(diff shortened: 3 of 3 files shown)


nfnty:
Could you provide an image comparison?

felf:
Here's a screenshot of both the default (first image) and solarized scheme (second image) that are shipped with 1.8.1. The terminal is KDE's Konsole. Notice the grey-on-blue directory in the left column in the first image, but in contrast the much more readable black-on-blue on the second image. The grey-on-green for an executable file is also telling.

I'll do the rebasing tomorrow, as I am on the road right now.
ranger-default
ranger-solarized

nfnty:
To me these changes look worse than before, you'll have to convince me with better before/after images in different terminals for me to accept this PR.

Also, you still haven't rebased against master.

felf:
Sorry about the delay and misunderstanding, I've been down with a bad cold the last few days. Those two screenshots were made _without_ my changes and use the default colour schemes that come with ranger, in order to show the problem.

The previous images were made on a Gentoo system. Here is how ranger looks _with_ my changes, taken on an Ubuntu 12. Again the default and the solarized theme.

patched-default
patched-solarized

I saw your first request for a rebase and out of reflex clicked the "Update branch" button, which did a merge instead. :/  (this is my first github interaction).

nfnty:
I still have nothing to compare against. You've posted two images without changes from `Konsole` and two images with changes from another terminal emulator. Please, annotate each before/after image pair with terminal and version, so that I can reproduce your issues.

felf:
It is also Konsole, KDE 4.14.2 on Ubuntu 12.04.5. The colour scheme set in Konsole is "Linux colours" which is shipped with the package. The differences to look for are:
- in the respective top image (Ranger's colourscheme "default") in the left column: the active directory is shown grey on blue. My patch would effectively change it to black on blue.
- both colorschemes: the selected file in the main column is shown grey on colour. My patch would remove the bold flag, making the text black.

On a sidenote: I tried to reproduce it in other terminal emulators available at my work place: xterm 271, lxterminal 0.11 and urxvt 9.14. Neither of those have this issue. I have no deep knowledge about terminal emulation and how colours are _supposed_ to be reproduced, but Konsole's colourscheme uses a bright variant of a colour if it is written bold, and so, "bold black" becomes grey. Perhaps this is even a Konsole issue. :/

nfnty:
Yeah, I can see the differences. I'm unable to test Konsole myself, so I'll leave this to you/others. It's very likely caused by Konsole.

nfnty:
I'll close this for now, feel free to request reopening if you have a better fix.

PR #763, opened 2017-01-12
Add blank line after class or function definition.

stepshal:


nfnty:
This will be fixed in #743. Please, check it out and leave your comments there.

PR #770, opened 2017-01-18
Use 'file' as a fallback file preview method

Ape:
Ranger displays useful previews for many types of files, but sometimes there is no preview available for the type. In this case we could use `file` as the fallback command for providing information about the file type. This feature is implemented in this pull request.

`file` supports a huge number of file types and often provides helpful information. For example, an executable binary:
```
ELF 64-bit LSB executable
 x86-64
 version 1 (SYSV)
 dynamically linked
 interpreter /lib64/ld-linux-x86-64.so.2
 for GNU/Linux 2.6.32
 BuildID[sha1]=12becb41c4c38114d1df17d25fed36f907998faa
 stripped
```

Or an unknown `mapping.db` file that is now identified to be an SQLite database:
```
SQLite 3.x database
 last written using SQLite version 3011001
```

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -118,4 +118,4 @@ case "$mimetype" in
         try mediainfo "$path" && { dump | trim | sed 's/  \+:/: /;';  exit 5; } || exit 1;;
 esac
 
-exit 1
+try file --brief "$path" && { dump | tr ',' '\n'; exit 0; } || exit 1;


nfnty:
This needs proper exit handling in case `file` fails.

Use long option names and do not use option `-L`.

Ape:
@nfnty Thanks for the feedback. The commit now uses long option names and doesn't use `-L`. The command `file` never fails (even if the file does not exist).

nfnty:
That's not true. From `file(1)`: `file returns 0 on success, and non-zero on error`

Ape:
I updated the commit.

nfnty:
Thanks! Realized that the long options may not be compatible with other platforms. Also `-L` is needed because `$path` is not the dereferenced path, my bad.

nfnty:
Also the output can contain `,` in other places where splitting is not wanted, so the splitting was removed.

PR #791, opened 2017-02-03
Custom config file option for rifle

swalladge:
This change adds a command line option to the `rifle` standalone script to allow specifying a custom configuration file instead of the default.

#### ISSUE TYPE

#### CHECKLIST

#### DESCRIPTION

- new `-c CONFIG_FILE` option added to `rifle.py` arg parsing
- move `conf_path` logic to after arg parsing to allow the `-c` option to affect the process (if `-c` is set, that file will be used with no questions asked and without attempting to fallback to another default config file)
- rifle man page updated to suite.

#### MOTIVATION AND CONTEXT

I use rifle all the time, both with ranger and from the shell as an `xdg-open` replacement. However I find there are generally 2 contexts I want to use it in - from the shell where preference is for opening in text applications (eg. vim, atool for viewing archives), and the other is from gui applications where I need it to open a graphical application (eg. gvim, ark for archives). Currently there doesn't seem to be a way to split between the two. This cli option would allow tuning multiple `rifle.conf` files to different contexts. (Although, if there could be a better way, please suggest one. 😃 )

#### TESTING

This only affects running `rifle` as the standalone script. I couldn't find any automated tests set up relating to rifle. Please advise if any tests are required for this.

Also it seems like it fails pylint, which I failed to notice before... If maintainers are interested in merging this feature, I'd be interested to hear your opinion on the tests it failed on (too many branching - likely because if the extra nested if/else blocks) and how much refactoring you'd like to see.

UPDATE: refactored so the code is cleaner as well as passes pylint.

### DIFF
--- a/doc/rifle.1
+++ b/doc/rifle.1
@@ -139,7 +139,7 @@ rifle \- ranger's file opener
 .SH "SYNOPSIS"
 .IX Header "SYNOPSIS"
 \&\fBrifle\fR [\fB\-\-help\fR] [\fB\-f\fR \fI\s-1FLAGS\s0\fR] [\fB\-l\fR] [\fB\-p\fR \fI\s-1KEYWORD\s0\fR]
-[\fB\-w\fR \fI\s-1PROGRAM\s0\fR] \fIfiles\fR
+[\fB\-w\fR \fI\s-1PROGRAM\s0\fR] [\fB\-c\fR \fI\s-1CONFIG_FILE\s0\fR] \fIfiles\fR
 .SH "DESCRIPTION"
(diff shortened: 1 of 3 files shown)


nfnty:
Thanks!

starun96:
Any updates to this? I would like to use my existing rifle.conf in my custom directory for ranger rather than the one in .config/ranger. Right now, I'm hard linking my .config/ranger/rifle.conf to my custom one, which doesn't seem ideal.

swalladge:
@starun96 this has been merged and ready to use. It only applies to using `rifle` on the command line though. A new issue/PR should be opened for any other purpose.

PR #797, opened 2017-02-07
fix the is_binary function

### LINKED ISSUE #161: Ranger attempts to preview binary files
With scope.sh disabled, ranger attempts to preview .odt and .apk files even though they are binary.  The effect is that the preview pane is opened (if it was collapsed before), but it stays blank, no preview is displayed.

In this snippet, the first file does not provoke a preview whereas the second does.
https://gist.github.com/SammysHP/764fcd3d1c474bd49645

swalladge:
Fix problems relating to the `is_binary` function and differences in string
handling between python 2 and 3. `is_binary` now works as expected in both
python versions, and avoids ranger attempting to preview a binary file.

#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Python version: 2 and 3

#### CHECKLIST

#### DESCRIPTION

- opens file in binary mode to avoid any decoding errors
- uses set of control characters as set of integers in python3, or a set of
  characters in python2, due to differences in string handling between the
  versions.

#### MOTIVATION AND CONTEXT

Fixes #161

#### TESTING

Tested with python 3.6.0 and 2.7.13

### DIFF
--- a/ranger/container/file.py
+++ b/ranger/container/file.py
@@ -4,12 +4,14 @@
 from __future__ import (absolute_import, division, print_function)
 
 import re
+import sys
+
 from ranger.container.fsobject import FileSystemObject
 
 N_FIRST_BYTES = 256
-# pylint: disable=invalid-name
-control_characters = set(chr(n) for n in set(range(0, 9)) | set(range(14, 32)))
-# pylint: enable=invalid-name
+CONTROL_CHARACTERS = set(list(range(0, 9)) + list(range(14, 32)))
+if sys.version_info[0] < 3:
+    CONTROL_CHARACTERS = set(chr(n) for n in CONTROL_CHARACTERS)
 
 # Don't even try to preview files which match this regular expression:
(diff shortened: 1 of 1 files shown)


nfnty:
Why are you matching a `str` object against a `bytes` object for Python 2?

I'll fix this.

swalladge:
@nfnty wait what? Can you explain what you mean?

nfnty:
Sorry, thought about it the wrong way. `bytes` objects don't exist in Python 2...

What I meant to say is that you're matching `int` against `bytes` objects for Python 3. Which means that `is_binary()` will always return `False`.

nfnty:
Disregard everything I've just said, seems like I'm way too tired to reason about Python 2 vs 3 `string`/`bytes`/`int` handling. 

Of course the `set()` operation converts the `bytes` object into a set of `int`, so this will work. I'll clean it up and merge it.

nfnty:
Thanks!

PR #809, opened 2017-02-16
Change DIRSTATUSES order.

m-ou-se:
- Untracked files should be less important than changes to tracked
  files.

- When there are both changed and deleted files, 'changed' seems to be a
  better summary of the situation than 'deleted', so prioritize changed
  over deleted.

- Staged files should be more important than non-staged files, since
  that indicates there's a commit being formed, but not yet committed.


nfnty:
This is very subjective and should be fixed with better symbols for composite statuses.

nfnty:
I'll close this for now, if you come up with a better implementation then you're free to open a new issue.

PR #853, opened 2017-04-11
rc.conf: use the correct confpath for scope.sh

### LINKED ISSUE #849: Hardcoded incorrect path to scope.sh
Ranger always looks for scope.sh inside ~/.config/ranger ([responsible line](https://github.com/ranger/ranger/blob/ad58add9e0e6cb4675b231d967ff2204eabad4c3/ranger/config/rc.conf#L46)). This is incorrect: ranger should instead look inside [$XDG_CONFIG_HOME](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.7.html)/ranger.

This is aggravated by the fact that `ranger --copy-config=scope` will (correctly) copy scope.sh to $XDG_CONFIG_HOME/ranger. One would imagine that ranger would be able to find config files it itself created.

#### ISSUE TYPE
- Bug report

#### RUNTIME ENVIRONMENT
- Operating system and version: Arch Linux 4.9.6-1
- Terminal emulator and version: irrelevant; the issue can be reproduced even when ranger isn’t attached to a pty.
- Python version: Python 3.6.0 (default, Jan 16 2017, 12:12:55)
- Ranger version/commit: reproduced on both ranger-stable 1.8.1 and current master (30553796)
- Locale: LANG=en_US.UTF-8, LC_COLLATE=C

#### STEPS TO REPRODUCE
1. `export XDG_CONFIG_HOME=~/not-the-default`
2. `ranger --copy-config=scope`
3. `strace -f ranger 2>&1 | grep scope.sh`

#### EXPECTED BEHAVIOR
Ranger opens and uses $XDG_CONFIG_HOME/ranger/scope.sh:

    stat("~/not-the-default/ranger/scope.sh", ...) = 0

#### CURRENT BEHAVIOR
Ranger looks for ~/.config/ranger/scope.sh and fails if it doesn’t exist:

    stat("~/.config/ranger/scope.sh", ...) = -1 ENOENT (No such file or directory)

#### CONTEXT

This issue prevents image previews from working. As a workaround, I created the directories ~/.config and ~/.config/ranger and copied my scope.sh in there; this lets image previews work, but is obviously not a proper solution.

#### POSSIBLE SOLUTIONS
Just change [this line](https://github.com/ranger/ranger/blob/ad58add9e0e6cb4675b231d967ff2204eabad4c3/ranger/config/rc.conf#L46) to use $XDG_CONFIG_HOME instead of hard-coding .config.

Grimy:
Change the default rc.conf to look inside $XDG_CONFIG_HOME rather than ~/.config, as required by the XDG Base Directory specification.

#### ISSUE TYPE
- Bug fix

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION

The change is implemented with the line ```eval cmd("set preview_script " + fm.confpath("scope.sh"))```. This re-uses the fm.confpath function to guarantee that we’re using the same path as `ranger --copy-config`, which is the desired behavior.

#### MOTIVATION AND CONTEXT

This fixes #849; see the issue for details.

### DIFF
--- a/examples/rc_emacs.conf
+++ b/examples/rc_emacs.conf
@@ -37,7 +37,7 @@ set confirm_on_delete multiple
 # Which script is used to generate file previews?
 # ranger ships with scope.sh, a script that calls external programs (see
 # README.md for dependencies) to preview images, archives, etc.
-set preview_script ~/.config/ranger/scope.sh
+eval cmd("set preview_script " + fm.confpath("scope.sh"))
 
 # Use the external preview script or display simple plain text or image previews?
 set use_preview_script true
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -43,7 +43,7 @@ set confirm_on_delete multiple
 # Which script is used to generate file previews?
 # ranger ships with scope.sh, a script that calls external programs (see
 # README.md for dependencies) to preview images, archives, etc.
-set preview_script ~/.config/ranger/scope.sh
+eval cmd("set preview_script " + fm.confpath("scope.sh"))
 
 # Use the external preview script or display simple plain text or image previews?
 set use_preview_script true


j-degreef:
Ok, it works (here), but that will surely complicate things for newbies, all this because one needed to run a strace instead at looking at its rc.conf see  #849.

Between, 
set preview_script ~/.config/ranger/scope.sh
and 
eval cmd("set preview_script " + fm.confpath("scope.sh"))
newbies will need a big hand if they face problem with their scope.sh :)
A configuration file is meant to be changed by the average user and should be self explanatory.
Plus we all use different environments and we can not count on the "Single Source of Truth"
On Linux Mint, for instance XDG_CONFIG_HOME is not set.

Issue #821, opened 2017-03-01
Hide files when called with --choosedir


Droggelbecher:
#### ISSUE TYPE
- Improvement/feature request

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 14.04 LTS
- Terminal emulator and version: rxvt-unicode (urxvt) v9.19 - released: 2013-10-27
- Python version: 2.7.6
- Ranger version/commit: ranger-stable 1.7.1
- Locale: de_DE.UTF-8

#### EXPECTED BEHAVIOR
When using ranger as a directory chooser (eg. from within another program such as for conveniently changing the working directory of a text editor), it feels very unnatural to have non-directory files visible and even being able to execute actions on them.
I would rather expect that:
a) Only directories are visible (that makes it clear to the user that he/she is currently choosing a directory)
b) The process of selecting a directory feels more like selecting something (currently its: entering the directory of your choice then quitting, which almost feels like aborting the whole selection process).

Alas, I have no more concrete proposal for b), but I think a) should be easy to implement.

#### CURRENT BEHAVIOR
Currently ranger is pretty much behaving as without --choosedir and writing out the last visited directory as a side effect (which possibly might be desirable for some applications, in that case I'd like to request a seperate command line switch to change behaviors).

#### CONTEXT
I'd like to write a script for the kakoune text editor to conveniently change its current working directory, right now "ranger --choosedir=..." works technically, but as mentioned above doesn't really "feel" like choosing a directory at all.

#### POSSIBLE SOLUTIONS

I recommend to either change the default behavior when called with --choosedir or to implement an additional flags such that non-directories are not displayed at all.
b) could probably be solved with a keybinding (which in turn is easily done in the config file), biggest problem here is to find one that feels natural (as return is entering a directory, what should we use for finally selecting it?).

nfnty:
Only expectation `a` is feasible and could be implemented with something similar to the command `filter_inode_type`, but globally.

PR #866, opened 2017-05-19
Show only directories when using --choosedir

randynobx:
#### ISSUE TYPE

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
Added global_inode_filter_type setting for global filtering of file types. Uses same logic as the filter_inode_type command.

Sets the `global_inode_filter_type` to `'d'` (for directory) when --choosedir flag is used

#### MOTIVATION AND CONTEXT
Addresses request for global filtering of files when using --choosedir= in issue #821 

#### TESTING
Have done basic testing by hand to ensure functionality

### DIFF
--- a/ranger/container/directory.py
+++ b/ranger/container/directory.py
@@ -95,6 +95,11 @@ def mtimelevel(path, level):
 
 class Directory(  # pylint: disable=too-many-instance-attributes,too-many-public-methods
         FileSystemObject, Accumulator, Loadable):
+
+    FILTER_DIRS = 'd'
+    FILTER_FILES = 'f'
+    FILTER_LINKS = 'l'
+
     is_directory = True
     enterable = False
     load_generator = None
@@ -252,11 +257,25 @@ class Directory(  # pylint: disable=too-many-instance-attributes,too-many-public
                         return False
                 return True
             filters.append(hidden_filter_func)
+        if self.settings.global_inode_type_filter or self.inode_type_filter:
+            def hidden_inode_filter_func(obj):
+                # Use local inode_type_filter is present, global otherwise
+                if self.inode_type_filter:
+                    inode_filter = self.inode_type_filter
+                else:
+                    inode_filter = self.settings.global_inode_type_filter
+                # Apply filter
+                if self.FILTER_DIRS in inode_filter and obj.is_directory:
+                    return True
+                elif self.FILTER_FILES in inode_filter and obj.is_file and not obj.is_link:
+                    return True
+                elif self.FILTER_LINKS in inode_filter and obj.is_link:
+                    return True
+                return False
+            filters.append(hidden_inode_filter_func)
         if self.filter:
             filter_search = self.filter.search
             filters.append(lambda fobj: filter_search(fobj.basename))
-        if self.inode_type_filter:
-            filters.append(self.inode_type_filter)
         if self.temporary_filter:
             temporary_filter_search = self.temporary_filter.search
             filters.append(lambda fobj: temporary_filter_search(fobj.basename))
--- a/ranger/container/directory.py
+++ b/ranger/container/directory.py
@@ -92,15 +92,15 @@ def mtimelevel(path, level):
         mtime = max(mtime, max([-1] + [os.stat(d).st_mtime for d in dirlist]))
     return mtime
 
-class InodeFilterConstants(object):
+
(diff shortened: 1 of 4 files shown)


ranger/container/directory.py
```diff
@@ -252,6 +252,19 @@ def hidden_filter_func(fobj):
                         return False
                 return True
             filters.append(hidden_filter_func)
+        if self.settings.global_inode_type_filter and not self.inode_type_filter:
+            # pylint: disable=too-many-boolean-expressions
+            def hidden_inode_filter_func(obj):
+                g_filter = self.settings.global_inode_type_filter
+                if (
+                        ('d' in g_filter and obj.is_directory) or
```
Vifon: Could we either use use a few if-else branches with one `return` in each or return a single expression? Right now it's a strange hybrid.

```python
if foo:
    return True
elif bar:
    return True
else:
    return False
```
or
```python
return foo or bar
```

Not:
```python
if foo or bar:
    return True
else:   # or even no else
    return False
```

I'd slightly prefer the first version.

Vifon: Also, it's almost the same function as the one used in `commands.py`, in `filter_inode_type.execute`. Maybe we can reuse it and not duplicate this code? Especially since that one uses the named constants instead of the magic strings (`self.FILTER_DIRS` vs `'d'`).

ranger/core/main.py
```diff
@@ -119,6 +119,9 @@ def main(
         FileManagerAware.fm_set(fm)
         load_settings(fm, args.clean)
 
+        if args.choosedir:
+            SettingsAware.settings.global_inode_type_filter = 'd'
```
Vifon: Please use a named constant instead of `'d'`.

randynobx: Where should these constants be declared/stored? The value being assigned will be the same, but I can understand wanted a constant for readability and maintenance purposes.

ranger/container/directory.py
randynobx: The first version would be my choice as well.

As for the code, my understanding is that the filter does not 'exist' in ranger's code until the command is run, which is a problem for this use case.

Vifon: Do you mean that it's a lambda function? Let's make it a regular function and use it in both these places. This file looks like a good place for it I'd like to hear @hut's opinion.

ranger/core/main.py
Vifon: Either `ranger.container.directory` or some new module like `ranger.core.constants`. Again, I'd like to hear @hut's opinion on this.

ranger/container/directory.py
randynobx: That sounds like a good idea. I can easily refactor both the local and global inode_filters into one block of code; which would actually take care of resolving any conflicts between the two.

I'll work on that code in case. But I too would like to hear @hut's opinion.

ranger/core/main.py
randynobx: AFAIK (and can find via google) the accepted way to name constants in python is to use all caps (such as `FILTER_DIR`). But I am getting pylint errors complaining that this is an 'invalid variable name'?

randynobx:
Refactored the inode_filter_func code and moved the logic for `:filter_inode_type` from commands.py to refilter() to reduce redundant code.

Still need to decide on where to store the constants: `FILTER_DIRS`, `FILTER_FILES`, and `FILTER_LINKS`

Vifon:
Looks good to me. @hut, please take a look when you'll have some time.

hut:
Looks good, thank you!

hut:
somehow the setting global_inode_type_filter is set to "d" by default even when not using --chosedir. I worked around this by unsetting it explicitly in rc.conf, but this is weird anyway. Got any idea why?

randynobx:
I'm unable to reproduce this behavior.

Just to clarify, the filter is enabled by default (and set to "d") even without the --choosedir flag?

hut:
Yes. Do you use the current master branch with the workaround, maybe? Try commit 88e0bc0540fb963b3cf638217dd578b127fd6a3a

randynobx:
I actually haven't applied the workaround patch. I'm using the master branch at commit 88e0bc0 right now

hut:
Ok, I will investigate later

PR #870, opened 2017-05-21
Add the :narrow command

### LINKED ISSUE #867: Filter command to only show marked files
#### ISSUE TYPE
- Improvement/feature request

#### RUNTIME ENVIRONMENT

skipped

#### EXPECTED BEHAVIOR

A command that filters the file view to only show marked files. In filtered view, unmarking a file/dir should not hide it until the command is run again. Should be re-settable to unfiltered view.

#### CONTEXT

Selecting files sparingly in a large-ish directory can quickly cause you to lose overview. By filtering to only show marked files, you can more easily grasp the context and potentially unselect marked files.

#### POSSIBLE SOLUTIONS

Probably as an argument to `scout`.

Vifon:
The feature name is derived from https://www.gnu.org/software/emacs/manual/html_node/emacs/Narrowing.html

Closes #867.

### DIFF
--- a/examples/rc_emacs.conf
+++ b/examples/rc_emacs.conf
@@ -406,6 +406,7 @@ map <C-x>zs    toggle_option sort_case_insensitive
 map <C-x>zu    toggle_option autoupdate_cumulative_size
 map <C-x>zv    toggle_option use_preview_script
 map <C-x>zf    console filter%space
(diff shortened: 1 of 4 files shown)


hut:
Thanks! Nice that you even added it to rc_emacs.conf :P

PR #879, opened 2017-05-28
Yank command

toonn:
Implement a yank command instead of the long mappings for yn, yd, yp and ys.

#### ISSUE TYPE

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
The command can use any of pbcopy, xclip or xsel, depending on what's available.
Behaviour for yn, yd and yp has changed. Instead of the highlighted file they operate on the selection, defaulting to the highlighted file if nothing's selected.

#### MOTIVATION AND CONTEXT
The yank command makes it easier to use multiple clipboard managers, xclip instead of xsel for example. It also groups the behaviour of several long mappings.

#### TESTING
Manually tested the command in all it's incantations.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1688,6 +1688,7 @@ class yank(Command):
 
     def execute(self):
         import subprocess
+
         def clipboards():
             from ranger.ext.get_executables import get_executables
             clipboard_managers = {
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1721,7 +1721,7 @@ class yank(Command):
             process.communicate(input=input_argument)
 
     def get_selection_attr(self, attr):
-        return [getattr(file, attr) for file in
+        return [getattr(item, attr) for item in
                 self.fm.thistab.get_selection()]
 
     def tab(self, tabnum):
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -91,7 +91,6 @@ import re
 import subprocess
 
 from ranger.api.commands import Command
-from ranger.ext.get_executables import get_executables
 
 
 class alias(Command):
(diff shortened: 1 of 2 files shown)


ranger/config/commands.py
```diff
@@ -88,6 +88,7 @@
 from collections import deque
 import os
 import re
+import subprocess
```
Vifon: Could you put that into `yank.execute` like `get_executables`?

ranger/config/commands.py
```diff
@@ -1670,3 +1671,62 @@ def execute(self):
         # Ask the browsercolumns to redraw
         for col in self.fm.ui.browser.columns:
             col.need_redraw = True
+
+
+class yank(Command):
+    """:yank [name|dir|path]
+
+    Copies the file's name(default), directory or path into both the primary X
```
Vifon: Please add a space before "(default)".

ranger/config/commands.py
```diff
@@ -1670,3 +1671,62 @@ def execute(self):
...
+            for manager in ordered_managers:
+                if manager in executables:
+                    return clipboard_managers[manager]
+
+            return []
+
+        clipboard_commands = clipboards()
+
+        selection = self.get_selection_attr(self.modes[self.arg(1)])
+        input_argument = "\n".join(selection)
```
Vifon: Rename `input_argument` to something specifying the intent, like `new_clipboard_contents`.

ranger/config/commands.py
```diff
@@ -1670,3 +1671,62 @@ def execute(self):
...
+                'pbcopy': [
+                    ['pbcopy'],
+                ],
+            }
+            ordered_managers = ['pbcopy', 'xclip', 'xsel']
+            executables = get_executables()
+            for manager in ordered_managers:
+                if manager in executables:
+                    return clipboard_managers[manager]
+
```
Vifon: Consider removing this empty line.

hut:
Hi toon, this looks good so far. There is some altered behaviour now though: when using `yd`/`yp`/`yn` on multiple files, it will copy multiple lines instead of just one. This has advantages but also disadvantages. I will think about it, in the meantime, comments by anyone would be appreciated.

Vifon:
In my opinion the new behavior is superior. If I have any files marked, operating on them is The Expected Behavior™ for me. We don't have the separate copy/cut commands for ignoring the selection either. I'd also want to point out that <kbd>yn</kbd> and <kbd>ys</kbd> now do the same thing so we might want to deprecate one of them (probably <kbd>ys</kbd>).

hut:
Good point. Merging. Thank you both!

PR #912, opened 2017-07-24
Make the pagewise scrolling symmetrical

Vifon:
Moving half a page down and back up should put the user in the original position. It was true only if the file count was even. If it was odd, the user ended up one file above the originally focused one.

### DIFF
--- a/ranger/ext/direction.py
+++ b/ranger/ext/direction.py
@@ -20,6 +20,8 @@ False
 
 from __future__ import (absolute_import, division, print_function)
 
+import math
+
 
 class Direction(dict):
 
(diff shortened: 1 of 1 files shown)


Vifon:
I am aware the rounded calculation may have some side effects such as not strolling a full page if the file count is odd (it will be full page -1) but in my opinion the benefits outweigh them.

nfnty:
I don't think disabling `no-else-return` is a good idea. Much better to just remove the `else` statement, the rest of the codebase follows this rule already.

nfnty:
Also you should squash this PR.

Vifon:
Fair enough, I've squashed them. I was hesitant to do so as I considered the previous commits an information/warning about what was already tried before and why it was not enough.

I think I found a solution for our little pylint problem that will satisfy both of us.

ranger/ext/direction.py
```diff
@@ -142,8 +144,16 @@ def move(self, direction, override=None, minimum=0,  # pylint: disable=too-many-
...
             self['_move_cycles'] = int(cycles)
-            return int(minimum + pos)
-        return int(max(min(pos, maximum + offset - 1), minimum))
+            ret = minimum + pos
+        else:
+            ret = max(min(pos, maximum + offset - 1), minimum)
+        # Round towards the direction we're moving from.
+        # From the UI point of view, round down. See: #912.
+        if direction < 0:
+            ret = int(math.ceil(ret))
```
Vifon: The documentation makes a point that `math.ceil()` returns a float, which doesn't make sense in our case. It will work but doesn't make sense.

nfnty: Ugh, apparently `math.ceil()` has changed from 2 to 3... Good catch; it's fine then.

nfnty: Always returns `int` on 3.

PR #978, opened 2017-11-27
Fix bug high ram usage when previewing images

Jeytas:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Manjaro
- Terminal emulator and version: urxvt v9.22
- Python version: 3.6.3
- Ranger version/commit: 1.9.0b6
- Locale: en_GB.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

I added these two lines so that the image preview automatically closes which prevents high RAM usage.

#### MOTIVATION AND CONTEXT

Looking at many images, ranger would get a high RAM usage and oftentimes slow down the system after looking at too many pictures.


hut:
Sadly this removes the advantage of caching the images by w3mimgdisplay. But this insane memleak definitely intolerable.

So, thank you very much. I will merge it with both commits squashed, to keep the git history of README.md clean.

Vifon:
@hut I would definitely consider making a counter instead and killing the `w3m` process only after every `n` images. Maybe even configurable `n`. It would give us the best of both worlds.

PR #996, opened 2017-12-14
rifle_sxiv.sh: be POSIX compat: single = for string eq

quite:


Vifon:
I believe this file is no longer used and its functionality is now included in ranger itself. @hut, could you confirm this? Maybe we should remove this file completely then?

quite:
Hm, yeah, that's probably correct. I think I read a note somewhere... I do use this myself, if you remove it I'll fork it to its own repo :)

Vifon:
How do you use it? With ranger/rifle or as an independent script? Well, if somebody is using it, I see no point in removing it.

quite:
Yeah, standalone actually. I stopped using ranger itself quite some time ago :> Forgot it, kind of. But perhaps I should look at it again, lots seem to have happened...

quite:
So, LGTM :)

Vifon:
Merged, thank you. :)

PR #1013, opened 2018-01-03
Fix column_ratios tab completion

Vifon:
The tab completion for `:set column_ratios <TAB>` previously used the literal list representation of this setting (e.g. `[1, 3, 4]`) which cannot be used in this syntax. Now this list is joined into a single comma-separated string that is actually usable (e.g. `1,3,4`).


hut:
=)

PR #1016, opened 2018-01-05
Changes to the draw_border setting

squigglezworth:
This changes the 'draw_borders' setting to a string, with possible values of none (default), outline, separators, or both (outline + separators)

Outline draws an outline around the columns with no vertical lines between the columns.
Separators draws vertical lines between the columns
Both draws both, which is how the current draw_borders setting works
None, obviously, draws none of these and is the default


doc/ranger.pod
```diff
@@ -755,9 +755,18 @@ Display the file size in the status bar?
...
 
-Draw borders around columns?
+Draw borders around or between the columns? Possible values are:
+
+ none           no borders of any sort
+ outline        draw an outline around all the columns
+ separators     draw only vertical lines between columns
+ both           both of the above
+ 
+=item draw_separators [bool]
```
Vifon: I believe this setting got removed. Let's remove it from the docs too.

ranger/config/rc.conf
```diff
@@ -123,8 +123,8 @@ set status_bar_on_top false
 # currently running tasks which support progress bars?
 set draw_progress_bar_in_status_bar true
 
-# Draw borders around columns?
-set draw_borders false
+# Draw borders around columns? (separators, outline, both, or none. Separators are vertical lines between columsn. Outline draws a box around all the columns. Both combines the two)
```
Vifon: Can we break this line into a few shorter ones?

ranger/container/settings.py
```diff
@@ -96,6 +96,7 @@
 ALLOWED_VALUES = {
     'cd_tab_case': ['sensitive', 'insensitive', 'smart'],
     'confirm_on_delete': ['multiple', 'always', 'never'],
+    'draw_borders': ['None', 'both', 'outline', 'separators'],
```
Vifon: Since `'None'` here is a string and not the true `None` value, let's keep it lowercase for consistency.

hut:
Thank you for the pull request, @squigglezworth. The new version of ranger is going to be released soon, and I will merge your patches after that. :)

toonn:
@squigglezworth, if you fix up the PR according to vifon's review it might make it in the next release.

toonn:
@squigglezworth, I applied the requested change myself because we're nearing a release.

The `ranger.1` and `rifle.1` files were conflicting with master and I didn't know how to resolve them so I rebased, resolved merge conflicts and then force-pushed. Then I realized I was rewriting *your* history, which is pretty bad. I did another force push to fix it. I'm sorry if I still screwed it up. I'll resolve the conflicts when merging into master.

PR #1018, opened 2018-01-06
Completion list preview when shell command typing

shadowxnex:
#### ISSUE TYPE
* Improvement/feature solution-proposal

#### CURRENT BEHAVIOUR
When you typing shell command in console, sometimes you want to see the variants how to complete this command on the list. Any bash terminal have such behaviour:
```
# fi<TAB>
fi           file         file-rename  filterdiff   findfs       firefox      fixcvsdiff   fix-qdf      
fiascotopnm  filefrag     file-roller  find         findmnt      fitstopnm    fixparts
```
But current behaviour is to iterate by round robin technique: you push the tab key at first time and have `file` option, double pushing affect `file-rename` one and etc:

```(ra)
user@ranger /home/user/
...
:shell file-rename 
```

#### EXPECTED BEHAVIOUR
It will be better if you can see on the preview list all available options before typing further.

```
user@ranger /home/user/
fi
fiascotopnm
file
file-rename
file-roller
filefrag
filterdiff
find
findfs
findmnt
firefox
fitstopnm
fix-qdf
fixcvsdiff
fixparts


:shell fi
```

### DIFF
--- a/ranger/container/settings.py
+++ b/ranger/container/settings.py
@@ -74,6 +74,7 @@ ALLOWED_SETTINGS = {
     'vcs_backend_svn': str,
     'xterm_alt_key': bool,
     'clear_filters_on_dir_change': bool,
+    'console_tab_preview': bool,
 }
 
 ALLOWED_VALUES = {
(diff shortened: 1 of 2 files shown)


toonn:
This is a cool idea. The window should be much smaller though, not replace the entire ranger interface. Like command hints or the show_bookmarks window.

PR #1036, opened 2018-01-22
:touch and :mkdir focus their argument

### LINKED ISSUE #1028: Touch command after creation of file don't point on created file
Hi! By default `:touch` command don't point to the just created file.
I think it will be really helpful, especially if in the directory where there are many files, so I don't need to search just created file.

toonn:
:touch now behaves more like the unix command in case the file already
exists, the notification is now info-level instead of error.
Both :touch and :mkdir now focus the new/existing file/directory.

Fix #1028

#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Not sure whether documentation should be updated.

#### MOTIVATION AND CONTEXT

Make touch behave more like the unix command and since that could just as well be called mkfile, give mkdir the same behaviour.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -838,12 +838,17 @@ class mkdir(Command):
     def execute(self):
         from os.path import join, expanduser, lexists
         from os import makedirs
+        from ranger.container.file import File
 
         dirname = join(self.fm.thisdir.path, expanduser(self.rest(1)))
         if not lexists(dirname):
             makedirs(dirname)
+            dir_new = File(dirname)
+            self.fm.thisfile = dir_new
+            self.fm.thisdir.pointed_obj = dir_new
         else:
             self.fm.notify("file/directory exists!", bad=True)
+            self.fm.thisdir.move_to_obj(dirname)
 
     def tab(self, tabnum):
         return self._tab_directory_content()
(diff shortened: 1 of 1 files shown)


ranger/config/commands.py
```diff
@@ -857,12 +862,19 @@ class touch(Command):
 
     def execute(self):
         from os.path import join, expanduser, lexists
+        from os import utime
+        from ranger.container.file import File
 
         fname = join(self.fm.thisdir.path, expanduser(self.rest(1)))
         if not lexists(fname):
             open(fname, 'a').close()
+            file_new = File(fname)
```
hut: it wouldn't work to just do a move_to_obj() here?

ranger/config/commands.py
```diff
@@ -838,12 +838,17 @@ class mkdir(Command):
     def execute(self):
         from os.path import join, expanduser, lexists
         from os import makedirs
+        from ranger.container.file import File
 
         dirname = join(self.fm.thisdir.path, expanduser(self.rest(1)))
         if not lexists(dirname):
             makedirs(dirname)
+            dir_new = File(dirname)
```
hut: This doesn't work with `:mkdir x/y/z`. Perhaps use `self.fm.select_file(dirname)`, it's made for this :)

toonn:
I'm currently stuck on this:
- `select_file` doesn't focus new paths in the current directory, it does in subdirectories and for existing files/directories
- `move_to_obj` doesn't focus new `File`s
- setting `thisdir.pointed_obj` and `thisfile` to File(path) doesn't work if path's in a subdirectory or it's an existing path

I need something that can focus a file/directory no matter where it is and whither it's new. Directories only in the current directory really.

PR #1037, opened 2018-01-22
Load system-wide configuration files.

### LINKED ISSUE #869: System-wide configuration file
Is there a system wide configuration file that would apply to all users?

`~/.config/ranger/rc.conf` is per-user configuration file and applies only to the current user.

I'm looking for something like `/etc/inputrc` to apply to all users of the system by default.

Thanks in advance.

toonn:
Both rc.conf and commands.py are now additionally loaded from
`/etc/ranger` if they exist.

Fix #869

#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Not sure where we document config file locations.

#### MOTIVATION AND CONTEXT

Issue #869 asks for system-wide configuration files. People usually expect this behaviour.

#### TESTING

I've tested under python2.7, further tests under all versions since 2.6 should be performed.

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -129,7 +129,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.9.0" "2018-01-28" "ranger manual"
+.TH RANGER 1 "ranger-1.9.0" "2018-05-08" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
 .\" way too many mistakes in technical documents.
 .if n .ad l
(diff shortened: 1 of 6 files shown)


toonn:
I've done minimal updates to documentation.
One further thing we need to decide is `--copy-config`:
- Add an `etccommands` and `etccommands_full`?
- Include it in `all`?
- If the user doesn't have permission to write to `/etc` should everything fail to copy or just the files
  that would go in `/etc`?

toonn:
@Vifon's pov on this is it's strictly for the package manager or the user to handle. I can see a case for providing the convenience command. I'll wait for a couple more opinions before going ahead without the extra `--copy-config` options.

PR #1041, opened 2018-01-29
Solve map conflict for `gt`.

### LINKED ISSUE #1040: gt is mapped double
It is mapped both to ``go /tmp`` and ``go next tab``:

https://github.com/ranger/ranger/blob/master/ranger/config/rc.conf#L390
https://github.com/ranger/ranger/blob/master/ranger/config/rc.conf#L463

This is confusing.
I vote to get rid of the ``next tab`` mapping, since that already has other keys.

toonn:
For consistency with vim `gt` still means _go to the next tab_ the map
for _go to /tmp_ is now `gp` (first available letter in `/tmp`).

Fix #1040

#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

The map for `/tmp` was useless, now it's not.
The map for `/run/media/$USER` was missing and fit in well with both `/media` and `/mnt`.
Maybe the map for `/media` should also include `$USER`?
Possibly we should handle the case where `$USER` is `root` specially, dropping the `$USER` part.
Since these "hardcoded" maps are getting somewhat cluttered, maybe we should consider having a new bookmarks namespace with system directories.

#### TESTING

Tested both maps.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -387,7 +387,7 @@ map gv cd /var
 map gm cd /media
 map gM cd /mnt
 map gs cd /srv
-map gt cd /tmp
+map gp cd /tmp
 map gr cd /
 map gR eval fm.cd(ranger.RANGERDIR)
 map g/ cd /


Vifon:
I'm not convinced `/run/media/$USER` deserves its own keybinding. How widely is it used?

PR #1054, opened 2018-02-04
Change bulkrename to execute a fixed command

### LINKED ISSUE #880: bulkrename doesn't work if open txt file in a new terminal
If I set ranger to open txt files in a new terminal by changing the relevant line in `rifle.conf`:
```
mime ^text,  label editor = $EDITOR -- "$@"
```
to
```
mime ^text,  flag t, label editor = $EDITOR -- "$@"
```
Then when doing a bulkrename, the newly opened text editor contains no file names in it.

#### ISSUE TYPE
- Bug report

#### RUNTIME ENVIRONMENT

- Operating system and version: Manjaro 17.0.1
- Terminal emulator and version: Gnome-terminal 3.22.1
- Python version: 3.6.0
- Ranger version/commit: ranger-stable 1.8.1
- Locale: en_US.utf-8

#### EXPECTED BEHAVIOR

In ranger, select multiple files, then `:bulkrename` opens the editor with selected file names.

#### CURRENT BEHAVIOR

If set to open editor in a new terminal, editor (`vim` in my case) opens but with nothing in it.

EgZvor:
Do not attempt to resolve mimetypes and labels to determine the right
call for the bulkrename command. We always need the same behaviour.

Fixes #880

#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
- Operating system and version: 4.14.15-1-ARCH
- Python version: 3.6.4 (default, Jan  5 2018, 02:35:40) [GCC 7.2.1 20171224]
- Ranger version/commit: ranger-master 1.9.0 (afdd6aafcc0654424f8901e8aa83948382ef7bd2)
- Locale: en_US.UTF-8

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
bulkrename command now uses a fixed command instead of using rifle to determine it. The command is simply `vim "$@"`, which may not be the best choice.
In `rifle.py` `execute` function was refactored to allow execution of a predetermined command.

#### MOTIVATION AND CONTEXT
#880

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1056,6 +1056,8 @@ class bulkrename(Command):
     After you close it, it will be executed.
     """
 
+    RENAME_CMD = 'vim "$@"'
+
     def execute(self):  # pylint: disable=too-many-locals,too-many-statements
         import sys
         import tempfile
@@ -1073,7 +1075,8 @@ class bulkrename(Command):
         else:
             listfile.write("\n".join(filenames))
         listfile.close()
-        self.fm.execute_file([File(listpath)], app='editor')
+        rename_command = "set -- '%s'; %s" % (File(listpath).path, self.RENAME_CMD)
+        self.fm.rifle.execute_command(rename_command)
         listfile = open(listpath, 'r')
         new_filenames = listfile.read().split("\n")
         listfile.close()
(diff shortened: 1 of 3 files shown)


SammysHP:
What if people use vi or emacs or nano?

EgZvor:
@SammysHP that'd be unfortunate. So, with terminal editors the problem could be solved simply by passing in `'FT'` flags to negate forks. I'm not sure what would behaviour be in case of `gui` editors that will always make forks.

toonn:
Fixed command is unacceptable. Furthermore, I don't like this fix. We need to rethink `bulkrename` so it can deal with processes that fork.

PR #1058, opened 2018-02-06
Deprecate :bulkrename

### LINKED ISSUE #880: bulkrename doesn't work if open txt file in a new terminal

Vifon:
:bulkrename is now an alias to vidir(1).

Advantages of vidir:
- proper resolution of rename cycles, whereas a cycle in :bulkrename would cause a data loss
- fewer other subtle bugs like this
- less code to maintain
- uses $VISUAL/$EDITOR instead of calling rifle (fixes #880)

Disadvantages of vidir:
- it's a new soft dependency
- it doesn't track the ranger tags of the renamed files, unlike :bulkrename
- it's no longer possible to replace "mv" with another command before executing the scenario (doubtful whether anyone actually would do this)

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1048,81 +1048,25 @@ class chmod(Command):
 class bulkrename(Command):
     """:bulkrename
 
-    This command opens a list of selected files in an external editor.
-    After you edit and save the file, it will generate a shell script
-    which does bulk renaming according to the changes you did in the file.
+    Use vidir(1) to rename the selected files. Please refer to
+    vidir(1) documentation for details.
 
-    This shell script is opened in an editor for you to review.
-    After you close it, it will be executed.
     """
 
-    def execute(self):  # pylint: disable=too-many-locals,too-many-statements
-        import sys
-        import tempfile
-        from ranger.container.file import File
-        from ranger.ext.shell_escape import shell_escape as esc
-        py3 = sys.version_info[0] >= 3
-
-        # Create and edit the file list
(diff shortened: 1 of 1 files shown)


NoSuck:
I always thought `:bulkrename` was commendable but unnecessary.

tkkcc:
vidir or mmv support rename cycles. Some images just accidentally lost using bulkrename. (；′⌒`)

toonn:
That sucks. We should add caveat emptors everywhere, man page, ranger's help, maybe shield bulkrename from use with an option that's disabled by default even. Until it is fixed at least.

PR #1062, opened 2018-02-07
Support cygwin UNC paths

laktak:
Allows users on Wndows/cygwin to browse UNC paths (network shares) in the form of `//server/path`.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
ranger version: ranger-master 1.9.0
Python version: 3.6.3 (default, Oct 31 2017, 19:00:36) [GCC 6.4.0]
Locale: en_US.UTF-8
Windows 10/cygwin

#### CHECKLIST

#### DESCRIPTION
`ranger/core/tab.py` is modified to allow paths starting with `//` (they were reduced to `/` by the old code).

#### MOTIVATION AND CONTEXT
Allows users to browse network shares on Windows/cygwin.

### DIFF
--- a/ranger/core/tab.py
+++ b/ranger/core/tab.py
@@ -144,7 +144,14 @@ class Tab(FileManagerAware, SettingsAware):  # pylint: disable=too-many-instance
         else:
             pathway = []
             currentpath = '/'
-            for comp in path.split('/'):
+            spath = path.split('/')
+
+            # support cygwin UNC paths
+            if path[:2] == '//':
+                currentpath = '//'
+                spath = spath[2:]
+
+            for comp in spath:
                 currentpath = join(currentpath, comp)
                 pathway.append(self.fm.get_directory(currentpath))
             self.pathway = tuple(pathway)


toonn:
Does this work? It looks like you're dropping the servername and sharename from `spath`. How do UNC paths work, is //server a valid path?

laktak:
Yes it works fine in cygwin. Windows UNC paths are usually `\\server\share[\path]` but cygwin uses `/` instead and is also able to `ls //server`.

The code inserts `//` into `currentpath` because the `join()` function would otherwise reduce it to `/server`. Ranger shows it as `server/path` at the top (from `pathway`).

toonn:
Still are you sure you want to slice `spath[2:]`? It looks like you think that'll drop `//` but `spath = path.split('/')` so there's no slashes at all in `spath`.

laktak:
Well the path that is used to list the directory is `currentpath` which will start with `//`. Removing the two empty entries from `spath` does not affect it. 

The difference is how ranger displays the path at the top - it's either `server/path` (current code) or `//server/path` if I remove the `spath = spath[2:]` line. Which do you prefer?

toonn:
Ah, my mistake, of course split leaves empty strings. Still I'd prefer the most accurate path so `//server/...`. Wouldn't `server/...` be confusing since it could be a path relative to the current directory? Even though that doesn't make sense because it would be relative to itself but that's not obvious at first sight. I don't know how windows would display such a path though so if it drops the leading `//` then I'll approve this PR as is.

laktak:
I tried both versions but the one with `spath = spath[2:]` is a lot faster. I haven't yet had time to figure out why though. I guess this could be improved with #1063.

toonn:
Have you had time to look at this? How does widows normally show such paths? Maybe a unc prefix would be even clearer, so `unc://server/...`, like URIs. Might be more confusing if that's not a valid URI though.

laktak:
No, I've been out of town and there aren't many Windows PCs with networks around ...

Anyway I think we should go with `//server/path` as that is what cygwin is using as well (WSL does not support UNC AFAIK and windows cmd.exe uses the `\\server\path` format).

hut:
multiple slashes are collapsed into one slash in linux for some reason. Try `ls /////usr`. With your patch, `ranger //usr` will produce a weird state where the column that normally would display the root directory remains empty. Without the patch, it will just go to `/usr` normally.

If we can fix that, I'm for it.

toonn:
@laktak, pinging because we're getting ready for a release and this could make it in imo.

toonn:
@laktak, why'd you close this? Still seems like a sensible PR to me, as long as unc paths would get a `unc://path` prefix. I'm sure we can come up with a way to workaround the weirdness, maybe by using `\\` paths?

laktak:
Sorry, deleted this by mistake while removing some test branches.

I don't have a Windows test environment at the moment ... so should I leave this open for someone else to pick up?

PR #1067, opened 2018-02-09
Fully encompass tag_toggle

NoSuck:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Linux 4.14.15-1-ARCH #1 SMP PREEMPT Tue Jan 23 21:49:25 UTC 2018 x86_64 GNU/Linux
- Terminal emulator and version: XTerm(330)
- Python version: Python 3.6.4
- Ranger version/commit: bfe1dbafaec8d79cb77fca93d4daa5a98e2cbf29
- Locale: ja_JP.utf8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

The `tag` parameter has been added to `tag_add` and `tag_remove`.

#### MOTIVATION AND CONTEXT

Previously, it was not possible to specify a tag through `fm.tag_add` or `fm.tag_remove`.  I noticed the issue because the `tag_toggle` command is insufficient for using IPC in a remote-like, broadcast fashion (as toggling would result in different output depending on the number of open ranger processes).

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -838,11 +838,11 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
         self.ui.redraw_main_column()
 
-    def tag_remove(self, paths=None, movedown=None):
-        self.tag_toggle(paths=paths, value=False, movedown=movedown)
+    def tag_remove(self, paths=None, movedown=None, tag=None):
+        self.tag_toggle(paths=paths, value=False, movedown=movedown, tag=tag)
 
-    def tag_add(self, paths=None, movedown=None):
-        self.tag_toggle(paths=paths, value=True, movedown=movedown)
+    def tag_add(self, paths=None, movedown=None, tag=None):
+        self.tag_toggle(paths=paths, value=True, movedown=movedown, tag=tag)
 
     # --------------------------
     # -- Bookmarks


toonn:
What exactly are you trying to accomplish with this? It sounds like you're trying to synchronize multiple local ranger instances by using IPC to have all of them toggle a tag. All local processes would use the same `tagged` file afaik so isn't this needless duplication of effort? Why doesn't `fm.tags.sync()` suffice?

NoSuck:
> What exactly are you trying to accomplish with this?

I want to tag files in ranger from an external program.

> All local processes would use the same `tagged` file afaik so isn't this needless duplication of effort?

No.  The external program does not simply write to `tagged`, as files tagged thusly would not immediately show in any running ranger instances.  Rather, the external program “informs” every ranger instance of the newly tagged files by running `tag_add` (or `tag_remove`).

> Why doesn't `fm.tags.sync()` suffice?

Please explain.

toonn:
Oh, so the problem is that the IPC call gets every ranger instance to execute the call? Well that sounds like a problem that should be fixed in the IPC code. This'd be a fragile patch that just hides that insane behavior.

Not saying this patch might not be useful still, just that it doesn't sound like the right solution to the problem.

NoSuck:
> Oh, so the problem is that the IPC call gets every ranger instance to execute the call?

No.  The external program *should* interact with every ranger instance, as every ranger instance shares the same tag information.  The problem is that ranger's `fm` API currently allows for specifying a tag symbol only when *toggling* tags, not when adding or removing them.

The workaround you mention is viable, but even `fm.tags.sync()` requires user input in order to reflect changes.

toonn:
I disagree that IPC should by default be multicast.

What user input does it require? It's what `:reset` now does so it shouldn't require input.
It still looks like using this could cause problems where multiple instances are trying to write to the `tagged` file simultaneously.

toonn:
I merged your changes because they're entirely reasonable, and were unnecessarily held up by my questions. I'd like to continue the discussion though.

NoSuck:
Thank you.

Would printing to `tagged` be recommended as a suitable method for tagging paths via an external program?

`fm.tags.sync()` seems to require keyboard input, for example `move up=1`, in order to reflect changes that have been printed to `tagged` from an external program.  `:reset` does not, perhaps because it reloads the view (moving the cursor to the current directory's first item, etc).  User input for `fm.tags.sync()` could likely be simulated by calling some other function (likely in `reset`).

The tagging issue is just an example, really.  In any case, it does not make much sense to have different sets of parameters for the same family of functions, so I appreciate the merge.

toonn:
You may be right probably these lines:
```
old_path = self.thisdir.path
self.enter_dir(old_path)
```
I imagine just calling `:redraw_window` should be enough.

PR #1068, opened 2018-02-11
Prevent unconditional symlink resolution

NoSuck:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Linux 4.14.15-1-ARCH #1 SMP PREEMPT Tue Jan 23 21:49:25 UTC 2018 x86_64 GNU/Linux
- Terminal emulator and version: XTerm(330)
- Python version: Python 3.6.4
- Ranger version/commit: bfe1dbafaec8d79cb77fca93d4daa5a98e2cbf29
- Locale: ja_JP.utf8

#### CHECKLIST

#### DESCRIPTION

Passing paths to `tag_toggle` now leaves symlinks unresolved, matching behavior when no paths are passed.

#### MOTIVATION AND CONTEXT

Previously, calling `tag_toggle(['/path/to/link'])` would result in ranger tagging `/path/to/original`, whereas highlighting `/path/to/link` and calling `tag_toggle` with no arguments would correctly link `/path/to/link`.

#### TESTING

All tests passed!

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -8,7 +8,7 @@ from __future__ import (absolute_import, division, print_function)
 import codecs
 import os
 from os import link, symlink, getcwd, listdir, stat
-from os.path import join, isdir, realpath, exists
+from os.path import join, isdir, realpath, exists, abspath
 import re
 import shlex
 import shutil
@@ -823,7 +823,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
         if paths is None:
             tags = tuple(x.realpath for x in self.thistab.get_selection())
         else:
-            tags = [realpath(path) for path in paths]
+            tags = [abspath(path) for path in paths]
         if value is True:
             self.tags.add(*tags, tag=tag or self.tags.default_tag)
         elif value is False:


NoSuck:
Steps to reproduce from the shell.  (Substitute CWD for `/path/to` in the ranger `eval` command.)

```mkdir temp ; cd temp ; mkdir orig ; ln -sr orig link ; cd orig ; touch TEST ; cd .. ; ranger```

Enter the `link` directory and execute the following:

```:eval -q fm.tag_add(['/path/to/temp/link/TEST'])```

toonn:
We're thinking we want the opposite, that tagging a link tags the target and all links pointing to the target.

NoSuck:
Why?

toonn:
There's two views, links are cheap but distinct copies of files and links *are* their targets. I'm not convinced either way. @Vifon, another member of the team leans towards the latter. This is still up for discussion, hence my update on where we stand rn.

Tbh I'm thinking more and more maybe tagging needs an option like `find` for whether to follow links or not.

NoSuck:
An option would be fine, but be warned:  Juggling links instead of *paths* is what leads to bugs like #1069.

First of all, if symlinks “are” their targets, then symlinks should also be tagged when their targets are tagged.  This is, of course, infeasible.

It gets hairier.  For example, what happens if a user tags a symlink that is later updated to point to another target?  Would the original target remain tagged?  Would the new target automatically be tagged--even if the symlink was updated outside of ranger?  And what about hard links?  Should a tagged hard link also tag every file system entry with that inode?

There are too many annoying questions for me, but if somebody else wants to answer them all, please at least leave an option for tagging *paths*.  That is all I ask.

toonn:
Those would be non-issues.
1. Only realpaths would be tagged, figuring out which files in a directory are tagged would also be done by mapping realpath over them, so not infeasible, not even hard.
2. The symlink doesn't point to the tagged realpath anymore so it wouldn't be tagged anymore.
3. Hard links would be treated the same.

I'm basically convinced we should mimick find(1) but that does make #1069 blocking for this issue.
Let's [ask for opinions](https://www.strawpoll.me/15673476).

NoSuck:
You said this:

> ... tagging a link tags the target and all links pointing to the target.

Now, you say this:

> Only realpaths would be tagged...

Hence, confusion.

The latter would not be as messy, but it would eliminate the ability to tag links.  At the same time, it could be considered convenient for users who do not want to distinguish between links and link targets.  However, this raises the question of why a link was even made in the first place, if not to be treated differently in some respect.

An option may be appropriate, as both views have arguable merit.  The option would make sense as a configuration value in `rc.conf`, as it would determine display of paths' tags.

NoSuck:
... Semantically equal?

Best of luck.

toonn:
Thanks.

hut:
I agree that an option would be best here

toonn:
Ok, so since we want the resolve links option, #1069 is blocking this. Are you interested in working on this, @NoSuck?

Issue #983, opened 2017-12-03
icat preview image method (suggestion/enhancement)


JackofSpades707:
#### ISSUE TYPE
- Improvement/feature request

#### EXPECTED BEHAVIOR

A picture is worth a thousand words, I've linked to one at the bottom of this post

#### CURRENT BEHAVIOR

Better image rendering within the terminal

#### CONTEXT

Better overall Terminal Experience

#### POSSIBLE SOLUTIONS

Modify the rc.conf setting to use icat

#### IMAGES / VIDEOS

![image](https://user-images.githubusercontent.com/17239750/33530989-18b1d5e4-d83c-11e7-9189-ae1e53b55c85.png)
You may also be interested in these links : 
https://github.com/kovidgoyal/kitty/issues/33
https://github.com/kovidgoyal/kitty/blob/master/graphics-protocol.asciidoc

soredake:
@hut
kitty recently gained support of specifying display region and clearing displayed images
https://github.com/kovidgoyal/kitty/commit/0b3a5a7d4568aacc07096b0acfd992ab3f3e0c25
https://github.com/kovidgoyal/kitty/commit/945613555a1ab413dc37209817936b12342b1cc4

https://github.com/kovidgoyal/kitty/issues/228#issuecomment-352173562

toonn:
@soredake Does icat work on terminals other than kitty, how about inside of tmux? If it does then implementing kitty's graphics protocol for ranger might be worth the effort. So far we've avoided doing image previews with anything more than a program called from `scope.sh` because there just hasn't been a universal method that just works. We don't want to get stuck maintaining some custom hack with similar failings to w3mimgdisplay.

mark-dawn:
I have spent some days implementing this for myself and I got it to work with reasonable performance.
I am using directly kitty's image protocol instead of kitty's icat, which misbehaves with some sizes.
![kittyranger](https://user-images.githubusercontent.com/36431389/36131892-819b9a7a-1074-11e8-80e3-a5cf97e4f380.jpg)
I think there are two main problems with this system: first, I used **PIL/pillow** to resize and convert images, which is not a dependency right now (w3m is external and other terminals accept different filetypes). However it doesn't seem unreasonable to ask for an image library if the user is interested in images, and I was planning of implementing a fallback to ImageMagick too.
[...]

PR #1077, opened 2018-02-16
Kitty image preview

mark-dawn:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- ranger version: ranger-master 1.9.0
 - Python version: 3.6.4 (default, Jan  5 2018, 02:35:40) [GCC 7.2.1 20171224]
- Locale: en_US.UTF-8
- Operating system and version:  Arch Linux (up to date)
- Terminal emulator and version: kitty 0.7.1 and XTerm(331)
- Python version: 3.64 - 2.7.14
- Ranger version/commit:  #0b4f6b2a 

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
- Added support for previews in [kitty](https://github.com/kovidgoyal/kitty.git) using its innate image protocol (similar to terminology or iTerm2). Support come in two flavors, one leveraging the local storage (faster) and one streaming the image alongside the draw commands (slower but network transparent). Manapages and `rc.conf` have been updated to advertise this new option.
- (Minor) Refactored some code in the `img_display.py` module that was used in three subclasses as an helper function.
- Fixed support and docs for terminology: during tests for the above change I notices that support for terminology was broken, independently from the changes (due to a missing `\n`). This has been fixed; also the lack of documentation for the terminology support was fixed too, adding it to both the manpages and `rc.conf`

#### Notes and issues
Since kitty only likes png or raw RGB data `PIL` is required as a dependency. I also plan on implementing an extra backend using the external `ImageMagick` toolkit, like kitty does for his own icat utility. I left it for a later time since this method has been shown to fail occasionally, and is potentially slower since it has to invoke external processes. In any case, `PIL` is only imported at the class level and not at the module level, leaving other options alone.
It is also possible to change `KittyImagePreview` to automagically select between streaming and local file, but due to a bug and my incompetence with `stdin` this feature is delayed, and the choice is made in `rc.conf`

#### MOTIVATION AND CONTEXT
The changes have been motivated by the fact that kitty has native support for images, meaning less edge cases, and in general w3m doesn't work on terminals that refresh independently of input, like kitty. This brings full image preview to kitty, basically.

#983 Relevant discussion

#### TESTING
The new branch has been used both on python2 and 3 (packages installed with pacman are `python-pillow` and `python2-pillow`).

The standard test suite has been run. 

The core of the codebase is mostly untouched, except for intialization of `ImageDisplay` classes in `fm.py` naturally. The `_get_image_displayer` function has been updated with the relevant branches and excluded from pylint `too-many-return-statements` since ""fixing"" the warning would amount to add a local variable and make the function one line longer and potentially slower.

The other touched part is obviously `img_display.py` with most changes being orthogonal to the original code, except for the excision of a small common helper function to temporarely move the cursor (now wrapped nicely as a context manager). Terminology, which had the most complex use case has seen no change to its (buggy and now fixed) behavior. I couldn't test iTerm2 graphically but the text output looks fine.

#### IMAGES / VIDEOS
kittyranger
kitty showing images for more than half a millisecond (you will have to believe me on the persistence of the output)


doc/ranger.pod
```diff
@@ -205,6 +205,11 @@ This feature relies on the dimensions of the terminal's font.  By default, a
 width of 8 and height of 11 are used.  To use other values, set the options
 C<iterm2_font_width> and C<iterm2_font_height> to the desired values.
 
+=head3 terminology
+
+This only works in terminology. It can render vectors graphics, but works only locally.
```
toonn: Small typo "vector graphics".

doc/ranger.pod
```diff
@@ -221,6 +226,21 @@ window.
 
 To enable this feature, set the option C<preview_images_method> to urxvt-full.
 
+=head3 kitty
+
+This only works on Kitty.
+It requires PIL or pillow at the moment to work. A nasty bug that can
+corrupt the terminal window when scrolling quicly many images, that can
```
toonn: Typo "quickly through". "by with" drop one.

ranger/config/rc.conf
```diff
@@ -85,13 +85,32 @@ set preview_images false
 #   width of 8 and height of 11 are used.  To use other values, set the options
 #   iterm2_font_width and iterm2_font_height to the desired values.
 #
+# * terminology:
+#   Previews images in full color in the terminology terminal emulator.
+#   Supports a wide variety of formats, even vector graphics like svg
```
toonn: End the sentence with a '.' : )

ranger/config/rc.conf
```diff
@@ -85,13 +85,32 @@ set preview_images false
...
 #   Preview images in full color using urxvt image backgrounds. This
 #   requires using urxvt compiled with pixbuf support.
 #
 # * urxvt-full:
 #   The same as urxvt but utilizing not only the preview pane but the
 #   whole terminal window.
+#
+# * kitty:
+#   Preview images in full color using kitty image protocol
+#   (https://github.com/kovidgoyal/kitty/blob/master/graphics-protocol.asciidoc),
```
toonn: Maybe move the url to a footnote?

ranger/config/rc.conf
```diff
@@ -85,13 +85,32 @@ set preview_images false
...
+#   Requires python PIL or pillow library.
+#   In experimental stage: tmux support is untested, and a scrolling too fast a folder with many images may glitch ranger;
```
toonn: Needs a rephrasing. "and scrolling too fast through a folder..."

ranger/config/rc.conf
```diff
@@ -85,13 +85,32 @@ set preview_images false
...
+#   Future improvements to kitty will ameliorate this issue, for now call the command 'redraw_window' to get rid of the garbage.
+#
+# * kitty-network
+#   Similar to base kitty, bit instead of local storage it uses kitty's protocol special feature to
```
toonn: "bit" -> "but"

ranger/config/rc.conf
```diff
@@ -85,13 +85,32 @@ set preview_images false
...
+#   stream the whole image over standard input. More error prone, and more intensive since it scales down images,
+#   producing also worse quality previews.
+#   However it makes possible to see previews froma ranger instance over the network,
```
toonn: "it makes _it_ possible to see previews _from_ ranger over the network" (emphasis for clarity not intended as emphasis)

ranger/container/settings.py
```diff
@@ -98,7 +98,8 @@
     'confirm_on_delete': ['multiple', 'always', 'never'],
     'line_numbers': ['false', 'absolute', 'relative'],
     'one_indexed': [False, True],
-    'preview_images_method': ['w3m', 'iterm2', 'urxvt', 'urxvt-full'],
+    'preview_images_method': ['w3m', 'iterm2', 'urxvt',
```
toonn: Still missing terminology setting?

ranger/core/fm.py
```diff
@@ -234,6 +236,10 @@ def _get_image_displayer(self):
             return URXVTImageDisplayer()
         elif self.settings.preview_images_method == "urxvt-full":
             return URXVTImageFSDisplayer()
+        elif self.settings.preview_images_method == "kitty":
+            return KittyImageDisplayer()
+        elif self.settings.preview_images_method == "kitty-network":
+            return KittyImageDisplayer(stream=True, resize_height=480)
```
toonn: Shouldn't the height be determined dynamicall?

ranger/ext/img_display.py
```diff
@@ -35,6 +38,29 @@
     '/usr/local/libexec/w3m/w3mimgdisplay',
 ]
 
+# Helper functions shared between the previewers (make them static methods of the base class?)
+
+
+@contextmanager
+def temporarly_moved_cursor(to_y, to_x):
```
toonn: Typo "temporarly".

mark-dawn:
Oh, wow, those are some embarrassing-grade typos. I even left in notes about the scrolling bug when it has been solved...I should have double checked this stuff. Thanks for the rundown.

>Shouldn't the height be determined dynamicall?

*dynamically :P  Jokes aside, that's a very bad name when taken out of context. It is not related to the viewbox rectangle, but it is the max tolerated height before the image gets scaled down internally to ease the transfer. I should just base it off the actual area now that I think of it.

toonn:
Even though it's technically not referring to the viewport it would seem to me like it's definitely dependent on the viewport. If you're scaling down anyway it doesn't make sense not to scale down all the way to a small viewport. Having a hard cap makes some sense but should imo be based on (10Gb) LAN so not a very hard cap ; )

mark-dawn:
Technically you are half right, but the issue is not only bandwidth, but also the fact that the payload in the APC is base64 encoded. Encoding a filename is easy, encoding a whole big image in RBG(A) is a minute worth of time, at least with the basic python implementation. I thought it would also benefit slow hdds when saving the temporary image when not in network mode.
However, turn out this point is moot, since I am saving to a temporary file, which python correctly creates in `/tmp`, so they are mapped to RAM! Indeed I am testing this on my laptop which has a slow hdd and I don't see any difference  when I get rid of the resize.

So for now I am going to execute the resize only in the case of `kitty-network`, but in a future commit I want to implement what you suggested, scaling down the image to the best viewport I have. Since kitty fills the viewport (which is expressed in columnsxrows) stretching the image (with a nearest-neighbor filter) you can get some artifacts despite my best efforts. If I could resize the image with a better filter the effect would be nicer, but for that I will have to find out the size in pixel of a character cell.

toonn:
Thank you for all the work on this. I'm waiting for a couple kitty users to test this. Or some more maintainer eyes on the patch, since it's significant.

tkerber:
I've been testing this PR for personal use and have noticed the following:

* When using python2 and ranger, no images display, stack traces with `stdout.buffer` not existing get outputted.
* The error message for PIL not being found is undescriptive on its own, and doesn't terminate in a new line.
* Images smaller than the preview area get scaled up, leading to unwanted artifacting.
* SVG images fail to display.

mark-dawn:
Thanks @tkerber for the tests!
> - When using python2 and ranger, no images display, stack traces with stdout.buffer not existing get outputted.

My mistakes, and this actually makes sense. What it doesn't make sense is that I tested and it worked for me _before_. Probably a typo when calling `python2`. Anyway it is an easy fix.

>  - The error message for PIL not being found is undescriptive on its own, and doesn't terminate in a new line.

I actually had no idea on how to show an error message, I thought that exception were silently ignored, but it turns out that if you have a message attached it gets printed in red to the status bar. I am moving it there in the next commit, possibly adding something like "please search your repositories for the python library pillow" (suggestion welcomed).

> - Images smaller than the preview area get scaled up, leading to unwanted artifacting.

This is "by design" in a way, but I do agree that it looks kinda of meh. I didn't notice because I was testing with a medium sized terminal in a folder with big images. Anyway since I was going to implement resizing the thumbnail to the viewport size (with some decent filter) for downsizing it should made upsampling a bit better. However I do agree that it might be good to want images sized as close as possible to the native size, so would it make sense to implement a config option to switch between this different behaviors?  It could possibly be back-ported to other protocols, too.
I would like some opinions on this.

> - SVG images fail to display.

svgs are just too different to be read by pillow. They fail on w3m too, and the only terminal that supports them seems to be terminology. Support might be coming if there is a simple way to generate a raster image from them. Maybe if the user has inkscape installed it could be possible to generate a png in a reasonable timeframe? Alternatively a native library could be the solution. This could also benefit other protocols, too.
TL; DR: if you make use of a lot of svg in your workflow I suggest to switch to the terminology, because they are not coming anytime soon™

toonn:
I'd make the error message something along the lines of "Image previews require PIL".

Having the scaling be configurable would be good with at least three options: no scaling, downscaling only and both. Backporting would be ideal but if that requires PIL it'd definitely have to be optional like most of ranger's dependencies, silently falling back to sane behavior if it's not available.

SVG support for other methods would also be great if the performance is good enough but this is definitely an enhancement, not of the highest priority. These last two things would be seperate PRs of course.

@tkerber, thank you for testing. Image previews are probably the most common reason for issue reports so we want to be careful adding new methods lest they introduce even more issue reports : )

kovidgoyal:
Just dropping in with some tips, that might be helpful:

1) You can display SVG images using rsvg-convert (that is how kitty icat itself does it)

2) Rather than creating files in /tmp (/tmp is not always mounted in RAM) you should either create files in /dev/shm, or if that does not exist (for example on non-linux platforms) use the kitty image protocol support for POSIX shared memory directly (see man shm_overview and the kitty graphics protocol page for how to use it). Though I dont know if that is possible to do using only python+ctypes or not. 

3) You can get the exact pixel size of a character cell trivially using termios, see https://github.com/kovidgoyal/kitty/blob/master/graphics-protocol.asciidoc#getting-the-window-size and icat.py inkitty for code to do it in python. Once you have that, it is perfectly possible to draw images using the graphics protocol avoiding all scaling. 

And let me finish by saying that it is cool to see the kitty image protocol being used in the wild. I am happy to take suggestions for improving the protocol based on actual usage.

doc/ranger.pod
```diff
@@ -229,14 +229,11 @@ To enable this feature, set the option C<preview_images_method> to urxvt-full.
 =head3 kitty
 
 This only works on Kitty. It requires PIL (or pillow) to work.
+It is able to automatically work over network,
```
toonn: Nitpick: I'd stick closer to the original phrasing. "Allows remote image previews, for example in an ssh session." I'm not sure the performance implications should be in the manpage but I could be convinced otherwise.

mark-dawn:
@kovidgoyal thanks for the tips! I think I am going to leave svgs to a later commit, but this is helpful to know. Talking about `/tmp` I am using in the sense that `NamedTemporaryFile` saves there by default on my system.  For know I will keep it simple and puth my faith in python choosing a suitable location.

@toonn Noted!

Also I found out that kitty display images in their native size if a rectangle is not specified, so the next commit will solve all issues about resizing artifacts.

mark-dawn:
I got working previews without artifacts. Smaller images are not scaled at all, while big images are fit in the best way to the preview pane. It was deceivingly simple to implement, contrary to my belief!

Anyway if the test from other users go well and everything else code-wise is fine, I would say this is a great point for merging, since future improvement are kinda worth they own pull request.

ranger/container/settings.py
```diff
@@ -98,7 +98,7 @@
     'confirm_on_delete': ['multiple', 'always', 'never'],
     'line_numbers': ['false', 'absolute', 'relative'],
     'one_indexed': [False, True],
-    'preview_images_method': ['w3m', 'iterm2', 'urxvt',
+    'preview_images_method': ['w3m', 'iterm2', 'terminology', 'urxvt',
                               'urxvt-full', 'kitty', 'kitty-network'],
```
toonn: Woops, looks like `kitty-network` snuck its way back in.

doc/ranger.pod
```diff
@@ -205,6 +205,11 @@ This feature relies on the dimensions of the terminal's font.  By default, a
 width of 8 and height of 11 are used.  To use other values, set the options
 C<iterm2_font_width> and C<iterm2_font_height> to the desired values.
 
+=head3 terminology
+
+This only works in terminology. It can render vector graphics, but works only locally.
+To enable this feature, set the option C<preview_images_method> to terminology.
```
toonn: Grammar nazi patrol : )
Missing new line before this one. Don't actually care about whether there's one or not but we should keep it consistent.

ranger/config/rc.conf
```diff
@@ -85,13 +85,26 @@ set preview_images false
...
 #
 # * urxvt-full:
 #   The same as urxvt but utilizing not only the preview pane but the
 #   whole terminal window.
+#
+# * kitty:
+#   Preview images in full color using kitty image protocol.
+#   Requires python PIL or pillow library.
+#   If ranger does not share the local filesystem with kitty
+#   the transfer method is switched to encode the whole image in
```
toonn: I'd drop "in the protocol that" then add "this" before allows. Something about the sentence seems off otherwise.

Lompik:
image previewed correctly on kitty as well here. Thanks !

NeoTheFox:
Got it working on my setup without any issues, it works, I've tried some edge cases and so far it works really well!

soredake:
Tested on gentoo, working good.

doc/rifle.1
```diff
@@ -129,7 +129,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RIFLE 1"
-.TH RIFLE 1 "rifle-1.9.0" "2018-01-25" "rifle manual"
+.TH RIFLE 1 "rifle-1.9.0" "02/16/2018" "rifle manual"
```
FichteFoll: Ah, my precious ISO dates NotLikeThis

doc/ranger.pod
FichteFoll: All other sections have this line in a separate paragraph.

doc/ranger.1
```diff
@@ -317,6 +323,23 @@ The same as urxvt but utilizing not only the preview pane but the whole terminal
...
+.IX Subsection "kitty"
+.PP
+This only works on Kitty.
+It requires \s-1PIL\s0 or pillow at the moment to work. A nasty bug that can
+corrupt the terminal window when scrolling quicly many images, that can
+be solved by with the \f(CW\*(C`redraw_window\*(C'\fR command.
+.PP
+To enable this feature, set the option \f(CW\*(C`preview_images_method\*(C'\fR to kitty
+.PP
+\fIkitty-network\fR
```
FichteFoll: Seems like this file needs to be regenerated because this section doesn't exist in the `ranger.pod` anymore (there are also some typos above).

mark-dawn:
@Lompik I have no control in what is considered an image; the interface I implemented just opens and display whatever file ranger asks, so if something that is not an image is getting sent to it,  the problem must lie in an incorrect preview script or ranger itself.
About exceptions I don't have to explicitly catch them since ranger does it for me (as long as `-d` is not specified) and displays the associated message in the status bar. You are seeing a crash because `PIL` is throwing two exception at once, which makes python terminate anyway. This actually seems a bug in `PIL` itself, since `OSError` looks like the exception that  `PIL` is trying to pass up the stack, but it raises it before having handled the original error. Anyway, I cannot get `PIL` into the same situation using non-images files (it indeed produces a single `OSError` exception) so it must be an edge case when the file resemble in some way a malformed image file.

@FichteFoll whops, I forgot to run `make man`. The dates are not completely my fault though, the Makefile is using `date +%x` to generate them instead of `date +%F` or `date --iso-8601`. Should I make another pull request to fix this?

Thanks everyone for testing this

mark-dawn:
- Updated the manpages.
- implemented docs corrections.
- I have silenced a warning pillow would output on very big images. In the future, it may be worth introducing a configuration option to toggle between hidden and actually stopping the image from loading. But at least for now it doesn't spill on the screen like before.
- tried to implement support for `tmux` and failed miserably, ranger will always freeze no matter what I try. For now I simply shutdown the image preview by raising the proper exception, so ranger is still usable if is used with `tmux` during an emergency tty-only session, while on the desktop folks can use the built-in tabbing/windowing capabilities of kitty itself.
   - changed `pager.py` so when a ImgDisplayUnsupportedException is caught from the image dispalyer, not only the previews are shut down but the error message is also printed to the statusbar like for other exceptions.

toonn:
@Lompik tried to leave a comment yesterday, guess I failed to click the button. Thank you for that report,  lead us to finding a circular reasoning bug I have yet to make a proper issue out of it.

@mark-dawn too bad concerning tmux but I'm not sure any of the preview methods currently supports it, maybe urxvt and terminology. And possibly mpv in the near™ future.

throwup:
I have one (possibly stupid question): Is it expected that the displayed picture is always vertically centered?
If so, do you plan to add an "alignment" feature? I often have terminals filling up my whole vertical screen space and find the centered preview uncomfortable to look at.

throwup:
I also seem to get this error message whenever I select an image file, however ranger does not actually crash:
```
ranger version: ranger-master 1.9.1
Python version: 2.7.14 (default, Sep 16 2017, 17:49:51) [GCC 7.3.0]
Locale: None.None
Current file: '/home/leo/2018-02-23-165447_1366x768_scrot.png'

  File "/nix/store/2xm11w6pbyj2zscd4car31pjcah80bj4-ranger-19e8039/lib/python2.7/site-packages/ranger/core/main.py", line 196, in main
    fm.loop()
  File "/nix/store/2xm11w6pbyj2zscd4car31pjcah80bj4-ranger-19e8039/lib/python2.7/site-packages/ranger/core/fm.py", line 418, in loop
    self.image_displayer.quit()
  File "/nix/store/2xm11w6pbyj2zscd4car31pjcah80bj4-ranger-19e8039/lib/python2.7/site-packages/ranger/ext/img_display.py", line 666, in quit
    os.remove(self.temp_paths.pop())
OSError: [Errno 2] No such file or directory: '/tmp/rgr_thumb_PufqdM.png'

```

toonn:
@throwup I'm not sure how the other image preview methods handle it but that's actually a good suggestion. Would probably need to be configurable since there's multiple options and not everyone'll like the same thing: centered, always in the upper left corner of the preview column, left aligned and centered as much as possible around the highlighted file, etc.

FichteFoll:
w3m is in the upper left corner.

mark-dawn:
I actually implemented  the centering because I liked it more, but I do agree that is inconsistent with w3m (but not with some other previewers). I can comment out the two lines that take care of it for now, but I do agree that it would be nice to have an options to set it. However this would need a change in the interface between `ImageDisplay` classes and ranger itself, since they should be independent from the fm itself. It could be just an argument during initialization.
Anyway I would move this to a new pull request, since it would require to adapt all the other `ImageDisplay` classes. I think I could do it for w3m, kitty and terminolgy easily, but no guarantee for the others.

@throwup whops, fixed. Python2 throws a different exception when the file is missing.

aenda:
I tested on Arch Linux 4.15.15-1 and everything works for me! Not sure if I missed a comment on this above but I could not get it to crash at all, even when scrolling very rapidly through images (I had seen that mentioned as a bug previously). Any updates on merging this?

Also, I was looking at the kitty APCs again, and believe that the dependency on Pillow could be dropped by using the c and r keys to scale the image within kitty. I also was looking at ImageMagick integration, but that would just require another import/library installation to work, so I disregarded it. I'm still looking at how resizing on kitty's end rather than ranger's might work - let me know if it seems viable and I will continue working on it!

noctuid:
> too bad concerning tmux but I'm not sure any of the preview methods currently supports it, maybe urxvt and terminology. And possibly mpv in the near™ future.

w3m image preview works fine with tmux. The images span splits, but they show up.

toonn:
Re merging: This'll be included in the next release, which is coming soon™ (weeks, not months).

mark-dawn:
@aenda yup, the crashing was fixed. Imagemagick would be just an alternative dependency since it ends up installed on most system anyway

Re: tmux. I did this some time ago, I think I said it was "broken" because the images were actually spanning the splits....I can take another look at that

crocket:
I built this pull request on Gentoo Linux and tested it on kitty-0.8.4.
It works a lot better than w3m on alacritty, st, and rxvt-unicode.
It seems persistent and reliable.

crocket:
Kitty is also waiting for a new keyboard API in glfw. https://github.com/glfw/glfw/issues/1140
Kitty doesn't support ibus through XIM.

If you want to use an input method in ranger right now, you can't use kitty for now.

mark-dawn:
@crocket ok, this is weird, I maged to reproduce the issue....but only a single time. After that the preview just disappears when pressing `r`, but the prompt behaves. This also seem to  persist between instances of ranger and kitty just fine. Does it keep happening on your end?

toonn:
I didn't know input methods needed support from programs. I thought the point was they didn't? It's good you've found a satisfying solution but let's not call it previews, you're just viewing the images at this point : )

mark-dawn:
crocket My bad, I am also getting the effect, it just takes a second to show up. Hopefully is not some weird bug between curses and kitty. I'll try to dig out the open_with code later and see if I find what's going on with it. I wonder if other terminal emulators with built-in image support get the same effect....

aenda:
@MindTooth Merge conflicts are trivial (dates in documentation iirc). Here's my diff from origin/master with image preview included - you should be able to just `patch -p0 < image-preview.txt`. (github is picky about file extensions and won't accept .patch)
[image-preview.txt](https://github.com/ranger/ranger/files/1976869/image-preview.txt)

mark-dawn:
@MindTooth I rebased everything on the last commit, tell me if the docs look alright

toonn:
@mark-dawn :+1: I'm gonna try to make time next week to review your awesome contribution properly and get it merged in master. I was trying to avoid merging such significant changes because I'm the whippersnapper maintainer but I'm just gonna take the responsibility. Worst case we can always roll master back. Just wanted to say thanks for not getting discouraged by the long wait.

MindTooth:
@mark-dawn:  Best case, we have images inside kitty 😇

doc/ranger.1
```diff
@@ -702,7 +717,7 @@ in ranger.
 .IP "automatically_count_files [bool]" 4
 .IX Item "automatically_count_files [bool]"
 Should ranger count and display the number of files in each directory
-as soon as it's visible?  This gets slow with remote file systems.  Turning it
+as soon as it's visible?  This gets slow with remote file sytems.  Turning it
```
MindTooth: A leif in you help text.

MindTooth:
Also, on macOS, maybe we need to state that you need to install the pillow library via the system Python.  Did not work with the Homebrew ones.

```python
sudo /usr/bin/easy_install pillow
```

mark-dawn:
@MindTooth Dang, I am bad at grammars.
I don't have access to MacOS, I guess that the availability of pillow depends if the homebrew library can be accessed by the python that is used for ranger, so it may depend on the system configuration.

@toonn I don't mind the wait, this is my first major contribution to a big public project, so I'd rather have a through review myself. 

Finally I had a look at the `open_with` menu issue, and it actually happens with any menu (like the one invoked with `z`), but it seem to exist with other methods (at least `w3m` exhibits the same effect). So the ideal way would be to disable the image rendering when invoking a menu, but I need to dig into the cod for that.

doc/ranger.1
toonn: That wasn't introduced by @mark-dawn I"m pretty sure. It's in master, and I already patched it in one of my PRs.

Rom1deTroyes:
Le dimanche 06 mai 2018 à 22:53:22 (+0000), Toon Nolten a écrit :
> toonn commented on this pull request.
> 
> > @@ -702,7 +717,7 @@ in ranger.
> -as soon as it's visible?  This gets slow with remote file systems.  Turning it
> +as soon as it's visible?  This gets slow with remote file sytems.  Turning it
> 
> That wasn't introduced by @mark-dawn I"m pretty sure.

Seems to be... 2 months ago in an other PR (-;
https://github.com/mark-dawn/ranger/blame/82a9ed96671e62bd69536d42c393351b1f22e824/doc/ranger.1#L720

> It's in master, and I already patched it in one of my PRs.

"Seuls ceux qui ne font rien ne font pas d'erreurs !"

BTY, thanx to all of you for the great job you're doing !

++
    Romain
-- 
Brave devant l'ennemi, lâche devant la guerre,
telle est la devise des vrais généraux.
	-+- Jean Racine -+-

toonn:
Hah, thanks @Rom1deTroyes. I deleted that comment pretty quickly because I started doubting myself so your confirmation is helpful. :+1:

ranger/ext/img_display.py
```diff
@@ -465,3 +464,205 @@ def _get_sizes(self):
...
+                    central_blk + b',m=1;' + payload_blk + \
+                    self.protocol_end
+            yield self.protocol_start + \
+                central_blk + b',m=0;' + payload + \
+                self.protocol_end
+        else:
+            yield self.protocol_start + central_blk + b';' + self.protocol_end
+
+    @staticmethod
+    def _resize_max_area(image, max_area, img_filter):
```
toonn: Hmm, is `max_area` the right thing to scale to? You could have a thin strip with a small area but a length that's out of "bounds."

ranger/ext/img_display.py
```diff
@@ -465,3 +464,205 @@ def _get_sizes(self):
...
+        with NamedTemporaryFile() as tmpf:
+            tmpf.write(bytearray([0xFF] * 3))
+            tmpf.flush()
+            for cmd in self._format_cmd_str(
+                    {'a': 'q', 'i': 1, 'f': 24, 't': 'f', 's': 1, 'v': 1, 'S': 3},
+                    payload=base64.standard_b64encode(tmpf.name.encode(self.fsenc))):
+                self.stdbout.write(cmd)
+            sys.stdout.flush()
+            resp = [b'']
+            while resp[-1] != b'\\':
```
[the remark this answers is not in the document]
mark-dawn: You are right, in general that part was a bit messy to work around differencies in string handling between python 2 and 3, but I found a way to express it much more cleanily that works in both using bytestring concatenations only instead of lists.

ranger/ext/img_display.py
```diff
@@ -465,3 +464,205 @@ def _get_sizes(self):
...
+        except ImportError:
+            raise ImageDisplayError("Images previews in kitty require PIL (pillow)")
+            # TODO: implement a wrapper class for Imagemagick process to
+            # replicate the functionality we use from im
+
+        # get dimensions of a cell in pixels
+        ret = fcntl.ioctl(sys.stdout, termios.TIOCGWINSZ,
+                          struct.pack('HHHH', 0, 0, 0, 0))
+        n_cols, n_rows, x_px_tot, y_px_tot = struct.unpack('HHHH', ret)
+        self.pix_row, self.pix_col = x_px_tot / n_rows, y_px_tot / n_cols
```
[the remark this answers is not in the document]
mark-dawn: I guess it works out since it is supposed to be an int in the end, but this numbers get further divided down the line and there a float is actually needed, so I do not exactly know if having them to be ints here makes sense. I guess that for logical typing it is more correct anyway, so I am going to make the change.

ranger/ext/img_display.py
```diff
@@ -465,3 +464,205 @@ def _get_sizes(self):
...
+            payload = base64.standard_b64encode(
+                bytearray().join(map(bytes, image.getdata())))
+        else:
+            # put the image in a temporary png file
+            # t: transmissium medium, 't' for temporary file (kitty will delete it for us)
+            # f: size of a pixel fragment (100 just mean that the file is png encoded,
+            #       the only format except raw RGB(A) bitmap that kitty understand)
+            # c, r: size in cells of the viewbox
+            cmds.update({'t': 't', 'f': 100, })
+            with NamedTemporaryFile(prefix='rgr_thumb_', suffix='.png', delete=False) as tmpf:
```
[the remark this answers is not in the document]
mark-dawn: Even PNGs get resized most of the time in my experience, so an in termediate file is generated in most cases. I do not think that the extra logic and code bloat deriving from it would justify the very few cases were a resize is not necessary.
It would maybe be worthy in space-saving optic to change a bit the behavior of the caching for the thumbnails. To put it bluntly, it doesn't work for now. The idea was to keep around those temporary files so only the first time visiting a big image the user would experience a slowdown, and then it would be smooth sailing for the rest of the session. On low end hardware this would be even more important. However, I am not currently checking if the filename already exist, so the calculations are repeated **and** the files stick around until ranger is closed. I will see if I can get it to work properly, and add some extra logic in the `clear` method to clean up old cached files if they start accumulating.
If it takes too much time with the next release approaching disabling the chaching is actually trivial, I can make `NamedTemporaryFile` actually delete the file at the end of the `with` block and move all the file sending logic inside said block.

ranger/ext/img_display.py
mark-dawn: Actually this methos is unused and survived my refactoring by luck.  It is gone now :stuck_out_tongue_winking_eye:  . The actual resizing logic is on line `576`.

mark-dawn:
@toonn I fixed most of the trivial stuff, and commented on some others issues. I was planning to submit the new revision soon™, but I am getting some errors during testing due to #1146. Curiosly it is in older code, so I am unsure if it makes sense to fix it separately in another pull request or to just fix them in this one. I guess the first makes more sense, but should I make a new pull request or I can somehow add commits to #1146?

On this last issue I think a piece of my code cheats a little. Basically I have one instance of similar code, but flake8 is failing to catch it because I used `\`. Something like this:
```
foo + \
bar + \
baz```
Personally, while I think the new style is horrible, I would rather not avoid adhering to the style of the rest of the codebase using some loophole. Should I refactor this code too or is it actually fine, and should I file some bug report to the flake8 guys?

**Comment by mark-dawn** [comment]
So I actually got the caching to work, but I am still unsure if I want to keep it in.
A quick rundown:
- Caching:
    -  pros:
        1. less computation
        2. faster browsing when moving up and down repeatedly
   - cons:
        1. if ranger crashes badly, thumbs won't be deleted until the next machine restart (or anything else that wipes `/tmp`)
        2. is making the code way more complex that it needs to be
        3. doesn't really help the first time visiting a file, which I is the most likely case for my browsing patterns
        4. might need extra config options to configure the max cache size
        5.  the max size invalidates benefits pretty easily
- No Caching:
     - pros:
        1. code is simpler
        2. more reliable, kitty is tasked with deleting the file and this happens on read
    - cons:
        1. visiting a very big file multiple times will slow down ranger each time
        2. probably less "green", which might be a concern for battery bound users

This said, @whoever_tested_this can you say if slowdown was perceptible with a large part of your files? Do you often revist images multiple times during everyday operations?

**Comment by noctuid** [comment]
> if ranger crashes badly, thumbs won't be deleted until the next machine restart (or anything else that wipes /tmp)

Unless the user changes the cachedir or uses tmpfs for the default cachedir, they'll be kept forever. I don't think that the caching with the w3m method has a max size (I checked just now, and there were over 6k thumbnails for me). Since you've already implemented it, I think it might be fine to keep the code and just do it the same way as the default method (no max size; maybe optionally disableable). I don't have a strong opinion either way.

> if ranger crashes badly, thumbs won't be deleted until the next machine restart (or anything else that wipes /tmp)

I'd actually prefer if the cache wasn't deleted when ranger exits.

> more reliable, kitty is tasked with deleting the file and this happens on read

Is this file created in `/tmp`?

**Comment by noctuid** [comment]
As for image preview in tmux, I tried it and it looks like this line in `_late_init` causes an infinite loop:
```
while resp[-1] != b'\\':
    resp.append(self.stdbin.read(1))
```

File: ranger/ext/img_display.py (hunk 20 above)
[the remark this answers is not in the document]
Comment (toonn) [reply ranger/ext/img_display.py]: I was thinking it might be more accurate since fractional pixels don't make sense but I agree maybe the conversion isn't worth it.

File: ranger/ext/img_display.py (hunk 21 above)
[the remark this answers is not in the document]
Comment (toonn) [reply ranger/ext/img_display.py]: If we hold off on caching I'd add the removal at the end of `NamedTemporaryFile`. I didn't consider PNGs would also have to be processed anyway.

**Comment by toonn** [comment]
@mark-dawn, keep to the old style for now, don't worry about the test failures, new style fails with the dual warning. You can add the proper ignore in `.flake8` for now. You're the first to prefer the old style btw.

About caching, I think it'd be nice but if you feel it's shoehorned in maybe we hold off on it until the next release. It would be awesome if we could figure out a preview method independent cache.

**Comment by mark-dawn** [comment]
@noctuid Regarding tmux, the thing doesn't get stuck in an infinite loop, it's actually stuck on the read call. Tmux seem to not be able to pass through communications from the terminal. Frankly I expected unconventional APCs to pass through wrapped in tmux "guards", like I wrap my outgoing commands. Maybe kitty need to add support for them (kind of backwards IMHO), or maybe tmux was never intended to do that and it would be worth to fire a feature request toits maintainers.
In any case I can add a temporary workaround for now, but I would like the issue to be solved in the long run, since it interferes with detecting the best transmission method and will sort of conflict with detecting the absence of kitty, which causes the same block and I wanted to solve before release. It would annoying to freeze ranger because you forgot to turn off image previews when working outside the X-server, for example.

**Comment by mark-dawn** [comment]
@noctuid Regarding caching, having a stable cache outside `tmp` is way more complex than it sounds, and probably worthy of an entire feature request of its own. For example now the temp files are exactly what the name implies, and are supposed to disappear after a single session. To be honest, the current implementation is already "broken" since the thumbnail that are saved are actually the final "display size", which is dependent on the ranger window size. So resizing the window should be enough to invalidate the current cache.

A proper stable cache need to have a stable place, an organization to keep track of which files still exist and which need new thumbnails, integration with ranger own indexer, which size is best for the thumbnail, if it is worthy to copy images smaller than this size, decision on how it interacts with other image previewers (which don't need it as much as kitty) and other previews in general, and a whole new range of options for the user to regulate its workings, since it taxes storage space and indexing times when accessing folders. And all of this for something that in my testing seem to not really have that big of an impact, except with >4K-tier sized images. However, my weakest machine is a laptop only 5yo, so people with X220 might find this a more pressing issue.
Regarding the cache of w3m that seem to be handled entirely by w3mimg, and frankly it might be even counterproductive, if for example w3mimg relies on the w3m browser to delete those when history gets deleted. After all, it is intended for a browser, so those might be full sized copies and not thumbnails, they might not get used if on our side we are not specifying to do so and are probably sticking around forever. How many of those 6K files do still exist on your machine, for example? How many are on mine, since I didn't know of this feature? I get that storage space is cheap, but one of the main advantage of linux is less (invisible) bloat, and I would like to give users control on what gets saved, especially if it is going to stick around.

For now I think I am going to revert the changes on my end, since it needs way more refinement than it currently has.

**Comment by noctuid** [comment]
It sounds reasonable to not add caching then (at least for now).

> And all of this for something that in my testing seem to not really have that big of an impact, except with >4K-tier sized images. However, my weakest machine is a laptop only 5yo, so people with X220 might find this a more pressing issue.

I have a *ton* of high quality images and use ranger regularly for browsing larger images. Any images in the 2-5mb+ range takes 1-3 seconds to preview (and I'm using a thinkpad p50). This include some 1080p images which take half a second to a second to show for me. The w3m image preview method is significantly faster for these files (close to instant or at least well under half a second) without caching. If there's not some other optimization that can be done, I think caching could potentially be a really useful feature. Are the images currently created in `/tmp`? Because if they aren't I'd like to have them created in tmpfs/RAM.

> How many of those 6K files do still exist on your machine, for example? How many are on mine, since I didn't know of this feature? I

Probably most of them. I only found out yesterday that I had so many cached images from ranger. My `.cache/ranger` folder was only ~500mb which isn't that bad considering the number of images, but it was suprising/annoying. I agree that if caching is ever added it should be configurable (and maybe off by default).

**Comment by mark-dawn** [comment]
@noctuid 
>Probably most of them. I only found out yesterday that I had so many cached images from ranger. My .cache/ranger folder was only ~500mb which isn't that bad considering the number of images, but it was suprising/annoying. I agree that if caching is ever added it should be configurable (and maybe off by default).

Strangely enough, this folder was completely empty for me. Did you use by chance any different preview method, like libcaca's ascii art?
>Are the images currently created in /tmp? Because if they aren't I'd like to have them created in tmpfs/RAM.

They are for me, but according to python docs it might be a different folder. Can you see if you run
```
python -c 'from tempfile import NamedTemporaryFile; print("test_me", file=NamedTemporaryFile(prefix="search_me", mode="w", delete=False))'
```
you get in `/tmp` a file named `search_me$somegarbage`? In the worst case I can create them in `/tmp` by myself.
>I have a ton of high quality images and use ranger regularly for browsing larger images. Any images in the 2-5mb+ range takes 1-3 seconds to preview (and I'm using a thinkpad p50). This include some 1080p images which take half a second to a second to show for me.

ok, you made me notice something _really_ weird. See, I was considering "big" the image in my home folder that takes 3'' to load. Turns out that's only 1.5MB! However, I have several images in the range 2-10MB that are instantaneus, and stuff in ~30MB area that barely take 1' to load. The only difference I see is that the image in my home folder has been exported froma svg, and has a lot of transparency. However I found in my folder another couple of images with transparency that are the same size and take less (but are a bit slower than normal).
Can you tell me if your slow images are transparent? Also which filesystem are they in? Are they also pngs? Could you possibly upload one for testing? I wonder if it has to do with the built-in compression in pngs... :thinking: 
In any case it is mostly a problem with pillow, since if I remeber correctly is written mostly in python.
>The w3m image preview method is significantly faster for these files (close to instant or at least well under half a second) without caching. 

All other image preview methods can feel faster because they do not block and call an external compiled utility. In my experience w3m isn't that _faster_, but if you are scrolling a lot it won't stop you while loading an image and go to the next.  Maybe if I switched pillow for openCV I could get there, but importing the entirety of it for image previews seems rather excessive.
I wanted to  try something like w3m, an external program that can run asyncronously and possibly faster to resize images. I was thinking of ImageMagick like it used in kitty's demo icat, but it seem very unreliable when resizing images (and still slow with my transparent image). I wonder if w3m can resize and save images?

**Comment by noctuid** [comment]
> Strangely enough, this folder was completely empty for me. Did you use by chance any different preview method, like libcaca's ascii art?

I've always been using w3m image preview. After deleting it though, no new images have been created. I hadn't updated ranger in a while until recently, so maybe there was some change wrt caching.

> you get in /tmp a file named search_me$somegarbage

Yep, that works for me.

> Can you tell me if your slow images are transparent? Also which filesystem are they in? Are they also pngs? Could you possibly upload one for testing?

Most or all of the images I've tested are not transparent. The fileystem is ext4. I have the issue with both jpgs and pngs. Most of the images in [this gallery](https://imgur.com/gallery/MCrBR) take close to a second or more to display for me. It's not directly related to size for me either. Some smaller images are slower, but transparency doesn't seem to be the issue.

> In my experience w3m isn't that faster

It's not just that scrolling is faster; the image actually displays a lot sooner for me. I wasn't trying to scroll fast when testing.

> I was thinking of ImageMagick like it used in kitty's demo icat, but it seem very unreliable when resizing images (and still slow with my transparent image)

I have no idea if it will make a difference, but it might be worth trying GraphicsMagick instead. [imgp](https://github.com/jarun/imgp) is pretty much dedicated to resizing images and claims to be fast. `ffmpeg` might also be worth trying. AFAIK w3m cannot resize images.

**Comment by mark-dawn** [comment]
>Most or all of the images I've tested are not transparent. The fileystem is ext4. I have the issue with both jpgs and pngs. Most of the images in this gallery take close to a second or more to display for me. It's not directly related to size for me either. Some smaller images are slower, but transparency doesn't seem to be the issue.
Wait, are they the jpgs you can get from right clicking the ones you have, or if you download the whole album the quality is different? The jpg I can get can be over the  >1MB of size, but none of the one I tried is slow.

**Comment by noctuid** [comment]
> Wait, are they the jpgs you can get from right clicking the ones you have, or if you download the whole album the quality is different? The jpg I can get can be over the >1MB of size, but none of the one I tried is slow.

I have the whole gallery downloaded, but the file size isn't any different if I right click on those. They all (even the ~300KB one) display much slower with kitty image preview for me. Any other ideas for why this might be happening for me but not for you?

> BTW, what distro, python version and pillow version are you using?

Arch, python 3.6.5, and pillow 5.1.0

**Comment by noctuid** [comment]
That's way better. A lot of large images are pretty instant. Some images are still slow to load compared to with w3m, but even the worst images (e.g. 7-8mb) take more like half a second to a second. It would be awesome if those were as fast, but the current speed is good enough (for me at least) without caching.

**Comment by noctuid** [comment]
I was curious how long resizing takes and tested resizing [an 8mb png](https://eoimages.gsfc.nasa.gov/images/imagerecords/74000/74393/world.topo.200407.3x5400x2700.png) with a few methods. This data isn't all that useful/carefully collected, but it looks like there's not much that can be done to improve resizing speed.

0.50 imagemagick (-scale and 0 png compression)
0.51 imgp (imgp also uses pillow)
0.61 ffmpeg
1.24 imagemagick  (-resize and 0 png compression)

I tried on a few other images and imagemagick was usually the fastest, but the quality was comparably pretty bad with `-scale`. Ffmpeg was usually the second fastest and had the highest quality thumbnails. I tried Graphicsmagick too, but it was always slower. `-thumbnail` is supposed to be faster but was 2-3x slower with Imagemagick for me.

I ran `time.time()` before and after the `image.resize` call in the current code. I don't know if this is an accurate method of timing, but it was generally twice as fast as calling any other program (~0.25 seconds for the above image). I'm thinking it's probably not worth using Imagemagick or another external program for resizing.

**Comment by toonn** [comment]
Does performance differ depending on the size you resize to/from? Maybe a small crop then a resize could be more performant?

**Comment by mark-dawn** [comment]
I think that part of the penality for the resize is actually in the decoding of the original image if that has heavy compression, but that should be the same for any other program any way. I tried switching the filter to `PIL.Image.BILINEAR`, but I couldn't see a significative difference.

**Comment by mark-dawn** [comment]
Also I had a look into the tmux issue, and even if I disable the check I cannot get to talk to kitty. I copied what the other classes that support tmux are doing, but it seem to fail on mine. On closer inspection the ""class"" of commands used by the other terminals seem different (they end in `\a` or something like that, while kitty's end with `\088\\`), so I might need another "wrapper" around the command, but I have no idea were to look into tmux docs.
It is also not reporting the cell size when using `TIOCGWINSZ`, but this hsould be an easy patch on their end and we could work around with some senible defaults (actually iterm2 doesn't too and we have settings for that, but this makes the previewer class not independent from fm. If more than one previewer is interested in having them we could extend the interface of `__init__` to have them as kwargs and pass them in, mantaining isolation easily)

**Comment by noctuid** [comment]
Yeah I don't think the actual resize operation is the issue. I tried using opencv, and the resize operation takes 0.009s, but reading the image takes ~0.23s.

**Comment by toonn** [comment]
@mark-dawn, just a heads up, we're gonna go ahead with #1146. You can use either style in this PR though, if need be I can easily fix up any flake8 issues when I merge.

**Comment by toonn** [comment]
@mark-dawn I hope to be merging this towards the end of the week. Did you revert the caching changes? (I didn't see any commits after that comment.) Any reason we shouldn't merge yet?

**Comment by mark-dawn** [comment]
Sorry about the radio silence, I had a couple of weeks full of work, but it seems over now :relieved:.

The branch could be merged actually, but I have a new version that should fix eventual pep8 issues. I can merge master back into my repository, rebase everything and check if everything is alright and push this evening hopefully, and then you can proceed!

File: ranger/ext/img_display.py (hunk 21 above)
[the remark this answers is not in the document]
Comment (Vifon) [reply ranger/ext/img_display.py]: Actually why not use the full "ranger_thumb_" prefix?

**Comment by mark-dawn** [comment]
Just an heads up, pylint was complaining about `core/main.py:374,25: Using deprecated method load_module() (deprecated-method)`.
I think it might not be fixable due to python2 compatibility, so I put an lint exception there to pass the test, but maybe it would be better to put a `# TODO` there just to mark tthat it mgiht need attention in the future?

**Comment by toonn** [comment]
I'm not sure how deprecation is handled in the python stdlib. I assumed it won't be removed from older versions of the language when I wrote that code. It's still deprecated in py3.8 though so let's just crash and burn on this ; )

**Comment by toonn** [comment]
It's in master 🎉
Next release is getting closer.

**Comment by sandersantema** [comment]
I've installed pillow with pip3 but I can't seem to get it to work. Ranger still gives an error and says I need pillow to work. 

- Mac OS 10.13.4
- In Bash and Fish
- pip 10.0.1 from /usr/local/lib/python3.6/site-packages/pip (python 3.6)
- Pillow:
    1. Version: 5.1.0
    2. Location: /usr/local/lib/python3.6/site-packages
- Path settings in fish (this actually might be the problem, but I couldn't see why)
    1. set -Ux fish_user_paths "usr/lib/python2.7/site-packages" $PATH
    2. set -Ux fish_user_paths "/usr/local/lib/python3.6/site-packages" $PATH

**Comment by Rom1deTroyes** [comment]
Le vendredi 15 juin 2018 à 00:51:33 (-0700), sandersantema a écrit :
> - Path settings in fish (this actually might be the problem, but I couldn't see why)
>     1. set -Ux fish_user_paths "usr/lib/python2.7/site-packages" $PATH

"/usr/..." ?
But if it's installed in /u/l/l/python3.6/ it must be something else...

++
    Romain
-- 
Je pardonne aux gens de n'être pas de mon avis,
je ne leur pardonne pas de n'être pas du leur.
	-+- Charles-Maurice de Talleyrand-Perigord -+-

**Comment by MindTooth** [comment]
I needed to install Pillow on the system Python, as Ranger seemed to force that.  I might be wrong, but I recall I needed to do just that.

**Comment by freed00m** [comment]
Does anyone has rendering issues? 

Some images just keeps blinking and preventing me to input command at the command line, making input very hard.

Just tried to rename a file, an extra line must be drawn killing the experience. I am using ranger from repository master branch. The second image is when I cancel the command.

2018-07-17-170749_1046x515_scrot
2018-07-17-170755_1046x515_scrot


Anyone has an idea what is causing this?

**Comment by mark-dawn** [comment]
@freed00m that's weird, it looks like the console got garbled up. Can you tell me if the files  that cause this have something in common? Size maybe, or an aspect ratio? Are they in a specific folder, or have weird filenames?
Also, are you on python 2 or python 3? Also if you send the command `redraw_window` dos it get fixed?

**Comment by freed00m** [comment]
@mark-dawn It's not file specific, here an example file https://ibb.co/bLpb3d .

This file can be renamed confortably at `~/Pictures/screenshotblur.png`  but whenever this file sits at `/tmp/screenshotblur.png` it triggers this bug. 

I am running Manjaro Testing branch so I am cca week behind Arch in terms of versions, currently  `Python 3.6.6`.

I use `redraw_window` to recover from this bugged state when focused on non image file.


I am thinking maybe the rendering of font Awesome is triggering this? Or maybe my font of choice "Fira font medium" causing this?

If you have some cool breakpoints I could debug I could :)

**Comment by freed00m** [comment]
I've tried disabling devicons and even changed fonts to no effect.
The weirdest part is when running ranger with vscode's debugger attached, the bug is not triggered..

:O what the heck is going on?

**Comment by mark-dawn** [comment]
@toonn The filenames are already hashed.
What triggers the bug is ranger refreshing way to quickly. You can get the same effect by browsing `$somedir` while running `while true; do touch $somedir/test; rm $somedir/test; done`. The effect is way bigger than normal since the file is getting linked/unlinked at way higher rates, but the principle is the same.
The weird part is that while refreshing the UI glitches only while previewing an image file, but not with other previews. It might happen due to escape codes from the image commands getting jumbled up in the pipes.
I tried using terminology in the same situation, and while it gives me still a nice epileptic flashing, the UI *doesn't* glitch.
The only way I see this happening is if ranger can stop a ongoing `draw` operation if some update happens.

I tried fixing it by using a fixed file that gets rewritten, instead of creating a new file every time. This indeed reduces the flickering, but the file updating still causes refreshes every now and then that corrupt the interface when trying to issue a command. 
Another thing I've tried is @kovidgoyal suggestion, and while it doesn't fix the problem, it should make it extremely unlikely (and I could still disable the previews in case some user wanders in `/dev/shm`). However I would like to check if a binding for `shm_open` exists in the standard library, as it would make the preview POSIX complaiant

**Comment by toonn** [comment]
Ah, didn't understand you're continuously deleting the thumbnails.

**Comment by mark-dawn** [comment]
I found the core of the issue, somehow when the cursor get moved to a new spot to draw the image it might "fail" to get back before the next refresh. In quotes because it might just be late.
The moving is handled in a context manager, and I am now sure that the call to `draw` is actually blocking, so the cleanup code must be called before ranger can go on with its business. Indeed, by wrapping the calls to curses with stderr debug statement I can see that the "store cursor" and the "load cursor" get called in pairs, even during the glitch.

This restrict the fuckup to curses, since it is the only component that might be asynchronous. So when the cleanup code is called, curses/terminal is instructed to put back the cursor where it was before (async), then the `draw` subroutine exit, and at this point if a redraw comes fast enough the cursor might not have been back.
I wonder if there is a way to wait for the cursor to be restored. I already flush stdout, so our only hope lies in the terminfo interface. Which might be a sign to look for a workaround like shm or a subfolder.

**Comment by toonn** [comment]
Why would a subdirectory help exactly?

**Comment by mark-dawn** [comment]
AH! FIXED! (sort of)
So, for the save and restore cursor stuff I was basically reusing the code that was around the other classes, that goes like this
```
curses.putp(curses.tigetstr("sc"))  # save cursor pos
sys.stdout.write(curses.tigetstr("cup"), y, x)  # move cursor
do_stuff()
curses.putp(curses.tigetstr("rc"))  # restore cursor pos
sys.stdout.flush()
```
The problem is that putp must not put stuff directly in stdout, so when python flushes the restore command might not be in stdout yet. The solution is simple, switch putp for a write to python stdout interface so everything goes along in the correct order.
The image still flicker, but at a much lower rate and without any slowdown or glitch, which is a great improvement.

@toonn Oh, I could just not preview anything in the subfolder. Not ideal, but we could avoid disabling previews in the entire /tmp. We could even snuck it in `$XDG_RUNTIME_DIR`, where it wouldn't even be out of place. The use of `/tmp` was entirely motivated by python standard library

**Comment by mark-dawn** [comment]
@toonn I have a new version ready to push that should clean some bugs, but when I run test pylint goes absolutely berserk and reports errors all over the source tree. Pytest and flake8 are fine and everything is rebased on the latest master. Do you have any idea on why that might be the case?
Maybe Arch has a version of pylint that's too new? Could you give me the version the project is currently using? Or is it broken on your side too and pylint is ignored atm on travis?

**Comment by FichteFoll** [comment]
@mark-dawn https://github.com/ranger/ranger/blob/8e9451e6648c57d8016dad280132bf65a5d919cc/Pipfile#L12

Just use `pipenv install --dev` to set up your environment.

**Comment by mark-dawn** [comment]
Thanks, that fixed up everything, except my blindness. I will probably make a new pull request since I am on a new branch.

**Comment by toonn** [comment]
@Moelf, yes. The kitty method only works in kitty so the default is still w3m of course. I've thought about adding an `auto` method that tries to detect the terminal in use. That's not exactly straightforward sadly, so it wouldn't always work. No work has been done on this though, it's not a priority tbh.

**Comment by kovidgoyal** [comment]
Just FYI, the kitty graphics protocol supports detection at runtime. See icat/main.py (the detect_support() function for how the kitty icat tool itself does it).

**Comment by mark-dawn** [comment]
@toonn we could at least merge the `preview_image` and `preview type` settings into one, so people don't get confused, but I don't know if it's worth it to mess with old settings.

@kovidgoyal that would be actually very useful, I still check at runtime if kitty is actually there but I ask the user to set an env variable, which is shitty. I'll take a look asap, thanks.

**Comment by Moelf** [comment]
> 
> 
> @Moelf, yes. The kitty method only works in kitty so the default is still w3m of course. I've thought about adding an `auto` method that tries to detect the terminal in use. That's not exactly straightforward sadly, so it wouldn't always work. No work has been done on this though, it's not a priority tbh.

Makes sense, but how can I tell ranger to use Kitty method since I always use Kitty anyways?

**Comment by toonn** [comment]
@Moelf, add/change `set preview_image_method kitty` in `rc.conf`. It's all in the man page.

**Comment by ThatOneCalculator** [comment]
Hey, I know this thread is kinda old, but does anyone know how to use this feature with Kitty? I'm kinda lost.

**Comment by freed00m** [comment]
@theonecalculator 
Did you setup in rc.conf ?

```
set preview_images true
set preview_images_method kitty
```

ThatOneCalculator:
> @theonecalculator 
> Did you setup in rc.conf ?
> 
> ```
> set preview_images true
> set preview_images_method kitty
> ```

Worked flawlessly! My bad.

PR #1093, opened 2018-02-24
Logfile arg

FichteFoll:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

 SKIP

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Added `--logfile` to the man page and made it default to `"-"` iff `--debug` was also provided.

#### MOTIVATION AND CONTEXT

SKIP

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -129,7 +129,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.9.1" "02/22/2018" "ranger manual"
+.TH RANGER 1 "ranger-1.9.1" "2018-02-24" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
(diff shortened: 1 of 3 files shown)


FichteFoll:
Setting logfile to `-` seems not to behave too well as I'm getting update collisions with log messages appearing on screen in the ranger file view, leading to broken rendering and the messages then not actually being there when closing ranger. Maybe https://github.com/ranger/ranger/pull/1093/commits/766aa3539cc7d8f123cb38077f83dfcc695d172f needs to be reverted (or scratched).

This is even worse with #1092 because chardet has quite some more DEBUG logging.

toonn:
I agree, ranger needs stdout and stderr is output to the same place by default. Can we make it log to the logging view `:display_log` by default?

FichteFoll:
It's been 5 years, I haven't needed it since and the status of whether the changes can make it in is also unclear, so I'm closing this. If something like this is desired again, it should be redone from scratch.

Issue #1094, opened 2018-02-26
Scope.sh can't handle text with nullbytes


gaddox:
#### Runtime Environment
ranger version: ranger-master 1.9.1
Python version: 3.6.4 (default, Jan  5 2018, 02:35:40) [GCC 7.2.1 20171224]
Locale: en_US.UTF-8
Kernel: 4.15.3-2-ARCH x86_64

Affects all terminal emulators, and stops tty from interacting with the file.

#### Current Behavior
If you're using the scope.sh file previewer, when you highlight a file, with a type of "text/plain", but which contains a trailing nullbyte, the output will still be sent to Python and cause it to crash. 

#### Possible Solutions
See #1095

#### Steps to reproduce
0. Uncomment/edit in `rc.onf`, `set use_preview_script true` and comment out `set preview_script ~/.config/ranger/scope.sh`
1. Obtain a file that's under 41 bytes, has a mime-type of `text`, and has a trailing nullbyte.
a.  `echo -en "FooBar\x00" > BREAK_ME; file -i 'BREAK_ME'`
2. Run `ranger` and set the cursor bar on top of `BREAK_ME`
3. Crash

#### Traceback
```
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/ranger/core/main.py", line 196, in main
    fm.loop()
  File "/usr/lib/python3.6/site-packages/ranger/core/fm.py", line 390, in loop
    ui.redraw()
  File "/usr/lib/python3.6/site-packages/ranger/gui/ui.py", line 338, in redraw
    self.draw()
  File "/usr/lib/python3.6/site-packages/ranger/gui/ui.py", line 365, in draw
    DisplayableContainer.draw(self)
  File "/usr/lib/python3.6/site-packages/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/usr/lib/python3.6/site-packages/ranger/gui/widgets/view_miller.py", line 100, in draw
    DisplayableContainer.draw(self)
  File "/usr/lib/python3.6/site-packages/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/usr/lib/python3.6/site-packages/ranger/gui/widgets/browsercolumn.py", line 176, in draw
    self._draw_file()
  File "/usr/lib/python3.6/site-packages/ranger/gui/widgets/browsercolumn.py", line 203, in _draw_file
    Pager.draw(self)
[...]

Issue #1086, opened 2018-02-23
Ranger crashes, if there is a 'octave-workspace'


theova:
#### Runtime Environment

- Operating system and version: Arch Linux
- Terminal emulator and version: termite v13
- Python version: Python 3.6.4
- Ranger version/commit: ranger-master 1.9.1
- locale: de_DE.UTF-8

#### Current Behavior
The file 'octave-workspace' generated from an interactive octave-cli session is stored in a directory. If I try to go into this directory with ranger, it crashes.

#### Traceback
```
Traceback (most recent call last):
  File "/home/theo/ranger/ranger/core/main.py", line 196, in main
    fm.loop()
  File "/home/theo/ranger/ranger/core/fm.py", line 390, in loop
    ui.redraw()
  File "/home/theo/ranger/ranger/gui/ui.py", line 338, in redraw
    self.draw()
  File "/home/theo/ranger/ranger/gui/ui.py", line 365, in draw
    DisplayableContainer.draw(self)
  File "/home/theo/ranger/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/home/theo/ranger/ranger/gui/widgets/view_miller.py", line 100, in draw
    DisplayableContainer.draw(self)
  File "/home/theo/ranger/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/home/theo/ranger/ranger/gui/widgets/browsercolumn.py", line 176, in draw
    self._draw_file()
  File "/home/theo/ranger/ranger/gui/widgets/browsercolumn.py", line 203, in _draw_file
    Pager.draw(self)
  File "/home/theo/ranger/ranger/gui/widgets/pager.py", line 101, in draw
    self._draw_line(i, line)
  File "/home/theo/ranger/ranger/gui/widgets/pager.py", line 132, in _draw_line
    self.addstr(chunk)
  File "/home/theo/ranger/ranger/gui/curses_shortcuts.py", line 37, in addstr
    self.win.addstr(*args)
ValueError: embedded null character

```

Thank you very much for your work! Ranger is super nice!

toonn:
When you go into that directory is this file highlighted or not?

theova:
`octave-workspace` was the first file/element in this directory and I wasn't able to see the content of the directory.

I'm sorry, I can't reproduce it anymore. If it will reappear, I'll try to find out how to reproduce it.

Issue #1082, opened 2018-02-18
Ranger crashes when selecting a rdp file


UnicodeAlt255:
#### Runtime Environment

- Operating system and version: Arch Linux x86-64, updated as of 2/17/2018
- Terminal emulator and version: konsole 17.12.2 and termite v13
- Python version: 3.6.4
- Ranger version/commit: af0ef6a233744121307fba3b11089f05e9369782
- Locale: en_US.UTF-8

#### Current Behavior
Selecting a file named `Default.rdp` causes ranger to crash.

#### Expected Behavior
Ranger should not crash.

#### Context

I was simply browsing my Documents folder and  it happened to have a file called `Default.rdp`. When my selection reached that file `Default.rdp`, ranger crashed and gave me a message to report the traceback.

#### Possible Solutions
When I open up `Default.rdp` in emacs or kate, it appears normal. However, `cat` shows that there are two strange characters in the beginning of the file, which may be causing the crash.

#### Steps to reproduce
1. run ranger
2. select `Default.rdp`

#### Traceback

```
ranger version: ranger-stable 1.9.0
Python version: 3.6.4 (default, Jan  5 2018, 02:35:40) [GCC 7.2.1 20171224]
Locale: en_US.UTF-8
Current file: '/home/daniel/Documents/Default.rdp'

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/ranger/core/main.py", line 196, in main
    fm.loop()
  File "/usr/lib/python3.6/site-packages/ranger/core/fm.py", line 390, in loop
    ui.redraw()
  File "/usr/lib/python3.6/site-packages/ranger/gui/ui.py", line 338, in redraw
    self.draw()
  File "/usr/lib/python3.6/site-packages/ranger/gui/ui.py", line 365, in draw
    DisplayableContainer.draw(self)
  File "/usr/lib/python3.6/site-packages/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/usr/lib/python3.6/site-packages/ranger/gui/widgets/view_miller.py", line 100, in draw
    DisplayableContainer.draw(self)
  File "/usr/lib/python3.6/site-packages/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/usr/lib/python3.6/site-packages/ranger/gui/widgets/browsercolumn.py", line 176, in draw
    self._draw_file()
[...]

PR #1091, opened 2018-02-24
Catch the right errors in addstr

toonn:
Python changed an error from `TypeError` to `ValueError` in version 3.5
now we catch both for backwards compatibility.
Relevant issue: https://bugs.python.org/issue22215

Fixes #990
Fixes #1045
Fixes #1079
Fixes #1082
Fixes #1086

PR #1095, opened 2018-02-26
Fix scope.sh crashing on text ending with nullbyte

gaddox:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
ranger version: ranger-master 1.9.1
Python version: 3.6.4 (default, Jan 5 2018, 02:35:40) [GCC 7.2.1 20171224]
Locale: en_US.UTF-8
Kernel: 4.15.3-2-ARCH x86_64

#### CHECKLIST

#### DESCRIPTION / MOTIVATION AND CONTEXT
Currently, scope.sh doesn't differentiate between charsets.
So, if a file with the mimetype `text/plain` is encoded in
binary, and contains a trailing nullbyte within the first 40
bytes, it will crash with `embedded null charcater`. This is
due to using `file --dereference --brief --mime-type --` on
line 170, instead of using the POSIX-compatible `file -i`,
which does the same thing, but also lists the charset.

The main culprit is the `file` utility, which lists any files
with 40 or less bytes, even if they contain control characters,
as `ISO-8859 text` or `ASCII text`. If the last byte is a
nullbyte, it will be sent to Python and crash the draw()
function.

If the `MIMETYPE` variable was defined with the second version,
all you would need to do is add a `*binary)` case to
`handle_mime`, right before `text/* | */xml)`, and just let it
fall through to the regular `fallback_handler`. Instead, the
most inexpensive thing you can do is retrieve the first 40 bytes
of any file with mimetype `text/plain` with `head -n 3`. Then
do a hex dump in named-character format, `od -t a`, on the output
and pipe it to `grep -q 'nul'` which will determine if it
contains any nullbytes. If it does, call `return` and exit the
case statement to the `fallback_handler`.

This is also is a very minor security concern.

_After Edit:_ This isn't only relevant to the specific cases I listed. Going over past issues with scope.sh, it solves others:

Resolves: #1094 
Resolves: #1082 
Resolves: #1086

#### TESTING
Isolated, manual tests on `scope.sh` with bash `set -x` debugging. Multiple files were tested, including the problem ones, nothing pops out. Full tests on `ranger` with `scope.sh` switched out for the hotfix, and nothing pops out. All files, except the problem ones, still display the same way and the problem files display without issue anymore.

Affects should be minimal, and the state-less design is maintained.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -130,6 +130,10 @@ handle_mime() {
     case "${mimetype}" in
         # Text
         text/* | */xml)
+            # Check for nullbyte in first 40 bytes
+            if head -n 3 "${FILE_PATH}" | od -t a | grep -q 'nul'; then
+                return
+            fi
             # Syntax highlight
             if [[ "$( stat --printf='%s' -- "${FILE_PATH}" )" -gt "${HIGHLIGHT_SIZE_MAX}" ]]; then
                 exit 2


toonn:
Patched in #1091 thanks for the PR, be sure to clone and test the master branch first next time so you don't lose time implementing something that's already been fixed : )

Feel free to continue the discussion despite me closing the issue if you think the patch is still necessary.

Issue #1108, opened 2018-03-11
Rifle flag t only works with urxvt


toonn:
Rifle's `flag t` only works with urxvt (because who cares about standards?).
Urxvt passes any arguments after `-e` verbatim, most other terminals consume them as arguments.
Some terminals use `-e`, others use `-x`.

toonn:
Summary
-------

<dl>
  <dt>`-c`</dt>
    <dd>tilda</dd>
  <dt>`-e`</dt>
    <dd>cool-retro-term, gnome-terminal, lilyterm, lxterminal, mate-terminal, pantheon-terminal, st, terminator, terminology, termite, urxvt, xfce4-terminal, xterm</dd>
  <dt>`-x`</dt>
    <dd>gnome-terminal, lilyterm, mate-terminal, terminator, xfce4-terminal</dd>
  <dt>`--command`</dt>
    <dd>lxterminal, mate-terminal, terminator, tilda</dd>
  <dt>`--command=`</dt>
    <dd>gnome-terminal, lxterminal, xfce4-terminal</dd>
  <dt>`--exec`</dt>
    <dd>terminology</dd>
  <dt>`--exec=`</dt>
    <dd>termite</dd>
  <dt>`--execute`</dt>
    <dd>gnome-terminal, lilyterm, mate-terminal, terminator, xfce4-terminal</dd>
  <dt>`--execute=`</dt>
    <dd>pantheon-terminal</dd>
  <dt>other</dt>
    <dd>Mac os is even worse in this regard: [Terminal.app](https://stackoverflow.com/a/28373404), [iterm2](https://gist.github.com/vyder/96891b93f515cb4ac559e9132e1c9086), kitty.app(?).</dd>
</dl>

References
----------

For alacritty: **?**

For cool-retro-term:
> `-e <cmd>`
> Command to execute. This option will catch all following arguments, so use it as the last option.

For gnome-terminal:
> `-e, --command=STRING`
> Execute the argument to this option inside the terminal. 
> `-x, --execute`
> Execute the remainder of the command line inside the terminal.

For guake: No option.

For kitty/kitty.app: **?**

For konsole:
> `-e command`
> Execute command instead of the normal shell.

For lilyterm:
> `-e COMMAND | -x COMMAND | --execute COMMAND`
> Run a command when starting up. Must be the final option.

For lxterminal:
> `-e STRING --command=STRING --command STRING`
[...]

PR #1109, opened 2018-03-11
Attempt to fix issue #1108 flag t problem

Xunius:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: manjaro 17.1.6
- Terminal emulator and version: gnome-terminal, xfce4-terminal, terminator, lxterminal, xterm, urxvt
- Python version: 3.6
- Locale: en_US.UTF-8

#### CHECKLIST

- [X ] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [X ] All changes follow the code style **[REQUIRED]**
- [X ] All new and existing tests pass **[REQUIRED]**
    - [ ] Tests have been updated

#### DESCRIPTION

In `ext/rifle.py`, starting from line 358:

* add an `else` block complementing `if 'TERMCMD' not in os.environ` block. 
* in the new `else` block, check if TERMCMD in a list of terminals that works with `-e` flag or a list that works with the `-x` flag, set `cmdflag` to `-e` or `-x` correspondingly.
* move the `if term not in get_executables()` block to after the new `else` block, and set `cmdflag` to `-e` to the fall-back term of `xterm`.
* finally compose the `cmd` as `cmd = [os.environ['TERMCMD'], cmdflag] + cmd`.

The list is not exhaustive, new terminals that still experience issue can be
added to the -e or -x group or make a new `elif` block

#### MOTIVATION AND CONTEXT

issue #1108: Rifle's flag t only works with urxvt (and xterm) (because
who cares about standards?).  Urxvt passes any arguments after -e
verbatim, most other terminals consume them as arguments.  Some
terminals use -e, others use -x.

#### TESTING

Only tested the named terminals in a single linux system.


ranger/ext/rifle.py
```diff
@@ -359,30 +359,30 @@ def execute(self, files,  # noqa: E501 pylint: disable=too-many-branches,too-man
...
+                        if term in ['gnome-terminal', 'xfce4-terminal',
+                                    'mate-terminal', 'terminator']:
+                            cmdflag = '-x'
+                        elif term in ['lxterminal', ]:
+                            cmdflag = '-e'
                         else:
-                            cmdflag='-e'
+                            cmdflag = '-e'
 
                     if term not in get_executables():
```
toonn: Let's put this after the if clause with the branches for urxvt, so line 372.

ranger/ext/rifle.py
```diff
@@ -359,30 +359,30 @@ def execute(self, files,  # noqa: E501 pylint: disable=too-many-branches,too-man
...
                             # Sometimes urxvt calls itself "rxvt-256color"
                             if 'rxvt' in get_executables():
                                 term = 'rxvt'
                             else:
                                 term = 'urxvt'
-                            cmdflag='-e'
+                            cmdflag = '-e'
                     else:
                         term = os.environ['TERMCMD']
-                        if term in ['gnome-terminal', 'xfce4-terminal', 'mate-terminal',\
```
toonn: Let's extract this to after we've determined `term` properly.

ranger/ext/rifle.py
```diff
@@ -355,23 +355,44 @@ def execute(self, files,  # noqa: E501 pylint: disable=too-many-branches,too-man
...
+                        self.hook_logger("Can not determine terminal command.  "
+                                         "Please set $TERMCMD manually.")
+                        # A fallback terminal that is likely installed:
+                        term = 'xterm'
+
+                    # Choose correct cmdflag accordingly
+                    if term.lower() in ['xfce4-terminal', 'mate-terminal',
+                                        'terminator']:
+                        cmdflag = '-x'
+                    elif term.lower() in ['xterm', 'xterm-256color', 'urxvt',
```
toonn: Did you check whether `xterm-256color`, `rxvt-256color` and `rxvt-unicode` are ever binary names? I'm asking because that's different from the previous behavior. I believe the command for urxvt is always named urxvt or rxvt, at least that's what the previous behavior indicates.

Xunius: Yes I overlooked this part, will use 'xterm' for ['xterm', 'xterm-256color'], and 'urxvt' for ['urxvt', 'rxvt-unicode'], and for 'rxvt' for 'rxvt-256color', if 'rxvt' found in executables, use 'rxvt', otherwise use 'urxvt'. Plan to change that in the next commit.

What do you think we should do with those not working with -e or -x or whatever (e.g. pantheon-terminal), fallback to 'xterm'?

Btw I don't remember did terminology work for you? And termite?

toonn: Don't forget to include "Fixes #1108" in a commit message so the issue gets linked and it doesn't have to be the commit where you satisfactorily solve the problem, it's a nice indication someone's working on it so people don't quietly perform redundant work. (Would also make it easier for me to refer back to it ; )

Both terminology, termite and pantheon-terminal work with `-e "vim file.txt"` for me. tilda works with `-c "vim file.txt"`. st works with `st -e vim file.txt`.
Haven't been able to test kitty because of driver woes.

ranger/ext/rifle.py
```diff
@@ -355,23 +355,59 @@ def execute(self, files,  # noqa: E501 pylint: disable=too-many-branches,too-man
 
                 cmd = prefix + [command]
                 if 't' in flags:
-                    if 'TERMCMD' not in os.environ:
-                        term = os.environ['TERM']
-                        if term.startswith('rxvt-unicode'):
+                    term = os.environ.get('TERMCMD', os.environ['TERM'])
+                    # we can do this, can't we?
+                    term = term.lower()
```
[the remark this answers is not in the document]
toonn: If you could address this we can finally merge your PR. If you're too busy let me know, then I'll find some time.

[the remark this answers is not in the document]
Xunius: I nearly forgot what we were doing...
I removed term.lower(), after merging the master, that should work, hopefully...

Xunius:
@toonn Hmmmmmmm. I don't quite get what's going on. Did you mean that if you set `TERMCMD` or `TERM` to, for instance, `mate-terminal`, but `term` in `rfile.py` is `xterm`?
`termite`, `st` and `urxvt are already covered.

toonn:
Nope, I didn't set anything. I listed what the value of `TERM` was for each of the terminals I tried.

PR #1123, opened 2018-03-27
Improve VCS symbols

chuanconggao:

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

I propose that we adjust the symbols for `untracked` and `changed` to `?` and `+`, respectively, as Git uses. Plus, now `changed` and `staged` have different symbols. As a result, we need to adjust the symbol for `unknown`, possibly using `!`. In summary, the adjusted symbols can be
``` python
    vcsstatus_symb = {
        'conflict': (
            '✗', ['vcsconflict']),
        'untracked': (
            '?', ['vcschanged']),
        'deleted': (
            '-', ['vcschanged']),
        'changed': (
            '+', ['vcschanged']),
        'staged': (
            '*', ['vcsstaged']),
        'ignored': (
            '·', ['vcsignored']),
        'sync': (
            '✓', ['vcssync']),
        'none': (
            ' ', []),
        'unknown': (
            '!', ['vcsunknown']),
    }
```

BTW, it seems the symbol used for sync is actually the left part of square root `√` (can verify by copying somewhere else and zoom). I also propose we use the proper `✓` instead of `√`. I also updated the symbol for conflict to the proper `✗` instead of letter `X`.

The symbols for remote are also adjusted slightly for matching style.

#### MOTIVATION AND CONTEXT

The VCS extension is awesome. However, some of its symbols like `untracked` and `changed` are very confusing to long-term Git users. Arguably, Git is dominating and using the same set of symbols makes great sense.


toonn:
I agree either corresponding more to git or actually having the symbols depend on the VCS you're using is good.
Not sure about the `X` and `√`, they're visually similar enough imo and the "proper" symbols seem more unicode-y so I'd only consider doing that if they're covered by all common fonts, not sure about coverage for sqrt so maybe that should be `V`.

chuanconggao:
I am okay with just changing symbols for `untracked` and `changed`. 

I think for simplicity, using the same set of symbols across different VCS makes sense. The change is minimal.

toonn:
Is yellow used already? Seems like a good color between red and green.

hut:
Looks good to me, thank you.

PR #1125, opened 2018-03-28
Added: shift tabs right/left

dbosst:
to shift tabs, use new shortcuts ALT-p and ALT-o

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 16
- Terminal emulator and version: GNOME Terminal 3.18.3
- Python version: Python version: 2.7.12 (default, Dec  4 2017, 14:50:18) [GCC 5.4.0 20160609]
- Ranger version/commit: ranger-master 1.9.1 / f855979587bd918f0d32c0caba79ab4b4aa531cf
- Locale: None.None

#### CHECKLIST

- [] All new and existing tests pass **[REQUIRED]**
    - [ ] Tests have been updated

#### DESCRIPTION

Shift the selected tab left or right.
Uses shortcuts added to rc.conf: ALT-p to shift tab right, ALT-o to shift tab left.

#### MOTIVATION AND CONTEXT

Let's you organize the tabs so you can place them in the order you want.

#### TESTING

Tested ok with different numbers of tabs open  (1 tab, 2 tabs  ... 9 tabs).
Tested ok with close tab/restore tab.
Tested ok with save_tabs_on_exit.

None.
Running the base/make test compared to this fork/make test: 0.00 difference in score

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -478,8 +478,8 @@ map <a-6>     tab_open 6
 map <a-7>     tab_open 7
 map <a-8>     tab_open 8
 map <a-9>     tab_open 9
-map <a-p>     tab_shift 1
-map <a-o>     tab_shift -1
+map <a-r>     tab_shift 1
+map <a-l>     tab_shift -1
 
 # Sorting
 map or set sort_reverse!
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -478,6 +478,8 @@ map <a-6>     tab_open 6
 map <a-7>     tab_open 7
 map <a-8>     tab_open 8
 map <a-9>     tab_open 9
+map <a-p>     tab_shift 1
+map <a-o>     tab_shift -1
 
 # Sorting
 map or set sort_reverse!
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1191,59 +1191,6 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 return self.tab_open(i, path)
         return None
 
-    def tab_shift(self, offset=0, pos=None):
-        """Shift the tab left/right
-
-        Shift the current tab to the left or right by either:
-        offset - changes the tab number by offset
-        pos - shifts the tab to the specified tab number
-        """
-
-        oldtab_index = self.current_tab
-        if pos is None:
-            assert isinstance(offset, int)
-            # enumerated index (1 to 9)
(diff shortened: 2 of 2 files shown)


toonn:
Are you sure you have all the dependencies for `make test`? Travis gives some errors related to `flake8`, we try to adhere strictly to PEP8.

I like this, do have a suggestion though. Could we do keyword arguments so you can either provide an offset or an actual index, starting from 1 just like tabs are numbered? Seems a more useful command to me, e.g. `:tab_shift to=1` to make the current tab the first tab.
Best case you'd implement the semantics of vim's `tabmove`, N is an indexed move, +N or -N are shifts. (And rename `:tab_move` to something better so you can call this `:tab_move` : )

Don't like the mapping though. What's the motivation for `alt-o` and why is `alt-p` a right shift? Our default mappings are inspired by vim so let's at least keep it mnemonic, `alt-l` for left and `alt-r` for right. Or alternatively not include default mappings at all.

dbosst:
Passes tests now (thanks, I was missing a dependency)

Added shift to position number argument (pos).  Can't use 'to' as the argument name so using 'pos', since 'to' fails tests (C:1194, 4: Argument name "to" doesn't conform to snake_case naming style (invalid-name))

Changed the shortcuts to alt-l alt-r as you suggested.

Can explicitly set tab number now for the tab by using offset/pos: If there are several closed tabs, tabs 1,5,6, with tab 1 selected, tab_shift 1 it will change the number to 2, instead of 5. I believe this is desirable as it is, so someone can change the number to whatever they feel like.

Also, trimmed the refresh for the tab to the minimum. Works ok in both miller view/multipane view.

toonn:
Foiled again by pep8 : ), I'd suggest maybe index instead of pos but either's fine. I'd actually consider just going with `to` and adding a no-warn comment. `:tab_shift to=5` just reads nice and is consistent with `:move`.

Tbh moving from 1 to 2 in your scenario is a little unexpected behavior imo but not renumbering when closing tabs is too so I guess it's fine. Could be useful if you want some tab to always be last I guess.

I'm going to wait on opinions for this, also whether it's a better fit for `actions.py` or `commands.py`.

toonn:
I was looking at merging this. I still think `pos` should be renamed to `to`. The limiting of the index I'm not sure about. Nothing stops you from creating tabs at higher indices so there's really no reason to limit this function. Shifting left can cause some tabs to be shifted to the end of the list so the check on offset not overflowing the index can go imo.

With these changes it'll be ready to be merged.

Issue #914, opened 2017-07-25
Mime-type rules is rifle.conf don't work for some files


maln0ir:
#### ISSUE TYPE

- Bug report

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch Linux
- Terminal emulator and version: rxvt-unicode 9.22
- Python version: 3.6.2
- Ranger version/commit: 1.8.1
- Locale: pl_PL.UTF-8

#### EXPECTED BEHAVIOR

Ranger should use `$EDITOR` to open any plain text file (mime-type: `text/plain`)

#### CURRENT BEHAVIOR

Ranger use `$EDITOR` to open only some plain text files. Some plain text files
cannot be opened even if there is rule is `rifle.conf` to do that.

#### STEPS TO REPRODUCE

    ~ $ cat $XDG_CONFIG_HOME/ranger/rifle.conf
    mime ^text,  label editor = $EDITOR -- "$@"
    ~ $ echo test > test.sql
    ~ $ echo test > test.txt
    ~ $ file --mime-type test*
    test.sql: text/plain
    test.txt: text/plain
    ~ $ ranger

Now, try to open both files from ranger.

When opening test.sql file ranger says: `No action found.`, but test.txt file
can be opened with `$EDITOR` without a problem.

toonn:
This has been confirmed. [Here's](https://github.com/khailey/sqltuning/blob/master/jonathan_lewis_vst_example.sql) an example sql file that ranger won't open with the editor even though the mime-type is `text/plain`.

Workaround in the mean time is to add a rule to rifle.conf similar to the following: `ext sql = ${VISUAL:-$EDITOR} -- "$@"`.

Vifon:
It seems to be caused by the usage of the `mimetypes` Python module before trying `file(1)`. That's indeed usually quite unexpected. The `.sql` files belong to the `application/x-sql` type according to this module, so we're missing the `mime ^text` rule. I'm yet to decide how to fix it without breaking for instance #44. See also: 9d1f7a09

toonn:
Sounds like inevitable breakage either way. I personally don't like depending on extensions for mime-types, that kind of makes either `mime` or `ext` redundant in `rifle.conf`. However, reading through those links seems to indicate there's more problems doing it the right way. Simplest solution would be to add `sql` and `json` to `data/mime.types`. I could see `.sql` being used for database files or something which might break this solution.

toonn:
[...]

PR #1127, opened 2018-03-28
Open .sql files with editor

nvllsvm:
Add default support for opening sql files with text editor. 

Confirmed working:
``
ranger version: ranger-stable 1.9.0
Python version: 3.6.4 (default, Mar  9 2018, 23:15:12) [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]
Locale: en_US.UTF-8
``


toonn:
There's discussion about this issue in #914, I'm closing this PR because it's a minor stopgap instead of a solution. You are of course free to use your custom `rifle.conf` and thank you for contributing : )

PR #1132, opened 2018-03-31
Rangerpath

toonn:
Added file selecting behavior in `Tab`'s `enter_dir` so positional arguments can be files instead of only directories. Added relevant documentation and ran the tests.


magnetophon:
I did ``python ranger.py /path/to/file``, and ranger opened in ``/path/to``, with ``file`` selected.
In other words: LGTM, thanks!

Issue #1137, opened 2018-04-12
`ranger --cmd "flat 1"` causes exception


FichteFoll:
#### Runtime Environment

- Operating system and version: Arch Linux
- Terminal emulator and version: alacritty, custom branch
- Python version: 3.6.4
- Ranger version/commit: a30b5e5b
- Locale: en_US.UTF-8

#### Current Behavior

Throws exception.

#### Expected Behavior

Runs specified command.

#### Context

Well, I want to open ranger in a directory and flatten it (by 1) immediately.

#### Steps to reproduce
1. `ranger --cmd "flat 1"`

#### Traceback

```
16:15:22 ERRO Notification: 'NoneType' object has no attribute 'unload'
16:15:22 ERRO 'NoneType' object has no attribute 'unload'
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/ranger/core/actions.py", line 251, in execute_console
    cmd.execute()
  File "/usr/lib/python3.6/site-packages/ranger/config/commands.py", line 1574, in execute
    self.fm.thisdir.unload()
AttributeError: 'NoneType' object has no attribute 'unload'
```

FichteFoll:
Fixed via #1137.

toonn:
Fixed in #1139 actually : p

FichteFoll:
Ah yes, of course. #1137 is this issue.

PR #1139, opened 2018-04-13
Enter directory before executing command arguments Issue #1137

mxovd:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Archlinux 4.15.15-1-ARCH
- Terminal emulator and version: termite 13-2
- Python version: 3.6.4
- Locale: en-US

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Ranger crashes when launching with --cmd "flat 1" command.
This commit set the directory before running the command arguments

#### MOTIVATION AND CONTEXT

There was a bug report about it.
#1137 

#### TESTING

I tested some other commands and it seems to be working fine. (--cmd "terminal", --cmd "grep")
I'm not the most knowledgeable about this project, could you confirm others areas of the codebase are not affected by this? Thanks

### DIFF
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -182,6 +182,7 @@ def main(
             fm.select_file(args.selectfile)
 
         if args.cmd:
+            fm.enter_dir(fm.thistab.path)
             for command in args.cmd:
                 fm.execute_console(command)
 


toonn:
One thing comes to mind. Behavior may be unexpected when starting ranger with mulitple paths. I'm not sure `thistab`'s the right tab at this point of startup.

sivering:
It seems that it will only apply flat to the first path (first tab) when using multiple paths.
I'll look into it to make it more inclusive.

mxovd:
In the case of multiple paths, the first specified path is set as `thistab` during initialization. Unless the user expect to set all the specified directories flat, it should be working as expected. (unless there's something i'm not seeing of course)
If this behavior is preferred, I can look into it. Thanks

PR #1146, opened 2018-04-21
MISSION CRITICAL update to adhere to pep8

### LINKED ISSUE #1145: Pep8 update regarding binary operators
Pep8 was [recently updated](https://github.com/PyCQA/pycodestyle/issues/498) to recommend breaking lines *before* rather than after binary operators.

This is how it should've been all along, therefore this is a **MISSION CRITICAL** issue with the ranger codebase ; )

toonn:
Pep8 was changed to recommend breaking lines *before* rather than after
binary operators. This patch makes everything right with the universe. ; )

Fixes #1145


rmzm:
Sounds good. Need to update `.flake8` though

toonn:
Added. Why are we ignoring E221 btw? Seems sensible imo. Only thing I can think of is alignment but that's not very popular in python as opposed to haskell.

rmzm:
Yes, we ignore E221 for alignment in assignments, like in https://github.com/ranger/ranger/blob/1687e0f41f172cef6923f75f4c6f1038f19671dc/ranger/gui/color.py#L53 . IMO the readability is worth it.

hut:
I messed up the merge via the web interface. Gonna redo it manually.

PR #1156, opened 2018-04-27
ext.rifle: Additional Check for mime type with mimetype command

mikeri:
#### ISSUE TYPE
- Breaking changes
Some MIME types are not found by rifle. With an additional test with the mimetype command after the file command returns 'application/octet-stream' rifle should now be able to recognize all known mime types in the system.

### DIFF
--- a/ranger/ext/rifle.py
+++ b/ranger/ext/rifle.py
@@ -262,10 +262,13 @@ class Rifle(object):  # pylint: disable=too-many-instance-attributes
             mimetype, _ = process.communicate()
             self._mimetype = mimetype.decode(ENCODING).strip()
             if self._mimetype == 'application/octet-stream':
-                process = Popen(["mimetype", "--output-format", "%m", fname],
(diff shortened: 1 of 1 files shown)


toonn:
Did you handle the case where `mimetype` isn't installed?
Travis is complaining because of pylint, if you install both pylint and flake8 you can run `make test` to run the linting locally, less of a wait ; )

mikeri:
Sorry for not testing. I corrected the indent and added a simple try except pass, it seems to work fine. Would that suffice?

toonn:
You probably corrected the indent with the previous pylint warning. Pylint wants you to align the continuation with the opening parenthesis.
`Popen(x...` You want to have the continuing line start in the same column as `x`.

mikeri:
Sorry again, the bare exception cause the test to fail. Added OSError, it works here with both Python2 and 3 when mimetype is not installed.

PR #1178, opened 2018-05-16
Actually check whether selection is a directory for rangerpath patch

### LINKED ISSUE #1177: Why did ranger stop opening files with 'l' (lowercase L)?
I'm using the git version on Arch Linux. Does any you have the same issue?

toonn:
`Actions.move()` used `enter_dir()` to determine whether the selection
was a directory. Now we explicitly check whether it's a directory.

Fixes #1177

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -129,7 +129,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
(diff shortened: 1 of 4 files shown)


Vifon:
Looks sensible.

freed00m:
Ok, after this i get err

'list' object has no attribute 'is_directory' after every attempt of opening directory or file??

What am I missing? I can't use ranger after this, do I have to configure something?

PR #1188, opened 2018-05-26
Boldisnotbright

toonn:
Stop lying and use bright instead of bold. Fall back to dim white instead of bright black in terminals that don't support bright colors.

### DIFF
--- a/ranger/colorschemes/default.py
+++ b/ranger/colorschemes/default.py
@@ -37,29 +37,35 @@ class Default(ColorScheme):
             if context.container:
                 fg = red
             if context.directory:
+                attr |= bold
                 fg = blue
                 fg += BRIGHT
             elif context.executable and not \
                     any((context.media, context.container,
                          context.fifo, context.socket)):
+                attr |= bold
                 fg = green
                 fg += BRIGHT
             if context.socket:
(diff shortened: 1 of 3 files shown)


Vifon:
Is it supposed to merge the draw borders PR too?

toonn:
No, I don't think it is. That's probably why the list of commits is so screwy? I based it off the wrong branch? I'll rebase I guess.

Vifon:
This way we won't have any differentiation between bold/bright and regular text on a terminal with no bright colors, correct? If that was your intention and you tested it's not a true problem, merge away.

toonn:
Yes, I don't really like it. BUT afaik that would've been the case with the current code too? Since bold is usually implemented as bright.

You should check out the PR so you can judge the colors. I'd like more contrast tbh but I"m not sure that's possible. I'm currently thinking of adding the bright instead of replacing the bold. Because in terminals that *do* support it it does make the difference clearer imo. But I'm not sure what'll happen in terminals that support bold as bright? Would we get out of range of the available colors?

One nice improvement that comes with this though now marked and marked & cut files *do* look different which didn't use to be the case, at least for me.

PR #1198, opened 2018-06-05
Introduce a helper function for versioning

toonn:
The `version_helper` function should make version changes simpler, upon a release simple bump `__version__` and change `__release__` to `True`, afterwards change `__release__` back to `False`. It also tries to call `git describe` so `ranger --version` reports a commit hash if available.
    
This should prevent slip-ups where released versions report ambiguously as being `ranger-master x.y.z` which is what we use for non-release versions.


toonn:
Would it be better if this happened at install time in `setup.py` so `ranger-git` installed from the AUR would report the commit hash too?

PR #1199, opened 2018-06-07
Support for RTL languages (Hebrew, Arabic) in the display of file names

esm7:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: Ubuntu 17.10
- Terminal emulator and version: urxvt-unicode
- Python version: 2.7.14
- Locale: EN-US

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
This pull request adds basic BIDI support to Ranger, allowing correct display of file names with Hebrew and Arabic characters on terminal emulators that display these characters properly.

All modern GUIs are BIDI-aware and reverse Hebrew and Arabic texts when they see one, but terminal emulators are not. They always display text in the direction it is sent, leaving the interpretation to the application.

Ranger so far was unable to handle BIDI text and for users who have file names in Hebrew mixed with English ones like I do -- this made life harder. This PR adds a configurable usage of the python-bidi Python package, which encapsulates the non-trivial logic needed to reverse BIDI text in a way that makes sense. This makes it very pleasant to go over a list of files in any language.

#### MOTIVATION AND CONTEXT
I've been using Ranger for a few years but always had to jump to Nautilus when handling directories with Hebrew file names. I can now handle these directories rather flawlessly in Ranger.

#### TESTING
I ran the unit tests, ran with and without the new configuration option and verified the output in multiple scenarios.
Additionally, I've been using Ranger with these changes on a daily basis for a few months now, so I'm quite sure there are no unexpected side-effects. Note that this only changes how names are visually displayed within Ranger, the internal representation of files remains unaffected.


toonn:
This is a really cool addition! I had no idea bidi could be such a simple addition. Does this work with things like `:rename` and `:yank name`, maybe even `:bulkrename`?

Travis complains about a pylint error, since you ran the tests and didn't get this, you're probably missing either pylint or flake8 or both. Personally I actually prefer the explicit `if...then...else` but the pylint gods don't so we blindly follow their guidance ; )

If this works flawlessly with rename and its ilk it's definitely going in the next release. Thanks a million : )

esm7:
I'm happy you like this, because it made my life with Ranger so much better
:)
Actually :rename, :bulkrename and :yank name worked before this too. Even
`@` and then typing a command to operate on the file name. The only problem
was with how these names are displayed on screen and that's the only scope
of my change.
One limitation though is that there's no fancy on-the-fly rendition of BIDI
strings while you type them for :rename or other commands. So if I type
`:rename <some-name-with-hebrew>`, the name is gonna appear reversed while
I type it, and then displayed correctly in the list (if BIDI is on, of
course). To me it looks intuitive because I'm not familiar with any modern
TUIs that do WYSIWYG BIDI handling in text fields anyway.

Regarding the pylint error, I fixed the place you mentioned, but I couldn't
reproduce the failure even after installing pylint and flake8 so I might
not be running the tests correctly. Please let me know if I should look
into it or if things pass now so there's no need.

toonn:
It passes now. Not sure why you didn't get the reported error. Are you running the tests with `make test`?

esm7:
Indeed I was not running with `make test` and that was my error! Been running just part of the tests all along. Glad it still passes.

toonn:
No problem. One thing I forgot to verify, is bidi compatible with python versions all the way back to 2.6? It's ok if it's not, since it's optional, but it would definitely be nice.

This is a great first contribution, keep 'em coming ; )

PR #1207, opened 2018-06-18
Allow optional file preview for root.

horakmar:
Add option of file preview for root user.


toonn:
Tbh I'm not convinced, this has come up before and we've held it off. Why? Because it's a giant hole in your security for some minor convenience. Why is having previews enabled from startup important for the root account? Previews are nice but not necessary and I never spend enough time logged in as root to reach for ranger let alone need previews. Add to that that activating them is only two keybindings - for a whopping total of 4 key presses - away (`zi` and `zv`) and the tradeoff just doesn't seem worth it.

To be clear, we're not saying no (yet), hence why the PR is still open. BUT I'm reluctant until you show me some good use cases and while I don't speak for the rest of the team their opinion is probably similar.

horakmar:
I agree with @NemesisRE. Users should have freedom, possibly they can be warned. I hate when EU protects people against their own foolishness. Please do not repeat that mistake, it's against natural selection.
On the other hand I'm pretty OK with `zp` `zv` - I didn't know about those option toggles before. Thanks!
My use case is editing many files in pretty complicated dir tree of the **reclass cluster model**. Files are owned by root.

Vifon:
I agree with @toonn, this option would mostly encourage bad practices.  For the brief moments running ranger as root is justified, <kbd>zp</kbd> and <kbd>zv</kbd> should be enough.  If one really needs to enable this option even for root, it's easy enough to disable the check in the code.  I'm considering adding an option with the "true" value of "I_am_aware_its_insecure" or something like this but it's getting silly at this point.  We're still considering the options on IRC, so I'll leave it open for a while.

hut:
Would it be an acceptable alternative for you to run ranger like this?

`sudo ranger --cmd="set preview_files=true"`

Maybe with a shell alias?

NemesisRE:
Yes absolutely anyway to override the default behavior would be appreciated.

toonn:
I'm closing this since our consensus didn't change and the workaround is apparently sufficient.

PR #1222, opened 2018-07-04
Implement the filter stack

Vifon:
Inspired by https://github.com/Fuco1/dired-hacks#dired-filter

It's still missing the documentation. Volunteers?

#### ISSUE TYPE

#### CHECKLIST
    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

It wasn't possible to compose multiple filters before. Now it's possible.

I'm aware we already have 2 or 3 filter mechanisms. This generic one should be able to cover all our cases and replace the other filters in the future.

#### TESTING

The added functionality works as intended and shouldn't affect the rest of the file manager.


ranger/config/commands.py
```diff
@@ -1545,6 +1545,54 @@ def execute(self):
...
+    """
+    def execute(self):
+        from ranger.core.filter_stack import SIMPLE_FILTERS, FILTER_COMBINATORS
+
+        subcommand = self.arg(1)
+
+        if subcommand == "add":
+            try:
+                self.fm.thisdir.filter_stack.append(
+                    SIMPLE_FILTERS[self.arg(2)](self.rest(3))
```
toonn: vifon, vifon, vifon...

ranger/config/commands.py
```diff
@@ -1545,6 +1545,54 @@ def execute(self):
...
+        elif subcommand == "show":
+            stack = map(str, self.fm.thisdir.filter_stack)
+            pager = self.fm.ui.open_pager()
+            pager.set_source(["Filter stack: "] + stack)
+            pager.move(to=100, percentage=True)
+            return
+        else:
+            self.fm.notify(
+                "Unknown subcommand: {}".format(subcommand),
+                bad=True
```
toonn: Do we use bad for warnings? I thought it was only errors.

ranger/container/directory.py
```diff
@@ -297,6 +299,7 @@ def inode_filter_func(obj):
         if self.temporary_filter:
```
toonn: Is this `zf`? And will `zf<CR>` still clear filtering? I'm assuming it won't. Does the filter stack and `zf` even combine?

ranger/core/filter_stack.py
```diff
@@ -0,0 +1,114 @@
...
+        # De Morgan's laws.
+        return not accept_file(
+            fobj,
+            ((lambda x, f=filt: not f(x))
+             for filt
+             in self.subfilters),
+        )
+
+    def __str__(self):
+        return "<Filter: {}>".format(" or ".join(map(str, self.subfilters)))
```
toonn: I wonder if we need parentheses? Otherwise I think it won't be clear what pop and decompose are going to work on?

doc/ranger.pod
```diff
@@ -620,6 +620,47 @@ A key chain that allows you to quickly change the line mode of all the files of
...
+
+=item .!
+
+Negate the topmost filter.
+
+=item .c
+
+Clear the filter stack.
+
+=item .*
```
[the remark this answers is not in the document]
Vifon: > Oh boy, won't this be hard to keep track of mentally?

In case of confusion, one can press `..` to view the filter stack state.

> What happens if you haven't used .! or .| or you used them way down the stack?

It's a nop then, it just does nothing as decomposing a simple filter returns a single-element list with this filter on its own. If they are way down the stack, you would need to make them topmost first.

doc/ranger.pod
```diff
@@ -620,6 +620,47 @@ A key chain that allows you to quickly change the line mode of all the files of
...
+
+Decompose the topmost filter combinator (e.g. C<.!>, C<.|>).
+
+=item .p
+
+Pop the topmost filter from the filter stack.
+
+=item ..
```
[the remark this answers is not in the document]
Vifon: I don't understand your point. You can use it beforehand too.

doc/ranger.pod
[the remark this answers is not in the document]
toonn: Hmm, do we need a stack rotate op? So you can decompose whichever filter you want?

ranger/config/commands.py
Vifon: I consider it an error.

ranger/container/directory.py
Vifon: `zf` works like an additional name filter on the stack. They combine and `zf<CR>` doesn't clear the stack, only this single name filter. In the future we could get rid of `temporary_filter`.

ranger/core/filter_stack.py
Vifon: If you mean the `<...>` angle brackets, yes, they add to the clarity of all of this, I'd rather keep them.

doc/ranger.pod
```diff
@@ -620,6 +620,47 @@ A key chain that allows you to quickly change the line mode of all the files of
...
+
+=item .f
+
+Apply the typefilter "file".
+
+=item .l
+
+Apply the typefilter "symlink".
+
+=item .|
```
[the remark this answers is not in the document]
Vifon: I wanted to keep it working consistently in an RPN fashion.

doc/ranger.pod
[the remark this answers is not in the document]
Vifon: Good idea, I'll add it in a minute.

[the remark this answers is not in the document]
Vifon: Done, rotation added.

PR #1228, opened 2018-07-08
Update plugin_pmount.py

natemaia:
Currently the `pmount` plugin uses static device letters and partition numbers. this change uses present devices and partitions instead.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version:  Arch Linux x86_64
- Terminal emulator and version: st 0.8.1
- Python version: 3.6.6 (default, Jun 27 2018, 13:11:40) [GCC 8.1.1 20180531]
- Ranger version/commit: ranger-master 1.9.1
- Locale: en_CA.UTF-8

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
- Changes the binds to remove the need for uppercase device letters, if a device has no partitions <alt><m or M><device letter> mounts/unmounts the device.
- With only one partition the same binds will handle it instead of the whole device.
- With greater than one partitions, map <alt><m or M><device letter><partition number> for each.

This is still not perfect/fully functional, devices not present when ranger is started *(usb drives)* won't have a map created unless ranger is restarted :| ... Is there a simple way to have ranger 'reload' this plugin, or use a different hook, etc.?

#### MOTIVATION AND CONTEXT
Personal use, avoiding mappings to non-existent devices.


Vifon:
Was it tested with both Py2 and Py3? (my standard question whenever I see operations on strings involving encode/decode)

I don't see any way to plug into `:reset` or `:reload`. Can you submit this file as a new file (just a new commit restoring the old version and moving this one, no need to rewrite history or create a new PR)? This way the end user can select which drawback they prefer and we can polish it all later.

natemaia:
Hey @Vifon 

In Python 2.7.15 I had no issues.

I've added the original back and renamed the PR version as you suggested.

Cheers

Vifon:
Great! Thanks!

PR #1229, opened 2018-07-10
Change the method of defining colorschemes

parmort:
Use dictionaries instead of method redefinition.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: `Arch Linux x86_64 4.14.51-1-lts`
- Terminal emulator and version: `termite v13`
- Python version: `3.6.5`
- Ranger version/commit: `ranger-master 1.9.1`
- Locale: `en_US.UTF-8`

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
- Moved logic from `ColorScheme` sub-classes to `ranger.gui.ColorScheme.use`
- Created a dictionary in `ranger.gui.ColorScheme` called `colors`
- Moved tag-color_tuple relationship into aforementioned dictionary `colors` in the sub-class

#### MOTIVATION AND CONTEXT
This makes defining colorschemes easier for someone just getting into making colorschemes. It also shields the logic from the colorscheme creator, making it seem less of a hack or an afterthought.

#### TESTING
All current tests have passed with no changes necessary to the rest of the codebase. Good job @hut for making ranger very decoupled!

### DIFF
--- a/ranger/colorschemes/default.py
+++ b/ranger/colorschemes/default.py
@@ -1,167 +1,104 @@
-# This file is part of ranger, the console file manager.
-# License: GNU GPL version 3, see the file "AUTHORS" for details.
-
 from __future__ import (absolute_import, division, print_function)
 
 from ranger.gui.colorscheme import ColorScheme
 from ranger.gui.color import (
-    black, blue, cyan, green, magenta, red, white, yellow, default,
-    normal, bold, reverse,
-    default_colors,
+    black, blue, cyan, green, magenta, red, yellow,
+    bold, reverse
 )
 
 
 class Default(ColorScheme):
     progress_bar_color = blue
+    default_colors = (None, None, None)
 
-    def use(self, context):  # pylint: disable=too-many-branches,too-many-statements
-        fg, bg, attr = default_colors
-
-        if context.reset:
-            return default_colors
-
-        elif context.in_browser:
-            if context.selected:
-                attr = reverse
-            else:
-                attr = normal
-            if context.empty or context.error:
-                bg = red
-            if context.border:
-                fg = default
-            if context.media:
-                if context.image:
-                    fg = yellow
-                else:
-                    fg = magenta
-            if context.container:
-                fg = red
-            if context.directory:
-                attr |= bold
-                fg = blue
-            elif context.executable and not \
-                    any((context.media, context.container,
-                         context.fifo, context.socket)):
-                attr |= bold
-                fg = green
-            if context.socket:
-                fg = magenta
-                attr |= bold
-            if context.fifo or context.device:
-                fg = yellow
-                if context.device:
-                    attr |= bold
-            if context.link:
-                fg = cyan if context.good else magenta
-            if context.tag_marker and not context.selected:
-                attr |= bold
-                if fg in (red, magenta):
-                    fg = white
-                else:
-                    fg = red
-            if not context.selected and (context.cut or context.copied):
-                fg = black
-                attr |= bold
-            if context.main_column:
-                if context.selected:
-                    attr |= bold
-                if context.marked:
-                    attr |= bold
-                    fg = yellow
-            if context.badinfo:
-                if attr & reverse:
-                    bg = magenta
-                else:
-                    fg = magenta
-
-            if context.inactive_pane:
-                fg = cyan
-
-        elif context.in_titlebar:
-            attr |= bold
-            if context.hostname:
-                fg = red if context.bad else green
-            elif context.directory:
-                fg = blue
-            elif context.tab:
-                if context.good:
-                    bg = green
-            elif context.link:
-                fg = cyan
-
-        elif context.in_statusbar:
-            if context.permissions:
-                if context.good:
-                    fg = cyan
-                elif context.bad:
-                    fg = magenta
-            if context.marked:
-                attr |= bold | reverse
-                fg = yellow
-            if context.frozen:
-                attr |= bold | reverse
-                fg = cyan
-            if context.message:
-                if context.bad:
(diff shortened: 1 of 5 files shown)


parmort:
I messed up my commit history. Sorry.

ranger/colorschemes/default.py
```diff
@@ -1,167 +1,104 @@
-# This file is part of ranger, the console file manager.
```
toonn: Why drop the GPL header?

ranger/colorschemes/default.py
```diff
@@ -1,167 +1,104 @@
...
-            elif context.vcsstaged:
-                fg = green
-            elif context.vcssync:
-                fg = green
-            elif context.vcsignored:
-                fg = default
-
-        elif context.vcsremote and not context.selected:
-            attr &= ~bold
-            if context.vcssync or context.vcsnone:
```
toonn: This is what I mentioned in the review.

parmort:
What sort of filetype were you thinking about? JSON would work, and if we use [PyYAML][1], YAML would also work.

[1]:https://pyyaml.org/wiki/PyYAMLDocumentation

toonn:
I don't think either JSON or YAML are powerful enough to fully describe colorschemes tbh.
Someone'd have to distill everything that can be colored, every way it can be colored and a reasonable portion of how you can determine how to color things. Right now we support all of python for that last part, giving up all of that seems too restrictive.

However I do agree that a simpler way to start making colorschemes is useful. So I suggest you implement a colorscheme that is just a class that takes a dictionary. Then you can write the colorschemes you want to write by just writing the dictionary then calling DictionaryColorscheme(dictionary) to create the colorscheme.
That way the current powerful colorschemes are unchanged and people can use your colorscheme to make new simple ones.

That would actually be a really good PR for our new colorschemes repo. We could further debate including it by default and translating our current colorschemes to use it in that PR.

parmort:
So I've started work on the class, but I ran into an error:
```
TypeError: __init__() missing 1 required positional argument: 'dictionary'
```

I couldn't find a fix, but I was wondering if anybody could look over my code? You can find it [here](https://github.com/parmort/ranger/tree/colorscheme_issue)

toonn:
Actually, you could just provide a default paramater for the directory. That way it's actually a valid colorscheme and you can make another valid colorscheme by passing a different dictionary.

toonn:
Give your `__init__` method `dictionary` parameter a default value, either something boring, something you'd like or the default theme values'll do.
That way you shouldn't get the error about a missing argument but people can still use it by creating a class that passes a different dictionary.

Actually this seems like a good application for inheritance. Just define the dictionary as a field instead of an argument to `__init__` then people can just inherit and override the field.

parmort:
I personally like the method of doing something like this:
```python
colors = {
    # Stuff in here
}

Scheme = DictionaryScheme(colors)
```

and having it automagically create the scheme. It keeps the scheme simplistic, and all the more advanced writers have access to writing their colorschemes with the old method of class overwriting.

toonn:
That's fine by me. Not much simpler than using inheritance though, imo:
```python
# NewDictionaryColorscheme.py

class NewDictionaryColorScheme(DictionaryScheme):
    colors = { #new stuff }
```

Advantage of the inheritance approach is that you could implement defaulting to `super`'s dictionary I think. So people could inherit from a `DictionaryScheme` and just overwrite a couple colors.

parmort:
Isn't that what I had earlier? The users could redefine/override a default dictionary, and then if needed/wanted, redefine `use(...)`.

```python
# NewScheme.py

class NewScheme(ColorScheme):
    colors = {...}

    # optional, a default function is provided in ColorScheme
    def use(...):
        ...
```

If you wanted, I could move all this to a wrapper class `DictionaryScheme` and keep the base class the same. To me, however, that seems like a whole lot of baggage when we could define defaults.

toonn:
No, it's not. In your example you have to replace the entire dictionary. In the inheritance case you can provide a subset of the required dictionary and have missing keys looked up in the superclass.

toonn:
We should definitely keep the base class intact btw. You're changes might break every colorscheme that already exists.

parmort:
Ok. So create a wrapper class in `ranger.colorschemes` looking something like this:

```python
# ranger/colorschemes/dictscheme.py

from ranger.gui.colorscheme import ColorScheme

class DictScheme(ColorScheme):
    colors = {}

    def use(...):
        .
        .
        .

# ~/.config/ranger/colorschemes/example.py

from ranger.colorschemes.dictscheme import DictScheme

class MyScheme(DictScheme):
    colors = {...}

    # optional
    def use(...):
        .
        .
        .
```

toonn:
Yes, and possibly factor out all the bits and pieces in `use()` so they can be individually overridden.

parmort:
Like some functions?
```python
def use(...):
    if context.directory:
        fg, bg, attr = directory()

def directory():
    return (fg, bg, attr)
```

Or maybe something else?

toonn:
I was thinking have the conditionals in there but then call `directory()` to determine the colors. That way you can change the way a directory is colored depending on its `ctime` or `size` or the levenshtein distance to your favorite rockband : )

toonn:
Labeled this as next-release but actually this should go into our colorschemes repo. Could you PR it over there? Feel free to include changing the README in the PR so it has a section "Easiest way to customize colorschemes" (or something along those lines) explaining how to implement a new one based on this work.

We might at a later time include selected colorschemes from that repository by default. (We probably will as long as there's no issues, not like a couple hundred lines of colorschemes are gonna inflate our install size or anything. We just haven't discussed this yet.)

PR #1231, opened 2018-07-15
Use Pipfile/pipenv to manage the linters

Vifon:
We don't want the linters' behavior to suddenly change like right now after the release of pylint 2.0.


Vifon:
Crap, I didn't mean to merge it yet! Sorry everybody!

PR #1235, opened 2018-07-18
Update PYTHON variable

hektr:
`which python3` will likely show the latest version of python installed on system, so, it should go the first.
I think, it's not necessary to write all of possible versions of python. You can just use `which python3 || which python`...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,16 @@ NAME_RIFLE = rifle
 VERSION_RIFLE = $(VERSION)
 SNAPSHOT_NAME ?= $(NAME)-$(VERSION)-$(shell git rev-parse HEAD | cut -b 1-8).tar.gz
 # Find suitable python version (need python >= 2.6 or 3.1):
-PYTHON ?= $(shell python -c 'import sys; sys.exit(sys.version < "2.6")' && \
-	which python3 || which python3.6 || which python3.5 || which python3.4 || \
-	which python3.3 || which python3.2 || which python3.1 || \
-	which python || which python2.7 || which python2.6)
+PYTHON ?= $(shell \
+	     (python -c 'import sys; sys.exit(sys.version < "2.6")' && \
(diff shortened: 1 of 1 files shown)


toonn:
Hmm, not sure why they're all specified. Wouldn't `which python` select the most recent python? At least on arch it should already be py3 I believe.
I'm not completely convinced we want to prefer py3 over py2 anyway, could lead to breaking behavior for people.
I'm gonna wait for input from someone else on the team on this.

hektr:
Yes, you are right. In my debian machine ranger works better with python3. When I've reinstalled  ranger with command `sudo python3 setup.py install --optimize=1 --record=install_log.txt` the issue with one-column previewing was disappear ^_^  In debian `which python` selects the second version of python. And for using third version you should use `which python3`. I think, with time, ranger dependencies will turn to third python and if you don't prefer 3 version of python, it can cause some problems in future ^_^

toonn:
What do you mean? The standard command installs ranger for `python` why would users install dependencies manually assuming ranger uses `python3`?

While most may prefer py3, debian clearly doesn't yet and it's undeniably the most influential distro imo. I'd stick with installing it for `python` and let the distros decide when to switch.

I'm still not clear on why all those versions are listed though, why would `python` be absent but `python3.3` be present for example? Sounds like a weird setup. The comment also says either `>= 2.6` or `>= 3.1` and only the former is enforced.

hektr:
so, you can delete all `which pythonN`s except of just `which python`.

toonn:
According to @Vifon we probably need `which python || which python3` at least, because ubuntu installs `python3` without the `python` name.

Vifon:
Don't we need to handle python 3.7 too? Apart from this, looks fine to me.

Vifon:
Agreed. I'd go with `python3 || python || python`.

PR #1240, opened 2018-07-30
Update rifle

hektr:
I added djview for opening djvu files.
I open my djvu file using djview. May be some other people too. So, it can be useful. 

You can also use okular or qpdfview for opening djvu files, if djvulibre is installed on your system ^_^ But I didn't try this, so I am not sure about this))


toonn:
Zathura opens djvu files too.

Vifon:
@hektr, so do you still consider djview useful after discovering Zathura? Just making sure whether I should close or merge this one. :)

hektr:
Today Zathura is not working(error: Unknown file type: 'image/vnd.djvu'). And it could print only all/even/odd pages, and in djview I can select exact pages for printing. I think, merging is better way)

PR #1244, opened 2018-08-06
Replace StopIteration with bare return

yannick-c:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT

- Terminal emulator and version: rxvt-unicode v9.22
- Python version: 3.7.0
- Locale: US

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

A bare `return` in generators raises `StopIteration` (this is backwards
compatible). Using `StopIteration` raised a warning in 3.6.x and raises an
Error in 3.7.0 (and probably upwards). So this is a compatibility patch that shouldn't be too controversial (even the `try/except` blocks on `StopIteration` should still work, since the same error is raised.

See https://docs.python.org/3/reference/simple_stmts.html#grammar-token-return_stmt

#### MOTIVATION AND CONTEXT

Ranger crashes on Python 3.7.0 currently.

The crashes.

#### TESTING

All (the pylint ones obviously give a lot of warnings).

They don't, as the bare return statement still raises a `StopIteration`. I could not find any other `StopIteration` in generators, but I might have missed some.

### DIFF
--- a/ranger/gui/widgets/pager.py
+++ b/ranger/gui/widgets/pager.py
@@ -234,7 +234,7 @@ class Pager(Widget):  # pylint: disable=too-many-instance-attributes
     def _generate_lines(self, starty, startx):
         i = starty
         if not self.source:
-            raise StopIteration
+            return
         while True:
             try:
                 line = self._get_line(i).expandtabs(4)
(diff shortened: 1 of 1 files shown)


toonn:
Argh python, why are you changing so much lately?

Thanks for the PR! It's a great first contribution :tada:

PR #1276, opened 2018-08-22
Implement image previewing for archive files

jakanaka-evan:
#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Implemented archive previews in scope.sh. It uses the first alphabetically
sorted image file in the archive or falls back to text-based preview methods if
there are none.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -129,24 +129,17 @@ handle_image() {
         #     application/x-xz|application/x-bzip2|application/x-gzip|application/x-tar)
         #     local fn=""
         #     local bsd=""
-        #     local zip=""; local rar=""; local z7=""; local tar=""
-        #     case "${MIMETYPE}" in
+        #     local zip=""; local rar=""; local tar=""
+        #     case "${mimetype}" in
         #         application/zip) zip=1 ;;
         #         application/x-rar) rar=1 ;;
-        #         application/x-7z-compressed) z7=1 ;;
+        #         application/x-7z-compressed) ;;
         #         *) tar=1 ;;
         #     esac
-        #     [ "$tar" ] && fn=$(tar --list --file "${FILE_PATH}")
-        #     [ $? != 0 ] && bsd=1 && fn=$(bsdtar --list --file "${FILE_PATH}")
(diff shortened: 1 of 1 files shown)


toonn:
Hmm, interesting. What archive formats does it support? And what are the optional deps?

Could be nice for comic book previews for example.

jakanaka-evan:
For now it only supports `.rar` and `.zip` files. Opt-deps are
 - for `.rar` files: `unrar OR bsdtar`,
 - for `.zip` files: `zipinfo (from unzip) AND atool OR bsdtar`.
I'll try to add `.7z`, `.tar`, and `.cbz`support today, and look into removing
the `atool` dep for `.zip` files.

`atool --list` is unfortunately not very suitable for this, since it lists
files differently for different formats.

toonn:
Don't forget `.cbr`, that's still popular for comics. `.tar.gz & .tar.bz2` would be nice, oh and `.xz`.

jakanaka-evan:
Here we go, added support for all of the requested formats and removed the `atool` dependency.
See jakanaka-evan/ranger@8fd92581dfcd126ce432b8832209f8cdf4a73ecd for more information on optional dependencies.

toonn:
Hmm, one more suggestion. Can you check mimetype  instead of using just the extension to determine something's a picture? Maybe not if you only get the filename. The python mimetypes module does mimetype by extension but maybe running python just for that is overkill.

Let's add at least `tiff, bmp and webp` though.

jakanaka-evan:
Changed `grep` into a `python -c "import mimetypes as m; ..."` command. It
obviously performs slower than a grep, but we are listing archive contents
prior to that anyway, so relatively speaking it's not that worse I think.

jakanaka-evan:
I've fixed up everything I could think of and made it pass shellcheck, so i
think it's ready for merge now.

Vifon:
It needs a comment describing what it is actually doing (i.e. previewing the first file in an archive). Apart from, if it works for you, @toonn, merge away!

toonn:
🎉 Thanks for the contribution!

Issue #451, opened 2016-01-16
w3mimgdisplay consumes a lot of memory


soredake:
After scrolling a 20-30 images it consumes up to 2gb+ memory.
![1452941014](https://cloud.githubusercontent.com/assets/5204968/12371966/fb2e6cf4-bc4e-11e5-8ecf-dcaa536598a7.png)
Related https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=218857

hut:
Yep, it keeps a cache. I think this should really be fixed in w3m rather than in ranger. :S

denzerdhard:
Add -gtk flag to w3m if using gentoo, this fixed it for me.

fenque:
Thanks, fixed consumtion.

Ambrevar:
Since I don't have a gentoo at hands just now, I had a look at what the -gtk USE flag does:
https://gitweb.gentoo.org/repo/gentoo.git/tree/www-client/w3m/w3m-0.5.3-r8.ebuild
I tried to compile with the `--with-imagelib=gtk2` configure option to no avail. Any suggestion?

denzerdhard:
No you need opposite without gtk but with imlib2. Try --with-imagelib=imlib2

Ambrevar:
That's what I had before, still the same issue.

mrogalski:
Gentoo's ebuild also sets the `--enable-image=imlib2` flag along with `--with-imagelib=imlib2`.

Ambrevar:
Tried different options, did not work for me. When only imlib2 is set, it crashes and nothing renders in ranger.
I'm currently running Arch Linux with the following compilation flags:

CPPFLAGS="-D_FORTIFY_SOURCE=2"
CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong"
CXXFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong"
LDFLAGS="-Wl,-O1,--sort-common,--as-needed,-z,relro"

mrogalski:
I'll include the full configure from my gentoo installation.  Maybe this will help You somehow?

[...]

Issue #859, opened 2017-04-26
ranger won't refresh preview column when previewing images and files


x4121:
Similar to #811, I get some strange effects when previewing images.
Bug stays even after deleting old ranger version (installed with apt) and building new version from current master.
Deleted `~/.config/ranger` directory, ran `ranger --copy-config all` and changed `preview_images` to true.

#### ISSUE TYPE
- Bug report

#### RUNTIME ENVIRONMENT
- Operating system and version: Ubuntu-Gnome 17.04
- Terminal emulator and version: konsole 16.12.3
- Python version: Python 3.5.3
- Ranger version/commit: ranger-master 1.9.0b5
- Locale: en_US.UTF-8

#### EXPECTED BEHAVIOR
- image preview redraws over previous images
- image preview "clears" after closing ranger

#### CURRENT BEHAVIOR
- some images are cropped
- small images are drawn over previous images
- image parts stay after closing ranger if terminal contains long texts

#### STEPS TO REPRODUCE
(see video)

#### TRACEBACK

```
Traceback (most recent call last):
  File "/usr/local/bin/ranger", line 41, in <module>
    sys.exit(ranger.main())  # pylint: disable=no-member
  File "/usr/local/lib/python3.5/dist-packages/ranger/core/main.py", line 222, in main
    fm.ui.destroy()
  File "/usr/local/lib/python3.5/dist-packages/ranger/gui/ui.py", line 171, in destroy
    DisplayableContainer.destroy(self)
  File "/usr/local/lib/python3.5/dist-packages/ranger/gui/displayable.py", line 291, in destroy
    displayable.destroy()
  File "/usr/local/lib/python3.5/dist-packages/ranger/gui/displayable.py", line 291, in destroy
    displayable.destroy()
  File "/usr/local/lib/python3.5/dist-packages/ranger/gui/widgets/pager.py", line 71, in destroy
    self.clear_image(force=True)
  File "/usr/local/lib/python3.5/dist-packages/ranger/gui/widgets/pager.py", line 60, in clear_image
    self.fm.image_displayer.clear(self.x, self.y, self.wid, self.hei)
  File "/usr/local/lib/python3.5/dist-packages/ranger/ext/img_display.py", line 146, in clear
    self.process.stdin.flush()
BrokenPipeError: [Errno 32] Broken pipe

```

#### IMAGES / VIDEOS
[...]

PR #1284, opened 2018-08-25
support ueberzug image displayer

seebye:
Überzug is a command line util which allows to draw images on terminals by using child windows.
It has some advantages over w3mimgdisplay, e.g.:
* it has no memory leak (/ unlimited cache) (related: https://github.com/ranger/ranger/issues/451)
* no race conditions as a new window is used to display images
* tmux support
* faster as the current w3mimgdisplay implementation as the communication goes only in one direction

More: https://github.com/seebye/ueberzug#%C3%9Cberzug

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Debian Buster
- Terminal emulator and version: xterm / alacritty
- Python version: 2.7.15 (default, Jul 28 2018, 11:29:29) [GCC 8.1.0]
- Ranger version/commit: ranger-master v1.9.1-173-gd37e4b2c
- Locale: None.None

#### CHECKLIST

    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

w3mimgdisplay does not work with alacritty, tmux is also not supported


toonn:
Looks pretty cool for a first contribution. Does überzug only support python 3.x? That's somewhat unfortunate. If it's not could you  run it with the same python that's running ranger, or just the python on the `PATH`?

seebye:
> Does überzug only support python 3.x? That's somewhat unfortunate. If it's not could you run it with the same python that's running ranger, or just the python on the PATH?

I used python3 features like asyncio to avoid dealing with thread safety, so it won't run with python2.

toonn:
Is the threading module an option, afaik that's been backported?

seebye:
@jealpark Termite is not in the debian testing repository, but I guess that termite does not support extracting the terminal size in pixels via termios.TIOCGWINSZ.  
Therefore the font size can't be determied.  
Another way to optain the terminal size isn't implemented currently, but it's easily fixable by using python-Xlib.

@toonn Honestly, I don't really wan't to support python2.  
Python3 should be installed on the most linux systems anyways and
if it's not it shouldn't be a problem as people were already willing to install an entire browser to display images with ranger.
So it would just increase the effort to maintain ueberzug in my opinion.

toonn:
Entire browser? w3m-img doesn't depend on w3m.

jaelpark:
@seebye I created a quick fix and opened a PR, see if it's something you want to merge. At least the problem with termite is fixed. This could be configurable, whether to get the dimensions with Xlib or using the original method.

toonn:
This doesn't work for me. `ueberzug` doesn't work on the command line either. Using a virtualenv with python  3.7 and the terminology terminal emulator.
ueberzug_corrupted_output

seebye:
In the newest version there are some changes for python3.7.  
(The preview worked for me after these modifications in terminology in a python3.7 virtualenv)  
However your error message looks like the window lacks the _NET_WM_PID property?  
Ueberzug can't work without it as it uses the pid to determine the window id.

toonn:
I have no idea why the property would be missing. No funny business on my end afaik.

seebye:
I think it should work with version 6 if it's not terminology itself which lacks the property.

seebye:
> I really think this should be fixed before we include the method in the release.
On the latest version of ranger the preview image also stays if you switch from terminology to e.g. w3m.
So I also implemented it for the other preview methods.

toonn:
For me the image never stayed when I switched from terminology to another method. If you did have that problem then that sounds like a good fix though.

seebye:
I enabled the redirection of stderr to /dev/null.  
So errors should be suppressed now.

toonn:
This is an awesome first contribution. Thanks for the work. I'm gonna allow for a little time for another team member to check out the changes but this is probably going in the next release. :+1:

Vifon:
Looks really interesting! Sadly I cannot get it to work. I'm using `set preview_images_method ueberzug` and I'm running ranger with Python 3. Anything else I need to know? Can a tiling window manager (XMonad) be the cause?

Vifon:
It also needs documentation in the rc.conf comment and a new allowed value [here](https://github.com/ranger/ranger/blob/cc6d9a0b19d250215b822342436e8e81e71c4b4d/ranger/container/settings.py#L106).

toonn:
@Vifon, you did actually install ueberzug first, right? And good catch about the documentation and the `ALLOWED_VALUES`, although that seems to be a misnomer because I was able to test and afair tab complete the ueberzug method without any problem : )

seebye:
> Sadly I cannot get it to work. I'm using set preview_images_method ueberzug and I'm running ranger with Python 3. Anything else I need to know?  

As toonn already mentioned you need to install ueberzug (`sudo pip3 install ueberzug`) and change preview_images_method to ueberzug, that's all.  
You can test whether ueberzug works by executing the following bash commands:
```bash
$ source "`ueberzug library`"
$ { ImageLayer::add [identifier]="example0" [x]="0" [y]="0" [path]="/some/path/some_image0.jpg" ; read ; } | ImageLayer
```

> Can a tiling window manager (XMonad) be the cause?

Works fine for me with xmonad in Xephyr & xterm.

@toonn I updated rc.conf and added ueberzug to preview_images_method in ALLOWED_METHODS.

Vifon:
@seebye I managed to get it working, no idea why it didn't work before. Works really well but has a few serious bugs too:

- When I closed ranger with <kbd>q</kbd>, the black ueberzug window stayed open and obstructed a part of my terminal.
- Even after I closed the terminal, the process persisted and resisted my SIGTERMs, I needed to use SIGKILL. The same happens when I run `ueberzug layer` (I tried it back then, when trying to use it manually, outside of ranger), all the signals are caught and dropped.

I'm happy to merge this image preview method in the future when I'll be more robust. Right now it needs some work. It's still pretty amazing and I'm sure it has some future. :)

toonn:
Just noting here ueberzug doesn't work on mac OS. This should be added to the documentation.

seebye:
> When I closed ranger with q, the black ueberzug window stayed open and obstructed a part of my terminal.
@Vifon 
1. Does this happen on each try? 
2. Does this happen if you try to view a really large image?  
If (1) is false it sounds like Xlib hangs on signal.signal
(I guess it as there are already workarounds to avoid it).  

> Even after I closed the terminal, the process persisted and resisted my SIGTERMs, I needed to use SIGKILL.

Sounds like a death lock so I guess my guess is correct.

rtng1:
OS: Debian GNU/Linux 9.5 (stretch) x86_64 
WM: i3 ; python2.7 ; rxvt-unicode (urxvt) v9.22
looked ok but next start it was w3m again

toonn:
@rtng1, yep that's how ranger's settings work. If you want to change it permanently you have to edit `rc.conf`. Thanks for testing : )

Vifon:
@seebye 

1. No, not on every try. I cannot reproduce it at the moment.
2. I don't remember which files I tested on, they may have been wallpaper-sized. I tried large and small right now, with no bug visible.
3. No, it didn't seem to freeze.

I'll open an issue in your repo if I get more details.

rtng1:
bildschirmfoto_2018-09-09_18-44-15
 it works with xfce-terminal and terminator smoothly! finally i see ranger in bright colours. nice nice!

toonn:
@rtng1 you can customize the colors most any terminal uses.

rtng1:
@toonn surely thats true, but as a not-developer and as a just-curious-user i didnt manage to intensify colours of urxvt (for example). but i would never want to miss those! i have to admit, when it comes to configurations, xfce and terminator seems to be more userfriendly to me.  :-) without urxvt and its possibilities i would not have fall in love with ranger i guess... nice work you folks do, thanx a lot.

remlap:
Am I correct in saying this wont work with zsh?

remlap:
> > Am I correct in saying this wont work with zsh?
> 
> No, the bash library seems not to work with zsh.

Thanks for getting back to me

remlap:
> > Am I correct in saying this wont work with zsh?
> 
> @remlap No, the bash library seems not to work with zsh.

@seebye Is there any chance of support in the future?

seebye:
>  Is there any chance of support in the future?

@remlap Well, at least I didn't plan to do so.  
Bash is only supported to make small simple scripts like showing images while using fzf.  
I choose to support bash as it's installed on (almost) every distribution.  
Also there seems to be conflicts as zsh uses a different syntax for associative arrays.  
```bash
declare -A foo=( [key1]=value1 [key2]=value2 )
```
vs
```zsh
declare -A foo=(key1 value1 key2 value2)
```
```shell
function foo {
    local -A bar="( $@ )" # works in bash
    local -A bar=( "$@" ) # works only in zsh
}
```

The library depends on it to  
1. make the code more readable  
```bash
# $1 $2 $3 ...
ImageLayer::add 55 79 20 /home/foo/bar.png
# vs
# $data[x] $data[y] $data[max_width] ...
ImageLayer::add [x]=55 [y]=79 [max_width]=20 [path]=/home/foo/bar.png
```
Shell code is nice to use on the go, but code like `$1`, `grep -oP '[0-9]'`, `ls -laZ` isn't readable at all.  
If you know what it does you know it only as your using it frequently.  
2. support some sort of keyword arguments (+ optional parameters) which can simply be passed to ueberzug.
3. make it simple creating this library as ueberzug parses the output of `declare -p associative_array`

toonn:
@remlap, note that this doesn't mean ueberzug image previews wouldn't work in ranger started from zsh.

remlap:
> @remlap, note that this doesn't mean ueberzug image previews wouldn't work in ranger started from zsh.

It works! Awesomeness! Thank you seebye for creating this a stable way to view images in mate-terminal with ranger and toonm for saying it would work with ranger with zsh.

seebye:
> When I closed ranger with q, the black ueberzug window stayed open and obstructed a part of my terminal.

I couldn't reproduce this till now, so I decided to wait for ueberzug to exit with a timeout of one second.  
If the timeout occurs ueberzug will be killed with SIGINT.  
I also updated the documentation and mentioned that a user needs to run X11 in GNU/Linux.

toonn:
I've lost track of this issue tbh. Could you recap whether you addressed all of @Vifon's remarks?

Why the timeout, is there any harm to sending SIGINT to a process that's already exiting? Also, wouldn't SIGTERM be more appropriate or is that rarely used?

seebye:
> I've lost track of this issue tbh. Could you recap whether you addressed all of @Vifon's remarks?

That's the only thing he mentioned:  
https://github.com/ranger/ranger/pull/1284#issuecomment-419544108  
And you stated that I should extend the documentation with the required operating system:  
https://github.com/ranger/ranger/pull/1284#issuecomment-419546507

> Why the timeout, is there any harm to sending SIGINT to a process that's already exiting? 

Ueberzug registers tmux hooks for the pane it runs in.  
These hooks will be unregistered as soon as ueberzug ends.  
(They are required to  for example notice new clients displaying ranger)  

> Also, wouldn't SIGTERM be more appropriate or is that rarely used?

After thinking about it you're right with the current implementation SIGKILL will also be send if ueberzugs queue is filled and needs more than one second to be processed.  
(E.g. user scrolls though a lot large images)  
With the latest commit I replaced it with a SIGTERM after a second and a SIGKILL after two in case the SIGTERM achieved nothing.  
I implemented it this way, because Vifon said that only SIGKILL was able to end ueberzug.  

> Even after I closed the terminal, the process persisted and resisted my SIGTERMs, I needed to use SIGKILL. The same happens when I run ueberzug layer (I tried it back then, when trying to use it manually, outside of ranger), all the signals are caught and dropped.

It seems like neither he nor I could reproduce this issue till now as he didn't opened an issue.
> No, not on every try. I cannot reproduce it at the moment.
> [...]
> I'll open an issue in your repo if I get more details.

https://github.com/ranger/ranger/pull/1284#issuecomment-419674815

seebye:
@toonn Sorry, I meant SIGKILL.

remlap:
> @jaelpark, @rtng1, @remlap, this feature is a about to be merged soon. Could you test this as your main preview method for a day or two? If no one reports an issue we can add a great new preview method for everyone to enjoy : )

My only problem is it goes out of frame ever so slightly with landscape pictures. I fixed it by changing to:

`x=start_x-1,`

In img_display.py

toonn:
@remlap, thanks for testing so quickly. These are exactly the sort of wrinkles I'd like to smooth out before merging.

toonn:
Hmm, I still get tons of tasks launched for `.svg` previews. It doesn't make ranger quite unusable but it's still a little annoying.
It's not limited to the `ueberzug` method though. I get almost the same result with the `terminology` method. Although that at least manages to actually flash a preview of the svg from time to time.

@seebye, any idea what the problem might be? Must be in some code common to more than one method.

It would be great if we could fix this but it's not a show stopper for this PR imo.

Zaeph:
Hello,

`ueberzug` is working pretty well on my end with `terminator`, and it even fixes `w3m`’'s issue of stacking previews mentioned in #859.

However, I’m encountering a similar problem (also mentioned in #859) with previews overlapping with the path of the file. I’m pretty sure it has to do with the way `w3m` and `ueberzug` get coordinates from `terminator`.
Here's a screenshot:
screen2

The problem is somewhat alleviated by setting `draw_borders` to `true`, but it’s still off:
screen3

And here’s a couple of screenshots with w3m for reference:
screen1

screen4

seebye:
> Not a problem this one, just to say its working fine for me, I am getting a warning of multi pane windows are not supported in tmux but I am seeing my pictures fine.

That's not a bug, tmux does not support as many events as ueberzug requires to support multi pane windows.  
E.g. it's impossible to detect pane swaps.  

> My only problem is it goes out of frame ever so slightly with landscape pictures. I fixed it by changing to: x=start_x-1,

The offset shouldn't be changed, it's better to reduce the image size.  
I don't know how rangers determines the size of the preview pager.  
E.g. whether it includes the border size etc.  
I just used the passed width & height.

> However, I’m encountering a similar problem (also mentioned in #859) with previews overlapping with the path of the file.

I don't know, so: is the titlebar part of the window?  
In this case the y position and the guessed font size will be wrong.

> Hmm, I still get tons of tasks launched for .svg previews. It doesn't make ranger quite unusable but it's still a little annoying.

Should be fixed in version 9.

Zaeph:
> I don't know, so: is the titlebar part of the window?
> In this case the y position and the guessed font size will be wrong.

And you were absolutely right: disabling `terminator`’s title-bar resolves the issue. Thanks!

I don’t think it’s worth the effort to find a way to account for those title-bars. I’m not sure if other terminal emulators feature them, but they’re a nice alternative to regular X-windows titles when you’re hiding those.

Here’s a couple of screenshot to confirm the proper behaviour:

screen5
screen6

toonn:
The width and height ranger passes to the draw method are the maximum number of columns respectively rows available for the preview. So as long as ueberzug doesn't go past those dimensions it should be fine.

@remlap, we recently added a `w3m_offset` setting because some terminals report borders differently, maybe check if that helps?

remlap:
> @remlap, we recently added a `w3m_offset` setting because some terminals report borders differently, maybe check if that helps?

I don''t think I have that option from where I downloaded ranger from https://github.com/seebye/ranger

toonn:
@remlap, sorry it's `w3m_offset` and it's in the PR branch because master was recently merged in.

seebye:
> The width and height ranger passes to the draw method are the maximum number of columns respectively rows available for the preview. So as long as ueberzug doesn't go past those dimensions it should be fine.

I think the latest improvements (in version 10) of the font size guessing solve this issue.  
(As the font size is guessed there's of course still room for incorrect sizes)

remlap:
Just grabbed version 10 from pip and all working nice and lined up for landscape photos.

toonn:
@seebye, maybe have a look at how other methods try to figure out the fontw and fonth?

seebye:
@toon I don't think that's possible.  
At least that's the best way I could think of till now.  
(Also the w3m implementation works in a similar way (https://github.com/ranger/ranger/blob/163c758f46d09e74c1da8f2d249fcd40e9b02afb/ranger/ext/img_display.py#L123), but ignores the padding)

This is how ueberzug guesses the font size currently:  
Known variables: xpixels, cols  

>      # let's assume the padding is the same on both sides:  
>      # let font_width = floor(xpixels / cols)  
>      # (xpixels - padding)/cols = font_size  
>      # <=> (xpixels - padding) = font_width * cols  
>      # <=> - padding = font_width * cols - xpixels  
>      # <=> padding = - font_width * cols + xpixels  

See: https://github.com/seebye/ueberzug/blob/master/ueberzug/terminal.py#L23
1. padding = 0: cols is a divider of xpixels
2. padding != 0:
2.1. padding * 2 >= font_width: font_width > real_font_width => xpixels / cols in [real_font_width + n, real_font_width + n + 1) where n in natural Numbers
2.2. padding * 2 < font_width: xpixels / cols in [real_font_width, real_font_width + 1) => font_width = real_font_width

This happens for the width and height, the maximum of both values will be assumed as the correct padding.  

So.. to guess the correct font_size and padding,  
1. the window needs to only display the text output  
2. the padding needs to be the same for all sides and  
3. the padding needs to be smaller than max(font_width, font_height).  

With the padding the font size will be guessed: (pixels - 2 * padding) / chars

toonn:
I was just waiting for some more testing. I'll go over it again tomorrow and merge unless I find something else. Are you sure the padding could be implemented from ueberzug's side though? The `w3m_offset` compensates for the terminal's padding so it shifts the preview left, out of the border.

jaelpark:
@toonn Been using it since it was announced, and with the exception of the one problem that was already fixed, it's been working great here. Thanks @seebye for your work, really nice to finally have an actually working image preview in a terminal emulator!

seebye:
> Are you sure the padding could be implemented from ueberzug's side though? 

@toonn Yes, ueberzug already determines the font size & padding by itself.  
The pull request code orders ueberzug to draw an image at (column c, row r) as ueberzug only supports chars as unit by now.  
So in the current version you couldn't implement an offset in ranger with pixels as unit.

@jaelpark You're welcome.

toonn:
@seebye, couple things, it seems like the previews don't use the entire available space? There's a bigger margin on the right of the images than the terminology method gives me.

And the svg problem wasn't fixed. Whenever I highlight an svg file tons of previewing tasks are spawned. I think this isn't specific to the ueberzug method but since you're looking at it anyway I figure you might be most suited to finding the problem right now.

seebye:
> it seems like the previews don't use the entire available space? There's a bigger margin on the right of the images than the terminology method gives me.

I removed the margin.

> And the svg problem wasn't fixed. Whenever I highlight an svg file tons of previewing tasks are spawned. 

I assumed that you meant processes with "tasks", so I catched the crash for unsupported file types and let ueberzug print an error message.

toonn:
Since the svg problem doesn't seem specific anyway I'm giving this my stamp of approval. 🎉 Thanks for the contribution! Hopefully ueberzug might replace the troublesome w3m for general use some day.

drozdowsky:
works great for me (st terminal, latest ranger with python 3.6)

linuxist:
> works great for me (st terminal, latest ranger with python 3.6)
> thanks

I am using st. Its not working for me. How to make it work?

toonn:
@linuxist, do you have both ranger and ueberzug installed for the same python?

aasril:
```
ranger --version
Python version: 3.5.2 (default, Nov 12 2018, 13:43:14) [GCC 5.4.0 20160609]
Locale: en_US.UTF-8
```

I'm using Ubuntu 16.04. Ueberzug is also installed for the same python version. I tried it with st and gnome-terminal and in both I get this error:
```
Task was destroyed but it is pending!
task: <Task pending coro=<prepare_display() running at /usr/local/lib/python3.5/dist-packages/ueberzug/xutil.py:47>>
```

toonn:
@seebye, is this an ueberzug issue?

seebye:
@aasril You should test whether e.g. (in bash)
```bash
source "`ueberzug library`"
{
    ImageLayer::add [identifier]="example0" [x]="0" [y]="0" [path]="/some/path/some_image.jpg"
    read
} | ImageLayer```
works.

**Comment by aasril** [comment]
@seebye 

Running the command print the same message as above (also the image is not displayed).

ueberzug_command

**Comment by seebye** [comment]
Does it print anything before this message?  
Sounds like ueberzug crashes, but this error message doesn't seem to describe the cause.  
Also how did you install ueberzug?

**Comment by aasril** [comment]
@seebye 

I should have scrolled further. Yes indeed there are messages before those repeated lines
```
  File "/usr/local/bin/ueberzug", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/ueberzug.py", line 225, in main
    main_layer(options)
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/ueberzug.py", line 181, in main_layer
    windows = batch.BatchList(window_factory.create(*window_infos))
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/ui.py", line 66, in create
    for info in window_infos]
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/ui.py", line 66, in <listcomp>
    for info in window_infos]
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/ui.py", line 164, in __init__
    self.create()
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/ui.py", line 233, in create
    self._set_click_through()
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/ui.py", line 285, in _set_click_through
    Xshape.SO.Set, Xshape.SK.Input, 0,
AttributeError: module 'Xlib.ext.shape' has no attribute 'SO'
```
I installed ueberzug by running `sudo -H pip install ueberzug`. It is updated, using Ubuntu 16.04 default python3.5.

**Comment by seebye** [comment]
It seems like you installed multiple versions of python-xlib some of them are outdated and all of them have the same name,  
so it doesn't matter if you installed the correct one.  
The right one is python-xlib.  
Likely python3-xlib (outdated) or / and xlib (outdated) is also installed on your system.  
You need to remove them (this could break other programs which uses them).

**Comment by aasril** [comment]
xlib is not installed but python3-xlib is. I removed it and than ran again the above command which result in this error:
```
  File "/usr/local/bin/ueberzug", line 6, in <module>
    from ueberzug.ueberzug import main
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/ueberzug.py", line 36, in <module>
    import ueberzug.xutil as xutil
  File "/usr/local/lib/python3.5/dist-packages/ueberzug/xutil.py", line 7, in <module>
    import Xlib
ImportError: No module named 'Xlib'
```

python-xlib is installed by package manager (v0.14). Then I removed python-xlib and installed again (v0.25) by using pip. Ran the command again and result in this error message:
[error-python-xlib.txt](https://github.com/ranger/ranger/files/2926462/error-python-xlib.txt)
Maybe because I'm running python3.5?

**Comment by seebye** [comment]
> Ran the command again and result in this error message: error-python-xlib.txt

@aasril Should be fixed in 18.1.3

**Comment by aasril** [comment]
@seebye You've already pushed it! :grin: 

Thanks a lot. I confirm that ueberzug now works for me both on terminal and on ranger.

**Comment by remlap** [comment]
Will ueberzug be supporting webp anytime soon?

**Comment by seebye** [comment]
@remlap Ueberzug supports the image formats which are supported by pillow-simd.

**Comment by sooriravindra** [comment]
I am using suckless terminal on Arch, ueberzug sample bash script displays image using the terminal. However ranger fails to display the image in preview.

Installed both using "sudo pip install ranger ueberzug"

**Comment by toonn** [comment]
You need the master version of ranger this is not in the release.

**Comment by sooriravindra** [comment]
I tried with the master version from github, same effect.

I also tried ranger-git in AUR, no luck :(

**Comment by toonn** [comment]
You did `set preview_images_method ueberzug`, right? Also, you're sure you were running the master version and not the installed one?

**Comment by sooriravindra** [comment]
Yes, my ~/.config/ranger/rc.conf has

```
set preview_images true
set preview_images_method ueberzug
```

And yeah I removed the earlier ranger and cloned master and did a make install. The preview window is blank.

**Comment by toonn** [comment]
@seebye, could you chime in? I have no idea why it'd work in the terminal but not ranger.

**Comment by ivomac** [comment]
@sooriravindra, I have the same setup as you, st, master version of git and ueberzug from pip. I can see preview images but only if they fit inside the preview pane. It seems like the images are not resized to fit.

Can you check if you can preview a low enough resolution image?

**Comment by sooriravindra** [comment]
@ivomac that seems like it, I was trying to preview images that don't fit in the preview window. Now when I tried a smaller image, the preview works.

The issue seems to be with resizing.

**Comment by toonn** [comment]
Hmm, maybe try uninstalling pillow-simd and installing pillow instead?

**Comment by ivomac** [comment]
I tried uninstalling pillow-simd and installing pillow but now ueberzug complains it is missing pillow-simd. Is there a way to tell it to use pillow?

This doesn't seem to be an issue with ranger. Opening an image with ueberzug on the terminal and forcing a resize using max_width and max_height gives a segmentation fault.

Used bash script:
```
#!/usr/bin/bash

source "`ueberzug library`"

ImageLayer 0< <(
	ImageLayer::add [identifier]="example" [max_width]="200"
[max_height]="200" [x]="0" [y]="0" [path]="$1"
	read
	ImageLayer::remove [identifier]="example"
)
```
images smaller than 200x200 are displayed correctly, larger ones give the error
```
/usr/lib/python3.7/site-packages/ueberzug/lib/lib.sh:
line 41: 11648 Segmentation fault (core dumped)
ueberzug layer -p bash "$@"
```

seebye:
> I tried uninstalling pillow-simd and installing pillow but now ueberzug complains it is missing pillow-simd. Is there a way to tell it to use pillow?

Try to install ueberzug by using pip it shouldn't complain then.  
(/ pillow and pillow-simd should be exchangeable if you install it with pip)
(Also someone made the effort to package ueberzug with pillow instead of pillow-simd.  
https://aur.archlinux.org/packages/python-ueberzug-nosimd-git/ )

ivomac:
python-ueberzug-nosimd-git worked perfectly. Thank you!

toonn:
@seebye, is there any chance you'd consider making `pillow` the default dependency? We're gonna get a bunch of issue reports about this when ueberzug makes it into the next release.

seebye:
@toonn Sure, I guess. I'm busy right now, so I'm likely doing it not until next week.

sooriravindra:
@seebye python-ueberzug-nosimd-git works great. However the default pip install complains if pillow-simd is not installed ( I had pillow installed).

Thank you!

eduardoejorqueras:
Hi, when I try to preview an image in ranger, at the bottom of the terminal appears
`[Errno 13] Permission denied 'ueberzug'`
Any hint? I have searched everywhere but I don't know what could I do.
I use gentoo, and if I do source "\`ueberzug library\`" before running ranger, nothing appears at the preview.

toonn:
What do you mean by source? Ueberzug is definitely not designed to be sourced by a shell. How did you install ueberzug?

eduardoejorqueras:
I installed following the [instructions](https://github.com/seebye/ueberzug#installation) from ueberzug's github, and by source, I mean [this](https://github.com/seebye/ueberzug#bash) in bash:
``` source "`ueberzug library`" ```

toonn:
Does that at least print the path to the library?

eduardoejorqueras:
Hey, suddenly, now it shows images whose resolution is smaller than the right side of the terminal, check this out:
image
But nothing happens with a bigger image:
image
And all I have as config is the [default config](https://github.com/ranger/ranger/blob/master/ranger/config/rc.conf) with ueberzug instead of w3m

sooriravindra:
@eduardoejorqueras Have you tried making Ueberzug to use pillow instead of pillow-simd?

eduardoejorqueras:
@sooriravindra thank you, I just uninstalled pillow-simd and works very nice :smile: , here the evidence:
image

remlap:
Hi I recently did a fresh install and now ueberzug fails to line up properly. 

Screenshot at 2019-08-03 12-24-08

seebye:
@remlap The font size & padding is guessed. See: https://github.com/ranger/ranger/pull/1284#issuecomment-434661477  
Your terminal doesn't fit the assumptions.  
You need to remove the menu bar & the scrollbar.

remlap:
It would work previously fine.

seebye:
@toonn I released a new version (on pypi/pip) which uses pillow instead of pillow-simd.

toonn:
@seebye, thank you a *lot* for this. I understand a more performant default is attractive so I would've totally understood if you said no. This'll probably save us a ton of issue reports ❤️

gxt-kt:
> Hi, when I try to preview an image in ranger, at the bottom of the terminal appears `[Errno 13] Permission denied 'ueberzug'` Any hint? I have searched everywhere but I don't know what could I do. I use gentoo, and if I do source "`ueberzug library`" before running ranger, nothing appears at the preview.

Can you share that how do you resolve the problem? I meet the same problem that show `[Errno 13] Permission denied 'ueberzug` in ranger. I see the [ueberzug](https://github.com/seebye/ueberzug) website has stoped maintained and no issues.

PR #1311, opened 2018-09-18
Image Positioning and img_preview bugfix/cleanup

mark-dawn:
#### ISSUE TYPE

- Bug fix
- (in future) Improvement/feature implementation

#### RUNTIME ENVIRONMENT

- __Kernel__: 4.18.6
- __Terminal__: kitty 0.11.3
- __Python__ 3.7/3.5/2.7
- __Ranger version__: rebased on last master `8e9451e6`
- __Locale__: en_US.UTF-8

#### CHECKLIST

- [X] The `CONTRIBUTING` document has been read
- [X] All changes follow the code style
- [X] All new and existing tests pass
    - [ ] Tests have been updated

#### DESCRIPTION
This commit is mostly groundwork to make way for two features for image previews, alongside a bugfix.

Bugfixing involved avoiding usage of curses `putchar` that could desync from python stdout, causing some mystical corruption bugs to appear every now and then. I noticed them during my time with kitty, and it might have caused issues with iTerm2 and possibly terminology.

One feature is the possibility to decide with options the positioning of the image preview, which surfaced during my last pull req.
This could be extended to all image previewers, and it could use some bikeshedding on how to expose this option to the user. I was thinking of two options, one for horizontal and one for vertical alignment, each with 3 possibilities `-1` `0` and `1`, for a total of nine possible positioning. Or maybe a float in this range.

This two changes lead me to look through the code of all previewers and unify common bits to interface with curses/terminal and move them into a common functions. This should also simplify writing image previewers in the future since the other implementations are less divergent between each other.

The other feature would be caching for thumbnails, and this would only help kitty atm, but could be extended to any other previewer that needs them. For now the behavior of kitty was changed to prefer `$XDG_RUNTIME_DIR` as the prime choice for thumbnails and `/tmp` as a fallback. If the implementation of caching is too complex it will be moved to its own branch/pull req

#### MOTIVATION AND CONTEXT
The change provides: 
- better organization of `img_preview.py`
- common functionality make sure all image previewer use the fastest and [bug-free](https://github.com/ranger/ranger/pull/1077#issuecomment-406092409) version of common code
- the possibility to decide the centering of image previews (soon™)
- slightly more stable previews with kitty in edge cases

#### TESTING
All test run, it should have 0 effect on all codebase outside `ext/img_preview.py`
I have used the new helper functions for a month with kitty, and did a couple of trial with w3m and terminology. However, I don't have a mac to test iTerm2.

__NEEDS TESTING__ : most importantly I sent this pull request "early" to get some people with different terminal emulators to test if the changes made affected functionality or performance in any negative way, before diving in adding the centering options, as if issues arises it will be hard to tell if it is due to code being moved around or the new feature.


toonn:
Can't resist the bikeshedding. I'd at least include alternatives in words, so left/center/right, top/middle(or center)/bottom. And have the order be vertical then horizontal so it'd read "bottom left" : )

freed00m:
Another thing I've tested, it works on kitty XWayland and pure wayland.

freed00m:
@mark-dawn Hi, I find issue.

Currently I am running kitty in sway 1.0 with  `KITTY_ENABLE_WAYLAND=TRUE`, but this issue is propably in xorg as well.

When command help window (or how it's called) is drawn it makes the immage blink, sometimes too fast that entire ranger became unresponsive has to be killed externaly. See screenshots.

2018-10-01-192725_grim
2018-10-01-192740_grim

toonn:
: s Ouch. I can confirm glitchiness when an image preview is combined with drawing the hints window.

mark-dawn:
Mh, that's weird, on my end 90% of the times the preview just disappear, but I can get the flickering to trigger sometimes. I wonder if it is tied to framerate...and if it is like this on master. Anyway, I'll see if there is a way to fix it, or else I think I will make a change to ranger itself to disable previews when opening the menu.

In that case, how do other previewers react when the menu is used? So they also clear the image, or do they draw over/under? Depending on how many of them act "correctly" it might be worthwhile to make this change to be generic instead of a special case for kitty, especially since it will have to be in ranger's drawing internals

toonn:
Terminology's native method handles this possibly even worse. First the image disappears, then the hints are draw (delayed), then the image is drawn over the hints and then the hints window disappears, even though you're still entering the sequence.

Ideally the hints window would just cover part of the image, with some methods this is impossible of course because they draw over the terminal window. Next best thing would be resizing the preview to the available space, this'd require 2 previews to be generated for many files, on a fast machine this probably doesn't matter. If that's not feasible we can just disable previews altogether whenever the hints window opens up (and probably bookmarks, and maybe the ask menu). I'd expect this to be annoying mostly for people who run ranger in a large window or fullscreen, cause it'll look like the preview disappears for no reason.

Issue #1302, opened 2018-09-09
Previews of images, PDFs, and videos misaligned on st


Sacules:
#### Runtime Environment
- Operating system and version:  Void Linux 64bit
- Terminal emulator and version: st 0.8.1
- Python version: 3.6.6
- Ranger version/commit: ranger-master 1.9.1
- Locale: en_US.UTF-8

#### Current Behavior
Preview of images, pdfs, and videos misplaced.
#### Expected Behavior
Previews to be in place.
#### Context
I'm using st with an inner padding of 12px, and previews end up looking like this:
![2018-09-09-174145_1366x768_scrot](https://user-images.githubusercontent.com/31902745/45268744-b369a600-b457-11e8-86df-f2c4aa2809d3.png)
The same happens when previewing videos and PDFs (so it's not a w3m-img issue), but other text-based previews and file descriptions are fine.

#### Steps to reproduce
1. Compile st with `borderpx` bigger than 0.
2. Use ranger with image/PDF/video preview enabled.

toonn:
The default pdftotext previews or viewing a rendered image of a pdf?

Sacules:
Rendered images of PDFs. 
![2018-09-09_1366x768](https://user-images.githubusercontent.com/31902745/45269096-3857be00-b45e-11e8-8996-7381ceb3a137.png)

toonn:
That's what I thought, those are still previewed with w3m or your image previewing method of choice.

LeLobster:
I switched from using urxvt to st yesterday and noticed this as well. Compiling `st` with a borderpx of 0 is fine but my standard of 29 causes w3m to misalign the preview.

Vifon:
I don't think st can report back what is its border. You'd need t hardcode this offset into your ranger too.

LeLobster:
I've been looking at [img_display.py](https://github.com/ranger/ranger/blob/master/ranger/ext/img_display.py), but I can't get ranger to show image previews when running `ranger.py` from a local repo clone.  
It does still read the config files in `XDG_CONFIG_HOME` but `set preview_images true` does nothing. Not sure what I'm missing.

Issue #1110, opened 2018-03-13
w3m preview appearing over text  when tiled vertically with another window in i3-gaps


The-Soulless:
#### Runtime Environment
- Operating system and version: i3-gaps Arch Labs GNU/Linux, is there a version to Arch? I just did `pacman -Syu` before this so I guess the latest version?
- Terminal emulator and version: st
- Python version: 2.7.14 and 3.6.4
- Ranger version/commit: ranger-master 1.9.1
- Locale: us_en

#### Current Behavior
Along with the problems in #856 the image preview overlaps the top text when tiled in i3 (Seen here: https://a.pomfe.co/cwnmcw.png )

#### Expected Behavior
It would work like this: https://a.pomfe.co/iefbbe.png

#### Steps to reproduce
1. Start i3-gaps (15 px for me) 
2. Start st
3. Tile two windows vertically (St being one of them)
4. Start ranger and go to an image

mafrazz:
Looks like Ranger ignores st margin border.
With `static int borderpx = 0;` the preview dont overlap the text.

Issue #1314, opened 2018-09-22
Image preview overlapping text


leewdch:
#### Runtime Environment

- Operating system and version: Arch Linux 4.18.9
- Terminal emulator and version: uxterm (336)
- Python version: Python 2.7.15 / Python 3.7.0
- Ranger version/commit: ranger-master 1.9.2
- Locale: en_US.UTF-8

#### Current Behavior
Image previews are overlapping ranger text

#### Expected Behavior
Image correctly wrapped in the preview area

#### Context
Just trying to preview images. The preview method set is w3m if you are wondering.

![image](https://user-images.githubusercontent.com/11262717/45919928-9e503680-be9d-11e8-91a0-3c2323a8f305.png)

Vifon:
Can you reproduce it in other terminals?

Vifon:
Then it's most likely a problem with UXTerm and not much can be done. My guess is that UXTerm doesn't support querying for the size of a single character so the calculations are off.

As for the black bars, you can try setting the `w3m_delay` option. It should be set to 0.02 by default in the new configs.

leewdch:
After using ranger-git and generating new configurations I found w3m_delay in rc.conf and now the black bar issue is fixed in URxvt. I'm going to use that and generate new configs for the other files. I don't know if you can do anything for UXTerm so I'll live closing or keeping the issue open to you. Thanks for the help

LeLobster:
@Vifon @toonn  I think PR #1327 would solve this issue as well, allowing @leewdch  to set the value to `12` as per his `UXTerm*internalBorder`  

Maybe I should consider changing the option name and man page description to a more generic one?  because apparently there are more terminals which would benefit besides `st`

PR #1327, opened 2018-10-04
Allow user to set a w3m offset

LeLobster:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
- Operating system and version: Debian 9.5 (stretch)
- Terminal emulator and version: st 0.8.1
- Python version: 2.7.13 (default, Sep 26 2018, 18:42:22) [GCC 6.3.0 20170516]
- Ranger version/commit: ranger-master v1.9.2-8-g2679b379
- Locale: None.None

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
Added a new config option to set a w3mimgpreview offset (both x and y) named `w3m_offset`  
Right before drawing the img we add this value to the calculated values of `start_x` and `start_y`
Added a comment about this new option in the man page, section PREVIEWS, below W3M

#### MOTIVATION AND CONTEXT
A simple fix for #1302 &  #1110 

#### TESTING
`make test`
No conflicts


Vifon:
Like you mentioned in #1314, let's make it `w3m_*` and not `st_*`, update the docs accordingly and it's good to go. Thanks!

toonn:
One problem with going by TERM is screen/tmux.

LeLobster:
I've never used screen or tmux, what is the problem with it?

Vifon:
I don't like d5d668a. If a user want's to set it, let them. That was one of the reasons for changing the name of this option. Or we could add another option for a list of terms. Well, clearly this is getting out of hand.

As for tmux/screen, they "clobber" the apparent TERM value. On the other hand, you shouldn't expect the preview to work well with tmux/screen. The window ID may change and then all the hell will get loose.

toonn:
I'd prefer dropping the TERM stuff. This is already a *very* niche setting you're adding, let's not also make it complicated. Do we *really* need this setting in the first place? If there's a way to automatically detect borders cross-terminal I'd prefer that tbh.

LeLobster:
Honestly I think it's beyond my capabilities to figure out a way to auto detect borders cross-terminal but I'd be willing to look into it. And although only `st` and `UXTerm` (which is just unicode supported xterm wrapper, and xterm itself is fine) seem to give issues currently there's no harm in future proofing for whatever other terms will show up.

For now I'll just revert d5d668a

toonn:
The first thing you'd need to figure out is whether these terminals somehow report on the padding? I have no idea how they'd do this tbh, maybe termcap?

LeLobster:
Should we move the discussion about that to a new issue?

toonn:
We can, sure. Recap the reason and next step(s) and label it 'enhancement'.

ranger/ext/img_display.py
```diff
@@ -219,9 +219,12 @@ def _generate_w3m_input(self, path, start_x, start_y, max_width, max_height):
             width = (width * max_height_pixels) // height
             height = max_height_pixels
 
+        start_x = int(((start_x - 0.2) * fontw) + self.fm.settings.w3m_offset)
```
toonn: Could you take the offset out of the cast? Imo it's clearer that way.

doc/ranger.pod
```diff
@@ -209,6 +209,9 @@ This does not work over ssh, requires certain terminals (tested on "xterm" and
 To enable this feature, install the program "w3m" and set the option
 C<preview_images_method> to w3m.
 
+When using a terminal with a border larger than 0 which is not automatically detected the w3m preview will be misaligned.
+Use the C<w3m_offset> option to manually adjust the image offset. This should be the same value as the terminals' border value.
```
toonn: terminals' -> terminal's

doc/rifle.1
```diff
@@ -1,4 +1,4 @@
-.\" Automatically generated by Pod::Man 4.10 (Pod::Simple 3.35)
```
toonn: You didn't change `rifle.pod` so please amend it out of the commit. (Check out the previous version and amend the commit.) We get lots of unnecessary merge conflicts on the generated docs.

doc/ranger.pod
```diff
@@ -209,6 +209,9 @@ This does not work over ssh, requires certain terminals (tested on "xterm" and
 To enable this feature, install the program "w3m" and set the option
 C<preview_images_method> to w3m.
 
+When using a terminal with a border larger than 0 which is not automatically detected the w3m preview will be misaligned.
```
toonn: Add a comma after detected. And are there terminals whose border gets detected automatically? If none are detected we need a comma before which. And I like "nonzero border" better than "border larger than 0," presumably there could be terminals with negative borders and if not it's still not wrong.

LeLobster: Auto detected is maybe not the right wording for it. But what I mean is that `w3mimgdisplay -test` returns different values between `st` and `urxvt` with the same border px and geometry which is what causes the misalignment. So technically `urxvt`'s padding gets 'auto detected'.  

What would you recommend changing it into?

doc/rifle.1
LeLobster: Still learning a lot about using git and I think I've done it correctly
```
git checkout /doc/rifle.1 origin/master
git commit --amend
```
git now says  `Your branch and 'origin/stborderw3mfix' have diverged,`  
Doing a force push to origin should be fine, right?

doc/ranger.pod
toonn: Auto detected because the terminal reports something different is fine imo. You could go with something more generic like "properly handled" but I'm not sure that's better.

doc/rifle.1
toonn: Yes, it's fine in this case because no one else is working on your fork. (At least I assume so.) You'll have to do the same for all the commits that had the file btw, not sure it was only in the last one.

doc/ranger.pod
toonn: No need to apologize. I just figure it's better I share whatever typos I notice so the end result is as high quality as we can make it : ) Myself, I have plenty of trouble with usage of commas in english.

doc/ranger.pod
LeLobster: I'll make it like this then `When using a terminal with a nonzero border, which is not automatically detected, the w3m preview will be misaligned.`

doc/rifle.1
LeLobster: Running a `git diff` between master and my updated branch doesn't mention rifle files anymore so we're good.

PR #1333, opened 2018-10-09
Add ShellCheck to tests

toonn:
This PR adds a rule to the `Makefile` to run ShellCheck on `scope.sh`.
I've already fixed up `scope.sh` to make this straightforward.

I'm hoping travis will deal with this. It seems silly to shellcheck for each version of python we test but I have no idea how to get around this with Travis.


toonn:
Turns out travis does simply deal with this. : )

Makefile
```diff
@@ -105,11 +108,19 @@ test_pytest:
...
+	@echo "Running shellcheck..."
+	# The $ for the range needs to be escaped because of make.
+	sed '2,$$s/^\( *\)#/\1/' ./ranger/data/scope.sh | shellcheck -a -
+
 test_other:
 	@echo "Checking completeness of man page..."
 	@tests/manpage_completion_test.py
 
-test: test_pylint test_flake8 test_doctest test_pytest test_other
+test: test_pylint test_flake8 test_doctest test_pytest test_shellcheck test_other
```
Vifon: Can we replace all the targets here with `test_py`?

Makefile
```diff
@@ -105,11 +108,19 @@ test_pytest:
 	@echo "Running py.test tests..."
 	py.test tests
 
+test_py: test_pylint test_flake8 test_doctest test_pytest test_other
+	@echo "Finished python and documentation tests!"
+
+test_shellcheck:
+	@echo "Running shellcheck..."
+	# The $ for the range needs to be escaped because of make.
```
toonn: This $ also needs escaping.

Makefile
```diff
@@ -105,11 +108,18 @@ test_pytest:
 	@echo "Running py.test tests..."
 	py.test tests
 
+test_py: test_pylint test_flake8 test_doctest test_pytest test_other
+	@echo "Finished python and documentation tests!"
+
+test_shellcheck:
+	@echo "Running shellcheck..."
+	sed '2,$$s/^\( *\)#/\1/' ./ranger/data/scope.sh | shellcheck -a -
```
Vifon: This regex won't work for tabs. I know we shouldn't mix them in but it won't detect it either unless `shellcheck` handles that. Let's use `\s` instead.

toonn:
@Vifon, here's a ping so you won't forget : ) (You *do* read these emails, don't you?! -_-)

PR #1334, opened 2018-10-09
Document w3m_offset in Settings section of manpage

toonn:
Travis was complaining about the setting not being documented.


toonn:
Merging without approval, because it's a minor change.

PR #1346, opened 2018-10-20
Broken preview of highlighted text with Python3

### LINKED ISSUE #906: Preview pane utf-8 display bug
ISSUE TYPE

    Bug report

RUNTIME ENVIRONMENT

    Operating system and version: Gentoo
    Terminal emulator and version: st 0.7
    Python version: 3.4.5
    Ranger version/commit: latest
    Locale: en_US.UTF-8

EXPECTED BEHAVIOR

    Text is properly displayed

CURRENT BEHAVIOR

    Garbage happens at some point

What I have in my scope.sh:
```
...
text/* | */xml)
            iconv -f "$(uchardet "${FILE_PATH}")" -t utf-8 "${FILE_PATH}" && \
                exit 5;;
....
```
https://files.catbox.moe/bnxnlt.png (terminal)
https://files.catbox.moe/knzzfa.png (ranger)

### LINKED ISSUE #967: Python3 vs python2 unicode display problem
Ranger has unicode problems in the preview pane (not something new) but I just discovered that they're only with python 3. Might make it easier to resolve.
I think my earlier report (https://github.com/ranger/ranger/issues/906) is the same problem since the oddities start to appear only after a certain point (even if the same character was displayed correctly before).

#### ISSUE TYPE

- Bug report

#### RUNTIME ENVIRONMENT

- Operating system and version: Gentoo
- Terminal emulator and version: st-9999 (tried with urxvt-9.22 too)
- Python version: tested with 2.7.14, 3.4.5 and 3.5.4
- Ranger version/commit: ranger-master 1.9.0b5
- Locale: 
```
> locale
LANG=en_US.utf8
LC_CTYPE="en_US.utf8"
LC_NUMERIC="en_US.utf8"
LC_TIME="en_US.utf8"
LC_COLLATE="en_US.utf8"
LC_MONETARY="en_US.utf8"
LC_MESSAGES="en_US.utf8"
LC_PAPER="en_US.utf8"
LC_NAME="en_US.utf8"
LC_ADDRESS="en_US.utf8"
LC_TELEPHONE="en_US.utf8"
LC_MEASUREMENT="en_US.utf8"
LC_IDENTIFICATION="en_US.utf8"
LC_ALL=
```

#### IMAGES / VIDEOS

From top to bottom: python3, python2_7, st:
https://files.catbox.moe/c7hpez.png

### LINKED ISSUE #1166: problems with previewing highlighted cyrillic text
#### Runtime Environment

- Operating system and version: 4.9.76-gentoo x86_64
- Terminal emulator and version: xterm-331
- Python version: 3.6.3 (default, Mar 28 2018, 14:50:40) [GCC 6.4.0]
- Ranger version/commit: ranger-master 1.9.1
- Locale:  en_US.UTF-8

#### Current Behavior

Both highlighters -- pygmentyzer and highlight, are showing rubbish instead of Cyrillic utf8 text sometimes.

If run them from console, i.e. `highlight -t 4 --out-format=xterm256 --style=pablo --force -- 1.md` all is ok. 
If look on text in rangers preview -- it looks like bad encoding, which sometimes resolves if put or remove extra symbol (e.g., ".") before Cyrillic text.

#### Expected Behavior

Show all utf8 characters correctly, if they are in current console font.

#### Steps to reproduce

look on  files in attached archive in rangers preview
[bug.tar.gz](https://github.com/ranger/ranger/files/1988449/bug.tar.gz)

baranovskiykonstantin:
ranger/core/loader.py:
CommandLoader.generate() gets text for preview panel from scope.sh
as the blocks of byrearray (max 512 bytes). This blocks immediately
decoded to utf-8. But in case when muli-byte unicode characters is
present they may be split and python will can't correctly to decode
it. For example: b'abc\xd0', b'\x9a...'
b'\xd0\x9a' - is cyrillic letter 'K' and it can't be decoded
separately (UnicodeDecodeError occurs).
So, received data blocks (bytearrays) must be collected and
only after that decoded as one whole byrearray.

Fixes #906
Fixes #967
Fixes #1166

### DIFF
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -194,6 +194,7 @@ class CommandLoader(  # pylint: disable=too-many-instance-attributes
                 selectlist.append(process.stdout)
             if not self.silent:
                 selectlist.append(process.stderr)
+            read_stdout = None
             while process.poll() is None:
                 yield
                 if self.finished:
@@ -210,10 +211,11 @@ class CommandLoader(  # pylint: disable=too-many-instance-attributes
                                 self.fm.notify(read, bad=True)
                         elif robjs == process.stdout:
                             read = robjs.read(512)
-                            if py3:
-                                read = safe_decode(read)
                             if read:
-                                self.stdout_buffer += read
+                                if read_stdout is None:
+                                    read_stdout = read
+                                else:
+                                    read_stdout += read
                 except select.error:
                     sleep(0.03)
             if not self.silent:
@@ -223,9 +225,12 @@ class CommandLoader(  # pylint: disable=too-many-instance-attributes
                     self.fm.notify(line, bad=True)
             if self.read:
                 read = process.stdout.read()
+                if read:
+                    read_stdout += read
+            if read_stdout:
                 if py3:
-                    read = safe_decode(read)
-                self.stdout_buffer += read
+                    read_stdout = safe_decode(read_stdout)
+                self.stdout_buffer += read_stdout
         self.finished = True
         self.signal_emit('after', process=process, loader=self)
 


toonn:
I'm looking into this but don't have a good way to test the fix. I'm trying the files from #1166 but since I don't read cyrillic I have no idea if the preview is correct or not. Do you have a better example?

It does look correct though, with ranger 1.9.1 and python 3.7. All of `1-3.md` look the same to me and none look like `1.png`.

baranovskiykonstantin:
@toonn, attached file contain 50x50 field of Unicode character 'FULL BLOCK' (U+2588). It is good example for testing.

[unicode_2588.txt](https://github.com/ranger/ranger/files/2498786/unicode_2588.txt)

toonn:
I still can't trigger the bug. It looks like a big white rectangle (white on black terminal).

baranovskiykonstantin:
Hmm. You are lucky :)
In my case I get the picture that shown on attached screenshot.
Used system is Archlinux x86_64.
Version of installed ranger is: ranger-master 1.9.2 (installed from aur - ranger-git 1.9.2.33.g21c2f54a-1)
Python version: 3.7.0 (default, Sep 15 2018, 19:13:07) [GCC 8.2.1 20180831]
Locale: en_US.UTF-8

screenshot

toonn:
But why isn't it broken for me?

baranovskiykonstantin:
Ensure that syntactic highlighting is enabled and available in scope.sh.
HIGHLIGHT_SIZE_MAX must be greater than file size, for example.

Maybe your ranger gets full 512-byte data blocks and it is don't cause any issue (256 unicode characters fully transfered). But what if we add simple one-byte character at start of file? Perhaps it breaks you preview. Try attached file, please.
[space+unicode_2588.txt](https://github.com/ranger/ranger/files/2499146/space%2Bunicode_2588.txt)

toonn:
Hmm, this warning from the python docs sounds ominous:

toonn:
Hmm, this warning from the python docs sounds ominous, maybe we should heed it?

>Warning Use communicate() rather than .stdin.write, .stdout.read or .stderr.read to avoid deadlocks due to any of the other OS pipe buffers filling up and blocking the child process.

And wouldn't passing `universal_newlines=True` to `Popen` let us just read unicode characters, rather than just bytes?

toonn:
Hmm, this warning from the python docs sounds ominous, maybe we should heed it?

>Warning Use communicate() rather than .stdin.write, .stdout.read or .stderr.read to avoid deadlocks due to any of the other OS pipe buffers filling up and blocking the child process.

And wouldn't passing `universal_newlines=True` to `Popen` let us just read unicode characters, rather than just bytes?
Yep, verified this works. See the attached diff (.txt because github filechooser doesn't like .diff): [Unl.txt](https://github.com/ranger/ranger/files/2500887/Unl.txt)

I like this approach better because it removes code instead of adding code and unicode handling is difficult to get right manually. Ideally you'd start from this and replace uses of `.stdout.read` and the like with `communicate()` as the warning states. Test on all the files that broke for you first of course, I only tested on your example and `1-3.md`.

👍 Awesome PR btw, keep them coming ; )

toonn:
Hmm, this warning from the python docs sounds ominous, maybe we should heed it?

>Warning Use communicate() rather than .stdin.write, .stdout.read or .stderr.read to avoid deadlocks due to any of the other OS pipe buffers filling up and blocking the child process.

And wouldn't passing `universal_newlines=True` to `Popen` let us just read unicode characters, rather than just bytes?
Yep, verified this works. See the attached diff (.txt because github filechooser doesn't like .diff): [Unl.txt](https://github.com/ranger/ranger/files/2500887/Unl.txt)

I like this approach better because it removes code instead of adding code and unicode handling is difficult to get right manually. Ideally you'd start from this and replace uses of `.stdout.read` and the like with `communicate()` as the warning states. Test on all the files that broke for you first of course, I only tested on your example and `1-3.md`.

👍 Awesome PR btw, keep them coming ; )

toonn:
Hmm, this warning from the python docs sounds ominous, maybe we should heed it?

>Warning Use communicate() rather than .stdin.write, .stdout.read or .stderr.read to avoid deadlocks due to any of the other OS pipe buffers filling up and blocking the child process.

And wouldn't passing `universal_newlines=True` to `Popen` let us just read unicode characters, rather than just bytes?
Yep, verified this works. See the attached diff (.txt because github filechooser doesn't like .diff): [Unl.txt](https://github.com/ranger/ranger/files/2500887/Unl.txt)

I like this approach better because it removes code instead of adding code and unicode handling is difficult to get right manually. Ideally you'd start from this and replace uses of `.stdout.read` and the like with `communicate()` as the warning states. Test on all the files that broke for you first of course, I only tested on your example and `1-3.md`.

👍 Awesome PR btw, keep them coming ; )

baranovskiykonstantin:
With Unl.txt files encoded with utf-8 is previewed correct but when a file with another encoding (in my case it iso-8859-1, see attachment) previewing stops working and shows error message "Loader work process failed...".

[io.asm.txt](https://github.com/ranger/ranger/files/2501514/io.asm.txt)

ranger/core/loader.py
```diff
@@ -170,62 +169,39 @@ def generate(self):  # pylint: disable=too-many-branches,too-many-statements
...
-                for line in process.stderr:
-                    if py3:
-                        line = safe_decode(line)
+            output = (None, None)  # (stdout, stderr)
+            try:
+                output = process.communicate(self.input)
+            except IOError as ex:
+                if ex.errno != errno.EPIPE and ex.errno != errno.EINVAL:
+                    raise
+            output = list(output)
```
toonn: Why a list? `(stdout, stderr) = communicate()` is easier to follow imo. And I don't like inlining `safe_decode` it's much less clear what's going on now.
Also, there's a problem with communicate, it waits until the entire output is available, that's not what we want, so I regret suggesting it.

baranovskiykonstantin: Tuple is immutable. It must be converted to some mutable type before decoding. I did choose list.

> ... that's not what we want, so I regret suggesting it.

No problem. I will remove last commit.

baranovskiykonstantin:
>The only real solution I can see is not taking bytes but characters but if we can't take a random chunk of bytes to decode and then take characters from that means we can't stream the output.

In every iteration of reading stdout-data, they is only collecting but don't showing in a preview widget. Only after terminating the process the signal 'after' emits and the got data from stdout is shows.
Please notice, the same behaviour was before my changes. 
I don't see the posibility to stream the output to a preview widget in real time. For realizing this feature we have to know the encoding of the text before starting to read stdout.
So I think that current solution is optimal.

toonn:
Then why do we read in 512 byte chunks? If what you say is true, we should be using communicate. I thought we only read chunks until we had enough to fill the preview or the pager. Then if you scrolled in the pager it'd read more. What is the yield even for if we don't actually read in chunks?

baranovskiykonstantin:
"Yield" does not return any value. It just interacts until process terminates. The gotten data processed only in 'after' signal handler. This signal emits only after terminating of the process.
I see single reason of reading stdout as 512 blocks - it is ability to terminate the porecess while stdout is reading ( see "if self.finished:" under "yield").

Just now found another solution of discussed bug. We can replace ".read(512)" with ".readline()" that also works in my case. But I don't sure that it will works in general case.

toonn:
Could you explain how appending the newly read undecoded data to the already decoded data then decoding the whole thing again works?

baranovskiykonstantin:
It isn't work in that way. All raw data is collects in local variable "read_stdout" step by step. After that, whole data decodes and stores in member of CommandLoader -  self.stdout_buffer, which in turn will processed in "after" signal handler.

baranovskiykonstantin:
Commit was updated. I hope I understand your remark correct.

toonn:
Thanks for the contribution! Encodings have been a problem point for a while.

PR #1348, opened 2018-10-22
Change scope.sh for compatibility with BSDs

toonn:
These changes should make `scope.sh` more POSIX compliant and more importantly compatible with the BSDs.

Everything but the `local` keyword is now POSIX compliant, however this keyword is available on all `sh` implementations that matter afaict. So I added a disclaimer. I'm open to rewording/shortening/removing it.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 set -o noclobber -o noglob -o nounset -o pipefail
-IFS=$'\n'
+IFS=$(printf '\n')
 
 # If the option `use_preview_script` is set to `true`,
 # then this script will be called and its output will be displayed in ranger.
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -109,7 +109,7 @@ handle_image() {
             orientation="$( identify -format '%[EXIF:Orientation]\n' -- "${FILE_PATH}" )"
             # If orientation data is present and the image actually
             # needs rotating ("1" means no rotation)...
-            if [[ -n "$orientation" && "$orientation" != 1 ]]; then
+            if [ -n "$orientation" && "$orientation" != 1 ]; then
                 # ...auto-rotate the image according to the EXIF data.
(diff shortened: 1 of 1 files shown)


me-ru-ru:
Testing running scope.sh with no arguments on FreeBSD I get the following errors:

POSIX sh: `set: Illegal option -o pipefail`
Text previews work, but everything else is crashing ranger.

mksh: `./scope.sh[36]: 1: parameter not set`
Works flawlessly in my testing.

tcsh: `set: Variable name must begin with a letter.`
I'm not getting any previews with tcsh.

me-ru-ru:
This is what FreeBSD uses as /bin/sh: https://www.freebsd.org/cgi/man.cgi?query=sh&sektion=1&manpath=freebsd-release-ports

toonn:
Since it works for `mksh` that error is probably just because `scope.sh` needs some arguments.

We didn't expect this to work with `tcsh`, I think csh is too different to be compatible with plain sh? I could be wrong but I expect it to take a lot more effort to make this happen.

The error about pipefail *is* very unfortunate though. `scope.sh` relies on the feature to keep things "simple." I'm not sure there's a simple way around this.

I've actually started to come around to implementing a `scope.py` because if ranger's running we can rely on a python interpreter being available.

ranger/data/scope.sh
```diff
@@ -100,7 +109,7 @@ handle_image() {
             orientation="$( identify -format '%[EXIF:Orientation]\n' -- "${FILE_PATH}" )"
             # If orientation data is present and the image actually
             # needs rotating ("1" means no rotation)...
-            if [[ -n "$orientation" && "$orientation" != 1 ]]; then
+            if [ -n "$orientation" && "$orientation" != 1 ]; then
```
subnut, line 112:
```suggestion
            if [ -n "$orientation" ] && [ "$orientation" != 1 ]; then
```

subnut (changes requested):

subnut:
> I've actually started to come around to implementing a scope.py because if ranger's running we can rely on a python interpreter being available.

IMO this would be really elegant and have no compatibility headaches.  
Python is python everywhere. The same code shall behave similarly no matter if it's Linux, *BSD or Solaris.

PR #1349, opened 2018-10-22
Change scope.sh shebang for BSDs

toonn:
This would make scope.sh fully compatible with the BSDs and as an added bonus on some systems should lead to a small (probably hardly noticeable) performance improvement (dash is faster than bash for example).

I also implemented changing the shebang when copying `scope.sh`.
I don't actually think that's necessary but here's my rationale:
>Apparently the differences between sh and bash are mystical to people and vifon'd rather keep it that way than field issue reports about bashisms not working and educate them on the matter.
Woops ^U^U^U

Basically, people expect bash so we give them bash in their personal editable copy of `scope.sh`.

### DIFF
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -301,14 +301,18 @@ class FM(Actions,  # pylint: disable=too-many-instance-attributes
             copy('data/scope.sh', 'scope.sh')
             os.chmod(self.confpath('scope.sh'),
                      os.stat(self.confpath('scope.sh')).st_mode | stat.S_IXUSR)
-        if which in ('all', 'rifle', 'scope', 'commands', 'commands_full', 'rc'):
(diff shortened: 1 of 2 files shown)


me-ru-ru:
Testing running scope.sh with no arguments on FreeBSD I get the following errors:

POSIX sh: `set: Illegal option -o pipefail`
Text previews work, but everything else is crashing ranger.

mksh: `./scope.sh[36]: 1: parameter not set`
Works flawlessly in my testing.

tcsh: `set: Variable name must begin with a letter.`
I'm not getting any previews with tcsh.

acid-bong:
~~Not enough work, `scope.sh` has bashisms (`local` vars, `pipefail` option, double-bracket `test`)~~
oh, you have a separate PR for that, why not unite them?

PR #1351, opened 2018-10-23
added support for DJVU text conversion

chakravala:
provides the ability to preview DJVU files with text preview, just like PDF files

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
Requires `djvutxt` program provided by `djvulibre` to convert

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
Very simple implementation on `scope.sh` that provides the ability to preview DJVU files with text preview, just like PDF files, only requires `djvutxt` to be installed on system

#### MOTIVATION AND CONTEXT
These changes help with previewing DJVU files with text conversion, so that they can be previewed just like PDF documents. This solves the problem of not having a text preview of DJVU files

#### TESTING
No tests have been run, as this doesn't affect any of the existing code base.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -94,6 +94,13 @@ handle_image() {
         #     convert "${FILE_PATH}" "${IMAGE_CACHE_PATH}" && exit 6
         #     exit 1;;
 
+		# DJVU
+        image/vnd.djvu)
+            # Preview as text conversion
+            djvutxt "${FILE_PATH}" && exit 5
+            exiftool "${FILE_PATH}" && exit 5
+            exit 1;;
+
         # Image
         image/*)
             local orientation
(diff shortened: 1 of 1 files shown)


toonn:
Could you add wrapping with `fmt` as is done for pdf previews? I think it's better to stay consistent.

chakravala:
Yes, I inserted it now. I made this modification many months ago, before `fmt` was added to the PDF.

ertwro:
Hi. It's working quite well but by any chances could you also add support for images? I remember you could do something like this:
ddjvu -page=1 -format=tiff "$filename" "${outputdir//.tif}" && convert "${ouputdir//.tif}" -quality 10 "${outputdir//.jpg}"

toonn:
That's in the works still at #1421.

PR #1352, opened 2018-10-24
File encoding guessing order changed.

### LINKED ISSUE #1350: Problems with decoding
Seems like the encoding doesn't always get detected properly? Haven't really had a chance to look into this.

```
21:45:03 ERRO Notification: Loader work process failed: Getting preview of /home/toonn/get-pip.py (Percent: 0)
21:45:03 ERRO decode() argument 2 must be str, not None
  File "/home/toonn/src/ranger/ranger/core/loader.py", line 411, in work
    next(item.load_generator)
  File "/home/toonn/src/ranger/ranger/core/loader.py", line 230, in generate
    self.signal_emit('after', process=process, loader=self)
  File "/home/toonn/src/ranger/ranger/ext/signals.py", line 268, in signal_emit
    fnc(signal)
  File "/home/toonn/src/ranger/ranger/core/actions.py", line 1125, in on_after
    text = self.read_text_file(path, 1024 * 32)
  File "/home/toonn/src/ranger/ranger/core/actions.py", line 1193, in read_text_file
    return codecs.decode(data, guessed_encoding, 'replace')
TypeError: decode() argument 2 must be str, not None
```

[And here's the file `get-pip.py.txt`](https://github.com/ranger/ranger/files/2503181/get-pip.py.txt)
, probably downloaded it with curl but I'm not certain of the provenance, had to append `.txt` to the name because of github's attachment rules, hope that doesn't magically fix the problem.

baranovskiykonstantin:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Archlinux x86_64
- Terminal emulator and version: rxvt-unicode (urxvt) v9.22
- Python version: 3.7.0 (default, Sep 15 2018, 19:13:07) [GCC 8.2.1 20180831]
- Ranger version/commit: ranger-master v1.9.2-39-g4d84abad
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Order of guessing an encoding of a file for previewing is changed to more universal. The situation, when chardet module can be loaded but can't detect encoding, is also handled.

#### MOTIVATION AND CONTEXT

Fixes #1350


baranovskiykonstantin:
PR modified according to your remark.

toonn:
Uhm, I'm kinda lost on the improvement in this PR? Raising an ImportError for something that's not an ImportError? I don't like exceptions as control flow and in this case you seem to be adding it somewhere it's not necessary?

baranovskiykonstantin:
Updated again. After all changes the order become the same. However, this PR handles case when chardet module loads but doesn't detect an encoding.

toonn:
Did you encounter chardet not being able to guess an encoding? According to the documentation `detect()` should be a total function, always providing a guess even though the confidence in that guess could be 0.

baranovskiykonstantin:
Why are you asking to me? Author of that pice of code is @FichteFoll. Ask him.
I just reworked his code in that way, that resolves issue #1350.

toonn:
I'm asking you because you're looking into it : )

baranovskiykonstantin:
Well, I don't know the reasons for which that function was realized in that way.
Should I find another solution for this issue?

Issue #87, opened 2013-03-25
Python process using high CPU


pencilcheck:
I don't understand why the python process are still running after opening a shell at a directory, shouldn't ranger just cd and quit itself?

That's my first impression when I figured that the Python process out of no where is from ranger after opening some shells from ranger.

After doing a small research I found out ranger spawns the shell, but not only I don't know how to resume ranger from the shell, I can't quit ranger easily, and what happen is that Python process would go crazy and uses 100% of CPU of some sort, and I have to go to activity monitor to force quit those Python processes, which would destroy everything, all the shell I have opened from ranger. This is very annoying.

Am I doing it wrong? Or is this some serious bug?

Until this bug is fixed, I don't know if I could still use ranger as directory jumper now :(

hut:
This issue is now 2 years old - does it still occur?

bebound:
It's still occur...
Open rager, using vim to edit file (vim is my default editor), quit vim, quit iTerm.
The a python processing will using 100% cpu.
Open rager, press q, and the problem will solved.

hut:
I'm afraid I don't know how to debug this.

nfnty:
If you're still experiencing this issue open a new issue and completely fill in the issue template.

Deluxo:
sorry for not complying, but this issue is still affecting me. I deleted scope file and problem seems to be solved.

Deluxo:
ok, scope file doesn't matter. This happens even without it.
Seems to happen randomly.
```
ranger-stable 1.8.1

Python 3.6.0 (default, Jan 16 2017, 12:12:55) 
[GCC 6.3.1 20170109]
```

ww7w:
This is still an issue. Removing scope.sh file does fix it.

toonn:
Only thing I can think of is an open issue/PR related to the ranger process not dying if stdin is closed.

dylan-chong:
This is still an issue and removing the scope file does not change anything.

toonn:
We need more info to even start looking into this. Does this happen while you're using ranger? Is ranger working on any previews, check with `w`. Check the log for errors `W`. Does this happen where you're in a network mount? What version of ranger are you running, what version of python, do you use bash or a different shell, does running ranger in a different terminal change anything? How about running `ranger --clean` to see if your config is at fault?

dylan-chong:
I figured out that disabling these options gets rid of the problem.
It only seems to happen on this one git repository which has 3600 files

```
[...]

PR #1368, opened 2018-11-03
Exit if STDIN is closed #1367

### LINKED ISSUE #1367: Ranger enters busy-loop if STDIN file descriptor is closed
I hope this will solve an old heisenbug. It doesn't seem to always reproduce naturally, but can be provoked using gdb. 

Potentially related, unresolved bug reports: 
- https://github.com/ranger/ranger/issues/87
- https://github.com/ranger/ranger/issues/525

#### Runtime Environment
- Operating system and version: Ubuntu 18.04.1 LTS
- ranger version: ranger-master v1.9.2-53-ge56c74bb
- Python version: 2.7.15rc1 (default, Apr 15 2018, 21:51:34) [GCC 7.3.0]
- Locale: None.None
- Shell: GNU bash, version 4.4.19(1)-release (x86_64-pc-linux-gnu)

#### Current Behavior

Ranger enters a busy-loop if STDIN is closed. 

#### Expected Behavior

Ranger exits if STDIN is closed. 

#### Background

Using ranger extensively, I occasionally notice that a ranger process is using 100% of the CPU. 

It is not uncommon for me to start ranger, navigate to a directory, spawn a shell to run a command and restart ranger again. 
I also usually close terminal windows using the window manager, rather than exiting each subprocess running in that window. 

#### Analysis

To the best of my understanding, this is how the bug manifests. 

When a shell exits (e.g. when closing a terminal), it sends a SIGHUP signal to its child processes. 
If the child process does not explicitly handle this signal (which ranger doesn't do), the default behavior is to terminate the process. 

When ranger spawns a shell which restarts ranger, the SIGHUP reaches the first ranger process, but not (always?) the second process (due to the fork-exec?). The terminal and the first ranger process exit, and the remaining ranger process continues to run, but becomes orphaned and adopted by init (pid 1). 

Ranger runs a 'while True' loop which repeatedly wait for user input, which happens here: https://github.com/ranger/ranger/blob/e56c74bb30ad2f363280e4d1a14bf2c20dff5b5f/ranger/gui/ui.py#L239

	key = self.win.getch()

https://docs.python.org/2/library/curses.html#curses.window.getch

If there is no input, getch() waits for some time, then returns -1. However, because the STDIN file descriptor was closed by the earlier exited processes, getch() presumably throws an error and returns -1 immediately, which causes ranger to enter a continuous busy-loop. 

#### Possible Solutions

If getch() returns -1 and os.isatty(0) returns False, ranger should exit. 

##### Natural way (not guaranteed to work)

In a terminal, start ranger: 

  > ranger

Press 'S' to spawn a shell
Start ranger again:

  > ranger

(Maybe press down a couple of times)
Close the terminal (e.g. Alt-F4)

The former ranger process exits, but the latter enters a busy-loop. 

##### gdb way

Force-close the STDIN file descriptor of a running ranger process

Start ranger
Attach to process with gdb: 

  > gdb -p PID_OF_RANGER

Close fd 0:

  > (gdb) p close(0)

Detach to let ranger keep running:

  > (gdb) detach

Ranger enters busy-loop.

igoyak:
See https://github.com/ranger/ranger/issues/1367

I'm not sure if this is the preferred way to exit ranger, suggestions welcome

### DIFF
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -238,7 +238,7 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
         for key in keys:
             self.handle_key(key)
 
-    def handle_input(self):
+    def handle_input(self):  # pylint: disable=too-many-branches
         key = self.win.getch()
         if key == curses.KEY_ENTER:
             key = ord('\n')
@@ -280,7 +280,6 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
             elif key == -1 and not os.isatty(sys.stdin.fileno()):
                 # STDIN has been closed
                 self.fm.exit()
-
 
     def setup(self):
         """Build up the UI by initializing widgets."""
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -277,6 +277,10 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
                 else:
                     if not self.fm.input_is_blocked():
                         self.handle_key(key)
+            elif key == -1 and not os.isatty(sys.stdin.fileno()):
+                # STDIN has been closed
+                self.fm.exit()
+
 
     def setup(self):
         """Build up the UI by initializing widgets."""


ranger/gui/ui.py
```diff
@@ -272,6 +272,9 @@ def handle_input(self):
                 else:
                     if not self.fm.input_is_blocked():
                         self.handle_key(key)
+            elif key == -1 and not os.isatty(0):
```
toonn: Could you use `sys.stdin.fileno()` rather than the hard coded `0`?

ranger/gui/ui.py
```diff
@@ -272,6 +272,9 @@ def handle_input(self):
...
+                # STDIN has been closed
+                sys.exit(1)
```
toonn: Proper way to exit is `self.fm.exit()`

toonn:
I'm pretty sure the other issues are the same problem. Feel free to include `Fixes #1367\nFixes #87...` in one of your commits.

igoyak:
Thanks for the feedback! I will check it out in a couple of days.

ranger/gui/ui.py
igoyak: Thanks, fixed.

ranger/gui/ui.py
igoyak: Good idea, fixed.

toonn:
Hmm, I actually can't reproduce the problem with the master version of ranger. @SuperShinyEyes, could you test this fix?

igoyak:
About propagating the SIGHUP: I've done some reading, but I'm not entirely sure how to do it, why it isn't happening currently or whether it's actually a good idea. It is related to the "job control" of shells and "process groups", where the "process leader" is responsible for forwarding signals (or rather translate a SIGHUP into SIGINTs for the children?). I could try to dig deeper, but maybe not in the foreseeable future. 

I do believe the behavior proposed in this PR is correct, though. Reasoning: ranger is a user-facing application. If the user cannot supply input, ranger has no reason to keep running.

igoyak:
> Hmm, I actually can't reproduce the problem with the master version of ranger. @SuperShinyEyes, could you test this fix?

Did you try the method using gdb? That is the only one that seems reliable. 
I rechecked both methods and can still produce the bug in commit 1ca1255a3cf26eb646fca75b5388b2c645fdf466

SuperShinyEyes:
@toonn @igoyak Is there a still need for me to test this fix? 

Sorry for the late reply. Last week was BUSY. I'd be more than happy to help fixing this. :)

SuperShinyEyes:
@toonn Got it. Will try tonight.

igoyak:
@toonn I have finally had more time to look deeper into this issue. 
I have updated the source branch of this PR, just rebased to use a modern commit as a parent. 
I have also written some more conclusions and a cross-platform way to replicate the issue: https://github.com/ranger/ranger/issues/1367#issuecomment-521773193

This issue still affects me (and others, it seems), and I think it would be a good idea to merge to master whenever appropriate.

toonn:
Will look into this tomorrow. Thanks for your effort!

@SuperShinyEyes, did you find the time to test this?

toonn:
Hmm, finally found time to look into this. Still can't reproduce by closing a terminal.
Apparently gdb won't run on macOS due to some code signing issues, at least without much effort. However, I was able to reproduce this with gdb on linux, had to guess at the type to cast the `close` call to though.

Now I can finally merge this 🎉 : )

PR #1394, opened 2018-11-30
Add bat support for code highlighting

leiserfg:
#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated


toonn:
Hmm, this doesn't work for me and I'm not sure why. In ranger I get just gray blocks: 
in ranger

On the command line it does work though: 
command line

leiserfg:
Try by exporting the environment variables manually and using the same arguments

Vifon:
I like it, thanks!

Before I merge it, what's the rationale behind the `COLORTERM` value you use? Their docs mention the values `truecolor` and `24bit` and it's neither of them here.

Vifon:
Ah, I only now saw your comment, @toonn. Well, we need to fix it beforehand of course.

toonn:
It works on the command line when I export the variables and then run your command almost verbatim, just dropping the `&& exit 5`.

Ah, it's because I run tmux. Should be fixable though since just invoking bat in tmux works.

leiserfg:
> I like it, thanks!
> 
> Before I merge it, what's the rationale behind the `COLORTERM` value you use? Their docs mention the values `truecolor` and `24bit` and it's neither of them here.

I am doing that 'cause I suspected that they use $COLORTERM as an optional replacement to $TERM and we need to force `bat` to use 256 colors.

toonn:
@leiserfg, do you mean *when* the terminal supports it? Because `pygmentize_format` can be just `terminal`. Also, wouldn't `truecolor` be fine if the terminal already defines it? You could maybe do something like `env COLORTERM=${COLORTERM :- ${pygmentize_format}} ...`.

leiserfg:
I don't know why it doesn't use the $TERM inside the bash script, using the COLORTERM is just a workaround.

toonn:
In the `bat` bash script? What I'm asking is why the usual settings of `COLORTERM`, like `truecolor`, aren't good enough?

leiserfg:
I don't know why, but using True color for COLORTERM make `bat` to use ansi.

toonn:
Ok, and the "bash script" you were referring to was bat?

leiserfg:
No, I was talking about scope.sh

leiserfg:
In theory, `bat` should use $TERM without problems inside of the scope.sh but it doesn't. That's what I tried  COLORTERM and it worked.

ranger/data/scope.sh
```diff
@@ -186,6 +186,9 @@ handle_mime() {
             fi
             highlight --replace-tabs="${HIGHLIGHT_TABWIDTH}" --out-format="${highlight_format}" \
                 --style="${HIGHLIGHT_STYLE}" --force -- "${FILE_PATH}" && exit 5
+                
+            # env COLORTERM=$pygmentize_format  bat --color=always  --style="plain"  "${FILE_PATH}" && exit 5
```
Vifon: There are unnecessary double-spaces in this line, please remove them.

ranger/data/scope.sh
```diff
@@ -186,6 +186,9 @@ handle_mime() {
             fi
             highlight --replace-tabs="${HIGHLIGHT_TABWIDTH}" --out-format="${highlight_format}" \
                 --style="${HIGHLIGHT_STYLE}" --force -- "${FILE_PATH}" && exit 5
+
```
Vifon: Remove the whitespace filling this line please.

leiserfg:
I played with my vars and apparently something is putting COLORTERM as truecolor but my TERM is a 256 colors one, so that is why I need to set COLORTERM to other value, so the code for calling bat is ok, but setting any value to COLORTERM  is needed only if you have the same problem I have (probably if you are using tmux within an truecolor terminal).

Vifon:
@leiserfg So we essentially want to force `bat` into the 8-bit color mode?

toonn:
I'd actually say the problem is not with bat but with your setup. I'd expect other programs to potentially break too.

leiserfg:
Til now only bat inside ranger

Vifon:
Only inside ranger?

leiserfg:
Yes, if I run bat outside ranger it works well.

Vifon:
That's quite interesting. Well then, the workaround is justified a bit. Would `COLORTERM=8bit` work here? It conveys our intent better.

toonn:
I don't like it. I still don't get why it's necessary. Bat works fine for me with `COLORTERM=truecolor&8bit` both in and out of tmux. All that changes is some colors are brighter with truecolor and bold is rendered as inverse when in tmux.

leiserfg:
So, there are only two ways: 
* Letting bat heuristics choose the color mode. (Can work or not and when it doesn't people will complain and will probably create another issue).
* Forcing 8bit mode that always works but is less fancy.
 
In both cases, one can document how to enable the other mode by removing the variable, but I think that the healthiest is the one that always works.

toonn:
It doesn't always work though, that's the problem. It needs to work inside of tmux. Bat by itself **does**, so there's no reason bat inside ranger can't.

toonn:
It doesn't make sense to force the COLORTERM variable to 8bit though, cause truecolor *works* in tmux, just not inside ranger. There must be something else going on.

leiserfg:
The thing is that ranger does not support truecolors, i just put 
```sh
awk 'BEGIN{                                                                         [python-3.7.1]           s="OoOoOoOoOo"; s=s s s s s s s s;
           for (colnum = 0; colnum<77; colnum++) {
               r = 255-(colnum*255/76);
               g = (colnum*510/76);
               b = (colnum*255/76);
               if (g>255) g = 510-g;
               printf "\033[48;2;%d;%d;%dm", r,g,b;
               printf "\033[38;2;%d;%d;%dm", 255-r,255-g,255-b;
               printf "%s\033[0m", substr(s,colnum+1,1);
           }
           printf "\n";
       }'
```
at the biggining of my scope.sh and the output is
2018-12-30-224609_916x42_scrot

instead of

2018-12-30-224727_1391x53_scrot

leiserfg:
I am using kitty too.

toonn:
Then why do you need `COLORTERM=8bit`? It works fine for me with `truecolor`. Sounds like something else is going on. Do you have `TERM` set to something other than `xterm-kitty` or `xterm-256color`?

leiserfg:
I'm using screen-256color but bat works for me outside ranger

leiserfg:
I tired with xterm-256color and got the same error.

toonn:
Ah, that's the reason it didn't work for me in tmux. Tmux uses `TERM=screen` by default. Would be great if it just worked without having to change that tbh.

The `env COLORTERM...` is still useless on my end though. I still don't see a good reason to include it.

leiserfg:
In tmux you can use screen-256colors and it will work, I use it by default cause even when I am not a tmux user any more cause xterm-kitty does not work with ssh.

leiserfg:
Anyway, without patching COLORTERM it doesn't work, patching COLORTERM works if you have a 256color  $TERM. So I think that as ranger doesn't support truecolor properly, the patch is the best thing we can do by now.

toonn:
You still seem to be ignoring what I'm saying. For me `COLORTERM` **makes no difference whatsoever**. What *does* make a difference is `TERM`.

leiserfg:
I do understand you, but I don't know if setting TERM in `scope.sh` is a good idea, probably the best one is fixing ranger to make it work with truecolor. Without it, you (with a no 256color $TERM) will have not `bat` with colors in ranger.

By the way, I know that I can copy the terminfo (or use `kitty +kitten ssh`) but the first is annoying if you have to deal with a lot of servers and the latter doesn't work with launchers that call `terminal ssh` as Albert or roffi (there are people crying to openssh for automatic terminfo copy).

leiserfg:
IMHO it is something related with ranger splitting the ansi codes when the output is truecolor. I ran `bat` inside of other programs with ncurses (nvim, ncurses) without patching the COLORTERM var and it had no problems.

toonn:
Then we should fix that.

leiserfg:
In `ranger/gui/ansi.py` the is where the splitting happens, there all ANSI values are parsed as 256color that's the problem.

leiserfg:
I was reviewing the code and I saw that the pager uses curses and apparently it does not exposes `init_extended_pair` (that function is used in ncurses for setting the colors on truecolor mode). Then I checked others python tui with truecolor support and they don't use curses. Take a look at https://github.com/prompt-toolkit/python-prompt-toolkit.

toonn:
Well that's a drag. Don't think anyone wants to port ranger from curses to any other toolkit.

Maybe you can report the lack of the `_extended_` functions to whoever maintains python's `curses`, maybe get it fixed for newer versions at least or find out a workaround?

Migacz85:
HI guys, super simple question how I can add this patch to playaround with it ?

Migacz85:
image
much better... 
:-) 
Thank You

annata83:
> image
> much better...
> :-)
> Thank You

Hey @Migacz85  may I ask How did you achive it?

toonn:
Hmm, this was working fine until I merged it into master. Which made me notice that highlight previews aren't working on master for me either. I guess we need to fix that regression first.

PR #1408, opened 2018-12-17
Added check if bookmark file is a symlink.

husjon:
If it is, the os.rename is performed on the original file path instead of overwriting the symlink.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: Linux workstation 4.19.8-arch1-1-ARCH #1 SMP PREEMPT Sat Dec 8 13:49:11 UTC 2018 x86_64 GNU/Linux
- Python version: Python version: 3.7.1 (default, Oct 22 2018, 10:41:28) [GCC 8.2.1 20180831]
- Ranger version/commit: ranger 1.9.2
- Locale: en_GB.UTF-8

#### CHECKLIST
    - [X] Tests have been updated

#### DESCRIPTION
When the bookmarks file is a symlink, it is overwritten on every save with a new file.
This pull requests implements a check to see if this is true, and performs the save at the link original location.
If the file is a plain file, the original behaviour remains.

#### MOTIVATION AND CONTEXT
I like having the bookmarks synced between my main computer and my laptop, where the bookmark file is located on a shared drive accessible by both.
This pull requests makes this possible.

#### TESTING
I've added a test for this implementation which runs successfully.
This change should not affect any other areas of the codebase.

### DIFF
--- a/ranger/container/bookmarks.py
+++ b/ranger/container/bookmarks.py
@@ -185,7 +185,13 @@ class Bookmarks(FileManagerAware):
             old_perms = os.stat(self.path)
             os.chown(path_new, old_perms.st_uid, old_perms.st_gid)
             os.chmod(path_new, old_perms.st_mode)
-            os.rename(path_new, self.path)
+
+            if os.path.islink(self.path):
+                link_path = os.path.realpath(self.path)
+                os.rename(path_new, link_path)
+            else:
+                os.rename(path_new, self.path)
+
         except OSError as ex:
             self.fm.notify('Bookmarks error: {0}'.format(str(ex)), bad=True)
             return
--- a/ranger/container/bookmarks.py
+++ b/ranger/container/bookmarks.py
@@ -187,8 +187,8 @@ class Bookmarks(FileManagerAware):
             os.chmod(path_new, old_perms.st_mode)
 
             if os.path.islink(self.path):
-                link_path = os.path.realpath(self.path)
-                os.rename(path_new, link_path)
+                target_path = os.path.realpath(self.path)
+                os.rename(path_new, target_path)
             else:
                 os.rename(path_new, self.path)
 
--- a/tests/ranger/container/test_bookmarks.py
+++ b/tests/ranger/container/test_bookmarks.py
@@ -64,12 +64,12 @@ def test_bookmark_symlink(tmpdir):
     bookmarkfile_orig = tmpdir.join("bookmarkfile.org")
 
     # Create symlink pointing towards the original plain file.
-    os.symlink(bookmarkfile_orig, bookmarkfile_link)
+    os.symlink(str(bookmarkfile_orig), str(bookmarkfile_link))
 
     # Initialize the bookmark file and save the file.
     bmstore = Bookmarks(str(bookmarkfile_link))
     bmstore.save()
 
     # Once saved, the bookmark file should still be a symlink pointing towards the plain file.
-    assert os.path.islink(bookmarkfile_link)
-    assert not os.path.islink(bookmarkfile_orig)
+    assert os.path.islink(str(bookmarkfile_link))
+    assert not os.path.islink(str(bookmarkfile_orig))
--- a/tests/ranger/container/test_bookmarks.py
+++ b/tests/ranger/container/test_bookmarks.py
@@ -61,7 +61,7 @@ def testbookmarks(tmpdir):
 def test_bookmark_symlink(tmpdir):
     # Initialize plain file and symlink paths
     bookmarkfile_link = tmpdir.join("bookmarkfile")
-    bookmarkfile_orig = tmpdir.join("bookmarkfile.org")
+    bookmarkfile_orig = tmpdir.join("bookmarkfile.orig")
 
     # Create symlink pointing towards the original plain file.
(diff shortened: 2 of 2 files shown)


toonn:
I think this is pretty reasonable and the extra checks aren't a big deal, imo.

Could you fix the problems travis reports though, seems like there's incompatibilities across python versions, we support all the way back to py2.6.

In the meantime you could consider using a bind mount rather than a symlink.

tests/ranger/container/test_bookmarks.py
```diff
@@ -56,3 +56,20 @@ def crash():
         secondstore.update_if_outdated()
     secondstore.update = origupdate
     secondstore.update_if_outdated()
+
+
+def test_bookmark_symlink(tmpdir):
+    # Initialize plain file and symlink paths
+    bookmarkfile_link = tmpdir.join("bookmarkfile")
+    bookmarkfile_orig = tmpdir.join("bookmarkfile.org")
```
Vifon: `.org` is a potentially misleading choice as it's used extensively by Emacs org-mode. I'd rather use `.orig`.

tests/ranger/container/test_bookmarks.py
```diff
@@ -56,3 +56,20 @@ def crash():
...
+
+    # Create symlink pointing towards the original plain file.
+    os.symlink(str(bookmarkfile_orig), str(bookmarkfile_link))
```
Vifon: Did you consider using `.decode()` instead of `str()`?

tests/ranger/container/test_bookmarks.py
```diff
@@ -56,3 +56,20 @@ def crash():
...
+
+    # Initialize the bookmark file and save the file.
+    bmstore = Bookmarks(str(bookmarkfile_link))
+    bmstore.save()
+
+    # Once saved, the bookmark file should still be a symlink pointing towards the plain file.
```
Vifon: As a formality, I'd check the symlink target path too, not only it being a symlink, since the comment suggests it is checked.

tests/ranger/container/test_bookmarks.py
```diff
@@ -56,3 +56,20 @@ def crash():
...
+def test_bookmark_symlink(tmpdir):
+    # Initialize plain file and symlink paths
+    bookmarkfile_link = tmpdir.join("bookmarkfile")
+    bookmarkfile_orig = tmpdir.join("bookmarkfile.org")
+
+    # Create symlink pointing towards the original plain file.
+    os.symlink(str(bookmarkfile_orig), str(bookmarkfile_link))
+
+    # Initialize the bookmark file and save the file.
+    bmstore = Bookmarks(str(bookmarkfile_link))
```
Vifon: The same as above.

tests/ranger/container/test_bookmarks.py
```diff
@@ -56,3 +56,20 @@ def crash():
...
+    bmstore.save()
+
+    # Once saved, the bookmark file should still be a symlink pointing towards the plain file.
+    assert os.path.islink(str(bookmarkfile_link))
+    assert not os.path.islink(str(bookmarkfile_orig))
```
Vifon: Once again I'd replace `str()` with `.decode()`.

tests/ranger/container/test_bookmarks.py
husjon: I agree, I'll make the change.

tests/ranger/container/test_bookmarks.py
husjon: Not possible unfortunately since `tmpdir` is a fixture from pytest.
It is throwing an AttributeError:
`AttributeError: 'LocalPath' object has no attribute 'decode'`

tests/ranger/container/test_bookmarks.py
husjon: The check happens after the bookmark file is saved, since the `bookmarkfile_orig` does not  exist until `bmstore.save()` has been called.

tests/ranger/container/test_bookmarks.py
Vifon: Ah, I thought it was initially a `bytes` object. If that's a path object, that's fine then.

tests/ranger/container/test_bookmarks.py
Vifon: Good point.

ranger/container/bookmarks.py
```diff
@@ -185,7 +185,13 @@ def save(self):
             old_perms = os.stat(self.path)
             os.chown(path_new, old_perms.st_uid, old_perms.st_gid)
             os.chmod(path_new, old_perms.st_mode)
-            os.rename(path_new, self.path)
+
+            if os.path.islink(self.path):
+                link_path = os.path.realpath(self.path)
```
toonn: Wouldn't `target_path` be a clearer name for this? It's the link's *target* we want to write to, right?

husjon: It would indeed.
`link_path` does kind of hints towards the path where is link is located.

husjon: I've made the change, travis should be done shortly.

PR #1412, opened 2018-12-20
Add a snap

evandandrea:
#### ISSUE TYPE

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION

It's currently hard to get an up to date version of Ranger on older versions of Ubuntu. 16.04 is at 1.7.1.

#### MOTIVATION AND CONTEXT

I've taken a bit of time to put a snap together (I work on Snapcraft) so ranger can be installed and upgraded from any version of Ubuntu, or one of many other distros (also version independent). As the Snap Store is [browsed](https://snapcraft.io/store) by millions, it'll also have the bonus of getting some more eyeballs on ranger.

#### TESTING

To build it, `snap install snapcraft` from Ubuntu, or [install Multipass](https://github.com/CanonicalLtd/multipass/releases) on MacOS then `brew install snapcraft`. Then run `snapcraft` to create the .snap file.

To publish, [create an account](https://snapcraft.io/account), then:
```
snapcraft login
snapcraft register ranger
snapcraft push --release=stable *.snap
```

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,24 @@
+name: ranger
+base: core18
+version: git
+summary: A VIM-inspired filemanager for the console
+description: |
+  Ranger is a console file manager with VI key bindings. It provides a
+  minimalistic and nice curses interface with a view on the directory
+  hierarchy. It ships with rifle, a file launcher that is good at
+  automatically finding out which program to use for what file type.
+
+grade: stable
+confinement: classic
+
+parts:
+  ranger:
+    plugin: python
+    python-version: python3
+    source: .
+
+apps:
+  ranger:
+    command: python3 $SNAP/bin/ranger
+    environment:
+      PYTHONPATH: $PYTHONPATH:$SNAP/lib/python3.6/site-packages


toonn:
It shouldn't be too hard actually, clone the repository then `make && make install` should be all you need. Ranger's also in nixpkgs so you could use nix to get it on ubuntu or any other supported platform. Brew also has support for both the latest stable release *and* `HEAD`.

Note that I'm not against adding a snap to snapcraft but I did label this with needs-opinions because we don't keep the nixpkgs expression in tree, nor the debian, arch, whatever packages. The closest thing is support for PyPI in our `Makefile` afaik. So we need to discuss how we want to deal with this.

Vifon:
@evandandrea Are the snaps usually maintained by the upstream or rather by the external maintainers (like with the regular repos)? I don't think any us use snaps regularly so I wouldn't count on us testing it properly internally.

evandandrea:
@Vifon they're typically maintained by the upstream to ensure users can trust and find support for what they install. I think it's also beneficial to you as publisher, as it'll give you install base metrics (anonymized) and we can help with [promotion](https://blog.ubuntu.com/2018/11/29/gitkraken-picks-up-community-demand-to-adopt-snaps).

toonn:
We're not all that interested in maintaining packaging. We only do it for PyPI. Why can debian/arch/nixos/and others maintain their own packages but not snapcraft? It'd be great if someone in the community assumes the mantle but like the others, it's not going in the main repo.

Vifon:
The main problem is that we'd need to have Snap installed to maintain the package and I don't think we have any interest in it as the users of our systems. Of course a VM would work too but it's an additional step. I'd be fine with a community-supported package.

toonn:
Closing this because of a lack of interest on our part. We would definitely point to a snap for ranger from the readme if it's maintained by someone interested in doing so.

Issue #1417, opened 2018-12-24
Feature request: Builtin plugin manager/installer


cjbassi:
I currently use a few plugins for ranger and I was curious about the idea of adding a plugin manager similar to those built on top of vim, like vim-plug. The main features would be to install, update, and remove plugins based on a git url, most likely specified in `rc.conf`. Also, commands specified in `commands.py` could be turned into a plugin that could be installed through the manger. And plugins would be installed to `$XDG_DATA_HOME/ranger/plugins` aka `~/.local/share/ranger/plugins`.

While the current plugin/commands.py system works, it feels rudimentary since you have to manually copy files into your `~/.config/ranger` directory and also it's difficult to update plugins once they're installed. This would also fix https://github.com/ranger/ranger/issues/1303. I realize this might take some work but I thought I would get some opinions first before starting anything. Thanks!

Vifon:
I'd rather have a separate script handle this. Ranger has no business in accessing the web and remote code execution is the last thing it should do. Some additional config file could control which plugins to download and enable (by copying them into `~/.config/ranger/plugins/`) and manage them this way. What do you think?

cjbassi:
Yah I've been playing around with the idea of having a separate script, but I think it would be more convenient and cleaner overall to have it built in. Could you elaborate on what you mean by remote code execution?

I'm also curious about the best way to handle installing different plugins, since it seems like you would need some sort of spec. I guess you could just require plugins to have a makefile with an install and uninstall command.

cjbassi:
So I thought about it some more and I think I came up with a good solution, so let me know what you think. So don't add a builtin plugin manager, but update ranger so that plugins are now loaded from `~/.local/share/ranger/plugins/{plugin-name}/plugin/*.py`. So create a spec requiring plugins to have a `plugin` folder containing the python files, and then you can just `git clone` repos into `~/.local/share/ranger/plugins/`. This is basically what vim-plug does and it works great.

And then either keep or remove the current functionality of also loading plugins from `~/.config/ranger/plugins/*.py`.

Thoughts on this?

toonn:
Does the plugin directory serve any purpose? It seems a little redundant.
And instead of loading all the python files in a repo, how about requiring a `plugin.py` or an __init__.py so the directory can be loaded as a module?

cjbassi:
[...]

PR #1418, opened 2018-12-24
Load plugin directories located in ~/.config/ranger/plugins

cjbassi:
#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Adds support for loading plugins located in `<ranger.args.datadir>/plugins/` (defaults to `~/.local/share/ranger/plugins/`). Plugins need to have a `__init__.py` file as an entrypoint.

#### MOTIVATION AND CONTEXT
Solves issues with plugin standardization, installation, uninstallation, and updating. It is intended that plugins will be `git clone`d into the new plugin location and managed there. Also makes it much easier to create a ranger-plugin-manager. Plugins will have to be updated to work with the new location.

### DIFF
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -407,8 +407,15 @@ def load_settings(  # pylint: disable=too-many-locals,too-many-branches,too-many
         except OSError:
             LOG.debug('Unable to access plugin directory: %s', plugindir)
         else:
-            plugins = [p[:-3] for p in plugin_files
-                       if p.endswith('.py') and not p.startswith('_')]
+            plugins = []
+            for path in plugin_files:
+                if not path.startswith('_'):
+                    if path.endswith('.py'):
+                        # remove trailing '.py'
+                        plugins.append(path[:-3])
+                    elif os.path.isdir(os.path.join(plugindir, path)):
+                        plugins.append(path)
+
             if not os.path.exists(fm.confpath('plugins', '__init__.py')):
                 LOG.debug("Creating missing '__init__.py' file in plugin folder")
                 fobj = open(fm.confpath('plugins', '__init__.py'), 'w')


cjbassi:
I updated this and it now works for me on python versions 2.7.15 and 3.4.5. Are there any concerns left about the structure? I removed the need to have a plugin folder, so now plugins just need a `__init__.py` file.

toonn:
Ah, I guess answers my comment on #1417 already.

ranger/core/fm.py
```diff
@@ -293,15 +293,15 @@ def copy(src, dest):
                     shutil.copy(self.relpath(src), self.confpath(dest))
                 except OSError as ex:
                     sys.stderr.write("  ERROR: %s\n" % str(ex))
-        if which == 'rifle' or which == 'all':
```
toonn: I don't disagree with these changes, just seems a little unfortunate to tie them up with a fairly major change to plugins.

.gitignore
```diff
@@ -17,3 +17,4 @@
 
 .idea
 .pytest_cache
+.mypy_cache/
```
toonn: What's this mypy? Could we benefit from using it?

ranger/core/main.py
```diff
@@ -471,7 +508,7 @@ def allow_access_to_confdir(confdir, allow):
         else:
             LOG.debug("Created config directory '%s'", confdir)
         if confdir not in sys.path:
-            sys.path[0:0] = [confdir]
+            sys.path.insert(0, confdir)
```
toonn: I like this. Way less fancy-shmancy, way more clear re intent.

ranger/core/main.py
```diff
@@ -436,6 +437,42 @@ def load_custom_commands(*paths):
...
+            if ex.errno != errno.EEXIST:
+                raise
+        sys.path.insert(0, plugindir)
+        try:
+            plugins = os.listdir(plugindir)
+        except OSError:
+            LOG.debug('Unable to access plugin directory: %s', plugindir)
+        else:
+            ranger.fm = fm
+            for plugin in sorted(plugins):
```
toonn: Fine for now, though I'd expect `os.listdir()` to return directory contents in a stable order?

Have you put any thought into topological sorting or something in case people want to be able to write plugins that'll only work if other plugins are already present?

How about filtering this in some way to weed out non-plugins? Seems more foolproof to me that way. Not a big deal though since this is still experimental™.

.gitignore
cjbassi: Mypy is basically the new builtin python type hinting, and vscode makes that cache folder for it for some reason. Mypy is pretty cool, especially for larger projects like ranger. You just add type hints primarily to function declarations and it helps your IDE debug typing issues. It doesn't enforce anything at run time, just helps your IDE. And you can add it incrementally. Although idk how well it would play with python2 or older versions of python3.

ranger/core/main.py
cjbassi: It seems to be in a stable order, just not sorted order.

The topological sorting is an interesting idea. Probably something to think about and come back to if we deem it important. 

Yah I was wondering if I should do any sort of weeding out. Should I filter out any non directories? Might be a good place to start.

cjbassi:
So the way this currently works, existing plugins would need to change their plugin file to a `__init__.py` file to work, but we could always change what we require that filename to be. We could also just import all `.py` files from each plugin which should make it so that plugins wouldn't' have to change anything. Let me know. Also, plugins should (optionally) add a `.gitignore` containing `*.pyc`.

So in this PR, plugins currently need to be put under `~/.local/share/ranger/plugins/`, but did we want to change it so the folders stay in the `~/.config/ranger/plugins/`? I think the first option more appropriately follows the XDG spec, but if we want to keep using the .config file since that's already used, then that makes sense too.

ranger/core/main.py
cjbassi: Yah I didn't know what that line did at first XD although it is a neat bit of syntax

.gitignore
Vifon: So like @toonn said in the other comment: nice but I'd rather see it as a separate change not tied to this PR. It makes reverting stuff easier if such need arises.

Vifon:
The best scenario: the plugins in `~/.config/ranger/plugins/` behave like they used to me and the new behavior is applied only to `~/.local/share/ranger/plugins/`. If it's not viable (code gets too hairy or something) we should be able to find some other compromise, for example we'd load `.../plugins/*/__init__.py` *AND* `.../plugins/*.py`.

toonn:
Well, I don't really think `XDG_DATA_HOME` is the right place for plugins, which are configuration. And yes, I agree we need to allow for both `__init__.py` style plugins and simple single file plugins.

Vifon:
@toonn If they are mostly being managed by a tool, I'm willing to admit they are quasi-data even if mostly for the sake of compatibility. They'd become not unlike `tagged` or `bookmarks` files.

toonn:
I'd argue that's up to the *external* tool to manage, with symlinks or some extra configuration for ranger. But people should be able to just copy `.config/ranger` to another machine and expect everything to just work, modulo system incompatibility or presence of optional deps.

Vifon:
> people should be able to just copy .config/ranger to another machine and expect everything to just work

Very good point. As a user that's definitely what I'd expect, though copying the plugins directly sounds dirty. I think we should have a new config in `~/.config/ranger/` for specifying the plugins the yet-to-be-written plugin manager to download. That would be the most important long-term improvement of this plugin overhaul: a full config would no longer contain the plugins, just a spec which ones to download.

Additionally, while I think the downloading of plugins should be done with an external utility, we could add a ranger command one could include in `rc.conf` to download the missing plugins on ranger startup *optionally* (opt-in).

toonn:
I still disagree much overhaul is necessary. Imo, we only need something like `plugin_path` or `plugin_paths`. Then people that want to could add an external plugin manager just by choosing a directory in the settings of both tools. And since ranger loads all plugins the manager could provide a plugin that reads another config and auto-downloads and loads the specified plugins.

I agree providing either a command in `commands.py` or a plugin in the examples directory or something could be provided by default. That way it's the only plugin people need to manually install. Not sure how we'd run a command on startup though, having to start ranger with `ranger --cmd=update_plugins` every time would probably be tedious. A command that installs the plugin manager plugin would be nice though. The command would copy the PiMP into `.config/ranger` so the PiMP is available as soon as you copy your config to another box and then the PiMP reads your `plugins.conf` or something, installing plugins to the datadir.

cjbassi:
So I like the idea of having a config file with the plugins you want to install, and just store that in ~/.config. I'm not sure about the idea of having ranger download/install the plugin manager tho, since that just adds a layer of indirection and I thought we weren't going to have ranger do any network stuff. What might be good is to have the plugin manager be in its own repo and have users clone that into the plugin folder, and that could be the only plugin that is manually cloned there. Then ranger could load it like a normal plugin, and it could provide some extra functions to help with plugin management through ranger. That's the idea behind vim plug basically :D

toonn:
Well, including the command doesn't mean you have to use it. I'd say it's fairly harmless. And I'm not opposed to including a PiMP plugin with ranger which would simply be copied from the examples folder, so no network access past installing ranger. This may sound similar but the only change to ranger it'd require is providing an extra command, which is really simple (and almost as convenient as having ranger do the plugin managing). : )

cjbassi:
So somehow I hadn't seen the examples folder XD but I checked it out and I see what you're saying now.

So few things:

1. I still think ranger should add support for plugins that are in separate folders under the main `~/.config/ranger/plugin` folder, so that you can just git clone a plugin and it'll work. This would make a plugin manager easier to implement too.

2. I don't think ranger should check for updates on every start. Just providing an update function in the plugin manager should be enough.

3. Copying the plugin manager from `examples` is definitely an interesting idea, but I would personally prefer to have a plugin manager in a separate repo owned by the ranger organization. I think this issue is a little more subjective, and it mainly comes down to how centralized you want ranger to be. I think vim has already solved the plugin problem in a good way by being decentralized, and I think separating out the plugin system is a good way of doing that. Otherwise you have to add extra logic to ranger core to handle copying and updating the plugin manager instead of treating the plugin manager like a normal plugin. And then you introduce issues in regards to updating the plugin manager, since it can only be updated after ranger has been updated. Having a separate repo just feels much cleaner to me.

cjbassi:
3. I think it also provides a more clear separation of concerns and is more clearly opt in or out. And installation shouldn't be more difficult since you're running one cli command versus 1 command from within ranger. And it's also a very familiar setup for people coming from the vim ecosystem. :D

Vifon:
> I still think ranger should allow plugins to be organized into separate folder under the main plugin folder, which makes it so plugins can be simply be managed with git and provides more plugin isolation.

No disagreement here, it should be simple to load both formats of plugins without breaking compatibility (everything that used to work will still work and some thing that didn't will now work).

> I don't think ranger/the-plugin-manager should check for updates on start. Providing an update function in the plugin manager should be enough.

Check for updates: no. Check for *missing* plugins: yes. Or rather call `our-manager install-missing` or something equivalent that would simply instruct the plugin manager to check if every specified plugin is there. No network communication necessary after the first run so it would be quite fast.


Where to keep the plugin manager is a secondary concern. Maybe it could be included as a separate installed program just like `rifle`, that's yet to be decided.

Vifon:
If you're unsure how to develop it git-wise, you won't go wrong with a separate repo. We can always use `git subtree` to incorporate it into the existing repository.

cjbassi:
>Check for updates: no. Check for missing plugins: yes. Or rather call our-manager install-missing or something equivalent that would simply instruct the plugin manager to check if every specified plugin is there. No network communication necessary after the first run so it would be quite fast.

Yah actually that sounds good to give users that option. Is it possible to specify commands in `rc.conf` to be run on ranger start? Otherwise it sounds like you might have to add a setting to toggle the command being run on start?

Vifon:
> Is it possible to specify commands in rc.conf to be run on ranger start?

It should be possible to call `:shell` or preferable wrap it in a new command.

cjbassi:
So I just made a plugin manager [here](https://github.com/cjbassi/ranger-plug) that is working pretty well. It requires this PR to work and also a couple of python packages. It's an early draft of course.

toonn:
RE

1. I don't see the issue. If users can specify `plugin_paths` then nothing is lost and they get to decide for themselves whether they want to keep the default `.config/ranger/plugins` or not.
2. Definitely what I would expect from a PiMP. Keeping your plugins, including the PiMP itself, up to date automatically should at least be an option.
3. I disagree here. I don't want a repo under the ranger umbrella because that implies we sanction it *and* will maintain it. Currently the team's not very interested in doing this afaict. What I'm talking about is including a super bare bones plugin in the examples directory that can do plugin install/update/self-update and a command that basically runs `cp examples/PiMP plugindir/PiMP`. This way getting started using this is just `:bootstrap_PiMP`, then `:shell vim .config/ranger/PiMP.cup` to add your desired plugins. Important features of this PiMP are simplicity/maintainability and reliability. Fancy stuff like progress bars and fancy configuration based on the environment/os/tmux/etc would not be features. This doesn't prevent anyone from implementing a far better PiMP, it's just the least effort way for the project to include a decent one by default. People could even use the included PiMP to install a better PiMP and then ditch the included one.

cjbassi:
Hi, any more thoughts on this PR? I realize this was a little messy but I was hoping we could at least add the ability to load plugin directories located in `~/.config/ranger/plugins` (which is what I changed this PR to), since it's a small change and solves several issues. The only consideration would be what the entry level python file should be called which is `__init__.py` by default. Thanks!

ranger/core/main.py
```diff
@@ -407,8 +407,14 @@ def load_custom_commands(*paths):
         except OSError:
             LOG.debug('Unable to access plugin directory: %s', plugindir)
         else:
-            plugins = [p[:-3] for p in plugin_files
-                       if p.endswith('.py') and not p.startswith('_')]
+            plugins = []
+            for p in plugin_files:
+                if not p.startswith('_'):
```
toonn: What's this about exactly? I know you didn't add it but I don't understand and wouldn't it prevent the `__init__.py`s to be loaded from the subdirectories?

ranger/core/main.py
```diff
@@ -407,8 +407,14 @@ def load_custom_commands(*paths):
...
+                    if p.endswith('.py'):
+                        plugins.append(p[:-3])
+                    elif os.path.isdir(os.path.join(plugindir, p)):
```
toonn: Why do we only need the name for files but that doesn't work for directories?

ranger/core/main.py
cjbassi: `plugin_files` is a list of files/directories in `~/.config/ranger/plugins` and the `__init__.py` files are located within one of the directories, so they're not getting ignored.

ranger/core/main.py
cjbassi: Not quite sure what you mean, but for python files, we import them using the name minus the `.py` extension, and for directories, we can just use the directory name since its treated like a module which autoloads the `__init__.py` file within it.

toonn: Why do you need the full path for directories, while just the name suffices for files?

cjbassi: Oh that's because to check if a file is a directory, we have to call `os.path.isdir` which requires a full path. While checking if a file is a python file just requires us to check if the filename is appended with `.py`.

toonn: Ah, right, didn't consider the plugindir isn't the pwd.

cjbassi:
Just fixed the variable name by renaming `p` to `path`.

ranger/core/main.py
```diff
@@ -407,8 +407,14 @@ def load_custom_commands(*paths):
         except OSError:
             LOG.debug('Unable to access plugin directory: %s', plugindir)
         else:
-            plugins = [p[:-3] for p in plugin_files
-                       if p.endswith('.py') and not p.startswith('_')]
+            plugins = []
+            for path in plugin_files:
+                if not path.startswith('_'):
+                    if path.endswith('.py'):
+                        plugins.append(path[:-3])
```
Vifon: What's this magical `-3` about?

cjbassi: It removes the trailing `.py`, since that gives the name of the python module. I just added a comment for that.

PR #1420, opened 2018-12-26
Silence KeyError for bookmarks while expanding macros

### LINKED ISSUE #1419: %any_path macro wreaks havoc when the character provided is not an existing bookmark
Try `"x` where `x` is a character not defined as a bookmark yet.

Possible solution: Make the macro values thunks that are only evaluated *if* necessary.

toonn:
The implementation for the `%any_path` macro relies on the bookmark
being defined, however when expanding the macro for uses of `%any` that
don't involve existing bookmarks you get error messages even though the
failure's expected.

Simplest solution is to just fail silently, like the existing `%any`
macro already did. Lazy evaluation would silence the error for all uses
of `%any`, since the value for `%any_path` would never be used, but that
would require altering all the generation and consumption of macro
values. Another solution is to only resolve the `%any_path` macro when
it's necessary but that would involve changing the way `wildcards` are
passed distinguishing between `any` and `any_path` macros.

Fixes #1419


Vifon:
Since it was already the behavior before introducing this feature&bug, looks sensible to me.

PR #1425, opened 2018-12-28
fix terminal flag in rifle.py

### LINKED ISSUE #1401: Open editor in separate window
#### Runtime Environment

- Operating system and version: Arch Linux, 4.19.4
- Terminal emulator and version: termite v14
- Python version: 3.7.1
- Locale:  en_US.UTF-8

#### Current Behavior
After opening a text file inside ranger, ranger is replaced by the editor.

#### Expected Behavior
Using some configuration, for example in `rifle.conf`, I can open a file with ranger in a new window, after which the window with ranger is still visible and usable.

#### Context

I'm trying to achieve the goal in "Expected Behavior" in the context of a tiled window manager, i.e., i3 version 4.16.

My `VISUAL` and `EDITOR` is vim 8.1.

#### Possible Solutions

I've tried both these configurations in `rifle.conf` but they cause ranger to not open the text file at all.

    mime ^text, flag t, label editor = ${VISUAL:-$EDITOR} -- "$@"

    mime ^text, flag f, label editor = ${VISUAL:-$EDITOR} -- "$@"

#### Steps to reproduce
1. Start ranger
2. Navigate to a text file
3. Press `l` to open the text file. The editor showing the file replaces ranger.

#### Other
Thanks a lot for ranger! 👍 ❤️

kmarius:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: 4.19.12-arch1-1-ARCH
- Terminal emulator and version: termite v14-1-gfad28a5
- Python version: Python 3.7.1
- Ranger version/commit: 1.9.2.81.g7f20f313-1
- Locale: en_US.UTF8

#### CHECKLIST

- [ ] All new and existing tests pass **[REQUIRED]** (pylint aborts on ranger.gui.widgets.statusbar)
    - [ ] Tests have been updated

#### DESCRIPTION

Firstly,  always invoke the shell with `sh -c 'cmd "$@"' _ file1 file2` instead of `sh -c 'set -- file1 file; cmd "$@"'`. This is alot easier to handle: If the terminal consumes all arguments after its exec-flag, give it this list, otherwise wrap it in quotes and concatenate it. (Note that _in theory_ this should also work with the old way `sh` was invoked, but failed on filesnames with spaces, even though all the quoting should take care of that)
This now also works with the root flag.

#### MOTIVATION AND CONTEXT

This should fix #1401, though I think there is some stuff left to do:
- I tested this with all hardcoded terminals except `lilyterm` which i couldn't build.
- On archlinux `pantheon-terminal`s executable has the name `io.elementary.terminal`. I don't know about other distros.
- Some refactoring: For example the function `build_command` used to build the `set -- file1 fil2; cmd`-string, but is pretty much useless now
- Fallback terminals in rifle.conf that take the command as a single argument don't work. Ranger can not know beforehand what the fallback is going to be, so it can't decide if it should wrap the command or not. One can, for example, give the wrapped command in an environment variable, or provide a script to wrap the args and call the terminal with `termite -e "$(quote "$@")"`.
- Launching a fallback terminal from within ranger in tmux will mangle tmux (try `TERMCMD=doesnotexist ranger.py` with one of the fallbacks that work, e.g. `xterm` or `urxvt`). This has to do with a missing hook (after the `finally` in `execute`) because we are returning from the function early.
- Some terminals don't set a unique `TERM` and just use some variant of `xterm`. This is irritating, because `xterm` will start instead of the preferred terminal. In this case the only way to have rifle start the right terminal is using `TERMCMD`. One could think about adding an option to ranger and call `rifle` with the appropriate `TERMCMD` set, so one does not have to set this variable globally.

### DIFF
--- a/ranger/ext/rifle.py
+++ b/ranger/ext/rifle.py
@@ -412,8 +412,7 @@ class Rifle(object):  # pylint: disable=too-many-instance-attributes
                     elif term in ['tilda']:
                         execflags = ['-c']
                     elif term in ['guake']:
-                        # put pwd manually here? this won't get expanded
-                        execflags = ['-n', '${PWD}', '-e']
+                        execflags = ['-n', os.environ['PWD'], '-e']
                     else:
                         execflags = ['-e']
 


ranger/ext/rifle.py
```diff
@@ -113,6 +113,11 @@ def squash_flags(flags):
     return ''.join(f for f in flags if f not in exclude)
 
 
+def shell_quote(string):
```
toonn: I'm sure shlex has a function for this, no?

ranger/ext/rifle.py
```diff
@@ -389,45 +394,39 @@ def execute(self, files,  # noqa: E501 pylint: disable=too-many-branches,too-man
...
-                            + files, flags='f',
-                            mimetype='ranger/x-terminal-emulator')
+                        self.execute(files=cmd, flags='f',
+                                     mimetype='ranger/x-terminal-emulator')
                         return None
 
                     # Choose correct cmdflag accordingly
                     if term in ['xfce4-terminal', 'mate-terminal',
-                                'terminator']:
-                        cmdflag = '-x'
```
toonn: Don't like the name, it's too generic. I want something that indicates these are the flags/arguments needed to start a command in the terminal.

ranger/ext/rifle.py
```diff
@@ -389,45 +394,39 @@ def execute(self, files,  # noqa: E501 pylint: disable=too-many-branches,too-man
...
+                        termflags = ['-e']
                     elif term in ['gnome-terminal', 'kitty']:
-                        cmdflag = '--'
-                    elif term in ['tilda', ]:
-                        cmdflag = '-c'
+                        termflags = ['--']
+                    elif term in ['tilda']:
+                        termflags = ['-c']
+                    elif term in ['guake']:
+                        # put pwd manually here? this won't get expanded
```
toonn: Why does it not get expanded? Might be a reason to stick with `set`?

ranger/ext/rifle.py
kmarius: Not in the `python2` version. I copied this from `shell_escape.py`.

ranger/ext/rifle.py
kmarius: I avoided `cmdflag` because these are not actually flags _for_ `cmd`, but I agree.  I think `execflags` fits well.

ranger/ext/rifle.py
kmarius: These arguments are passed to `Popen` and not to a shell, so they can't get expanded. I didn't change this part, and wouldn't want to wrap the whole command  in `sh -c ...` again. Just putting `os.environ['PATH']` there should work.

The `-n` flag in `guake` was fixed only recently, so no matter the argument, the `PWD` of the command was always set to `HOME`.

ranger/ext/rifle.py
toonn: Let's take [this approach](https://stackoverflow.com/questions/26790916/python-3-backward-compatability-shlex-quote-vs-pipes-quote).

ranger/ext/rifle.py
toonn: cmdflag as in the flag for executing a command. Execflags is fine with me though.

ranger/ext/rifle.py
toonn: Do you see an easy way to use the directory of the files rifle was triggered on instead of the `PWD`? Ranger doesn't change the `PWD` while navigating afair.

kmarius: The fist thing I tried was taking the directory of the first file. While that worked, I realized that this does not match the behaviour of the other terminals. They all inherit `PWD` from ranger, only `guake` needs it for the `-n` flag or it will start in `HOME`.

toonn: Hmm, maybe we should do `cd /path/to/selected/files; command-found-in-rifle.conf`?
I'm thinking this would mostly be nice with editors, like vim. You'd open a file in a repo, with flag t, then expect to be able to do `:e other-file`, rather than `:e /path/to/repo/other-file`.

kmarius: This would make sense when rifle is started from within ranger, but it's going to break things when used as a standalone script. The Files don't have to be in the same directory anyway and more importantly, relative paths miss after cd'ing.

toonn: Maybe we should make the paths absolute. I still think it'd be nice to rifle some files and have vim's working directory be set to that of either the first or last file passed, I prefer first but whatever as long as it's consistent.

Do you disagree the vim example is compelling?

Vifon: > Maybe we should make the paths absolute.

I didn't read the whole discussion but if you want to make all the paths passed to `rifle` absolute, it's a Really Bad Idea™. For some utilities, like `tar(1)`, they way the path gets passed is crucial.

kmarius:
Is there still interest in getting this fixed, or is somebody working on it? I rarely use `ranger` nowadays and would otherwise close this PR. I gained some insight while rewriting `rifle` as a personal project. The (in my opinion) easiest way is not hardcoding any terminals whatsoever and instead rely on terminals in `rifle.conf`. This would absolutely break many people's setups.

To make this work with terminals that only accept the command as a single argument (such as `termite`) one can introduce a new flag (say `e`) that escapes the given files so they can be used with `"$*"`. The only way around this currently is doing the exact same thing, but inside `rifle.conf` (or putting something in a script). I found this awkward line in my old `rifle.conf`:

    terminology -e "$(for a;do printf "'";printf -- "$a"|sed "s/'/'\\\\''/g";printf "' ";done)"

toonn:
Hmm, I'm not exactly sure about current status. I think I changed some of this.
I came to the same conclusion though, just have a list of terminals in `rifle.conf` the way we have for everything else and have `flag t` select the first terminal that matches. Then people can just put their preferred terminal at the top if they don't like the default order.

PR #1432, opened 2019-01-06
Fix the Kitty previews in fullscreen mode

### LINKED ISSUE #1431: Kitty previews don't cleanup after themselves in fullscreen mode
#### Runtime Environment

- Operating system and version: Gentoo Linux
- Terminal emulator and version: Kitty
- Python version: 3.6.5
- Ranger version/commit: e06e12ef
- Locale: en_US.utf8

#### Current Behavior

With `preview_images_method` set to `kitty`, when I fullscreen the image preview with <kbd>i</kbd> and then press <kbd>i</kbd> again, the display gets screwed up. Specifically the file listing gets shown below the space that used to be occupied by the image. Pressing <kbd>Ctrl+L</kbd> fixes it. @toonn mentioned that it happens only when the file listing is shorter than the whole terminal height.

#### Expected Behavior

Kitty previews should work in fullscreen without breaking the rest of ranger.

#### Possible Solutions

Refresh the display after clearing the Kitty preview. Fix is on the way already.

#### Steps to reproduce
1. `:set preview_images_method kitty`
2. <kbd>z</kbd><kbd>i</kbd>
3. Focus an image in a directory with not too many files.
4. <kbd>i</kbd><kbd>i</kbd>

Vifon:
Fixes #1431.

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -484,7 +484,7 @@ class URXVTImageFSDisplayer(URXVTImageDisplayer):
         return self._get_centered_offsets()
 
 
-class KittyImageDisplayer(ImageDisplayer):
+class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
     """Implementation of ImageDisplayer for kitty (https://github.com/kovidgoyal/kitty/)
     terminal. It uses the built APC to send commands and data to kitty,
     which in turn renders the image. The APC takes the form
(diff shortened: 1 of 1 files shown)


mark-dawn:
I was back in my hometown without a *nix machine for some time, sorry.
I tried replicating this in various terminal sizes and file listing lengths but to no avail.
I will try to see if I can see something outside my dev branch, maybe I did something weird in it.
At least we have a working fix for people experiencing the issue.

toonn:
@mark-dawn, note that the issue I mentioned you for is *different* from the one this PR fixes.

mark-dawn:
I don't get the flickering either, I should have made that clearer.
I can get a single slow "flicker" by pressing any input while the fullscreen preview is active, but only _once_. Nothing continuous thought.

toonn:
It's intermittent for me so it's probably hard to debug.

PR #1442, opened 2019-01-14
make Kitty image preview run in separate thread

sammoth:
This is based on PR #192. It speeds up scrolling in directories of image-previewable files.

Seems to work well for me, but I don't really know what I am doing


toonn:
It seems like you ripped out the necessary parts from the PR that would do this for *all* the image preview methods? I don't see a reason to do this only for the kitty preview method tbh.

toonn:
You think it's possible to rethink the ImageDisplayer implementation to just be asynchronous? No one in the team wants to deal with the problems that come with asynchronous processing really, so killing all the birds with one stone would be hugely more likely to be accepted.

toonn:
Keep in mind that whatever your solution it needs to be compatible all the way back to python 2.6.

sammoth:
I will give that a go, though I am not experienced with Python or writing threaded code, so if you are worried about introducing problematic async code someone with more expertise than me should review.

toonn:
Don't worry, we'll definitely review and test it before merging.
Keep in mind to run `make test` before pushing, we're strict about linting, you can find the necessary tools in `HACKING.md`.

sammoth:
Yep, at the moment the tests are crashing for me locally (rather than failing) but I will get to them

toonn:
@sammoth, do you have time to pick this up again in the near future? Otherwise I'll let this slip to next release.

sammoth:
No, afraid I haven't been able to look at this more yet.

seebye:
@toonn It would be great if the delay could be disabled for ueberzug by default as frames will already be dropped if ueberzug can't keep up with the drawing requests.

toonn:
@seebye, does ueberzug inherit from the kitty image displayer class? This PR only affects kitty. Ideally we'd implement asynchronous previews for all the methods generically. So this PR will only be accepted if it's completely finished.

seebye:
> @seebye, does ueberzug inherit from the kitty image displayer class?

@toonn No, it doesn't, but you suggested to implement it for all image displayer.  
So that's why I thought I should ask to exclude ueberzug from these changes.

toonn:
@seebye, woops, now I see.

PR #1452, opened 2019-01-25
fix Ctrl+Space crash

s-n-g:
When in input mode (e.g. search, :mkdir, :rename, etc.), pressing Ctrl+Space makes ranger crash
This will fix this issue

Crash example:
```
Python version: 3.7.2 (default, Jan 10 2019, 23:51:51) [GCC 8.2.1 20181127]
Locale: el_GR.UTF-8
Current file: '/home/spiros/1. "aaa" \'aaa\''

  File "/usr/lib/python3.7/site-packages/ranger/gui/curses_shortcuts.py", line 36, in addstr
    self.win.addstr(*args)
ValueError: embedded null character

  File "/usr/lib/python3.7/site-packages/ranger/core/main.py", line 195, in main
    fm.loop()
  File "/usr/lib/python3.7/site-packages/ranger/core/fm.py", line 394, in loop
    ui.redraw()
  File "/usr/lib/python3.7/site-packages/ranger/gui/ui.py", line 338, in redraw
    self.draw()
  File "/usr/lib/python3.7/site-packages/ranger/gui/ui.py", line 365, in draw
  File "/usr/lib/python3.7/site-packages/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/usr/lib/python3.7/site-packages/ranger/gui/widgets/console.py", line 111, in draw
    self.addstr(0, len(self.prompt), str(line[x:]))
  File "/usr/lib/python3.7/site-packages/ranger/gui/curses_shortcuts.py", line 44, in addstr
    self.win.addstr(*_fix_surrogates(args))
ValueError: embedded null character

```

### DIFF
--- a/ranger/gui/curses_shortcuts.py
+++ b/ranger/gui/curses_shortcuts.py
@@ -14,7 +14,7 @@ REVERSE_ADDCH_ARGS = sys.version[0:5] == '3.4.0'
 
 def _fix_surrogates(args):
     return [isinstance(arg, str) and arg.encode('utf-8', 'surrogateescape')
-            .decode('utf-8', 'replace').replace('\u0000', '') or arg for arg in args]
+            .decode('utf-8', 'replace').replace(u'\u0000', '') or arg for arg in args]
 
 
 class CursesShortcuts(SettingsAware):
--- a/ranger/gui/curses_shortcuts.py
+++ b/ranger/gui/curses_shortcuts.py
@@ -14,7 +14,7 @@ REVERSE_ADDCH_ARGS = sys.version[0:5] == '3.4.0'
 
 def _fix_surrogates(args):
     return [isinstance(arg, str) and arg.encode('utf-8', 'surrogateescape')
-            .decode('utf-8', 'replace') or arg for arg in args]
+            .decode('utf-8', 'replace').replace('\u0000', '') or arg for arg in args]
 
 
 class CursesShortcuts(SettingsAware):


Vifon:
Your patch is failing for Python 2.7. Try replacing `'\u0000'` with `u'\u0000'`, which should work for both Py2 and Py3.

toonn:
@Vifon, imo that only makes sense if we're changing the other place where we catch this error to use this method. So either this needs to catch the `ValueError` or we should change the other location to do the replace too.

s-n-g:
@Vifon oh, sorry...
I was under the impression that ranger targeted py3 only

Vifon:
It should be fixed in a more general manner as we already patched a similar problem with NULs in previews but it'll certainly do for a time being. Thanks a lot!

toonn:
I fear if you don't open an issue on it, it won't, @Vifon : p

Vifon:
Wait… It *didn't* actually fix it for me! My bad for not checking.

toonn:
Hasn't fixed it for me either, let's revert and use the error catching approach.

toonn:
Hmm, seems to insert a space now. Kinda weird because I don't see any replacing in the code but I actually like the behavior I think.

Vifon:
For me it actually blanks the whole line (try typing something and then pressing <kbd>Ctrl+Space</kbd>) but it's 100x better than crashing!

toonn:
Oh, that's still terrible : /
It blanks the line, adds what looks like a space and then when you press return you get an error message about an embedded null. We should just catch the error and drop the null imo.

PR #1454, opened 2019-01-25
container.bookmarks: Remove invalid bookmarks only on access

### LINKED ISSUE #1365: ranger 1.9: all autofs mounts triggered on start?
Since the switch to 1.9 all autofs mounts seem to be triggered when starting ranger, as well as every time the bookmark menu is requested. As not all mounts are available, and even though timeouts are defined, ranger stalls indefinitely. Disabling autofs allows fluid use as before.

Vifon:
Previously ranger was validating the bookmarks (whether they are existing
directories) when loading the bookmark file.  It caused the remote
filesystems like autofs to wake up unnecessarily and in general caused
more problems than solved.

Fixes #1365.

### DIFF
--- a/ranger/container/bookmarks.py
+++ b/ranger/container/bookmarks.py
@@ -12,7 +12,7 @@ from ranger.core.shared import FileManagerAware
 ALLOWED_KEYS = string.ascii_letters + string.digits + "`'"
 
 
-class Bookmarks(FileManagerAware):  # pylint: disable=too-many-instance-attributes
+class Bookmarks(FileManagerAware):
     """Bookmarks is a container which associates keys with bookmarks.
 
     A key is a string with: len(key) == 1 and key in ALLOWED_KEYS.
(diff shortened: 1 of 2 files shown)


toonn:
Hmm, should we really delete a bookmark just because an nfs volume can't be reached?

Vifon:
I kept it mostly for backwards compatibility, just made it lazy. Now that I checked, it was introduced only in 2017, so after re-evaluating I agree, let's drop this part.

eMPee584:
Mhh, while this indeed did fix #1365, now the stalling (due to invocation of autofs mounts) happens when I change the sort order?.. very odd.. tried bisecting it, but didn't come to a sensible conclusion :thinking:

toonn:
Is that a regression or did that happen before too? Hard to test I guess, since it's probably gone once they're mounted the first time... : /

Vifon:
@toonn, it should be possible to test it with `strace -e %file -f ranger`.

eMPee584:
Ok I tried bisecting again, but again not quite sure.. could this be related?
abacb776a7aab93ccab66512975a465d4c28a5ee `is the first bad commit`
before this, autofs mounts seem not to be triggered on sorting (but on initial load and on bookmark selection) ... However I do not understand why this change would cause that..
`strace` shows only that `stat` is triggered on the autofs paths.. tried to find out again where that is invoked from with `gdb -p $(pgrep -n ranger)` (`catch syscall stat`) but that doesn't give nice backtraces..

toonn:
@eMPee584, could you open a new issue on that? Please be detailed cause these bugs aren't exactly easy to find.

PR #1475, opened 2019-02-14
[wip] Granular VCS interrupt

### LINKED ISSUE #902: Ranger takes long to exit if vcs_aware is true
If I enable `vcs_aware`, then in a git repo with a large amount of untracked files, it takes very long to quit ranger (5 seconds +/-). 

With it disabled, when exiting, it's pretty much instantaneous. 

Thank you for this amazing software!

megheaiulian:
Granular VCS interrupt
Allows earlier cancellation of VCS tasks
#### ISSUE TYPE

- Bug fix
fixes #902

### DIFF
--- a/ranger/ext/vcs/vcs.py
+++ b/ranger/ext/vcs/vcs.py
@@ -256,9 +256,11 @@ class VcsRoot(Vcs):  # pylint: disable=abstract-method
         self.updatetime = time.time()
         return True
 
-    def _update_walk(self, path, purge):  # pylint: disable=too-many-branches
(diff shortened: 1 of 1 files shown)


toonn:
I'm not a big fan of this tbh. It's incredibly messy and you can't continue an interrupted call. Making this async would be a better solution and could allow picking up an interrupted process if done right.

*If* we add async it should be done right, i.e. pervasive in all of ranger. That's not an easy task but feel free to come to us with designs.
Keep in mind you can't use any async methods that don't go back to 2.6.

Issue #1473, opened 2019-02-13
:paste <path>


zzzdeb:
Feature Request:
It would be nice if paste function take path as argument. For example I select multiple files and move them with short_cut f.e. with ',mD' to Documents. For now I achieve this behaviour through shell command. But then I lose features of ranger like task, status bar ...

toonn:
Mark some files, using space, then `yy` to yank them or `dd` to cut them, move to `~/Documents` and then `pp` to paste them.

zzzdeb:
1. that is exactrly the reason why I wish this feature. For every frequently used dirs&files, I have shortcuts to do basic actions like opening, opening in new tab, moving or coping files to them. Making this actions doesnt need to make me leave current dir to come back after paste, I think. I just want to do this with one shortcut. I achieved with chain command better result. Only it erases my last visited dir also cant use '' after this action. Still feels hacky for basic action.
```
map Ya chain copy; cd  ~/hiwi/ACPLT-DevKit-linux64/acplt; paste; cd -
map ma chain cut; cd  ~/hiwi/ACPLT-DevKit-linux64/acplt; paste; cd -
```
2. For paste function to be able to paste only in current dir seems to be too harsh restriction.
3. As I see in source code
``` python
class CopyLoader
def __init__
self.original_path = self.fm.thistab.path
```
defines where to paste. Seems easily modifieble.

toonn:
Hmm, seems like you're just looking for the `:mv` or `:cp` commands?

zzzdeb:
exactly

toonn:
👍

zzzdeb:
I am new in github. I dont understand what this means. Does it mean that was it?

toonn:
@zzzdeb I ask whether `:mv` or `:cp` was what you wanted and you said yes...

You want to do `:paste %s /to/some/path`, right? That sounds like `:mv %s /to/some/path` would replace your cut/paste and `:cp ...` would replace your copy/paste.

zzzdeb:
mv cp functions implemented? I couldnt find them and understand what :move does. No documentation for :move.  Should I open another issue for these functions?

toonn:
Oh, you're right, i apologize. I assumed we had wrappers for them since we do for some other thing like `:cd` and `:touch`, etc.

I suggest either creating custom commands or just mapping something along these lines: `map pm console shell mv %s%space` or `map pc console shell cp %c%space`.

[...]

PR #1482, opened 2019-02-24
Implementing destination argument for the built in paste command

palikar:
This PR is directly addressing #1473 . I've implemented a _dest_ argument for the built in _paste_ command.

#### TYPE

#### RUNTIME ENVIRONMENT
- Debian GNU/Linux busterr/sid
- Urxvt 9.22
- Python 3.6.6 
- ranger-master 1.9.2  
- Locale: en_US.UTF-8

    - [ ] Tests have been updated

#### DESCRIPTION

I've just added one additional argument to the `paste` method in the `Actions` class. The given value gets passed to the `CopyLoader` object and if it is not `None` and given it is a proper path, the it will be used as destination path.  

#### MOTIVATION AND CONTEXT
Everything discussed in #1473 and several personal problems with ranger. I really wanted to be able to quickly copy/paste/cut files in some of my bookmarks. The most of the functionality can be implemented as custom function/plugin and the accompanying keybindings but the problem with the paste stays. Programmatically or not, one can only paste in the current directory and nowhere else. With the made changes, I can now have custom function like 

```python
class copy_to_bookmark(Command):
    """
    Copy/cut file(s) to bookmark
    """    

    def execute(self):
        key = self.arg(1)
        if key in self.fm.bookmarks:
            destination = str(self.fm.bookmarks[str(key)])

            if self.arg(2) == '-c':
                self.fm.copy(mode='set')
            elif self.arg(2) == '-d':
                self.fm.cut(mode='set')    
            self.fm.paste(dest=destination)
            self.fm.uncut()
        else:
            self.fm.notify('The key isn\'t bound to bookmark!')
```

I also have several key bindings for the function:
```
map mp<any>     copy_to_bookmark %any -p
map mc<any>     copy_to_bookmark %any -c
map md<any>     copy_to_bookmark %any -d
```
which allow me to quickly perform the action I want.

- <kbd>mp</kbd> - pastes the files in the copy_buffer in the chosen bookmark
- <kbd>md</kbd> - cuts the currently selected file in the chosen bookmark 
- <kbd>mc</kbd> - copies the currently selected file in the chosen bookmark

### DIFF
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -51,11 +51,11 @@ class Loadable(object):
 class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instance-attributes
     progressbar_supported = True
 
-    def __init__(self, copy_buffer, do_cut=False, overwrite=False):
+    def __init__(self, copy_buffer, do_cut=False, overwrite=False, dest=None):
         self.copy_buffer = tuple(copy_buffer)
         self.do_cut = do_cut
         self.original_copy_buffer = copy_buffer
-        self.original_path = self.fm.thistab.path
+        self.original_path = dest if dest is not None else self.fm.thistab.path
         self.overwrite = overwrite
         self.percent = 0
         if self.copy_buffer:
@@ -102,7 +102,7 @@ class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instan
                         tag = self.fm.tags.tags[path]
                         self.fm.tags.remove(path)
                         self.fm.tags.tags[
-                            path.replace(fobj.path, self.original_path + '/' + fobj.basename)
+                            path.replace(fobj.path, path.join(self.original_path, fobj.basename))
                         ] = tag
                         self.fm.tags.dump()
                 n = 0
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -398,8 +398,8 @@ class Loader(FileManagerAware):
                 else:
                     break
             except IndexError:
-
                 return
+
         item.unpause()
 
         self.rotate()
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -398,8 +398,8 @@ class Loader(FileManagerAware):
                 else:
                     break
             except IndexError:
-                return
 
+                return
         item.unpause()
 
         self.rotate()
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -476,6 +476,9 @@ map pl paste_symlink relative=False
 map pL paste_symlink relative=True
 map phl paste_hardlink
 map pht paste_hardlinked_subtree
+map pd console paste dest=
+map p`<any> paste dest=%any_path
+map p'<any> paste dest=%any_path
 
 map dD console delete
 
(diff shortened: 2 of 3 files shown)


ranger/core/loader.py
```diff
@@ -51,11 +51,11 @@ def destroy(self):
 class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instance-attributes
     progressbar_supported = True
 
-    def __init__(self, copy_buffer, do_cut=False, overwrite=False):
+    def __init__(self, copy_buffer, do_cut=False, overwrite=False, dest=None):
         self.copy_buffer = tuple(copy_buffer)
         self.do_cut = do_cut
         self.original_copy_buffer = copy_buffer
-        self.original_path = self.fm.thistab.path
+        self.original_path = dest if dest is not None else self.fm.thistab.path
```
toonn: I'm wondering if we could not just define the default for `dest=self.fm.thistab.path`?

Vifon: I don't think it'd work with a "dynamic" value as they are only assigned *once*. (that's also the reason why a mutable default is a very bad idea most of the time)

palikar:
I can't quite figure out why is `make test` failing. It fails on my machine too (as in the Travis builds) but it seems that the failing part is the _pylint_ . Locally, when I run _pytest_ every test passes. When I run pylint, however, it dies with some Exception from some `node_classes` class from the `astroid`. And yes, I am not sure what that means and how am I suppose to fix it :/ .

ranger/core/loader.py
palikar: Yep, probably not a good idea. I've just read [this](https://docs.python-guide.org/writing/gotchas/) and I think what it will happen is that the default `dest? will always be the directory of the first tab in ranger.

toonn: Thanks for pointing this out, @vifon.

@palikar, the current failure in travis is because pylint found some trailing whitespace somewhere. We're strict about applying pep8 and pylint, so even if the tests pass you'll have to pass the linting as well. Make sure you have all the testing dependencies, flake8, pylint, etc.

palikar: Ok, I cleared up the code and all checks on the Travis build now pass.

ranger/core/loader.py
```diff
@@ -398,8 +398,8 @@ def work(self):
                 else:
                     break
             except IndexError:
-                return
```
toonn: This seems like a weird whitespace change, pretty sure you didn't intend it?

toonn:
Great first contribution! 🎉 

I added default bindings for pasting with a path argument `pd` and for pasting to a bookmark `p'<any>` and ``p`<any>``. I didn't really understand your `mc` and `md` mappings. Why do you have to provide a bookmark for those? It seems to me like they just copy/cut files from the current directory, am I missing something? If not then you don't need that custom command to achieve this anymore : ) Yay reduced patch maintenance! 😁

palikar:
@toonn 

I don't think you are missing anything. My mapping are really just tailored to my personal tastes. In my mind, I keep everything "bookmarky" under my <kbd>m</kbd> key. The <kbd>mc</kbd> ("c" for copy) and the <kbd>md</kbd> ("d" for... cut) are just there to allow me to quickly send the file under the cursor to a specific bookmarked directory. I do understand how this my be a bit overkill for the general user and your solution with `p'<any>` is probably better default than my monstrosity.

toonn:
@palikar, oh no, it makes kinda sense to put them under `m`. The part I didn't understand is why your `mc` and `md` have the `<any>` bit. You have to pass a bookmark, one that exists nonetheless, otherwise it won't work but they don't actually do anything with the bookmark.

Issue #1466, opened 2019-02-09
%confdir not set in rc.conf


SilverFishCat:
#### Runtime Environment

- Operating system and version: 4.20.7-arch1-1-ARCH
- Terminal emulator and version: alacritty 0.2.7
- Python version: 3.7.2 (default, Jan 10 2019, 23:51:51) [GCC 8.2.1 20181127]
- Ranger version/commit: ranger 1.9.2
- Locale: en_US.UTF-8

#### Current Behavior
When %confdir is used in rc.conf an error is printed to stdout.
The error is:
```
Error in line `echo %confdir':
  'NoneType' object has no attribute 'thisfile'
```
Ranger continous normally without executing that line

#### Expected Behavior
The %confdir macro should expand to $XDG_CONFIG_HOME/ranger or whatever else was given by --confdir

#### Context
I want to use source %confdir/<extraconf> to be able to change a set of key bindings with as little changes to rc.conf as possible (specifically, changing J and K), so these remapping can be shared easily.

#### Steps to reproduce
1. Add `echo %confdir` to rc.conf
2. Run ranger
3. Close ranger

toonn:
I didn't think sourcing works in `rc.conf`?

SilverFishCat:
Seems to work just fine for me.
I have the default rc.conf (or at least I didn't change anything that I can remember) with the line `source ~/.config/ranger/sometext` at the end.
That file only contains these lines:
```
copymap <UP>       j
copymap <DOWN>     k
```
This works, and the longer file works too as long as I don't use %confdir.

toonn:
@Vifon, did you implement sourcing of configs? I thought we didn't have this?

Vifon:
Huh? No, I don't think so.

toonn:
@SilverFishCat, wouldn't `.` work like `%confdir` actually?

SilverFishCat:
@toonn no, unless I run ranger with `~/.config/ranger/` as my working directory

toonn:
I'd expect relative paths in `rc.conf` to be relative to its location tbh.

`%confdir` probably doesn't work because `rc.conf` is executed very early on startup.

palikar:
Ok, I've looked into this issue and I've figured it out.

@toonn  sourcing __is__ supported in `rc.conf` as `source` is a function defined in `action.py`. Everything in `actions.py` can be used as a console command or command in `rc.conf`. The sourcing is not the problem here.
[...]

PR #1492, opened 2019-03-11
Fixing macro expansions and fixing #1466

palikar:
Fix for #1466.

#### TYPE

#### RUNTIME ENVIRONMENT
- Debian GNU/Linux busterr/sid
- Urxvt 9.22
- Python 3.6.6 
- ranger-master 1.9.2  
- Locale: en_US.UTF-8

    - [ ] Tests have been updated

#### CHECKLIST
- [x ] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [ x] All changes follow the code style **[REQUIRED]**
- [ x] All new and existing tests pass **[REQUIRED]**
    - [ ] Tests have been updated

#### DESCRIPTION
Several checks make sure that the needed parameters for the checked macros are actually there and are not None.

#### MOTIVATION AND CONTEXT
Without the changes in this PR, I am pretty sure that macros cannot be properly expanded while sourcing `rc.conf`. The problem in the original issue was that `%confdir` was need in `rc.conf` so that additional file in the configuration directory could be sourced. It's important to note that the bug comes into play while one needs maco WHILE sourcing the `rc.conf` and not later while calling some defined function (in form of a key binding)  in `rc.conf`.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -308,7 +308,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
         else:
             macros['f'] = MACRO_FAIL
 
-        if self.fm.thistab.get_selection:
+        if self.fm.thistab and self.fm.thistab.get_selection:
             macros['p'] = [os.path.join(self.fm.thisdir.path, fl.relative_path)
                            for fl in self.fm.thistab.get_selection()]
             macros['s'] = [fl.relative_path for fl in self.fm.thistab.get_selection()]
@@ -321,7 +321,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
         else:
             macros['c'] = MACRO_FAIL
 
-        if self.fm.thisdir.files:
+        if self.fm.thisdir and self.fm.thisdir.files:
             macros['t'] = [fl.relative_path for fl in self.fm.thisdir.files
                            if fl.realpath in self.fm.tags or []]
         else:
@@ -338,6 +338,9 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 tab = self.fm.tabs[i]
             except KeyError:
                 continue
+            if not tab:
+                continue
+
             tabdir = tab.thisdir
             if not tabdir:
                 continue
@@ -356,6 +359,13 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 macros[i + 'f'] = MACRO_FAIL
 
         # define D/F/P/S for the next tab
+
+        if not self.fm.tabs:
+            macros['D'] = MACRO_FAIL
+            macros['F'] = MACRO_FAIL
+            macros['S'] = MACRO_FAIL
+            return macros
+
         found_current_tab = False
         next_tab = None
         first_tab = None
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -245,12 +245,16 @@ class FM(Actions,  # pylint: disable=too-many-instance-attributes
         return ImageDisplayer()
 
     def _get_thisfile(self):
+        if not self.thistab:
+            return None
         return self.thistab.thisfile
 
     def _set_thisfile(self, obj):
         self.thistab.thisfile = obj
 
     def _get_thisdir(self):
+        if not self.thistab:
+            return None
         return self.thistab.thisdir
 
     def _set_thisdir(self, obj):


toonn:
Does this actually fix the problem? Seems like macro expansion in `rc.conf` would just fail silently now.

palikar:
@toonn The macros that are not supposed to work will of course not work. On startup there is no way of getting the currently selected file as such file simply does not exist. The same with the current dir or tabs or whatever. The things that will do work are %confdir, %datadir, %space and %rangerdir

toonn:
Then why was ranger accessing the `thistab` or `thisdir`? Was it trying to resolve all macros, rather than the ones needed?

palikar:
@toonn yes, exactly. For some reason, when substituting a given macro - in `substitute_macros` in `actions.py` - the whole macro list gets generated - through `get_macros` in `actions.py`. I guess I could make it a little bit smarter and generate only what is needed but I would require a little bit more rewriting. Are you interested in such bigger change or the, what is essentially a quick fix, changes in the current PR are ok for now?

Vifon:
@palikar If you can make it lazy, it'd be awesome. Would wrapping each macro in a lambda help?

palikar:
@Vifon Yes. I've manged to do the whole thing lazy by wrapping each macro in a lambda or a function. Some macros are pretty big and putting everything in a lambda is unpractical and unreadable. The whole dict of lambdas is generated only once and then the corresponding functions are being called only when needed.

ALSO, the change is backward compatible. If the thing in the macro dict is not something callable but a value, it won't get called and it will work just as before. This way we can define the "dynamic" macros (the ones that change based on file manager state) as functions and the "static" ones (the ones that do not change) as values. Everything works and it should be more efficient than before as we are not generating everything on every substitution. 

The one "drawback" now is that a regular expression is used to extract the macro-parts of the string, the macros of which should be expanded. The macro-parts are assumed to match the expression
```
<MACRO_DELIMITER>\w* 
```
which is the same as

```
<MACRO_DELIMITER>\[a-zA-Z0-9_] 
```
The macro-parts can thus be comprised only of letters, digits and underscores.

ranger/core/actions.py
```diff
@@ -49,6 +49,8 @@ class _MacroTemplate(string.Template):
 class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-methods
         FileManagerAware, SettingsAware):
 
+    _macros = None
+    _macro_reg = re.compile(ranger.MACRO_DELIMITER + r"\w*", re.IGNORECASE)
```
toonn: Why do we need this regex now?

examples/plugin_new_macro.py
```diff
@@ -17,7 +20,8 @@
 # Define a new macro function
 def get_macros_with_date(self):
     macros = GET_MACROS_OLD(self)
-    macros['date'] = time.strftime('%m/%d/%Y')
+    # date is just a value so we can just put it in a lambda
+    macros['date'] = lambda: time.strftime('%m/%d/%Y')
```
toonn: If it's just a value do we need the lambda? You said values don't have to be wrapped. Let's only introduce the laziness where we need it/it's useful.

ranger/core/actions.py
```diff
@@ -294,7 +307,12 @@ def substitute_macros(self, string,  # pylint: disable=redefined-outer-name
             raise ValueError("Could not apply macros to `%s'" % string)
         return result
 
-    def get_macros(self):  # pylint: disable=too-many-branches,too-many-statements
+    def get_macros(self):
```
toonn: What does the indirection to `_get_macros` buy us? Are you just memoizing the call, maybe a decorator is more appropriate?

ranger/core/actions.py
```diff
@@ -303,91 +321,128 @@ def get_macros(self):  # pylint: disable=too-many-branches,too-many-statements
...
+        macros['d'] = lambda: self.fm.thisdir.path if self.fm.thisdir else '.'
 
         # define d/f/p/s macros for each tab
+        tabdir = None
         for i in range(1, 10):
             try:
                 tab = self.fm.tabs[i]
             except KeyError:
                 continue
+            if not tab:
```
toonn: How can this occur exactly if there's no `KeyError`?

ranger/core/actions.py
```diff
@@ -303,91 +321,128 @@ def get_macros(self):  # pylint: disable=too-many-branches,too-many-statements
...
+            macros[i + 'd'] = lambda: tabdir.path
+
+            macros[i + 'p'] = lambda: ([os.path.join(tabdir.path, fl.relative_path)
+                                        for fl in tabdir.get_selection()]
+                                       if tabdir.get_selection() else MACRO_FAIL)
+
+            macros[i + 's'] = lambda: ([fl.path for fl in tabdir.get_selection()]
+                                       if tabdir.get_selection() else MACRO_FAIL)
+
+            macros[i
```
toonn: Weird split here, please split inside the parens instead.

ranger/core/fm.py
```diff
@@ -245,12 +245,16 @@ def _get_image_displayer(self):  # pylint: disable=too-many-return-statements
         return ImageDisplayer()
 
     def _get_thisfile(self):
+        if not self.thistab:
```
toonn: I'm not sure these should fail silently. Is this necessary?

ranger/core/actions.py
palikar: Because we want to extract all macros from the given string. For example, if the string to substitute macros in is "shell echo %confdir > %f", we would like to extract "%confdir" and "%f" and generate the macro expansions only for those.

examples/plugin_new_macro.py
palikar: You are right here. I had made this change before I figured out that we can support callables as well as values.

ranger/core/actions.py
toonn: Oh, because previously whatever the substitute needed would be present anyway? I think this'll lead to `KeyError`s. Maybe we should check whether each key is in the string? Though I'm not sure that's possible for the variable keys like `%5d`.

ranger/core/actions.py
palikar: Yes, but the function `get_macros`existed before that and was returning the dict of all macros. The function can be used in plugins so I assume it is used by people, i.e. it is used by code that we do not control. The macros dict is, however, meant to be generated only once - because of the whole lazyness that we want. Originally I had the memoization in the `substitute_macros` function but this leaves the problem that `get_macros` can be called from the outside world, thus generating the dict again. Putting the dict generation in `_get_macros` lets us keep the "generate dict only once" idea  even if the function `get_macro` is called from wherever.

ranger/core/actions.py
palikar: Yep. Probably of that one utility I use for automatic PEP8. Oh, well

ranger/core/actions.py
palikar: I am too not sure why I added this in the original commit. Macros seem to be ok without it, I just tested so yeah, I'll remove it.

ranger/core/fm.py
palikar: Well, right now the check is also being made in the corresponding macro lambdas so I can remove those. However, if some part of the application calls any of those function at startup time, before initialization, a error will be raised "NoneType does not have attribute thisfile". But you are probalby right. If this is happening something is most likely gone wrong or ranger is being used wrong.

ranger/core/actions.py
palikar: Yes, the previous macro system was generating the whole dict each time the macros of a string were being expanded. If there is a macro in the string that ranger does not define, it just won't get expanded (it was so and now it is also so). No `KeyError`s will be generated so I don't quite understand what you mean.

palikar:
@toonn , I've made a new commit where I fix one big problem that kinda flew over my head. You should be able to understand the issue where you see the code in the new commit.

ranger/core/actions.py
toonn: Oh, I'd missed the if test. I know the regex is more efficient but are we sure we don't want other characters in future macros, and does it catch *all* the edge cases? Going through all possible macros every time, checking if they're a substring is terrible too tbh especially because macros could be substrings of other macros : /

I think the most pythonic way might be to override the get methods of the dict, through subclassing for example. [As noted here](https://stackoverflow.com/q/3387691) we'd have several methods to override — off the top of my head there's `__getitem__, get, pop` and maybe `_process_args` — but it'd be the most transparent solution I can think off. No values are unnecessarily evaluated and no fragile regex to determine the macros we really need.

@Vifon, since this is quite some work, what's your input on this?

ranger/core/actions.py
toonn: I still don't understand this tbh. You're *just* memoizing `get_macros`, right? And the only place you call `_get_macros` is in `get_macros`? So I still think a memoizing decorator is the way to go, it's more explicit about what's going on *and* more reusable.

ranger/core/actions.py
```diff
@@ -303,91 +322,161 @@ def get_macros(self):  # pylint: disable=too-many-branches,too-many-statements
...
+                return MACRO_FAIL
+
+            if not tab.thisdir:
+                return MACRO_FAIL
+
+            tabdir = tab.thisdir
+            return (tabdir.pointed_obj.path
+                    if tabdir.pointed_obj else MACRO_FAIL)
+
+        for i in range(1, 10):
```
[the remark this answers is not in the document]
palikar: What do you mean exactly with `return lambda: whatever`. I Tried writing the macros as:
```
for i in range(1, 10):
            macros[str(i) + "d"] = lambda: get_tab_dir(i)
            ...
```
and this does not work. I think(I am not sure though) it will probably work if `get_tab_dir`  returns a function that takes an integer and return the dir of the corresponding tab. Is this what you are suggesting?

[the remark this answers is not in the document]
toonn: No, I'm suggesting `get_tab_dir` returns a lambda.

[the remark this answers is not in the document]
palikar: Ok. Would't that make the levels of function indirection bit too much too handle. The functions `get_tab_*' contain some multi line logic themselves so returning a labmda would require to create one more function in the corresponding function and return it. The dependency introduced by partial is with the functools package which is in the standard python library even for python 2.7, so is there actually a dependency? And the use of partial is pretty clear for anyone that knows what partial does so is there a problem there?

ranger/core/actions.py
palikar: Does this memoizing decorator exists or do we have to write it?

ranger/core/actions.py
palikar: If you would like to, I can make the regex a little bit more complicated and it will catch everything. For example
```
%[^(?!%\s)]*
```
matches "%" then everything that is not "%s" and whitespace. It essentially matches up to another "%" or space.

ranger/core/actions.py
toonn: We don't have one yet, so you may have to define one.

ranger/core/actions.py
toonn: The problem is I don't trust anyone, not even myself to come up with a regex that'll catch everything.

ranger/core/actions.py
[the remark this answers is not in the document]
toonn: I don't think one extra level's too much to handle. It's basically just manually doing what `partial` does. And I disagree it's obvious that `partial` on a single argument function returns an argumentless function, there's nothing partial about providing a function with all its arguments. Functools in general isn't all that pythonic, that's too bad but this is too obscure a use to justify it imo.

And since the bodies are multiline you can just return a nested function definition, we just need it to close over the argument.

palikar:
@tonttu Ok, I'll address all of your points here

- To the memoization thing - you were wrong. There is in fact a `cached_function* decorator defined in `ranger/ext/cached_function`. I've used it to make `get_macros` get called only once.

- To the partial thing - it seems that your problem is that `partial` (like the word "partial") implies that only some of the arguments are bound and if it is used to create a argumentless function, this would be confusing.  There is, however, no way that am going to do the work of `partial` for each function that I want to define and let the whole thing absolutely UNresuable, so I've written a function `bind_arguments` that does exactly the same thing as `partial` but it has different name. This function can now be used everywhere where a argumentless function has to be created by binding the arguments in advance.

- To the regex thing - look at the class `_MacroTemplate` in `actions.py`. There you will see that `ranger` *is already* using a regex to determine what kind of words can get expanded to macros. The regex there is essentially a `\w` so only things that contain letters, numbers and underscores can *currently* be macros. In this sense, my system does not impose further restrictions on what can be a macro and what cannot. If you still have doubts about my regex - which is currently `[^%\s]*` - you can test it [here](https://regex101.com/) with arbitrary text.

examples/plugin_new_macro.py
```diff
@@ -1,7 +1,7 @@
 # Compatible with ranger 1.6.0 through 1.7.*
 #
 # This plugin adds the new macro %date which is substituted with the current
-# date in commands that allow macros.  You can test it with the command.
+# date in commands that allow macros.  You can test it with the command:
```
toonn: This still reads weird. Either rephrase or don't change it.

ranger/core/actions.py
```diff
@@ -396,6 +391,7 @@ def get_tab_selection(i):
                     if tabdir.get_selection() else MACRO_FAIL)
 
         def get_tab_file(i):
+            self.notify(i)
```
toonn: Is this for debugging purposes?

ranger/ext/bind_arguments.py
```diff
@@ -0,0 +1,14 @@
+# This file is part of ranger, the console file manager.
+# License: GNU GPL version 3, see the file "AUTHORS" for details.
+
+from __future__ import (absolute_import, division, print_function)
+
+
+def bind_arguments(func, *part_args):
```
toonn: Imo this is seriously overengineered and a lot *less* clear than `partial`, just do the wrapping manually in the few functions that actually need it, way more obvious and understandable.

examples/plugin_new_macro.py
palikar: I haven't changed it with the exception of the colon at the end. I just added a few lines  below to clarify how macros can be functions or values. Should I remove those?

toonn: No, this one sentence is the problem. "The command" and the colon don't mix well. Either drop the colon or do something like "You can test it with: "

ranger/core/actions.py
palikar: Yep, just removed it.

ranger/ext/bind_arguments.py
palikar: I agree that the solution is impracticable in the sense that `partial` already exist and there is no point in rewriting it. With that being said, I simply cannot see how writing something manually several times is better than using a design construct that is there and exists specifically to free us from the need of doing something manually.  I guess my problem here is on the principle and I am sorry but.... no. I cannot purposely write a less maintainable code that ignores a solution that out there and with that forces any other future developer to write the same thing manually, if they wish to extend the macro dict with a macro on their own.

toonn: @Vifon, gonna let you decide on this. As you know I prefer FP. In this case however I don't think the unfamiliar style for the average python programmer is worth the effort saved over manually wrapping, what like 3 functions? I can make the changes if I have just need a decision.

@palikar, then the only bit that still needs your attention is the regex. We need to bring the two in line. Also though, to drive my point home about the brittleness of regexes. `[^%\s]*` doesn't work because it'll include parens in the macro for example. That's just a small oversight and an easy fix sure, but why not make it an impossible mistake to make? Though let's await @Vifon's opinion on that lazy dict thing too.

It's a little weird to me how you consider what `partial` does to be too menial to repeat manually but what the regex does (which is far more brittle and a common source of problems across languages even) to be perfectly fine duplicated in two places, in different ways that should be kept in sync with future changes even though there's no indication in either location that there even is something to be kept in sync.

palikar: Oh, no.  I mean, if both regexes can be kept as one\at one place, I think that it should definitely be done. Albeit I am not sure if this is possible without some hackery. The way that the regex is stored in `_MacroTemplate` is in two separate pieces - a delimiter and the a thing that should be matched after the delimiter (or at least this is how I think string.Template work). In the macros extraction thing we need to search with the regex as a whole. I suppose something of the sort is possible

``` python
MACRO_REG = re.compile(_MacroTemplate.delimeter + _MacroTemplate.idpattern, re.IGNORECASE)
``` 
Or even probably better would be to keep all parts in a separate places and the construct the class (`_MacroTemplate`) and compile the regex with only what is needed. This would probably be the cleanest solution, if you agree to use the regex approach in extracting the macros. As I implied, `_MacroTemplate` will substitute only certain things that match the current regex. My idea - let's just use the same regex to extract the things that will potentially be substituted.

Issue #1381, opened 2018-11-20
Document %any_path macro.


toonn:
Just noticed the `%any_pathX` macro isn't documented in the macros section of the manpage.

hut:
Fixed in #1496

PR #1496, opened 2019-03-18
Documenting the use of <any>, %any and %any_path

palikar:
I just added a few sentences in the man page that describe the use of the macrs %any and %any_path. I saw #1381 and though that is pretty easy to clear this up.


doc/ranger.pod
```diff
@@ -302,6 +302,20 @@ The macro %space expands to a space character. You can use it to add spaces to
 the end of a command when needed, while preventing editors to strip spaces off
 the end of the line automatically.
 
+The macros %any and %any_path can be used when binding keys to
+commands. The macro %any will exapnd to the key pressed in the place of
```
toonn: Typo "exapnd". Maybe rephrase something along the lines of "the placeholder `<any>` will be replaced with the key...". Same thing for the next sentence.

doc/ranger.pod
```diff
@@ -302,6 +302,20 @@ The macro %space expands to a space character. You can use it to add spaces to
 the end of a command when needed, while preventing editors to strip spaces off
 the end of the line automatically.
 
+The macros %any and %any_path can be used when binding keys to
```
toonn: Could you try summarizing better what it might be used for? Maybe "When mapping keys you can use the placeholder `<any>`, the key entered in that position can be used through the `%any` and `%any_path` macros...

doc/ranger.pod
```diff
@@ -302,6 +302,20 @@ The macro %space expands to a space character. You can use it to add spaces to
...
+commands. The macro %any will exapnd to the key pressed in the place of
+<any>. The macro %any_path will exapnd to the path of the bookmark which
+is mapped to the key that is pressed in the place of <any>. For example:
+
+	map c<any> echo %any
```
toonn: Examples are a great idea but could you use more useful examples? These do show how it works but not why you might want to use them.

palikar: Yes. There are actually a few examples in the default configuration file where %any and %any_path are used so I've used those to illustrate them.

palikar:
@toonn  , I just made the commit with my changes but I've noticed something that undermines half of the work. Several lines above there is actually an explanation of what `%any` (and %any0, %any1,...) does :D . Should I somehow merge the two explanations (mine and the old one) or just leave one of them?

toonn:
No idea why I didn't respond to this immediately, sorry for slipping up. I think you should merge them, probably in the location of the existing explanation, no reason to rejigger the order really.

toonn:
I've merged this in b5d8d064, not sure why github's not picking up the status change for this PR.

PR #1497, opened 2019-03-19
Make w3mimgdisplay image autoscale to fill preview pane #1

slackhead:
This makes images previewed via w3m scale up to fill the preview pane.

#### ISSUE TYPE

ranger version: ranger-master v1.9.2-180-g7d0901d9
Python version: 3.7.2 (default, Mar 18 2019, 09:01:12) [GCC 5.5.0]
Locale: en_GB.UTF-8
st, urxvt and framebuffer in/outside screen/tmux

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
When using w3m as image preview method, this will scale up images to fill the preview pane.

#### MOTIVATION AND CONTEXT
When browsing images in a directory it is useful to have them scale to fit the available area. It's possible to perhaps add zoom key maps and/or config options for preview/zoom behaviour too.

#### TESTING
Tested in 3 terminals + inside screen and tmux.
This affects calculating the image size for w3mimgdisplay.


toonn:
I'm ambivalent with regards to this. It was a deliberate choice when implementing previews. Scaling down images that are too big to fit is great. Scaling up small thumbnails and the like can make them look like shitty low-res pictures. Imo, when looking for a thumbnail a smaller image would make it stand out more. If you need to look at them closely you should open an image viewer anyway.

Since I'm fine with the status quo and wouldn't mind if this changed I'm asking for opinions. If this is something you want or really don't want, speak up!

slackhead:
My motivation for this is that a couple of days ago I downloaded all the images from the irssi themes page into a directory and found some of them were too small to get a good impression of them, even on a 20" monitor.

Perhaps it needs a threshold or some option for the scale mode then.  Most image viewers have something like that.  I'm quite happy to work on something like that.

Other than that, the offsets that I used for screen and tmux help to stop images overlapping the edges of the preview pane.

toonn:
Like I said, there's two valid choices here. Going through a bunch of theme images to see which you prefer sounds like a job for a proper image viewer tbh. That doesn't invalidate your preferred solution but we don't change ranger willy nilly based on the preferences of single users : )

That's why I'm asking for opinions. If people generally prefer upscaling small images then that's what we'll go with. If no one speaks up I'll assume no one except for you cares and accept the PR.

slackhead:
I think you're right about using a proper image viewer.  I only ever use ranger on the rare occasions when I want to preview a bunch of files in a directory, even though I've been maintaining it for 9 years or so.  It might be a better idea if I took the parts I need to make a kind of previewer fork and drop most of the command interface.  Perhaps add options for dealing with images or other viewed files.  I don't know of any terminal image viewers around.

rabite0:
Just randomly found myself in here, you may safely disregard my opinion. :)

I actually think this is a good idea. If it's so small that it looks like crap when upscaled it's probably so small that you can barely see what it is anyway, *especially* when you're just quickly browsing through the files while looking for the right image. It might look worse, but it's going to be easier to recognize.

Since ranger decidedly isn't an image viewer it doesn't have to look good, but if it makes recognition easier that's an improvement for its use case. IMHO, at least.

toonn:
I'm actually ok with accepting your patch since no one answered for nearly a month.

slackhead:
OK, thanks.  I may have a look at a setting perhaps for zooming or keeping original size for very small images.  Some kind of threshold perhaps.  But as you say, a dedicated file previewer application might be more suitable.

toonn:
@slackhead, if you fix up the flake issues, run `make test` to check locally, make sure you have the testing dependencies installed, pytest/flake8/pylint, we can get this merged.

toonn:
@slackhead, I'm gonna let this slip to next release.

Issue #1514, opened 2019-03-31
Assert use


frankm42:
There are several areas where the assert statement is used by ranger outside of tests
https://github.com/ranger/ranger/blob/1ca1255a3cf26eb646fca75b5388b2c645fdf466/ranger/container/settings.py#L144

Others:
https://github.com/ranger/ranger/search?q=assert&unscoped_q=assert&type=Code

Quote from https://hackernoon.com/10-common-security-gotchas-in-python-and-how-to-avoid-them-e19fbe265e03?gi=a611705ec561

> [...] by default Python executes with __debug__ as true, but in a production environment it’s common to run with optimizations. This will skip the assert statement [...]

Since assert is used by ranger mostly outside of tests, it is not garanteed these asserts will execute, especially when run by distributions that use optimization by default.
This is only an issue if assert is used outside of debugging. 

In this line https://github.com/ranger/ranger/blob/1ca1255a3cf26eb646fca75b5388b2c645fdf466/ranger/container/settings.py#L144
It looks more like ranger relies on the assert statement for filtering "Allowed Settings" in this line

This might imply ranger is affected by number 3 of these security issues 
 https://hackernoon.com/10-common-security-gotchas-in-python-and-how-to-avoid-them-e19fbe265e03?gi=a611705ec561 

I tried to find a way to exploit the number 1 issue (input injection) as well, but so far was not successful. Are there any intended measures to prevent the interpretation of filenames?
I haven't found any use of shlex.quote for secure input sanitation.
Are there any areas I should look at more precisely when trying to assess this issue in ranger?

toonn:
The assert does look suspicious. They're basically `is not None` and `is in` tests anyway, right? No clue why they're asserts.

We're generally careful (i.e. we think about what we do for a few minutes : )) but there's no explicit mitigations. So feel free to check out the codebase and fix any potential issues.

hut:
The reasoning behind that was to catch such errors when developers test the software through use, but not unnecessarily slow down the execution.

But I guess I'm the only one who uses ranger without optimization (it's on by default for ranger), and this programming pattern is prone to failure in general, so I recommend doing conventional checks with `if`s and proper error handling.

File paths are quoted where possible. We use ranger.ext.shell_escape for this.

frankm42:
I ran bandit on ranger to find all potential issues, here is the result
https://gist.github.com/frankm42/5efe257cf20156de08d202bc33aad589

Summary:

toonn:
[...]

PR #1530, opened 2019-04-13
Replace assert

frankm42:
This is a first attempt to replace the use of assert (outside of tests) in ranger.
Fixing https://github.com/ranger/ranger/issues/1514

Any comments/improvements?

### DIFF
--- a/ranger/container/settings.py
+++ b/ranger/container/settings.py
@@ -177,16 +177,16 @@ class Settings(SignalDispatcher, FileManagerAware):
     def set(self, name, value, path=None, tags=None):
         if name not in ALLOWED_SETTINGS:
             self.fm.notify("No such setting: {0}!".format(name), bad=True)
-            return
+            raise
         if name not in self._settings:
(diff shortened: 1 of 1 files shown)


toonn:
Hmm, shouldn't failing asserts result in exceptions rather than fairly silent failure?

frankm42:
Yes it should, however I am not quite sure about the best way to do this.
Maybe someone with actual experience in python can show the way.

frankm42:
This is not the correct fix.
If someone tells shows the way, I am willing to fix it right on all areas where ranger uses assert

toonn:
I think the only way is to figure out what errors will happen when the asserts fail and then catch and handle those in the correct way. Keeping in mind that some asserts may check boolean equations which would be replaced with if-statements.

toonn:
Next time feel free to keep pushing/force pushing to the same branch, no need to close the PR, any and all history attached to decisions can be valuable.

PR #1532, opened 2019-04-14
Execute preview in sandbox

valoq:
ISSUE TYPE

`Improvement/feature implementation`

Tested on Fedora, Arch, Debian Buster, Ubuntu 18.04
Feature is not enabled by default.

The second commit also increases the work time to avoid the throbber to always rote when sandboxing is enabled. I have not noticed any other changed due to this change, so it should be fine to have it generally increased.

Note: I had a disagreement with git, therefore the repeated pull request


toonn:
You can force push to a PR branch. It's a bit nicer than closing and reopening PRs ; )

valoq:
This feature is better implemented through something like [apparmor.d](https://github.com/roddhjav/apparmor.d)
Please see the apparmor profiles for preview tools in question.

PR #1540, opened 2019-04-21
File accessibility can change after `stat`ing

### LINKED ISSUE #1537: High CPU usage when selecting a not accessible directory
#### Runtime Environment

- Operating system and version: Linux Mint 19.1
- Terminal emulator and version: mate-terminal (1.20.0)
- Python version: 3.6.7 (default, Oct 22 2018, 11:32:17) [GCC 8.2.0]
- Ranger version/commit: ranger-master 1.9.2
- Locale: en_US.utf8

#### Steps to reproduce
1. `$ ./ranger.py -c /tmp`.
2. Create a directory from another terminal `$ sudo mkdir /tmp/test` 
and make it not accessible `$ sudo chmod 700 /tmp/test/`.
3. Select the directory `test` in ranger.

#### Possible Solutions
The following workaround works fine for local filesystem, but doesn't work for remote `sshfs`.
../ranger/container/fsobject.py:
```diff
-from os import lstat, stat
+from os import lstat, stat, access, R_OK

-        self.accessible = True if new_stat else False
+        self.accessible = True if access(path, R_OK) else False
```

toonn:
Catch errors due to inaccessibility of directories. We relied on
permissions not changing on directories between `stat`ing to determine
accessibility and first trying to list their contents.

Fixes #1537

### DIFF
--- a/ranger/container/directory.py
+++ b/ranger/container/directory.py
@@ -373,7 +373,6 @@ class Directory(  # pylint: disable=too-many-instance-attributes,too-many-public
                     filenames = filelist
                     if not filelist and eacces['occurred']:
                         self.accessible = False
-                        self.size__reset()  # pylint: disable=no-member
                         self.unload()
                         return
                     self.load_content_mtime = mtimelevel(mypath, self.flat)
@@ -384,7 +383,6 @@ class Directory(  # pylint: disable=too-many-instance-attributes,too-many-public
                         # PermissionError is undefined in python 2
                         if ex.errno == errno.EACCES:
                             self.accessible = False
-                            self.size__reset()  # pylint: disable=no-member
                             self.unload()
                             return
                         else:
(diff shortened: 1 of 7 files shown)


ranger/container/directory.py
```diff
@@ -350,7 +351,15 @@ def load_bit_by_bit(self):
...
-                    filelist = os.listdir(mypath)
+                    try:
+                        filelist = os.listdir(mypath)
+                    except (OSError, IOError) as ex:
+                        # PermissionError is undefined in python 2
+                        if ex.errno == errno.EACCES:
+                            self.accessible = False
+                        else:
+                            raise
+                        filelist = []
```
Vifon: Can we move this assignment into the if block? I feel it'd make it more clear of our intent. It's not being done regardless of the if result, it just gets ignored in the else branch due to the exception.

PR #1542, opened 2019-04-23
manpage: FLAGS are set in rc.conf, not rifle.conf.

tkunic:
#### ISSUE TYPE
- Bug fix, documentation.

#### RUNTIME ENVIRONMENT
Not relevant.

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
Manpage, FLAGS are set in rc.conf, not rifle.conf.

#### MOTIVATION AND CONTEXT
Typo, wrong file referenced in manpage.

#### TESTING
None.

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -133,7 +133,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.9.2" "2019-05-20" "ranger manual"
+.TH RANGER 1 "ranger-1.9.2" "2019-06-18" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
 .\" way too many mistakes in technical documents.
 .if n .ad l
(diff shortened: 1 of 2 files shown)


toonn:
You must be confusing this with something else. These flags are definitely set in `rifle.conf`, there's plenty of examples of the use of `flag f` in the default `rifle.conf`.

Thank you for taking the time to PR this though, there's probably plenty of cases where we *did* mess up.

tkunic:
Thanks for your reply, @toonn ! I was looking for "how to get `:shell` to show me the output of a shell command?" Then I went to the man page, saw the FLAGS section, and:

```
       There are some additional flags that can currently be used only in the "shell" command: (for example ":shell -w df")

        p   Redirect output to the pager
        s   Silent mode.  Output will be discarded.
        w   Wait for an Enter-press when the process is done

       By default, all the flags are off unless specified otherwise in the rifle.conf configuration file.
```

And I thought the appropriate place to set defaults for the `:shell` command key binding is in `rc.conf`?

```
map !  console shell -w%space
```

I guess the paragraph above that does reference flags for rifle, so my edit was incorrect, but as it stands now, I think the part of the man page referencing default flags for `:shell` could be clarified?

toonn:
Oh, thank you for pointing this out, you're right, what fresh eyes won't do : ) Let's move the bit on flags for `:shell` to the bottom of that section and mention `rc.conf` there.

Maybe the scout flags should also be mentioned?

hut:
How about

> By default, all the flags are off, although you can set defaults for ":open_with" and "\<enter>"-presses in the rifle.conf configuration file.  One way to set the default flags for ":shell" would be to override the "!" key binding in rc.conf.

toonn:
I took your input, made the simplest rewording I could that included both confs.

PR #1543, opened 2019-04-25
Change the = map

toonn:
`=` (`:chmod` without an argument) made my ranger crash. The new mapping
opens the console with `:chmod 000<cursor>` this way it's easy to change
what permissions sets. It defaults to 000 because = in a chmod(1)
argument sets the mentioned bits and unsets all others, so `chmod =`
would unset all bits afaiu.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -461,7 +461,7 @@ map yn yank name
 map y. yank name_without_extension
 
 # Filesystem Operations
-map =  chmod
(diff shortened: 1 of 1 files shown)


toonn:
`=` is apparently meant to be used with a quantifier `777=` for example. I'll close this because it doesn't dramatically differ from the usage in this PR and people are probably used to it by now. We do need to mention that more explicitly somewhere though. And prevent `=` from crashing ranger.

PR #1550, opened 2019-05-07
Add installation using Homebrew

Aracki:
Add an example of how the package manager of your operating system can help you install `ranger`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,10 @@ You can also install ranger through PyPI: ```pip install ranger-fm```.
   </a>
 </details>
 
+### Installing using Homebrew
+You can use [Homebrew](brew.sh) to install ranger.
+`brew install ranger`
+
 ### Installing from a clone
 Note that you don't *have* to install ranger; you can simply run `ranger.py`.
 


toonn:
We can only accept a PR that adds every package manager under the sun, we don't want to play favorites. There's apt, yum, pacman (mention ranger-git in the AUR), nix and probably others.

If you add all of those we can accept the PR. If you can't/don't want to add all of them, feel free to add a page to the wiki for installation methods and mention just the homebrew method. Good luck : )

Aracki:
I don't see you have an "installation section" in wiki pages.

About PR, I think Homebrew is not "played" as a favorite. It's just for that osx users can faster find out how to install `ranger`. Every tool that can be installed with brew should contain that command in README.

toonn:
We don't mention linux specific package manager instructions either. This is on purpose btw, ranger's at tool that attracts people who are generally fairly comfortable on the command line so there's not much need for package manager instructions. Another reason is there's so many package managers you'd have to write instructions for, just on OSX there's 3, linux has endless package managers.

You could create the page on the wiki.

hut:
Seems a bit too obvious to me... I'd rather not add a list of 100 instructions like "apt-get install ranger" and "pacman -S ranger".

PR #1557, opened 2019-05-15
Scope no longer interferes with users' configuration

### LINKED ISSUE #1556: improve highlight flexibility in scope.sh
Currenlty, file `scope.sh` hard-codes all its colorization configurations, including the options being sent to the external `highlight` program for colorization. That program, `highlight`, offers a zillion [colorization styles](http://www.andre-simon.de/doku/highlight/en/theme-samples.php) and makes available an environment variable `HIGHLIGHT_OPTIONS` so that the style can be set without needing to copy and modify `scope.sh` just for that. Also, because `scope.sh` explicitly calls `highlight` with a hard-coded style, any user-selected style in that environment variable `HIGHLIGHT_OPTIONS` is ignored.

My recommendation is to delete the assignment of the highlight_style (currently line 39), modify the call to `highlight` (currently line 232) to not pass a style argument, and possibly add to the documentation that a particular highlight style can be set using the environment variable. For people using highlight in other contexts, this gives them allows them to uniformly change the setting across all uses of `highlight` just as the developers intended.

Also, I see that the project made allowance to choose `pygmentize` instead of `highlight`, but commented out the actual invocation. If you want to offer the choice, set a variable COLORIZER near line 39, and perform a case/esac on it around line 232. Better would be to have a config option in file `rc.conf` to set that variable.

If you want, I could make this a pull request, but it's quite trivial so your option.

toonn:
Our setup for syntax highlighting in `scope.sh` was actively preventing
people's configuration for `highlight` from working. Now
`HIGHLIGHT_OPTIONS` takes precendence over the default and over the
`HIGHLIGHT_STYLE` environment variable.

Users can specify `HIGHLIGHT_STYLE` or `PYGMENTIZE_STYLE` enviroment
variables now to customize the syntax coloring theme without having to
edit *and keep up to date* `scope.sh`.

Fixes #1556

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -36,8 +36,9 @@ FILE_EXTENSION_LOWER="$(printf "%s" "${FILE_EXTENSION}" | tr '[:upper:]' '[:lowe
 # Settings
 HIGHLIGHT_SIZE_MAX=262143  # 256KiB
 HIGHLIGHT_TABWIDTH=8
-HIGHLIGHT_STYLE='pablo'
-PYGMENTIZE_STYLE='autumn'
+HIGHLIGHT_STYLE=${HIGHLIGHT_STYLE:-pablo}
+HIGHLIGHT_OPTIONS="--style=${HIGHLIGHT_STYLE} ${HIGHLIGHT_OPTIONS:-}"
+PYGMENTIZE_STYLE=${PYGMENTIZE_STYLE:-autumn}
 
 
 handle_extension() {
@@ -228,9 +229,11 @@ handle_mime() {
                 local pygmentize_format='terminal'
                 local highlight_format='ansi'
             fi
-            highlight --replace-tabs="${HIGHLIGHT_TABWIDTH}" --out-format="${highlight_format}" \
-                --style="${HIGHLIGHT_STYLE}" --force -- "${FILE_PATH}" && exit 5
-            # pygmentize -f "${pygmentize_format}" -O "style=${PYGMENTIZE_STYLE}" -- "${FILE_PATH}" && exit 5
+            env HIGHLIGHT_OPTIONS="${HIGHLIGHT_OPTIONS}" highlight --replace-tabs="${HIGHLIGHT_TABWIDTH}" \
+                --out-format="${highlight_format}" \
+                --force -- "${FILE_PATH}" && exit 5
+            pygmentize -f "${pygmentize_format}" -O "style=${PYGMENTIZE_STYLE}"\
+                -- "${FILE_PATH}" && exit 5
             exit 2;;
 
         # DjVu


Boruch-Baum:
I like the fix; and it's superior to my original suggestion in #1556.

Missing is documentation, so that users become aware of the feature. In `doc/ranger.1`, line ~1648, you could add something like 
```
..IP "HIGHLIGHT_OPTIONS" 8
.IX Item "HIGHLIGHT_OPTIONS"
When the external program "highlight" exists, it is used to colorize output. You can use this environment variable of that program to customize how that's done, for example "--style=foo". See "highlight --list-scripts=themes."
```

You may also want to mention that the code still is hard-coding three options that can't be over-ridden by a user using the environment variable method, ie. `--replace-tabs=`, `--out-format`, and `--force`.

While I was viewing that file, I also noticed a nitpick to suggest to sneak in. The description of variable `W3MIMGDISPLAY_PATH` should explicitly mention that it is "owned" or is a feature of `w3m`.

There may be other related issues that I haven't noticed. For example, `pygmentize` and all the other external programs used by `ranger` may have the same situation as `highlight`.

Also, on line ~238 of the same `ranger.1` mentioned above, mention of `pygmentize` is omitted.

Performing `apt-cache show ranger` from a debian command-line, I see that there is a mis-match between the suggested/recommended/depends and the list of external tools mentioned in the man page (same location in source, line ~238; in the groffed output it's ine ~120 ymmv).

I manually made the corrections of the commit and based upon that, ie. without `git cloning` etc., your commit seems to work; however, I was expecting to be able to change the values on the fly from within ranger by using ranger's shell command `! export HIGHLIGHT_OPTIONS="--style=foo"`. That doesn't seem to be working.

Boruch-Baum:
The current interplay of components means that with the default `scope.sh` it is impossible for a user to use `pygmentize` once `highlight` is also installed. This is more likely to happen on a multi-user system than on a single-user one, say a school where different teachers might make different wacko OCD demands on their students (not to be too specific...). Maybe have a `set` option in `rc.conf` to define the colorizer program?

toonn:
Re the documentation changes, you're right thanks for pointing it out!
I didn't find a `HIGHLIGHT_OPTIONS` equivalent for pygmentize so I figured I'd leave it mostly as is. (Though I *think* the style var can be used to inject arguments to the command as it is written rn. Though this may change in the future.)

When you use ranger's `:shell` or any of `S`, `!`, `#` or `s` you're getting a subshell. Ranger can't execute processes in its parent's context. No easy way around this. So the export *is* probably technically working, a nested ranger would pick it up. You could try `:eval import os; os.environ["HIGHLIGHT_OPTIONS"]="--style=autumn".

Re the options that are still hardcoded. If you could check whether they can be overridden by repetition (like `--style`) or an inverse flag (thinking something like `--noforce`) I could basically do the same change for those.

Re debian packaging you'll have to contact the package maintainers, we only maintain the PyPI package.

Re interplay, that's correct. Main reason is to be backwards compatible. People who are used to `highlight` still need to get `highlight` previews but also it's a less popular tool than pygementize afaik and the "rarer" tools get precedence. There's two ways to work around this. You could install `ranger-fm` and `pygments` in a virtualenv, I think binaries installed in the venv take precedence over your `PATH`. The other way is just altering `scope.sh`, we consider it a configuration file. Disadvantage is only you won't get changes we make when you update so you should periodically check whether there's any important changes in `scope.sh` (most people just don't until it stops working : ))

toonn:
#1559 is ready for review. I didn't add to `W3MIMGDISPLAY_PATH` documentation because we don't mention for any of the variables whether they're specific to ranger. Also, I'm not sure that's important information, only time it might matter is if people want a certain setting to only apply to ranger, not general w3m use?

Boruch-Baum:
We've generated many threads within this issue, so here's a stab at replying:

1) The hard-coded `highlight` option `--replace-tabs` can be over-ridden by subsequent invocations, so it is ripe for being handled as you did for style. On the other hand, the hard-coded options `--force` and `-out-format` seem sensible to stay hard-coded.

2) The debian source tarballs don't include the problem code that I made issue of. It must have been on my system as cruft or some relic of a previous version.

3) Environment variable `W3MIMGDISPLAY_PATH` doesn't appear in my copy of the `w3m` man page and doesn't return anything from a `grep` on its documentation directory! OTOH, I am able to display images, and seem to be using the `w3m` method for it.

4) I can't speak to the relative popularity of `pygmentize` over `highlight`, but my memory is that  when I evaluated `pygmentize` (a few years ago?), it was agonizingly slow to load and display. That may be due to python being slow to load, which BTW is also the case with `ranger`; however `ranger` only needs to be loaded once per session, and `pygmentize` needed to be loaded for each file view.

5) Re: #1559, it's asking for trouble to explicitly list the current styles bundled with `highlight`, because they can change at any time, and a user or distribution or site can also add, subtract or modify the list. Better, simpler, and shorter to just refer the user to `highlight --list-scripts=themes`. That way, no one in the future will ask you to modify the list.

toonn:
1. Will do.
2. ✔️
3. Yes, it *is* specific to ranger, was confused by your earlier comment. Still a non-issue imo. It's just a way to change the path ranger uses to invoke w3m. `w3mimgdisplay` is really an internal tool of the `w3m` browser so it's not always made available on the `PATH`.
4. ✔️ 
5. Was only trying to be helpful but you're right people can run their own commands.

toonn:
Could you check #1559? Mostly it'd be great if you could test the tabwidth change.

Boruch-Baum:
Done, and it seems to work fine. I specifically tested the tab width, which for me (a two-space-tab person) was a rare joy to behold. The only caveat I should mention is that I performed the test by manually patching my local scope.sh, but I was careful and the patch isn't a major deal. Thanks for the great work! My plan is to slowly and incrementally ramp up he ranger learning curve and consider it as a replacement for midnight commander, so you may hear back from me in spurts of activity punctuated by long silences.

As a side note, it was a while before I learned that ranger supported both multipane and what you call miller view modes. You might get a spike in user interest if you made that more obvious. I think most people associate ranger with exclusively the miller view mode, and on that basis they drop ranger from consideration as a command-line tool.

toonn:
I guess we could include a screenshot of the multipane mode in the readme. And mention it more explicitly.

Why would that disqualify ranger as a cli tool? It's obviously the best mode to use in a terminal, otherwise it wouldn't be the default ; p

Boruch-Baum:
> I guess we could include a screenshot of the multipane mode in the readme. And mention it more explicitly.

I support both ideas.

> Why would that disqualify ranger as a cli tool?

That's a harsh way to characterize what I said, but yes, since the subject is other peoples' opinions, many are looking for a multi-pane workflow and presentation. That's what they're used to, and that's what they want. BTW, even in that regard, ranger is superior to midnight commander in the sense of enabling a three-panel side-by-side view, a feature not available in midnight commander, and which is nice for comparing data among three machines. Also, the ranger scripting macros are much more powerful, eg. the ability to have say five tabs for five devices and to do something like `tar -czf here.tgz %2t %3t %5t`.

toonn:
Hmm, seems like maybe it'd be great if you wrote a "compared to mc" page for our wiki : )

Boruch-Baum:
I know very little about `ranger`, have barely used even a small subset of its features, and couldn't at this point do it justice; I just don't know enough about it.

toonn:
Our wiki is for community contribution, no matter how small. It's ok if articles are works in progress. That single paragraph you wrote could already be a valuable piece of knowledge for someone familiar with mc and considering ranger.

PR #1574, opened 2019-05-29
Added the ability to escape ';' delimiters used by the 'chain' command with a '\'.

jakem72360:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
- Terminal emulator and version: urxvt v9.22
- Python version: 3.7.3 (default, Mar 26 2019, 21:43:19) [GCC 8.2.1 20181127]
- Locale: en_AU.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
The proposed patch allows the chain command's semi-colon delimiters to be escape with a backslash.

#### MOTIVATION AND CONTEXT
The proposed change addresses a important design flaw, in which a large number of commands cannot be given as arguments to a chain command. Examples include; renaming files with semi-colons, shell commands including semi-colons, etc.

#### TESTING
I attempted to run `make test` on my current machine, but it fails with the runtime exception `StopIteration`. Not sure if it's associated with #1244, but I wasn't able to successfully run the tests even with the current master branch and no changes. Whilst I couldn't run the automated tests, I did test the `chain` command with numerous possible inputs. The change is so small, doesn't affect the rest of the codebase, and is unlikely to fail the automated tests.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -283,11 +283,30 @@ class chain(Command):
     Calls multiple commands at once, separated by semicolons.
     """
 
+    def _split(self, text):
+        result = []
+        token = ''
+        state = 0
+        for c in text:
+            if state == 0:
+                if c == '\\':
+                    state = 1
+                elif c == ';':
+                    result.append(token)
+                    token = ''
+                else:
+                    token += c
+            elif state == 1:
+                token += c
+                state = 0
+        result.append(token)
+        return result
+
     def execute(self):
         if not self.rest(1).strip():
             self.fm.notify('Syntax: chain <command1>; <command2>; ...', bad=True)
             return
-        for command in [s.strip() for s in self.rest(1).split(";")]:
+        for command in [s.strip() for s in self._split(self.rest(1))]:
             self.fm.execute_console(command)
 
 


toonn:
Have to say I'm not a fan of the manual implementation of escaping. I can usually work around this limitation by putting the thing with semicolons second, only the first semicolon is parsed as a separator, right?

We usually argue that you're better off implementing a custom command once you run into issues like this.

jakem72360:
@toonn By manual implementation, are you referring to the `_split()` function inside the class? Python isn't my area of expertise, but I couldn't find a function in any of the common libraries that does the same job (feel free to correct me if I'm wrong). Unless there's a better way, I wouldn't say that single function alone is the end of the world in terms of LoC.

Problem is, that the current implementation of the `chain` command is fundamentally broken. It splits the string into a list, so no, the the first semicolon isn't the only one parsed. Furthermore, putting command containing a semicolon afterwards, is not a valid solution, as many use-cases require commands to be in a specific sequence, unless you're suggesting padding the chain with useless echo commands, which is also not a valid solution.

For the time being, I've re-implemented `chain` as a custom command on my own setup, but this is still a bug and needs to be resolved.

toonn:
Something like shlex might be suitable. If we have to do it manually we'd want the parsing for commands in general to be improved or even replaced with a real parser implementation.

I disagree `chain`'s fundamentally or otherwise broken. It doesn't do what you expect but I don't think I want it to do what you expect. I can already imagine the endless amount of bug reports along the line of "chain's broken" because people mess up shell-like escaping.
At least now we can say (just like the documentation says) "Chain *doesn't* deal with semicolons."

To be clear, this is the sort of confusion I'm talking about: `:chain shell some sequence that needs\; for some reason and 'filewith\;.txt'; echo 'o.O'` It's weird to have to escape in regular shell commands and double escaping is more often than not very confusing.

Maybe we can pull a vim trick out of our hat? Make the seperator an argument, i.e., `chain:command 1;command 1 continued:command 2:...` but also `chain/command1;command 1 cont./command 2`. And then we'd have space default to the current semicolon behavior to maintain backwards compatibility. The only sequence that'd still be impossible is one that lists literally every character that could be a seperator which isn't a big deal, imo. Then we can add backslash/shellquotes escaping, I'd be fine with that if there's a simpler alternative.

jakem72360:
I'm not sure if I'll have the time to get around to it in the near future, but we'll see what happens. Replicating vim/sed-like behaviour and allowing arbitrary delimiters is an interesting idea, but may require modification to the command parser itself (I don't know yet, I haven't checked the parser). The other option, could also be allowing the delimiter to be specified as a flag to chain, though that'd require complex parsing by chain to determine the end of flags and the beginning of commands.

>the endless bug reports
I don't see the potential for confusion. The current fix, as it stands, can easily be implemented without breaking compatibility with existing setups and configs. An un-escaped semi-colon behaves exactly the same way it did before.

PR #1576, opened 2019-05-31
Add preview support for .rtf, .doc, .docx, .xlsx, .xls in scope.sh

prosoitos:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: 5.1.5-arch1-2-ARCH
- Terminal emulator and version: rxvt-unicode 9.22-7
- Python version: 3.7.3-1
- Ranger version/commit: 1.9.2-1
- Locale: 
LANG=en_CA.UTF-8
LC_CTYPE=en_CA.UTF-8
LC_NUMERIC="en_CA.UTF-8"
LC_TIME=en_DK.UTF-8
LC_COLLATE=C
LC_MONETARY="en_CA.UTF-8"
LC_MESSAGES="en_CA.UTF-8"
LC_PAPER="en_CA.UTF-8"
LC_NAME="en_CA.UTF-8"
LC_ADDRESS="en_CA.UTF-8"
LC_TELEPHONE="en_CA.UTF-8"
LC_MEASUREMENT="en_CA.UTF-8"
LC_IDENTIFICATION="en_CA.UTF-8"
LC_ALL=

#### CHECKLIST

I don't know how to preform tests. I am not a software developer :slightly_frowning_face: 
    - [ ] Tests have been updated

#### DESCRIPTION

The title says it all.

#### MOTIVATION AND CONTEXT

Not required. Add functionality.

Office GUI take a long time to open. Having access to a preview of the file without having to open it in an external program can be a huge time saver.

n/a

#### TESTING

None. I simply made sure it is running on my system.

It shouldn't affect anything.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -83,6 +83,7 @@ handle_extension() {
         odt|ods|odp|sxw)
             ## Preview as text conversion
             odt2txt "${FILE_PATH}" && exit 5
+	    # Preview as markdown conversion
 	    pandoc -s -t markdown "${FILE_PATH}" && exit 5
             exit 1;;
 
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -87,7 +87,7 @@ handle_extension() {
 
 	## RTF and DOC
 	rtf|doc)
-	    ## Preview as markdown conversion
+	    ## Preview as text conversion
 	    ## note: catdoc does not always work for .doc files
 	    catdoc "${FILE_PATH}" && exit 5
 	    exit 1;;
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -85,6 +85,31 @@ handle_extension() {
             odt2txt "${FILE_PATH}" && exit 5
             exit 1;;
 
+	## RTF and DOC
+	rtf|doc)
+	    ## Preview as markdown conversion
+	    ## note: catdoc does not always work for .doc files
+	    catdoc "${FILE_PATH}" && exit 5
+	    exit 1;;
+
+	## DOCX
+	docx)
+	    ## Preview as markdown conversion
+	    pandoc -t markdown "${FILE_PATH}" && exit 5
+	    exit 1;;
+
+	## XLSX
+	xlsx)
+	    ## Preview as csv conversion
+	    xlsx2csv "${FILE_PATH}" && exit 5
+	    exit 1;;
+
+	## XLS
+	xls)
+	    ## Preview as csv conversion
+	    xls2csv "${FILE_PATH}" && exit 5
+	    exit 1;;
+
         ## HTML
         htm|html|xhtml)
             ## Preview as text conversion
@@ -303,7 +328,6 @@ handle_fallback() {
     echo '----- File Type Classification -----' && file --dereference --brief -- "${FILE_PATH}" && exit 5
     exit 1
 }
-
 
 MIMETYPE="$( file --dereference --brief --mime-type -- "${FILE_PATH}" )"
 if [[ "${PV_IMAGE_ENABLED}" == 'True' ]]; then
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -138,7 +138,6 @@ handle_extension() {
         json)
             jq --color-output . "${FILE_PATH}" && exit 5
             python -m json.tool -- "${FILE_PATH}" && exit 5
-	    pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
             ;;
 
         ## Direct Stream Digital/Transfer (DSDIFF) and wavpack aren't detected
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -89,6 +89,7 @@ handle_extension() {
 	rtf|doc)
 	    ## Preview as text conversion
 	    ## note: catdoc does not always work for .doc files
+            ## catdoc: http://www.wagner.pp.ru/~vitus/software/catdoc/
 	    catdoc "${FILE_PATH}" && exit 5
 	    exit 1;;
 
@@ -101,12 +102,15 @@ handle_extension() {
 	## XLSX
 	xlsx)
 	    ## Preview as csv conversion
+	    ## Uses: https://github.com/dilshod/xlsx2csv
 	    xlsx2csv "${FILE_PATH}" && exit 5
 	    exit 1;;
 
 	## XLS
 	xls)
 	    ## Preview as csv conversion
+	    ## xls2csv comes with catdoc:
+            ##   http://www.wagner.pp.ru/~vitus/software/catdoc/
 	    xls2csv "${FILE_PATH}" && exit 5
 	    exit 1;;
 
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -96,7 +96,7 @@ handle_extension() {
 	## DOCX
 	docx)
 	    ## Preview as markdown conversion
-	    pandoc -t markdown "${FILE_PATH}" && exit 5
+	    pandoc -s -t markdown "${FILE_PATH}" && exit 5
 	    exit 1;;
 
 	## XLSX
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -47,7 +47,6 @@ PYGMENTIZE_STYLE=${PYGMENTIZE_STYLE:-autumn}
 OPENSCAD_IMGSIZE=${RNGR_OPENSCAD_IMGSIZE:-1000,1000}
 OPENSCAD_COLORSCHEME=${RNGR_OPENSCAD_COLORSCHEME:-Tomorrow Night}
 
-
 handle_extension() {
     case "${FILE_EXTENSION_LOWER}" in
         ## Archive
@@ -84,46 +83,9 @@ handle_extension() {
         odt|ods|odp|sxw)
             ## Preview as text conversion
             odt2txt "${FILE_PATH}" && exit 5
-	    # Preview as markdown conversion
-	    pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
+            ## Preview as markdown conversion
+            pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
             exit 1;;
-
-	## ODT (using pandoc)
-        # odt)
-        #     # Preview as markdown conversion
-        #     pandoc -s -t markdown "${FILE_PATH}" && exit 5
-        #     exit 1;;
-
-	## RTF and DOC
-	rtf|doc)
-	    ## Preview as text conversion
-	    ## note: catdoc does not always work for .doc files
-            ## catdoc: http://www.wagner.pp.ru/~vitus/software/catdoc/
-	    catdoc -- "${FILE_PATH}" && exit 5
-	    exit 1;;
-
-	## DOCX, ePub, FB2 (using markdown)
-	## You might want to remove "|epub" and/or "|fb2" below if you have
-        ## uncommented other methods to preview those formats
-	docx|epub|fb2)
-	    ## Preview as markdown conversion
-	    pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
-	    exit 1;;
-
-	## XLSX
-	xlsx)
-	    ## Preview as csv conversion
-	    ## Uses: https://github.com/dilshod/xlsx2csv
-	    xlsx2csv -- "${FILE_PATH}" && exit 5
-	    exit 1;;
-
-	## XLS
-	xls)
-	    ## Preview as csv conversion
-	    ## xls2csv comes with catdoc:
-            ##   http://www.wagner.pp.ru/~vitus/software/catdoc/
-	    xls2csv -- "${FILE_PATH}" && exit 5
-	    exit 1;;
 
         ## HTML
         htm|html|xhtml)
@@ -131,8 +93,8 @@ handle_extension() {
             w3m -dump "${FILE_PATH}" && exit 5
             lynx -dump -- "${FILE_PATH}" && exit 5
             elinks -dump "${FILE_PATH}" && exit 5
-	    pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
-	    ;;
+            pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
+            ;;
 
         ## JSON
         json)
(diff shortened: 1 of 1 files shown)


ranger/data/scope.sh
```diff
@@ -77,13 +76,39 @@ handle_extension() {
...
+	    exit 1;;
+
+	# DOCX
+	docx)
+	    # Preview as markdown conversion
+	    pandoc -t markdown "${FILE_PATH}" && exit 5
+	    exit 1;;
+
+	# XLSX
+	xlsx)
```
toonn: Why not xlsx|xls?

ranger/data/scope.sh
```diff
@@ -77,13 +76,39 @@ handle_extension() {
...
 
+	# RTF and DOC
+	rtf|doc)
+	    # Preview as text conversion
+	    # note: catdoc does not always work for .doc files
+	    catdoc "${FILE_PATH}" && exit 5
+	    exit 1;;
+
+	# DOCX
+	docx)
```
toonn: Can't pandoc handle rtf and doc? Maybe we can add pandoc after catdoc? Check the pdf handling with mutool and pdftotext to see how we implement a fallback.

ranger/data/scope.sh
prosoitos: Because they use different programs (xls2csv and xlsx2csv). The 2 types of files are very different.

ranger/data/scope.sh
prosoitos: No, it cannot. Here again, despite the similar names, they are very different types files.

prosoitos: > Check the pdf handling with mutool and pdftotext to see how we implement a fallback.

pdftotext does not handle .docx files. I don't have mutool, but according to its man:

`The supported input document formats are: pdf, xps, cbz, and epub.`

prosoitos:
> What's the performance impact though? If there's noticeable delay when scrolling through files in ranger we comment the section out by default.

I do not notice any difference while scrolling. That said, I don't know how to quantify it through some test or benchmark.

I can comment it out if you prefer.

ranger/data/scope.sh
prosoitos: I just added some info on how to get those applications. It took me some time to look for and test applications which work for those conversions.

ranger/data/scope.sh
prosoitos: All the applications I mention are quick and work reliably, except for catdoc with .doc. It should work and does work for some files, but not for others. It may be because I am testing it on .doc files which I create with LibreOffice rather than MS Word. But that is not a big deal since .doc and .xls are archaic formats that very few people use now that they have been replaced with their "x" version years ago.

prosoitos: All these MS formats are a total nightmare anyway :laughing: and I doubt ranger users love them. But unfortunately we sometimes have to use them when working with the rest of the world :expressionless:

toonn: You're right, pandoc only handles rtf as an output format. Maybe we should add all the input formats pandoc accepts to the case though? Not really a reason not to support all the formats the tool supports I think.

You misunderstand what I said about the pdf previews. Those tools only do pdf to plaintext conversion. What you should have a look at is how both those programs are called from the same case. This is useful when there's multiple programs that could handle a single format, then you can add one that's unlikely to be installed at the top and if it's not installed scope'll fall back to the next one.

ranger/data/scope.sh
toonn: You're right, I hadn't noticed the extra `x`.

Btw, does scope already handle .odt and the other openoffice/libreoffice formats?

ranger/data/scope.sh
prosoitos: > You're right, pandoc only handles rtf as an output format.

Well, the problem here is that pandoc does not accept .doc nor .rtf as an input format. It does however accept .docx

prosoitos: > You misunderstand what I said about the pdf previews. Those tools only do pdf to plaintext conversion. What you should have a look at is how both those programs are called from the same case. This is useful when there's multiple programs that could handle a single format, then you can add one that's unlikely to be installed at the top and if it's not installed scope'll fall back to the next one.

I think that we are both misunderstanding each other: there is no fallback for .rtf and .doc files.

There are other tools, besides catdoc, which can handle those files (unoconv for instance), but as far as I know, they cannot output to the terminal: they can only create a new file. So they did not seem suitable for the preview.

prosoitos: Having a temp text file created by unoconv and then pipe it to get the preview seemed heavy and twisted. Particularly since catdoc works flawlessly for .rtf files and .doc files, for which it is a bit hit or miss, is a disappearing format.

prosoitos: > Maybe we should add all the input formats pandoc accepts to the case though? Not really a reason not to support all the formats the tool supports I think.

I'll have a look at that.

toonn: I was thinking of having pandoc be a fallback for all the formats it supports.

prosoitos: One of the formats that pandoc accepts is .odt and there is already a tool (odt2txt) in scope.sh (and it is not commented).

If I add .odt to the pandoc paragraph, I don't know what would happen if both are uncommented. Will this create a conflict? If so, then I could comment the pandoc paragraph, but it is a bummer for all the other formats.

prosoitos: Of course, the odt2txt paragraph could be commented instead, but pandoc is a much heavier program and having it as the default for .odt files may not be the right decision? People can of course customize their scope.sh.

Or maybe it doesn't matter if both paragraphs are uncommented??? I just wonder what would happen if someone has both programs.

prosoitos: A lot of pandoc input formats are markup languages. So those would be irrelevant (no point converting from one markdown format to another for instance).

And some of the formats are already covered elsewhere in scope.sh (odt2txt I mentioned earlier, but also .html, .json, etc.).

And finally I am totally unfamiliar with some of the formats and I don't know whether they are text (so irrelevant here) or not.

So, I am not comfortable to add more to pandoc. If someone wants to do it of course, feel free to jump in.

toonn: Things like this is exactly what I described the fallback for, look at mutool/pdftotext to see how you can handle this. (Newer bash syntax has support for matching several clauses but let's not stray *further* from posix sh.)

ranger/data/scope.sh
prosoitos: It does accept odt|ods|odp|sxw

But none of the formats that I added.

ranger/data/scope.sh
toonn: Relevant formats I'd say are docx, epub, fb2 and odt. You're right that most of the markup languages are reasonably readable and people may not want to have them be rendered.

prosoitos: > Things like this is exactly what I described the fallback for, look at mutool/pdftotext to see how you can handle this. (Newer bash syntax has support for matching several clauses but let's not stray further from posix sh.)

I understand. I just wasn't sure I had to add a pandoc line for all of those already covered formats (since pandoc overlaps with several tools).

toonn: Yes, that's not very satisfying. We *could* merge all those cases and use the fallthroughs, it'd probably work but that seems worse than the duplication tbh. Another alternative is wrapping pandoc in a function but I'm not a fan of that either.

prosoitos: Well, because of the multi-overlaps of tools, it is unclear to me how to write it:

Currently, there is:

```sh
        # OpenDocument
        odt|ods|odp|sxw)
            # Preview as text conversion
            odt2txt "${FILE_PATH}" && exit 5
            exit 1;;
```

I can't add pandoc here because pandoc does not accept .ods, .odp, .sxw.

I could take .odt out of this and put it elsewhere, where I can add pandoc, but that changes the current default for .odt files.

Seems that the only way out is to add some commented paragraph.

prosoitos: And if I add .odt with .docx (which would make sense), then I have to comment that out and that kills the option for .docx unless the scope.sh is edited.

That's what I was trying to explain. It gets a bit messy and complicated.

I will add .odt by itself with pandoc and comment it out.

prosoitos: > We could merge all those cases and use the fallthroughs

The complication is that pandoc overlaps several categories, but does not cover all the subcategories in those categories.

Anyhow, I will submit something in 2 min and you can re-asses.

prosoitos: > Another alternative is wrapping pandoc in a function but I'm not a fan of that either.

I guess, that would be the clean way to do this. But might be overkill for this purpose.

toonn: The `odt|ods|odp...` isn't an issue. If odt2txt fails, it falls through to pandoc, if that fails no preview is shown. I think that's what the `exit 5` means (it's at the top of `scope.sh`).

prosoitos: Something inspired by rifle might be best :slightly_smiling_face:

prosoitos: Oh, I see.

prosoitos: OK, that simplify things already. I'll change this.

prosoitos: Well, I guess, instead of putting fb2 and epub with docx as I did, I could add a separate paragraph for them with pandoc next to where they were initially covered. That would be much less messy.

prosoitos: I'll change that too.

toonn: Yes, keep all the original clauses, just add the fall through with pandoc.

prosoitos: I think the current version is more sensible than my first attempts at adding more pandoc input formats. How does it look?

ranger/data/scope.sh
```diff
@@ -147,6 +176,12 @@ handle_image() {
         #     ebook-meta --get-cover="${IMAGE_CACHE_PATH}" -- "${FILE_PATH}" > /dev/null \
         #         && exit 6 || exit 1;;
 
+	# ePub, FB2 (using pandoc)
```
toonn: Oh, wait. This doesn't work. You need to output a jpeg in the handle_image() function I think.

toonn: So let's just add these formats to the docx clause maybe. We can treat it as the general pandoc clause for now.

ranger/data/scope.sh
prosoitos: Note though that I haven't tested pandoc with the new formats I added... (I don't have any files with those formats).

So I don't know whether it works and whether it adds a delay.

prosoitos: If you have epub and fb2 files and want to test it out, that would be best.

ranger/data/scope.sh
prosoitos: But if it needs an extra function, it won't work there either, will it? So why not get rid of it for now?

prosoitos: I added some comments about not being able to test pandoc with those formats in a thread you marked as resolved. So I don't know if you saw my messages.

prosoitos: Oh, I see, I moved it to another function.

prosoitos: I moved it quickly and I did not realize that I moved from the `handle_extension()` to the `handle_image()` function :grimacing: 

Will move it back with .docx.

toonn: I saw, no worry. I trust pandoc implicitly : ) (getting inundated with emails from all your comments so I'm not missing *anything* : ))


What do you mean extra function? It's just adding `|epub|fb2` to the `docx` clause. It'll give nicer epub previews without needing image preview support.

prosoitos: I deleted that comment. It was a misunderstanding before I realized my mistake.

I moved it back in the extension fonction. Have to leave now, but will check later if you would like more changes.

ranger/data/scope.sh
```diff
@@ -40,7 +40,6 @@ HIGHLIGHT_STYLE=${HIGHLIGHT_STYLE:-pablo}
 HIGHLIGHT_OPTIONS="--replace-tabs=${HIGHLIGHT_TABWIDTH} --style=${HIGHLIGHT_STYLE} ${HIGHLIGHT_OPTIONS:-}"
 PYGMENTIZE_STYLE=${PYGMENTIZE_STYLE:-autumn}
 
-
```
toonn: I think we want these double whitespaces back, here and at l266. Otherwise this looks ok for now.

I have a PR in the works to add shellcheck to our CI, that might flag some issues with your commands, things like not putting ` -- ` in front of the `FILE_PATH` arguments so they won't get parsed as options/flags. So I'm gonna hold off on merging this until after.

prosoitos: Last commit:
- add `--` to all the apps I had added (after testing that it works)
- re-add empty lines
- add fallback pandoc support to html and json

ranger/data/scope.sh
```diff
@@ -75,20 +75,53 @@ handle_extension() {
...
-            ;; # Continue with next handler on failure
+	    pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
+	    ;; # Continue with next handler on failure
+
         # JSON
         json)
             jq --color-output . "${FILE_PATH}" && exit 5
             python -m json.tool -- "${FILE_PATH}" && exit 5
-            ;;
+	    pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
```
toonn: I don't think json's an input format to pandoc, only an output format. Unless you tested it and it works?

prosoitos: I have not tested it, but the manual says it is:

https://pandoc.org/MANUAL.html#general-options

toonn: Must've read over that because that's what I checked. *Still* though, it only works for json that represents a pandoc AST. That's not a useful general json preview.

prosoitos: Will take it out then.

Vifon:
@toonn, let's wait with this one until #1333 will be ready. This one will need to use the double comments now, correct?

toonn:
@Vifon, yes. I intend to update it myself.

prosoitos:
Please let me know if I can help. I am very much of a noob for this kind of contribution, but would be happy to learn and do more if given some tips and info.

toonn:
@prosoitos, I'll give you a ping when I merge the shellcheck stuff.

prosoitos:
Thanks :slightly_smiling_face:

toonn:
@prosoitos, just merged the shellcheck PR. I think the simplest course of action would be to rebase and then read the comments at the top of scope to understand what changed, the 3rd paragraph in particular.

prosoitos:
Oh, sorry, I missed this. I will do a rebase tomorrow when I have a little time. Thank you @toonn!

ranger/data/scope.sh
```diff
@@ -75,20 +75,52 @@ handle_extension() {
         odt|ods|odp|sxw)
             # Preview as text conversion
             odt2txt "${FILE_PATH}" && exit 5
+	    # Preview as markdown conversion
```
toonn: Weird indentation.

ranger/data/scope.sh
```diff
@@ -75,20 +75,52 @@ handle_extension() {
...
 
+	# RTF and DOC
+	rtf|doc)
+	    # Preview as text conversion
+	    # Note: catdoc does not always work for .doc files
+	    catdoc -- "${FILE_PATH}" && exit 5
+	    exit 1;;
+
+	# DOCX, ePub, FB2 (using markdown)
+	# You might want to remove "|epub" and/or "|fb2" below if you have uncommented other methods to preview those formats
```
toonn: Maybe you should add a pandoc line to the section for epub and fb2 instead? Like you did for html?

ranger/data/scope.sh
```diff
@@ -75,20 +75,52 @@ handle_extension() {
...
+	    exit 1;;
+
         # HTML
         htm|html|xhtml)
             # Preview as text conversion
             w3m -dump "${FILE_PATH}" && exit 5
             lynx -dump -- "${FILE_PATH}" && exit 5
             elinks -dump "${FILE_PATH}" && exit 5
-            ;; # Continue with next handler on failure
+	    pandoc -s -t markdown -- "${FILE_PATH}" && exit 5
```
toonn: More weird indentation.

ranger/data/scope.sh
```diff
@@ -75,20 +75,52 @@ handle_extension() {
...
+	    ;; # Continue with next handler on failure
+
         # JSON
         json)
             jq --color-output . "${FILE_PATH}" && exit 5
             python -m json.tool -- "${FILE_PATH}" && exit 5
-            ;;
+	    ;;
```
toonn: More weird indentation. Make sure to be using spaces for indentation, not tabs.

Issue #1600, opened 2019-06-19
AttributeError: 'NoneType' object has no attribute 'pointer'


kdheepak:
I'm getting the following error when I open ranger on an image file

```
$ ranger ~/Desktop/Screen\ Shot\ 2019-06-19\ at\ 3.38.29\ PM.png
ranger version: ranger 1.9.2
Python version: 3.7.3 (default, Mar 27 2019, 16:54:48) [Clang 4.0.1 (tags/RELEASE_401/final)]
Locale: en_US.UTF-8

Traceback (most recent call last):
  File "/Users/$USER/miniconda3/lib/python3.7/site-packages/ranger/core/main.py", line 195, in main
    fm.loop()
  File "/Users/$USER/miniconda3/lib/python3.7/site-packages/ranger/core/fm.py", line 394, in loop
    ui.redraw()
  File "/Users/$USER/miniconda3/lib/python3.7/site-packages/ranger/gui/ui.py", line 339, in redraw
    self.finalize()
  File "/Users/$USER/miniconda3/lib/python3.7/site-packages/ranger/gui/ui.py", line 394, in finalize
    DisplayableContainer.finalize(self)
  File "/Users/$USER/miniconda3/lib/python3.7/site-packages/ranger/gui/displayable.py", line 264, in finalize
    displayable.finalize()
  File "/Users/$USER/miniconda3/lib/python3.7/site-packages/ranger/gui/widgets/view_base.py", line 60, in finalize
    - self.main_column.scroll_begin
AttributeError: 'NoneType' object has no attribute 'pointer'

ranger crashed. Please report this traceback at:
https://github.com/ranger/ranger/issues
```

#### Runtime Environment

- Operating system and version: 
MacOS Sierra - 10.12.6
- Terminal emulator and version: 
iTerm2 - Build 3.3.0beta12
- Python version: 
Python 3.6.5 |Anaconda, Inc.| (default, Mar 29 2018, 13:14:23)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
- Ranger version/commit: 1.9.2

Any suggestions on what might be going on? This is my first time using ranger, so I might be missing something.

I've also updated the rc.conf file with the settings for iTerm2

```
set preview_images true
set preview_images_method iterm2
```

kdheepak:
If I try to use ranger to open the folder, I get the broken image icon.

[...]

PR #1601, opened 2019-06-20
Fix Python3 image preview in iTerm2

kdheepak:
Fixes image previews in iTerm2 when using Python3. See #1600 for some additional information.

#### ISSUE TYPE

- Bug fix

#### Runtime Environment

    Operating system and version:
    MacOS Sierra - 10.12.6
    Terminal emulator and version:
    iTerm2 - Build 3.3.0beta12
    Python version:
    Python 3.6.4 
    Type "help", "copyright", "credits" or "license" for more information.
    Ranger version/commit: 1.9.2

#### CHECKLIST

Screen Shot 2019-06-20 at 3 04 34 AM

#### DESCRIPTION

I compared the output in the function that generates the image, of Python2 and Python3 versions of ranger. I found that the output of the `_encode_base_image` returned a bytestring. This was being string interpolated into a string. In Python2, this will result in an implicit conversion of the bytestring to a unicode string and will be interpolated into a unicode string. In Python3, this will result in the representation of the bytestring interpolated into the unicode string. See following example for differences between Python2 and Python3:

Screen Shot 2019-06-20 at 3 12 02 AM

The fix that I issued in this case is to decode the base image from a bytestring to a unicode string. This will do the right thing in Python2 and Python3.

Screen Shot 2019-06-20 at 3 17 39 AM

There's probably other places this change needs to be made as well but this is the only environment I'm able to test on.

Additionally, I'm not sure if decoding the bytestring from utf-8 to a unicode string is the right thing to do. Maybe it should be `.decode('ascii')` instead? Let me know what you think.

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -308,7 +308,7 @@ class ITerm2ImageDisplayer(ImageDisplayer, FileManagerAware):
     def _encode_image_content(path):
         """Read and encode the contents of path"""
         with open(path, 'rb') as fobj:
-            return base64.b64encode(fobj.read())
+            return base64.b64encode(fobj.read()).decode('utf-8')
 
     @staticmethod
     def _get_image_dimensions(path):


Xanxus1111:
屏幕快照 2019-08-27 下午5 28 11

Still have the same problem. 
Python3.6
iTerm2 3.3
macOS Mojave 10.14.6
ranger 1.9.2

toonn:
@Xanxus1111, is that *with* the patch?

Xanxus1111:
> @Xanxus1111, is that _with_ the patch?

nope It is just a traditional image file. There is something wrong with my image preview.

Xanxus1111:
@toonn 
nope It is just a traditional image file. There is something wrong with my image preview.

toonn:
@Xanxus1111, you're misunderstanding. Did you or did you not manually apply the fix in this PR to your ranger? The fix isn't in the latest release. We're working on the next release.

Xanxus1111:
@toonn  Thank you for your patient. I fixed it 👍😁

toonn:
Thanks for the fix, great first contribution! 🎉 
Apologies for the long wait.

PR #1604, opened 2019-06-27
Decode bytes into str for image display of iTerm2

### LINKED ISSUE #1546: iTerm2 image preview doesn't work with python 3
#### Runtime Environment

- Operating system and version: macOS 10.13.6
- Terminal emulator and version: iTerm2 3.3
- Python version: 3.6.8 installed by pyenv
- Locale: en_US.UTF-8

#### Current Behavior
`ranger` installed by `pip3 install ranger-fm` doesn't work with iTerm2 image preview. A placeholder image is displayed but not the real image.

After removing the `pip3` installed `ranger` and reinstalled it with Homebrew, the problem disappeared.

The shebang of `ranger` from the brew installed one is `#!/usr/bin/python -O` (python 2). I could reproduce the problem by changing it to `#!/usr/local/bin/python3 -O`.

Besides, iTerm2 image preview works with shebang `#!/usr/local/bin/python -O` (python 2) as well.

~~The reason I tried to use `ranger` with Python 3 is that, the keybindings from the wiki for tmux integration doesn't work Python 2.~~

~~Error `'module' object has no attribute 'quote'` occurred on Python 2, which is already reported by #1335~~

**Update**: the tmux integration is solved.

laggardkernel:
`base64.b64encode()` returns bytes in Python 3. We need to decode bytes into str to make image display for iTerm2 works under Python 3. This would fix #1546

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -308,7 +308,10 @@ class ITerm2ImageDisplayer(ImageDisplayer, FileManagerAware):
     def _encode_image_content(path):
         """Read and encode the contents of path"""
         with open(path, 'rb') as fobj:
-            return base64.b64encode(fobj.read())
+            r = base64.b64encode(fobj.read())
+            if isinstance(r, bytes):
+                r = r.decode('utf-8')
+            return r
 
     @staticmethod
     def _get_image_dimensions(path):


laggardkernel:
Damn, I didn't notice this is duplicated with GH-1601. 😨

toonn:
Since I can't confirm incompatibility with python 2, i.e., it seems to work just fine, I'm going with the simpler PR. Thanks for the contribution and the cross-referencing!

PR #962, opened 2017-10-23
Show current date and time in status bar

stuarthowieson:
#### ISSUE TYPE
- Improvement/feature implementation

#### RUNTIME ENVIRONMENT
- Operating system and version: macOS High Sierra (10.13)
- Terminal emulator and version: macOS Terminal.app 2.8 (400)
- Python version: 2.7.10
- Ranger version/commit: ranger-master 1.9.0b5
- Locale: None.None

#### CHECKLIST
- [x] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [x] All changes follow the code style **[REQUIRED]**
- [x] All new and existing tests pass **[REQUIRED]**
- [ ] Changes require config files to be updated
    - [ ] Config files have been updated
- [ ] Changes require documentation to be updated
    - [ ] Documentation has been updated
- [ ] Changes require tests to be updated
    - [ ] Tests have been updated

#### DESCRIPTION
A date and time indicator has been added to the right of the status bar. It displays in the same format as the date and time indicator already used for the _ctime_ and _vcs commit_ displays. It updates every minute.

#### MOTIVATION AND CONTEXT
The addition was motivated by the lack of a current date and time display in ranger. I often work with the terminal fullscreen and, having come across ranger recently and having integrated it into my workflow, could do with seeing the current date and time from within ranger.

#### TESTING
The supplied tests have been performed and passed. The changes are very self-contained therefore the changes do not have far-reaching effects.

#### IMAGES / VIDEOS
screen shot 2017-10-23 at 19 41 54

PR #1619, opened 2019-07-10
Added a configurable option to display date and time in the status bar

cibinmathew:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: CentOS-7
- Terminal emulator and version: GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)

- Python version: 3.7.3
- Locale: None

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Added the changes requested in https://github.com/ranger/ranger/pull/962 to add a configurable option to enable/disable displaying date and time in the status bar on top of the existing changes..
By default, date and time is not displayed.

#### TESTING

Ran tests as per the guidelines


toonn:
Have you considered adding a setting to allow specifying a format string? Would make it a lot more flexible.

cibinmathew:
> Have you considered adding a setting to allow specifying a format string? Would make it a lot more flexible.
>
In **rc.conf**, string with quotes not supported like below?
_set display_time_in_status_bar_format "%Y-%m-%d %H:%M"_

toonn:
You don't need the quotation marks. They end up literally in the string so your example actually results in the value `display_time_in_status_bar_format="\"%Y-%m-%d %H:%M\""`.

cibinmathew:
> You don't need the quotation marks. They end up literally in the string so your example actually results in the value `display_time_in_status_bar_format="\"%Y-%m-%d %H:%M\""`.

For **set** command, it kicks in cmd.resolve_macros() method and this throws error. Can resolve_macros be made false for set commands?

toonn:
I think you'll have to escape them by doubling up the %'s : /

cibinmathew:
Even with escaping %,  when trying to resolve macros(since python timestamp format contains '%'), it tries to access **thisfile** attribute which is non-existent in this case
https://github.com/ranger/ranger/blob/0cf33f2f7256429c616917c370675cc39091b35a/ranger/core/actions.py#L306


        if self.fm.thisfile:
            macros['f'] = self.fm.thisfile.relative_path
        

Shouldnt it be like

        if hasattr(self.fm, 'thisfile'):
            macros['f'] = self.fm.thisfile.relative_path

toonn:
#1629 should fix this particular problem.

toonn:
Let's split this into two settings, `time_format` defaults to `%X` and `date_format` defaulting to `%x`. Base it on #1715 to see the places that need changing.

toonn:
Closing this in favor of #1715.

hut:
I believe this was closed by mistake. The reason was that #1715 covers this already, but that PR doesn't actually add the current time to the status bar. This is a fine PR, I'll reopen it.

Issue #1616, opened 2019-07-07
Smarter `pp` paste for existing files


laktak:
When you paste a file that already exists (e.g. `yy` followed by `pp`), ranger will append a `_` to the name.

At the moment this breaks the file extension (e.g. `foo.txt` becomes `foo.txt_`).

Could this be changed to insert the `_` before the extension? Would you accept a PR for this?

toonn:
I don't like it. Extensions are *really* hard to guess and it's a lot less obvious that something's amiss. I'd rather fail harder, faster. You *could* write rifle rules so you can still open such files easily.

*If* we changed it we'd change it to a prefix, that way you don't have to worry about extensions.
A problem either way is that it's not clear *which* are the renamed files after such a paste if there's files with multiple affixes afterwards. So maybe asking for confirmation for each file would be better?

Another solution would be to hash all the clashing files, then create a directory containing that hash, if it already exists forget about those files, otherwise put them in there. Ugly? Yes but also less ambiguous. Still not ideal if you paste tons of clashing files though, "Which hashed directory did the last paste go into?"

laktak:
Hmm, I would argue that there is nothing amiss. By using `pp` I expect ranger to rename conflicting files instead of overwriting them with `po`.

What do you mean with extensions being hard to guess. Do you mean long extensions like `.tar.gz`? But even only taking the short extension would improve the situation (so you would get `foo_.md` instead of `foo.md_`). A prefix (`_foo.md`) is OK but it would also move the file away from the original in the list.

toonn:
I know, it'd group all the conflicting files. A half-assed solution wouldn't be an improvement imo. And dots don't always separate extensions `Masterpiece.2019.07` would be a perfectly valid name for a text file.
The fact that the suffix "breaks" the extension is a feature, not a bug. It's how you know something's amiss.

laktak:
OK, I see your point. I still don't like this feature.

toonn:
I'm reopening this because I could see prefix v. suffix as a setting or better yet a flag passed to the paste function. (Maybe a function argument instead so you could pass a "Insert _ before best guess at extension" function even.)

laktak:
The path is generated here: https://github.com/ranger/ranger/blob/master/ranger/ext/shutil_generatorized.py#L106-L119

I've modified it so it uses Python's splitext function to get to the extension of the file:

```
def get_safe_path(dst, keep_ext=False):
    if not os.path.exists(dst):
        return dst

[...]

PR #1624, opened 2019-07-14
Added a make_safe_path parameter to paste

laktak:
<del>It allows you to specify 'overwrite', 'rename', or 'rename_ext' (=default, same as the current behaviour) in case of a file name conflict. This obsoletes the overwrite parameter.</del>

It allows you to specify your own function to generate the path for conflicting files.

See https://github.com/ranger/ranger/issues/1616

#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

When you paste a file that already exists (e.g. yy followed by pp), ranger will append a _ to the name.

At the moment this breaks the file extension (e.g. foo.txt becomes foo.txt_).

This PR makes it possible to specify how the conflict should be resolved.


toonn:
I'd still rather have the more flexible solution I described. Rather than having a long if statement that picks a function based on a string, why not just pass in a function? It's infinitely more useful. I know you can't pass a function from `rc.conf` easily. But I look at this as something you'd customize using `commands.py` instead.

laktak:
I forgot to fix the default arg for CopyLoader but it does take a function as an argument. I described the call chain in the issue and thought that is what you meant.

Can you override actions in `commands.py` or how would you customize it?

ranger/core/actions.py
```diff
@@ -1591,14 +1592,29 @@ def _recurse_hardlinked_tree(self, source_path, target_path):
                 link(source_path,
                      next_available_filename(target_path))
 
-    def paste(self, overwrite=False, append=False, dest=None):
+    def paste(self, conflict='rename_ext', append=False, dest=None, overwrite=False):
```
toonn: Don't change the argument order. There's no good reason to.

toonn: I also don't like the string argument, too error prone. Let's just pass a function.

ranger/core/actions.py
```diff
@@ -1591,14 +1592,29 @@ def _recurse_hardlinked_tree(self, source_path, target_path):
...
         """:paste
 
         Paste the selected items into the current directory or to dest
         if provided.
         """
+        # overwrite is obsolete, it was replaced by conflict
+        if overwrite:
+            conflict = 'overwrite'
+
+        if conflict == 'overwrite':
```
toonn: This whole if statement needs to go, too complicated.

toonn: This is why we're using a function argument.

ranger/core/loader.py
```diff
@@ -51,12 +52,12 @@ def destroy(self):
 class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instance-attributes
     progressbar_supported = True
 
-    def __init__(self, copy_buffer, do_cut=False, overwrite=False, dest=None):
+    def __init__(self, copy_buffer, do_cut=False, resolve_conflict=get_safe_path, dest=None):
```
toonn: `overwrite`'s not going anywhere. It should take precedence over the new argument.

ranger/ext/safe_path.py
```diff
@@ -0,0 +1,41 @@
+# This file is part of ranger, the console file manager.
+# License: GNU GPL version 3, see the file "AUTHORS" for details.
+
+import os
+
+APPENDIX = '_'
+
+
+def get_safe_path_classic(dst):
```
toonn: The original `get_safe_path` is definitely not gonna change names.

examples/rc_emacs.conf
```diff
@@ -335,8 +335,8 @@ map <A-d>  console rename%space
 map <C-e>  eval fm.open_console('rename ' + fm.thisfile.relative_path)
 map <C-a>  eval fm.open_console('rename ' + fm.thisfile.relative_path, position=7)
 
-map <C-y>y  paste
-map <C-y>o  paste overwrite=True
+map <C-y>y  paste conflict=rename_ext
```
toonn: The default bindings are certainly not gonna change.

laktak:
> The intent is good but I feel like we're not getting closer to what I've been suggesting.

It's not like I dislike your suggestions - I did not add a function argument to paste because I have no idea how to configure it. Some help please.

Is this how you would do it? https://github.com/ranger/ranger/blob/master/examples/plugin_new_macro.py

toonn:
No, you'd just write a new command in `commands.py` that's just a thin wrapper over `paste` and passes in a different function for conflict resolution. `commands.py` is 100% a config file from our pov.

toonn:
And don't worry, we'd probably include the "rename_ext" in the default `commands.py` so you won't even actually need to keep track of your own.

laktak:
While looking into this I thought of another approach - how about we simply allow the user to replace the `get_safe_path()` function with his own implementation? E.g. a `set_safe_path_function(...)` that he can call in `commands.py`?

That way we could keep all functions as they are and only need to change the code in `shutil_generatorized.py`.

toonn:
Nope, that's worse. What's wrong with the way I suggested?

laktak:
> No, you'd just write a new command in `commands.py` that's just a thin wrapper over `paste` and passes in a different function for conflict resolution. `commands.py` is 100% a config file from our pov.

Can you tell me how to pass functions from this wrapper over `paste`?

```
class paste_ext(Command):
    def execute(self):
        paste_cls = self.fm.commands.get_command("paste")
        paste = paste_cls("overwrite=False")
        return paste.execute()
```

toonn:
There is no paste "Command", you just call the paste function defined in `actions.py`, you can handle arguments to your custom command if you want but you'd pass those in addition to a new locally defined conflict handling function.

laktak:
Took me a while to see that fm inherits from Actions. I've added `paste_ext` to ` ranger/config/commands.py`.

doc/ranger.pod
```diff
@@ -527,8 +527,8 @@ also "da", "dr" and "dt" shortcuts equivalent to "ya", "yr" and "yt".)
 Paste the files which were previously copied or cut, like pressing Ctrl+V in
 modern GUI programs.
 
-Conflicts will be renamed by appending an _ to the file extension. To change
-this see the paste command.
+Conflicts will be renamed by appending an _ to the full file name. To change
+this and insert a _ before the file extension see commands.py.
```
toonn: Please check elsewhere in the file, I think we use some sort of formatting for files. `F<file.ext>` maybe? Maybe we should also quote the underscores while we're at it? Look for other examples of single literal characters in the manual to check.

ranger/config/commands.py
```diff
@@ -1859,3 +1859,34 @@ def tab(self, tabnum):
...
             if mode
         )
+
+
+class paste_ext(Command):
+    """
+    :paste_ext
+
+    Like paste but renames conflicting files so that the
+    file extension stays intact.
```
toonn: Maybe admit here that it's best-effort rather than claim it always works : )

ranger/core/actions.py
```diff
@@ -1591,15 +1591,15 @@ def _recurse_hardlinked_tree(self, source_path, target_path):
                 link(source_path,
                      next_available_filename(target_path))
 
-    def paste(self, overwrite=False, append=False, dest=None):
+    def paste(self, overwrite=False, append=False, dest=None, make_safe_path=None):
```
toonn: Can't we just assign `get_safe_path` by default rather than `None`? Should simplify some of the other code as well.

laktak: Yes but at the moment `shutil_generatorized.py` appears to be 'private' (it's only imported in the `generate` function of `CopyLoader`). Should we move `get_safe_path` to a new file?

ranger/config/commands.py
laktak: OK

doc/ranger.pod
laktak: fixed

laktak:
@toonn ready to merge?

toonn:
I think so, still have to review thoroughly and actually test. Have been busy, will ping you when I get around to it, if necessary.

laktak:
hey @toonn can I help to get this merged? Maybe buy you a coffee? :-)

toonn:
Thanks for the ping, oversight on my part. Don't hold your breath though. Slow going lately.

laktak:
@toonn thank you! Wish you happy holidays! Let me know if you have a donate link (I only found the one for bitcoin which I don't have).

toonn:
We have a bitcoin donation link? That's news to me : )
Afaik the ranger team doesn't currently take donations. Thanks for the gesture anyway!
🎄 Enjoy the festivities!
(There won't be a new ranger under the tree but the new release isn't far off : ))

Issue #1618, opened 2019-07-09
Need a better way to show copy buffer contents


rayx:
The built-in way is `:echo %c`, but it displays all files in a single line and doesn't work well if the length of `%c` exceeds the line length.

Toon suggested a workaround in a separate discussion: `:shell echo %c | sed 's/ /\n/g' | less`, but unfortunately this doesn't work if file path contains space. 

It seems to be me that a dedicated command is needed to list copy buffer contents.

rayx:
I think there are two issues here:

1) The first issue is with `%c` and all other macros that may contain multiple files. They use space as separator. This doesn't work if file name or path contains space also. This is common issue with these macros and is not specific to listing copy buffer conent.

2) Due to the above issue, it's impossible to use `%c` to show copy buffer contents reliably, so a dedicated command is needed. Ideally it should display copy buffer in multiple lines, one for each file. But I'm not sure if ranger UI supports that.

nQ8hzkZw:
You can add a new command to `commands.py`, something like this:

```python
class show_copy_buffer(Command):

    def execute(self):
        for fobj in self.fm.copy_buffer:
            self.fm.notify(fobj.path)
```

`:show_copy_buffer` will show a contents of the copy buffer in message log `W`.

Or add a new mapping to `rc.conf`, for example:
`map b chain save_copy_buffer; shell -p cat ~/.config/ranger/copy_buffer`

toonn:
A general way to view macro contents isn't a bad idea actually. Might make it easier to get a feel for when to use which macros.

@nQ8hzkZw's workaround is definitely an improvement. One-liner version `:eval for f in self.fm.copy_buffer: self.fm.notify(f.path)`

rayx:
@nQ8hzkZw Thanks for the nice workarounds.

@toonn I'll see if I can work out a patch to add a command for viewing macro contents. Will let you know if I have more questions.

rayx:
So I looked at the code and found `:shell` command does quote file name or path (see line 282-287 in https://github.com/ranger/ranger/blob/master/ranger/core/actions.py). That leads to a simpler workaround:

    :shell -p ls %c 

Of course, this assumes that contents in macros are valid arguments for ls (meaning: they are either full pathnames or are filenames relative to cwd of sub-process). I'm not sure if that's true. So I settle down with another command that doesn't depend on `ls` and is reliable for all macros (Note the second `bash` argument in the command line is just for information purpose. See `-c` option in bash man page): 

[...]

PR #1625, opened 2019-07-14
Add display_macro command

rayx:
The change adds `display_macro` command to shows contents of a specified macro in the pager. 

Syntax of the command: 

    :display_macro <macro>

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: Ubuntu 18.04.2
- Terminal emulator and version: GNOME Terminal 3.28.2
- Python version: 3.6.7 (default, Oct 22 2018, 11:32:17) [GCC 8.2.0]
- Ranger version/commit: ranger-master v1.9.2-278-g89f61236
- Locale: en_US.UTF-8

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
The command supports all macros except %any and its friends because they are valid only when action functions are executed. Man page is also updated.

#### MOTIVATION AND CONTEXT
Ranger doesn't have a general command to view macro contents. See issue https://github.com/ranger/ranger/issues/1618.

#### TESTING
I tested the following scenarios:
- Run the command with a macro which contains multiple files (i.e., %c, after I added a few files to it). 
- Run the command with a macro which contains a single file (i.e., %d, or %rangerdir)
- Run the command with a macro which contains nothing (i.e, %c, before I added any files to it). The command shows "Macro %c is empty!" in the pager in this case.
- Run the command with an invalid macro name (i.e., %x, or "abc"). The command prints "Invalid macro error" message in status bar in this case.
- Run the command with incorrect number of arguments (i.e., with no arguments or more than 1 arguments). The command prints "Syntax: display_macro <macro>" message in status bar in this case.

The code change doesn't affect existing code. 

#### IMAGES / VIDEOS
N/A


ranger/config/commands.py
```diff
@@ -1859,3 +1859,18 @@ def tab(self, tabnum):
...
+    """:display_macro <marco>
+
+    Show macro contents in the pager.
+    """
+
+    resolve_macros = False
+
+    def execute(self):
+        if not self.arg(1) or self.arg(2):
+            self.fm.notify('Syntax: display_macro <marco>', bad=True)
```
toonn: Polo!

toonn: : )

ranger/core/actions.py
```diff
@@ -1651,3 +1651,30 @@ def rename(self, src, dest):
...
+            return False
+        macros = self.get_macros()
+        try:
+            value = macros[name[1:]]
+            # Macro value returned by get_macros() can be:
+            # - String: if the macro contains a single file
+            # - List: if the macro contains multiple files
+            # - MACRO_FAIL: if the macro hasn't been set
+            if value == MACRO_FAIL:
+                value = []
```
toonn: Maybe we want to display some placeholder along the lines of "This macro's empty."?

ranger/core/actions.py
```diff
@@ -1651,3 +1651,30 @@ def rename(self, src, dest):
...
+            if isinstance(value, str):
+                value = [value]
+        except KeyError:
+            self.notify('Invalid macro name: %s' % name)
```
toonn: Extract these two formatting strings into a variable. Less room for error.

ranger/core/actions.py
```diff
@@ -1651,3 +1651,30 @@ def rename(self, src, dest):
...
+            return False
+
+        # Display the value in the pager.
+        pager = self.ui.open_pager()
+        if value:
+            pager.set_source(["Macro %s:" % name] + value)
+        else:
+            pager.set_source(["Macro %s is empty!" % name])
```
toonn: If we do the placeholder we can drop this special case.

ranger/config/commands.py
rayx: Just curious, what does "Polo" mean? I googled it but didn't find an answer :)
That said, when I relooked at the code, I found `Command` class had `args` attribute, from which I could get argument length directly, so I modified the above code to improve the check in the if statement a bit.

ranger/core/actions.py
rayx: Done.

ranger/core/actions.py
rayx: Done.

ranger/core/actions.py
rayx: Done. Please see the updated PR.

ranger/config/commands.py
toonn: It's a pool game, someone's blindfolded and yells "Marco!" and then everyone else has to yell "Polo!" and the first one needs to try to find them. Just a joke because you had typo `<marco>` ; )

rayx: Changed. And thanks for the introduction to the game. It's interesting to know about it.

PR #1629, opened 2019-07-18
Catch exceptions when resolving macros in rc.conf

toonn:
While parsing the settings in `rc.conf` `self.fm.thisfile` doesn't exist
yet.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -21,21 +21,20 @@ from sys import version_info
 from logging import getLogger
 
 import ranger
-from ranger.ext.direction import Direction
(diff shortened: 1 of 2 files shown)


cibinmathew:
There are few more I could see further down which raises exception while resolving macros.

cibinmathew:
@toonn can you please review the other scenarios also added.

toonn:
@cibinmathew I dove in and implemented what I think is necessary to ease my concerns. Much less duplication and I believe I've retained the semantics re `None`.
It'd be great if you could test this branch.

vifon:
@toonn Looks good to me. Resolve the conflicts and I think it's good to go.

PR #1630, opened 2019-07-18
Handle exceptions while resolving macros from rc.conf

cibinmathew:
#### ISSUE TYPE

- Bug fix
Added exception handling for other attribute references (https://github.com/ranger/ranger/pull/1629) 
(While parsing the settings in _rc.conf_ `self.fm.thisfile doesn't exist yet`).


cibinmathew:
@toonn Can you please merge if this looks good.

toonn (commented):
I'm worried about two things, the first is semantics, `if attribute` means "if the attribute is a value other than none," we lose this with try-except.
The other thing is the repetition, maybe we should introduce a helper function for this?

toonn:
I think #1629 handled all of these cases, am I mistaken?

PR #1636, opened 2019-07-23
support wayland condition for rifle

### LINKED ISSUE #1635: Support wayland condition + integration with X condition (rifle)
#### Current Behavior

Currently, there is the `X` flag for rifle.conf that allows only showing a certain command when an X display is running (eg. for firefox). This obviously doesn't work for wayland.
I can work around this for now by adding duplicate entries like so:

```
ext x?html?, has firefox,        X, flag f = firefox -- "$@"
ext x?html?, has firefox,        !X, env WAYLAND_DISPLAY, flag f = firefox -- "$@"
```

The `!X` in the second entry is to avoid having duplicate entries when running wayland and xwayland is enabled.

I can't put both in the same entry, but the conditions are AND'd together.

#### Expected Behavior

Some condition or easier method to specify that a command can run under wayland as well as X.

#### Context

Need an easy way to specify that a command should be shown if wayland is running ($WAYLAND_DISPLAY is set), that plays nice with `X`.

#### Possible Solutions

Add a `W` condition that checks for `WAYLAND_DISPLAY` (to be the counterpart to `X`. This should play nice with the `X` condition so it logically works out. Eg.

```
# only shows if X available
X = firefox
# only shows if Wayland available
W = firefox
# only shows if Wayland or X available
X, W = firefox
```

swalladge:
Expands `X` and `t` flags to support Wayland as a graphical environment (to add to the current checks for darwin and Xorg).

fixes #1635

#### ISSUE TYPE

- Breaking changes (backwards compatible, but the XW condition in rifle.conf is not backwards compatible

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

~This introduces the extra `W` condition. This checks if
`WAYLAND_DISPLAY` is set. It can be used with the `X` condition to mean
"X or Wayland is available". For example:~

```
# firefox supports X and wayland
ext x?html?, has firefox,          XW, flag f = firefox -- "$@"
# example program that only runs on wayland
ext test, has example,          W = example -- "$@"
# backwards compatible (just X still works)
ext x?html?, has chromium,       X, flag f = chromium -- "$@"
```

#### MOTIVATION AND CONTEXT

This is to support users running wayland. Users may be running pure
wayland (without xwayland available), or with xwayland for running
X-only applications within wayland. The `X` condition is not enough to
neatly support these scenarios without needing to add duplicate entries
to rifle.conf

#### TESTING

Manually verified this. Did not add unit tests as there are no existing
tests found for the `X` condition behaviour.

### DIFF
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -27,6 +27,8 @@
 #   number <n>     | change the number of this command to n
 #   terminal       | stdin, stderr and stdout are connected to a terminal
 #   X              | $DISPLAY is not empty (i.e. Xorg runs)
+#   W              | $WAYLAND_DISPLAY is not empty (i.e. Wayland runs)
+#   XW or WX       | Wayland OR Xorg runs
 #
 # There are also pseudo-conditions which have a "side effect":
 #   flag <flags>  | Change how the program is run. See below.
@@ -70,7 +72,7 @@ ext x?html?, has chromium-browser, X, flag f = chromium-browser -- "$@"
 ext x?html?, has chromium,         X, flag f = chromium -- "$@"
 ext x?html?, has google-chrome,    X, flag f = google-chrome -- "$@"
 ext x?html?, has opera,            X, flag f = opera -- "$@"
-ext x?html?, has firefox,          X, flag f = firefox -- "$@"
+ext x?html?, has firefox,          XW, flag f = firefox -- "$@"
 ext x?html?, has seamonkey,        X, flag f = seamonkey -- "$@"
 ext x?html?, has iceweasel,        X, flag f = iceweasel -- "$@"
 ext x?html?, has epiphany,         X, flag f = epiphany -- "$@"
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -26,9 +26,7 @@
 #   directory      | $1 is a directory
 #   number <n>     | change the number of this command to n
 #   terminal       | stdin, stderr and stdout are connected to a terminal
-#   X              | $DISPLAY is not empty (i.e. Xorg runs)
-#   W              | $WAYLAND_DISPLAY is not empty (i.e. Wayland runs)
-#   XW or WX       | Wayland OR Xorg runs
+#   X              | A graphical environment is available (darwin, Xorg, or Wayland)
 #
 # There are also pseudo-conditions which have a "side effect":
 #   flag <flags>  | Change how the program is run. See below.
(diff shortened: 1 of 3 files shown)


toonn:
I think we'd rather have `X` mean X11 or Wayland or Darwin.

swalladge:
@toonn this doesn't really make sense since there are some applications that support X but not wayland, others that support both, and potentially others that only support wayland (although this is probably a rare case). If I'm on a wayland system with xwayland disabled, I only want to see applications suggested by rifle that actually support wayland.

toonn:
Same story for darwin. As maintainers I'm not sure we care. The semantics of the `X` flag are "graphical applications can be run." We have to maintain the default `rifle.conf` and I'm not exactly looking forward to having to verify a certain program runs on all of X11, Wayland and Cocoa.

You can already be more specific using `env` and I'm not opposed to adding more specific explicit flags or maybe even a way to specify custom flags. But that's a separate issue.

swalladge:
OK fair enough. Would you merge if the 'X' flag was simply updated to check for Wayland as well?

The only downside is if someone is on pure Wayland they will get suggestions for X only apps too (which will be a lot). I guess they can always just use 'env WAYRAND_DISPLAY' in their own conf files if it becomes annoying. At least supporting Wayland in the X flag means that the default conf will still show graphical apps on pure Wayland.

July 23, 2019 12:58:52 PM UTC Toon Nolten <notifications@github.com> kirjoitti:
>Same story for darwin. As maintainers I'm not sure we care. The
>semantics of the `X` flag are "graphical applications can be run." We
>have to maintain the default `rifle.conf` and I'm not exactly looking
>forward to having to verify a certain program runs on all of X11,
>Wayland and Cocoa.
>
>You can already be more specific using `env` and I'm not opposed to
>adding more specific explicit flags or maybe even a way to specify
>custom flags. But that's a separate issue.

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

toonn:
Yes, I would. And I'd probably accept a PR that implements triples along the lines of `XWD|X_D|X__|WDX`. Only real issue I see is it eats into our flag namespace, but ¯\_(ツ)_/¯

I'll ask whether anyone'd rather see custom flag definitions along the lines of: `flag XW = X, env WAYLAND_DISPLAY`.

swalladge:
@toonn How about this now - simply expanding `t` and `X` flags to include wayland when checking for a graphical environment. Then later if a design for custom flags or more finegrained control is developed, it can be reworked further.

swalladge:
@toonn  any updates on what you'd like to see with this PR?

PR #1639, opened 2019-07-23
doc: Mention necessary macro escaping for chain

Zaeph:
Hi there,

We've had someone on #ranger mention that s/he was having problems with the `chain` command, namely that it wasn't respecting the `%space` inside the commands to be chained.  This is because `chain` also expands macros, so `%space` has to be escaped to `%%escaped`.

This patch mentions this in the doc.

The documentation has *not* been updated.  I'd rather wait for feedback on the language.

HTH.


doc/ranger.pod
```diff
@@ -328,6 +328,14 @@ the end of the line automatically.
 
 To write a literal %, you need to escape it by writing %%.
 
+For adding a space character after a chained command, you also need to escape
```
toonn: I think it'd be more appropriate trailing the `%space` explanation. Maybe change the order so the %% has already been explained.

"Note that macros are expanded twice when using chain. For example, to use `%space` with a chain command you'd write `%%space`."

I think that's clearer. Also, I don't like your example because `%space` is just a workaround for editors trimming trailing whitespace. So you'd normally only need it at the very end of commands you put in `rc.conf`.

Zaeph: I've implemented your comments in the next commit.

doc/ranger.pod
```diff
@@ -328,6 +328,11 @@ the end of the line automatically.
 
 To write a literal %, you need to escape it by writing %%.
 
+Note that macros are expanded twice when using chain.  For example, to insert
+a space character in a chained command, you would write %%space. This is
+because the chain command is a macro-expanding command itself.
```
toonn: Drop the itself.

doc/ranger.pod
```diff
@@ -328,6 +328,11 @@ the end of the line automatically.
 
 To write a literal %, you need to escape it by writing %%.
 
+Note that macros are expanded twice when using chain.  For example, to insert
```
toonn: Don't double-space : )

toonn:
I merged this in  6fee63a, not sure why github's not closing automatically.

toonn:
Did I fuck up the merge?

PR #1642, opened 2019-07-25
add update_icon_title option

### LINKED ISSUE #1637: Update taskbar title together with window title when set update_title true
#### Runtime Environment

- Operating system and version: `Linux es 5.2.1-arch1-1-ARCH #1 SMP PREEMPT Sun Jul 14 14:52:52 UTC 2019 x86_64 GNU/Linux`
- Terminal emulator and version: `rxvt-unicode (urxvt) v9.22 - released: 2016-01-23`
- Python version: `3.7.3 (default, Jun 24 2019, 04:54:02) [GCC 9.1.0]`
- Ranger version/commit: `ranger 1.9.2`
- Locale: `en_US.UTF-8`

#### Current Behavior
Ranger updates the title (`WM_NAME`) but not the taskbar title (`WM_ICON_NAME`) when `set update_title true` is set in `rc.conf`.

#### Expected Behavior
Both the title and taskbar title should be changed.

#### Context

I want to be able to see from the taskbar what I have open in ranger.

#### Possible Solutions
Change [this line](https://github.com/ranger/ranger/blob/669100aad83e3a51f140b6085d836a8260eda8d0/ranger/gui/ui.py#L385) to `'\x1b]0;',` (source: http://tldp.org/HOWTO/Xterm-Title-3.html). Running
```
curses.setupterm()
curses.tigetstr('tsl')
```
returns `b'\x1b]2;'`, which only changes the window title but not the icon name.

It would be preferred of course if there was a way in `curses` to set the taskbar title without hard-coding this string.

#### Steps to reproduce
1. Open ranger with `set update_title true`
2. Navigate to the previous directory
3. Window title changes but the taskbar title doesn't

#### Traceback

```

```

siikamiika:
resolves #1637

Add an option for updating the icon title, similar to `update_title`.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: `Linux es 5.2.1-arch1-1-ARCH #1 SMP PREEMPT Sun Jul 14 14:52:52 UTC 2019 x86_64 GNU/Linux`
- Terminal emulator and version: `rxvt-unicode (urxvt) v9.22 - released: 2016-01-23`
- Python version: `3.7.3 (default, Jun 24 2019, 04:54:02) [GCC 9.1.0]`
- Ranger version/commit: git master
- Locale: `en_US.UTF-8`

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

On UI draw, check if the option `update_icon_title` is true in addition to the previous `update_title`, and if either is set, continue generating the escape sequences. The tuples that contain the required escape sequences are stored in a list and a for loop is used to print them to stdout.

#### MOTIVATION AND CONTEXT

When using a window manager that shows a taskbar, it can be useful to see what directory ranger is in when the window titlebar is not visible.

#### TESTING

Ranger was run from `ranger.py` after I added `set update_icon_title true` to `rc.conf`.


siikamiika:
I tried to find one, and found a comment saying it's not supported in NeoMutt source:
https://neomutt.org/code/terminal_8c.html#a887af9a8f3e9ab92c7d065fbe9416003

siikamiika:
Should the comment about the reasoning behind hardcoding the string be added to this PR too? I'm also wondering if it would be better to store it in a constant with a more expressive name, but I'm not sure about a proper place for storing the constant

toonn:
That's a good idea. I think `ui.py` is an "ok" place for the constant.

siikamiika:
Remembered that this existed and synced with master, are there things needed before approve?

toonn:
Yes, if you could send over a pint of extra time, that'd be stellar ; )
(No action needed on your part currently, just have to review and test.)

siikamiika:
Sure, have all the time you need :beer: :hourglass: :smile:

PR #1649, opened 2019-08-02
Warn when nesting ranger

### LINKED ISSUE #1645: Warn when RNGR_LEVEL is greater than 0 at startup
Some users accidentally nest ranger sessions after forgetting they're in a subshell spawned by ranger, `S`. We can't kill this parent subshell when ranger is started nested but we *can* warn we're nesting.

Implement a boolean setting to turn off the warning but warn about a non-zero `RNGR_LEVEL` at startup.

toonn:
Users frequently request that launching ranger in a subshell started
from ranger (using `S`) instead exits the subshell so they end up in
their original ranger instance without nesting, which is rarely useful.
This isn't possible because a process can't easily kill its parent
shell.

To at least avoid such confusion we warn about nesting by default.
A new setting `nested_ranger_warning` is added to `rc.conf`. The warning
can be either disabled or the severity increased so the message is more
visible.

Fixes #1645

I think we could also add a `quit` option that would immediately stop ranger if it's started nested. Not sure if we want to do this though.

Another thing we could change, increase the default setting to `bad` so the warning's red and stands out more.

I added some variations on "true/false" to be lenient in what we accept but we don't really do that for the other options so maybe we should just drop it?


toonn:
Oh, need to document this in the manpage.

Vifon:
> I added some variations on "true/false" to be lenient in what we accept but we don't really do that for the other options so maybe we should just drop it?

Let's drop it. I'd rather see it as some more generic abstraction over the settings values in the future. Let's not do this at this level.

Additionally let's replace `bad` with `error`, it's more clear what it does this way.

ranger/core/main.py
```diff
@@ -180,6 +180,16 @@ def main(
             for command in args.cmd:
                 fm.execute_console(command)
 
+        if int(os.environ[level]) > 1:
+            warning = 'Warning:'
+            nested_warning = "You're in a nested ranger instance!"
+            nrw = fm.settings.nested_ranger_warning.lower()
```
Vifon: Let's name this variable `warning_level` or actually almost anything that exists in a dictionary…

PR #1651, opened 2019-08-03
Make the wiki more visible

Zaeph:
Hello,

The goal of this PR is to make the wiki more visible.

To do this, I've added a 'Going further' section at the end of `README.md` which mentions the [Official User Guide](https://github.com/ranger/ranger/wiki/Official-User-Guide), the [FAQ](https://github.com/ranger/ranger/wiki/FAQ%3A-Frequently-Asked-Questions), and the [wiki](https://github.com/ranger/ranger/wiki).  The FAQ isn't too furnished, but having it on `README.md` might incentivise us to improve it.

This was prompted by two things:
1. Based on my experience, people do not have the habit of checking out the wiki on GitHub pages, especially when they're not used to GitHub.
2. People often ask us questions about the image-previews over at `#ranger`, even though we're addressing the common issues in the wiki.  This seems to indicate that those wiki articles do not have a good discoverability.

Also, please note that I've changed the capitalisation on a couple of headings in 
8cf531a.

HTH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,4 +154,4 @@ Community
 ---------------
 For help, support, or if you just want to hang out with us, you can find us here:
 * **IRC**: channel **#ranger** on [freenode](https://freenode.net/kb/answer/chat)
-* **Subreddit**: [r/ranger](https://www.reddit.com/r/ranger/)
+* **Reddit**: [r/ranger](https://www.reddit.com/r/ranger/)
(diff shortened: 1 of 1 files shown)


toonn:
Like it, would it be appropriate to include the irc channel and the subreddit in this section? Oh, I see you already thought so : )

Zaeph:
Yup, but I owe the idea to vifon. :)

toonn:
Yes, this was a ninja review, misdirection to keep you on your toes : >

Zaeph:
I don't have a particular rationale for going for the minimal capitalisation.  Even though a lot of English native-speakers tend to capitalise most words in their title, from my experience, it's not as frequent in academia.

I think they already stand out enough as it is, but if you'd prefer more capitals, I can fix that quickly.

toonn:
Yep, let's capitalize all the things!

PR #1652, opened 2019-08-03
fixes #1566

### LINKED ISSUE #1566: :setlocal does not work on some directories
#### Runtime Environment

- Ranger version/commit: ranger-master 1.9.2

#### Current Behavior
`:setlocal` command does not change settings if directory name (or path) contains a character from the list: `()[]{}$^,\';:"`.

#### Steps to reproduce
1. Create a directory `:mkdir test(1)` and some files inside or rename any other directory by adding a character from the list: `()[]{}$^,\';:"`.
2. Enter the created/renamed directory.
3. The `:setlocal sort_reverse True` command does not sort files and subdirectories. Any other settings also have no effect.

user062:
now _setlocal_ works with directories whose name has regex metachars.

### DIFF
--- a/ranger/container/settings.py
+++ b/ranger/container/settings.py
@@ -276,7 +276,7 @@ class Settings(SignalDispatcher, FileManagerAware):
         if path:
             if path not in self._localsettings:
                 try:
-                    regex = re.compile(path)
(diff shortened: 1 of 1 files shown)


user062:
Thanks!

PR #1653, opened 2019-08-03
Add logo at the top of README.md

Zaeph:
Hi,

Since hut has added the logo to [github.io](https://ranger.github.io/) page, we might as well add it to the top of our `README.md`.

Note that I'm using the same version that hut picked, which isn't the same as the one that is in `ranger-assets`.  This is why I'm using an external URL rather than adding the logo to the repo.  I'd rather hold off until we have a final version for the logo.

HTH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 ranger 1.9.2
 ============
 
+![logo](https://ranger.github.io/ranger_logo.png)
+
 [![Build Status](https://travis-ci.org/ranger/ranger.svg?branch=master)](https://travis-ci.org/ranger/ranger)
 <a href="https://repology.org/metapackage/ranger/versions">
   <img src="https://repology.org/badge/latest-versions/ranger.svg" alt="latest packaged version(s)">
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ranger 1.9.2
 ============
 
-![logo](https://ranger.github.io/ranger_logo.png)
+<img src="https://ranger.github.io/ranger_logo.png" width="150">
 
 [![Build Status](https://travis-ci.org/ranger/ranger.svg?branch=master)](https://travis-ci.org/ranger/ranger)
 <a href="https://repology.org/metapackage/ranger/versions">


toonn:
Thing about logos is they stick. You never really get rid of the association. And it's not like logos are essential for TUI programs. For stickers, yes.

Zaeph:
Your call.  It could have been a nice placeholder until we can get the *Logo of Our Dreams™*, but I understand your worry about having it stick.  I could make a case for the potential similarity between the current version and the final version, but since it's all just cosmetics, we might as well wait.

Vifon:
> Hmm, I'm not too sure of this. Mostly because vifon didn't like the logo afair and then I didn't like the version where it was clearly an r.

LIES! I liked these logos a lot!

Vifon:
As far as I'm concerned, let's add the logo used on our website to our readme since it's *already* associated with ranger one way or another.

hut:
We can wait for the perfect logo, but the project is over 10 years old now. Let's just choose something until the perfect logo will appear.

And it doesn't have to be this one forever. We had the logo with the black/white face with the ranger hat, and it was not a problem to change that either. And we can tweak minor things later on.

That said, I think the logo is really nice. It looks ok, is unique, fits great on stickers, scales up and down fairly well, and has the connection to the vim logo.

Let's go for it.

hut:
A couple more things:

* @Zaeph, many thanks for the final design of the logo :)
* Wouldn't it be prettier if the logo shows up after the CI info tho?

Zaeph:
Hello,

>     * @Zaeph, many thanks for the final design of the logo :)

My pleasure, glad to see my work on the homepage!

>     * Wouldn't it be prettier if the logo shows up after the CI info tho?

I originally looked at the `README.md` for [Tomb](https://github.com/dyne/Tomb), since I remembered that I liked that layout.  I've just tried it in another branch, and you're right, I think it looks better with the logo *after* the CI info.

I'll make a quick PR to adjust it.

PR #1656, opened 2019-08-05
Apply debian patch 0004 desktop entry lacks keywords

toonn:
One of the Debian patches our `ranger.desktop` to include `Keywords`. I
figure if we apply this downstream need not bother and many people can
benefit.


toonn:
No, forgot a comma, AMENDING!
"if we apply this, downstream need not bother"

PR #1658, opened 2019-08-06
Move logo after CI info

zaeph:
Hello,

This is a quick follow-up PR to #1653.

It moves the logo *after* the CI info.

HTH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 ranger 1.9.2
 ============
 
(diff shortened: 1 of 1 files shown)


toonn:
Hmm, I kinda appreciated the seperation from the text. Now it looks a little like a badly typeset image in word imo.

How about above the title as I'd suggested before? Or a rule or heading between logo and text? Or flowing it into the text like caligraphic first letters in manuscripts?

PR #1661, opened 2019-08-07
Fix fluke command line & status bar position error

### LINKED ISSUE #1646: After resizing ranger to 1 line height, status and command line never again find correct position
After having resized ranger to a height of a single line (which regularly happens due to my use of tmux with heavy pane shuffling), the status bar and command line never go back to their position at the bottom causing drawing glitches (as they are drawn at position 0,0 just as the title bar)..
I tried to trace it with winpdb and .. got that to crash, tOo 😅
Seems within `gui/displayable.py:resize()`, `self.win.mvderwin(y, x)` fails every time after having resized so small, and `self.win.getparyx()` always returns `0,0` ...

eMPee584:
When setting f.e. status bar position fails, it will be
retried after triggering a same-size resize event.

Fixes #1646

#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
- Operating system and version: Debian GNU/Linux bullseye/sid
- Terminal emulator and version: guake 3.4.0-1 and tmux 2.9a-2
- Python version: 3.7.4 [GCC 8.3.0]
- Ranger version/commit: 1.9.2-4
- Locale: en_GB.UTF-8


toonn:
Can we be sure one level deeper is all we need?

eMPee584:
valid question :thinking:

eMPee584:
> Can we be sure one level deeper is all we need?

(it's magnitudes better than before with that only even…)

eMPee584:
> Can we be sure one level deeper is all we need?
so I guess we can't, but I haven't encountered this to cause a problem..

PR #1662, opened 2019-08-08
Move logo to the very top

hut:
This is another PR follow-up to #1653.

Here, the logo is at the very top, and enlarged by 33%.

Compare it with #1658 and #1659 to see which one your prefer.


toonn:
My preference is for centering.

PR #1666, opened 2019-08-15
add support for 3D files including STL and DXF

sigboe:
This Pull Request leverages the free OpenSCAD software (available in a repository near you) to generate preview images for 3D model files. The two main file formats that most people will use are STL and DXF. And also include off, scad and csg. 

image

### DIFF
--- a/doc/ranger.pod
+++ b/doc/ranger.pod
@@ -1974,6 +1974,17 @@ I<highlight> will pick up command line options specified in this variable. A
 C<--style=> option specified here will override C<HIGHLIGHT_STYLE>. Similarly,
 C<--replace-tabs=> will override C<HIGHLIGHT_TABWIDTH>.
 
+=item OPENSCAD_COLORSCHEME
+
+Specifies the colorscheme used by I<openscad> while previewing 3D models. Read
+I<openscad> man page for colorschemes. Ranger will default to Tomorrow Night.
+
+=item OPENSCAD_IMGSIZE
+
+Specifies the internal resolution I<openscad> will use for rendering 3D models.
+The image will be downscaled to fit the preview pane. This resolution will
+default to "1000,1000" if no value is set.
+
 =item XDG_CONFIG_HOME
 
 Specifies the directory for configuration files. Defaults to F<$HOME/.config>.
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -67,8 +67,10 @@ handle_extension() {
         ## PDF
         pdf)
             ## Preview as text conversion
-            pdftotext -l 10 -nopgbrk -q -- "${FILE_PATH}" - | fmt -w "${PV_WIDTH}" && exit 5
-            mutool draw -F txt -i -- "${FILE_PATH}" 1-10 | fmt -w "${PV_WIDTH}" && exit 5
+            pdftotext -l 10 -nopgbrk -q -- "${FILE_PATH}" - | \
+              fmt -w "${PV_WIDTH}" && exit 5
+            mutool draw -F txt -i -- "${FILE_PATH}" 1-10 | \
+              fmt -w "${PV_WIDTH}" && exit 5
             exiftool "${FILE_PATH}" && exit 5
             exit 1;;
 
@@ -95,28 +97,6 @@ handle_extension() {
         json)
             jq --color-output . "${FILE_PATH}" && exit 5
             python -m json.tool -- "${FILE_PATH}" && exit 5
-            ;;
-
-        ## 3D models
-        ## OpenSCAD only supports png image output, and ${IMAGE_CACHE_PATH} is
-        ## hardcoded as jpeg. So we make a tempfile.png and just move/rename it
-        ## to jpg This works because image library is smart enough to handle it
-        stl|off|dxf)
-            [[ "${PV_IMAGE_ENABLED}" != 'True' ]] && exit 1
-            openscad --colorscheme="${OPENSCAD_COLORSCHEME}" \
-                    --imgsize="${OPENSCAD_IMGSIZE/x/,}" \
-                    -o "/tmp/$(basename "${FILE_PATH}").png" \
-                    <(echo "import(\"${FILE_PATH}\");")
-            mv "/tmp/$(basename "${FILE_PATH}").png" "${IMAGE_CACHE_PATH}" \
-                    && exit 6
-            ;;
-        scad|csg)
-            [[ "${PV_IMAGE_ENABLED}" != 'True' ]] && exit 1
-            openscad --colorscheme="${OPENSCAD_COLORSCHEME}" \
-                    --imgsize="${OPENSCAD_IMGSIZE/x/,}" \
-                    -o "/tmp/$(basename "${FILE_PATH}").png" "${FILE_PATH}"
-            mv "/tmp/$(basename "${FILE_PATH}").png" "${IMAGE_CACHE_PATH}" \
-                    && exit 6
             ;;
 
         ## Direct Stream Digital/Transfer (DSDIFF)
@@ -247,6 +227,28 @@ handle_image() {
         #     [ "$rar" ] || [ "$zip" ] && rm -- "${IMAGE_CACHE_PATH}"
         #     ;;
     esac
+
+    # openscad_image() {
+    #     TMPPNG="$(mktemp -t XXXXXX.png)"
+    #     openscad --colorscheme="${OPENSCAD_COLORSCHEME}" \
+    #         --imgsize="${OPENSCAD_IMGSIZE/x/,}" \
+    #         -o "${TMPPNG}" "${1}"
+    #     mv "${TMPPNG}" "${IMAGE_CACHE_PATH}"
+    # }
+
+    # case "${FILE_EXTENSION_LOWER}" in
+    #     ## 3D models
+    #     ## OpenSCAD only supports png image output, and ${IMAGE_CACHE_PATH}
+    #     ## is hardcoded as jpeg. So we make a tempfile.png and just
+    #     ## move/rename it to jpg. This works because image libraries are
+    #     ## smart enough to handle it.
+    #     csg|scad)
+    #         openscad_image "${FILE_PATH}" && exit 6
+    #         ;;
+    #     3mf|amf|dxf|off|stl)
+    #         openscad_image <(echo "import(\"${FILE_PATH}\");") && exit 6
+    #         ;;
+    # esac
 }
 
 handle_mime() {
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -44,6 +44,8 @@ HIGHLIGHT_TABWIDTH=${HIGHLIGHT_TABWIDTH:-8}
 HIGHLIGHT_STYLE=${HIGHLIGHT_STYLE:-pablo}
 HIGHLIGHT_OPTIONS="--replace-tabs=${HIGHLIGHT_TABWIDTH} --style=${HIGHLIGHT_STYLE} ${HIGHLIGHT_OPTIONS:-}"
 PYGMENTIZE_STYLE=${PYGMENTIZE_STYLE:-autumn}
+OPENSCAD_IMGSIZE=${RNGR_OPENSCAD_IMGSIZE:-1000,1000}
+OPENSCAD_COLORSCHEME=${RNGR_OPENSCAD_COLORSCHEME:-Tomorrow Night}
 
 handle_extension() {
     case "${FILE_EXTENSION_LOWER}" in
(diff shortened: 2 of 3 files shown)


ranger/data/scope.sh
```diff
@@ -94,6 +94,18 @@ handle_extension() {
             jq --color-output . "${FILE_PATH}" && exit 5
             python -m json.tool -- "${FILE_PATH}" && exit 5
             ;;
+	## 3D models
+	stl|off|dxf)
+	    [[ "${PV_IMAGE_ENABLED}" == 'True' ]] && \
+            openscad --colorscheme="Tomorrow Night" --imgsize="1000,1000" -o "/tmp/stl.png" <(echo "import(\"${FILE_PATH}\");"); mv "/tmp/stl.png" "${IMAGE_CACHE_PATH}" && \
```
toonn: Hmm, using `stl.png` for any and all previews will wreck the cache. Instead, base the name for the cached image on the path to the file being previewed.

sigboe: @toonn fixed, I now use `"/tmp/$(basename "${FILE_PATH}").png"`

ranger/data/scope.sh
```diff
@@ -94,6 +94,20 @@ handle_extension() {
...
+	stl|off|dxf)
+	    [[ "${PV_IMAGE_ENABLED}" == 'True' ]] && \
+		    openscad --colorscheme="Tomorrow Night" --imgsize="1000,1000" -o "/tmp/$(basename "${FILE_PATH}").png" \
+	       <(echo "import(\"${FILE_PATH}\");")
+	    mv "/tmp/$(basename "${FILE_PATH}").png" "${IMAGE_CACHE_PATH}" && \
+	    exit 6 
+	    ;; 
+        scad|csg)
+	    [[ "${PV_IMAGE_ENABLED}" == 'True' ]] && \
+            openscad --colorscheme="Tomorrow Night" --imgsize="1000,1000" -o "/tmp/stl.png" ${FILE_PATH}; mv "/tmp/stl.png" "${IMAGE_CACHE_PATH}" && \
```
toonn: Same issue here.

sigboe: I fixed it and squashed so there is not two seperate ops commits, or do you want me to squash it even further.

Openscad does not support exporting to Jpeg, and openscad needs the output file to be .png (only image format).

I just googled for example files, but STL is most important for me. Here are some open source stl files https://www.prusa3d.com/prusa-i3-printable-parts/ there are loads of more on thingiverse.

Vifon:
> @Vifon, our cache only handles jpegs currently, right? And, does moving the previews in there actually help or will the preview code just cache them as jpegs again?

Our cache expects a file with an `.jpg` extension while, according to what @sigboe said on IRC, OpenSCAD cannot work with JPEG but can with PNG. So essentially it puts there a PNG file and renames it as `.jpg`. It works but is essentially a hack.

sigboe:
@toonn the mv is required, because openscad requires the png name in the -o, unless I can opt to `${IMAGE_CACHE_PATH}` being png.

toonn:
Ok, then it's a fine workaround for a shortcoming in openscad. So two things left:
  1. Use DEFAULT_SIZE, entirely or in part.
  2. Make the styling align with our handling of both highlight and pygments.

sigboe:
@toonn ${DEFAULT_SIZE} is local to handle_image() while I put this handling in handle_extension() to avoid dealing with immature mimetypes of 3d model files.

We can move ${DEFAULT_SIZE} out of handle_image(), or I can catch it by mimetype which will not be accurate.

> Make the styling align with our handling of both highlight and pygments.

I'll certainly do this in the evening. Just need to browse around and figure out were these are set.

toonn:
Ugh, you're right. #iwishscopewasposix
So that leaves just the style thing, and actually, probably give the imgsize the same treatment. So just declare them as variables that inherit the value from the environment in the `## Settings` section.

Vifon:
And comments. Please add some comment describing our little hack.

sigboe:
@toonn okay ready for review again, tell me if you want more changes or if I misunderstood anything.

Also added the openscad colorscheme out to ## Settings

ranger/data/scope.sh
```diff
@@ -44,7 +44,8 @@ HIGHLIGHT_TABWIDTH=${HIGHLIGHT_TABWIDTH:-8}
 HIGHLIGHT_STYLE=${HIGHLIGHT_STYLE:-pablo}
 HIGHLIGHT_OPTIONS="--replace-tabs=${HIGHLIGHT_TABWIDTH} --style=${HIGHLIGHT_STYLE} ${HIGHLIGHT_OPTIONS:-}"
 PYGMENTIZE_STYLE=${PYGMENTIZE_STYLE:-autumn}
-
+RNGR_OPENSCAD_IMGSIZE=${RNGR_OPENSCAD_IMGSIZE:-1000,1000}
```
toonn: Maybe we can drop the `RNGR` prefix? I think the vars are specific enough. Would be ideal if openscad already read some variables for this, then we could reuse those.

ranger/data/scope.sh
```diff
@@ -44,7 +44,8 @@ HIGHLIGHT_TABWIDTH=${HIGHLIGHT_TABWIDTH:-8}
...
+RNGR_OPENSCAD_COLORSCHEME=${RNGR_OPENSCAD_COLORSCHEME:-Tomorrow Night}	#("Cornfield" "Metallic" "Sunset" "Starnight" "BeforeDawn" "Nature" "DeepOcean" "Solarized" "Tomorrow" "Tomorrow 2" "Tomorrow Night" "Monotone")
```
toonn: I understand wanting to document the options but we decided against this last time. [See here how we did it instead](https://github.com/ranger/ranger/blame/master/doc/ranger.pod#L1820).

sigboe: removed

ranger/data/scope.sh
sigboe: done

sigboe:
@toonn comment revised.

toonn:
@sigboe, make sure *all* the lines you're adding fit in the 80 character limit.

sigboe:
@toonn There, followed the new styleguide. Please say if there are any other style changes or other changes you want.

sigboe:
@toonn as per request in IRC i have updated the documentation to reflect the changes made. Please review those.

I also changed the `${OPENSCAD_IMGSIZE/x/,}` used in the openscad command to substitute `x` for `,` in case a user tries to set the sensible format which openscad doesn't accept.

ranger/data/scope.sh
```diff
@@ -94,6 +95,28 @@ handle_extension() {
             jq --color-output . "${FILE_PATH}" && exit 5
             python -m json.tool -- "${FILE_PATH}" && exit 5
             ;;
+	## 3D models
+	## OpenSCAD only supports png image output, and ${IMAGE_CACHE_PATH} is
+        ## hardcoded as jpeg. So we make a tempfile.png and just move/rename it 
+	## to jpg This works because image library is smart enough to handle it
```
Vifon: A very minor nitpick: There is a missing period after "jpg" and before "This works".

toonn: Good spot. Another is that the line above that is indented more.

doc/ranger.pod
```diff
@@ -192,7 +192,8 @@ scripts by setting the option C<use_preview_script> and C<preview_files> to true
 This default script is F<%rangerdir/data/scope.sh>. It contains more
 documentation and calls to the programs I<lynx> and I<elinks> for html,
 I<highlight> for text/code, I<img2txt> for images, I<atool> for archives,
-I<pdftotext> or I<mutool> for PDFs and I<mediainfo> for video and audio files.
+I<pdftotext> or I<mutool> for PDFs, or I<mediainfo> for video and audio files
+and I<openscad> for 3D models.
```
toonn: Hmm, the listing here isn't complete anyway. I'm not sure we should add openscad since it's relatively obscure.

doc/ranger.pod
```diff
@@ -1840,6 +1841,19 @@ I<highlight> will pick up command line options specified in this variable. A
 C<--style=> option specified here will override C<HIGHLIGHT_STYLE>. Similarly,
 C<--replace-tabs=> will override C<HIGHLIGHT_TABWIDTH>.
 
+=item OPENSCAD_COLORSCHEME
+
+Specifies the colorscheme used by I<openscad> while previewing 3d models. The
+available colorschemes are. Cornfield, Metallic, Sunset, Starnight, BeforeDawn,
```
toonn: Please list instructions for how to find out the options rather than listing the options. That way there's nothing to get out of date.

doc/ranger.pod
```diff
@@ -1840,6 +1841,19 @@ I<highlight> will pick up command line options specified in this variable. A
 C<--style=> option specified here will override C<HIGHLIGHT_STYLE>. Similarly,
 C<--replace-tabs=> will override C<HIGHLIGHT_TABWIDTH>.
 
+=item OPENSCAD_COLORSCHEME
+
+Specifies the colorscheme used by I<openscad> while previewing 3d models. The
```
toonn: I think 3D should be capitalized everywhere.

ranger/data/scope.sh
toonn: Hmm, looking at it in another diff view it doesn't look indented differently. Probably because of tabs. Make sure all the indentation is spaces.

doc/ranger.pod
sigboe: I agree, funny thing is that the man page of openscad has an outdated list. Not our problem though.

ranger/data/scope.sh
sigboe: `/%s/\t/        /g` fixed this good, as I was the only one who put dirty tabs into the file.

doc/ranger.pod
sigboe: `%s/3d/3D` done

doc/ranger.pod
sigboe: `%s/3d/3D` done

doc/ranger.pod
sigboe: Well, need a decision here. If its up to me, I'd list it. But you decide :smile:

doc/ranger.pod
toonn: Let's not make it our problem then. Say, "check the openscad manpage" or "check this or that wiki page," whatever you did to find them out, rather than listing things that will get outdated.

doc/ranger.pod
toonn: I'd go with no unless @Vifon feels differently.

doc/ranger.pod
```diff
@@ -1840,6 +1841,17 @@ I<highlight> will pick up command line options specified in this variable. A
...
+=item OPENSCAD_COLORSCHEME
+
+Specifies the colorscheme used by I<openscad> while previewing 3D models. Read
+I<openscad> man page for colorschemes. Ranger will default to Tomorrow Night.
+
+=item OPENSCAD_IMGSIZE
+
+Specifies the internal resolution I<openscad> will use for rendering 3D models.
+The image will be downscaled to fit the preview pane. This resolution will
+default to 1000,1000 if no value is set.
```
toonn: Maybe put the value in quotes, I think we do double quotes to indicate literals.

doc/ranger.pod
Vifon: It might be good for "marketing". Showing neat things ranger can do won't hurt. Let's go for it.

toonn: @Vifon, does this mean you'd like us to put *everything* up there? Marketing's for the README imo, this is the manpage, people would've already installed ranger by this point, no? It just seems little silly to make it harder to find how to get openscad previews as more mentions means manpage search shows more irrelevant results.

Vifon: Ah, it's a manpage? Still, previewing 3D models in particular is pretty neat. Definitely not everything but I'd include this one.

Vifon: OK, let's move it to the README.

toonn: @Vifon it already was/is in the README : >

ranger/data/scope.sh
```diff
@@ -94,6 +96,28 @@ handle_extension() {
             jq --color-output . "${FILE_PATH}" && exit 5
             python -m json.tool -- "${FILE_PATH}" && exit 5
             ;;
+        ## 3D models
+        ## OpenSCAD only supports png image output, and ${IMAGE_CACHE_PATH} is
+        ## hardcoded as jpeg. So we make a tempfile.png and just move/rename it 
+        ## to jpg This works because image library is smart enough to handle it
```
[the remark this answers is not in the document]
toonn: Oh, and let's comment out stl and scad/csg by default. File I tried took *ages* to preview.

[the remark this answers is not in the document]
sigboe: I think I used the OpenSCAD GUI to export a file as csg, and it worked. 

I'll add the periods in. 

What do you mean the image block?

Yeah I can comment the previews out by default.

I'll do the requested changes, today, or within a few days, I am traveling now, don't have much time on my laptop :slightly_smiling_face:

PR #1680, opened 2019-08-31
Make setlocal path=... argument a regex again

jakanaka-eva:
#### ISSUE TYPE

- Bug fix

#### CHECKLIST

    - [ ] Tests have been updated

From the `setlocal` section of the man page:
```
path is a regular expression.  This means that "path=~/dl" applies to
all paths that start with ~/dl, e.g. ~/dl2 and ~/dl/foo. To avoid
this, use "path=~/dl$".

```
#1652 caused the `path` argument to be regex escaped. This commit
fixes that.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -506,7 +506,7 @@ class setlocal(set_):
         if path is None:
             path = self._re_shift(self.PATH_RE_UNQUOTED.match(self.arg(1)))
         if path is None and self.fm.thisdir:
-            path = self.fm.thisdir.path
+            path = "^" + re.escape(self.fm.thisdir.path) + "$"
         if not path:
             return
 
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -507,13 +507,13 @@ class setlocal_(set_):
         # We require quoting of paths with whitespace so we have to take care
         # not to match escaped quotes.
         self.path_re_dquoted = re.compile(
-            r'^set.+?\s+{arg}="(.*?[^\\])"'.format(arg=self._arg())
+            r'^set.+?\s+{arg}="(.*?[^\\])"'.format(arg=self._arg)
         )
         self.path_re_squoted = re.compile(
-            r"^set.+?\s+{arg}='(.*?[^\\])'".format(arg=self._arg())
+            r"^set.+?\s+{arg}='(.*?[^\\])'".format(arg=self._arg)
         )
         self.path_re_unquoted = re.compile(
-            r'^{arg}=(.+?)$'.format(arg=self._arg())
+            r'^{arg}=(.+?)$'.format(arg=self._arg)
         )
 
     def _re_shift(self, match):
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -482,36 +482,93 @@ class set_(Command):
         return None
 
 
-class setlocal(set_):
-    """:setlocal path=<regular expression> <option name>=<python expression>
+class _setlocal(set_):
+    """Shared class for setinpath and setinregex
 
-    Gives an option a new value.
+    By implementing the _arg abstract propery you can affect what the name of
+    the pattern/path/regex argument can be, this should be a regular expression
+    without match groups.
+
+    By implementing the _format_arg abstract method you can affect how the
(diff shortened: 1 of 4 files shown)


ranger/config/commands.py
```diff
@@ -502,7 +502,7 @@ def execute(self):
         if path is None:
             path = self._re_shift(self.PATH_RE_UNQUOTED.match(self.arg(1)))
         if path is None and self.fm.thisdir:
-            path = self.fm.thisdir.path
+            path = "^" + re.escape(self.fm.thisdir.path)
```
toonn: Let's add the "$" too.

toonn:
Actually, as far as UX goes I think allowing paths with regex special characters is important. So I suggest adding an `re = None` default argument, if it's specified then it overrules path, if it's not, path's used literally.

ranger/config/commands.py
jakanaka-eva: Done. However this is a slightly breaking change, since `setlocal` has
always included all subdirectories in contrast to what the man page
says.

jakanaka-eva:
I imagine that some people (well at least me) have `setlocal`s in
their `rc.conf`, expecting the `path` argumet to behave like a regex,
so I resorted to just editing the man page.

toonn:
I know this'd be a backwards incompatible change but imo the UX improvement is important.

@vifon, care to weigh in?

s3rb31:
This bug still exits. The current manpage doesn't reflect the programs functionality correctly.

The code should be fixed to allow regex in the path as the manpage currently states, while regex metachars should have to be escaped (preserving backwards compatibility) or the manpage should be updated to correctly reflect the fact that the `path` argument to `setlocal` is NOT a regular expression anymore (breaking backwards compatibility) and thus everybody who has this in their `rc.conf` is now screwed.

As @toonn mentioned earlier IF the "UX improvement" is deemed so important to warrant configuration breaking changes, what should have been done in #1652 is to implement it via a _different_ option, e. g. `re_path` or `re=False` (per default).

Personally I don't think enabling the use of regex metachars in setlocal paths is a big enough improvement to justify configration breaking changes, but thats just my opinion.

toonn:
Thank you for pinging. This slipped through the cracks!

vifon:
In my opinion the sane default would be to treat it as a plain string and having it as regexp as optional. Backwards compatibility is important, but sometimes it needs to be sacrificed for the better UX.

toonn:
Ok, so we're in agreement. @jakanaka-eva, are you interested in picking this up again? Otherwise I'll take over, because the current situation is no good.

toonn:
@Euren, @manujchandra, I have implemented a `setinpath` with `setlocal`'s current behavior and a `setinregex` with `setlocal`'s old behavior, it would be great if you could test this PR since we don't want to introduce even more breakage. : )

Euren:
Having made no change to my config, running with --debug crashes with
`ValueError: The option named 'false' does not exist`
when evaluating the setting   
`setlocal path=/mnt/nas preview_files false`                                                 

I think the change to shifting is having an off-by-one interaction with `set_option_from_string`

toonn:
Can you try the other commands and quoting too?

Euren:
The newly added commands also fail in the same way and quoting makes no difference.
Digging a bit, this is `[self.arg(n)]` before and after parsing arg in `set_local_.execute()`:

Before: `['setlocal', 'path=/mnt/nas', 'preview_files', 'false']`
After: `['preview_files', 'false']`

and `parse_setting_line()` grabs `self.arg(1)`. Compared to the behavior in master, we are shifting an additional time.

toonn:
Thanks for the digging. It should be an easy fix. I do wonder why I was experiencing the off-by-one error during testing unless I added it though 😕

toonn:
I think my fix is simpler than the `range(len(...` stuff. I'm still clueless as to why I needed the extra `1 +` increment though but I can still reproduce it.

Euren:
Works well on my end now 👍

toonn (approved):
LGTM

manujchandra:
Let me review and revert.

manujchandra:
I have tested it. It seems to be working 👍🏽

PR #1683, opened 2019-09-01
Add example to copymap manpage entry

### LINKED ISSUE #1678: copymap command -NOT AN ISSUE- More like Help~
Hello there everyone!
I am going deeper in understanding Ranger so expect me often here.
I found this copymap command:

> 
> **copycmap key newkey [newkey2 ...]**
> See copymap
> 
> **copymap key newkey [newkey2 ...]**
> Copies the keybinding key to newkey in the “browser” context.  This is a deep copy, so if you change the new binding (or parts of it) later, the old one is not modified.
> 
> To copy key bindings of the console, taskview, or pager use “copycmap”, “copytmap” or “copypmap”.
> 
> **copypmap key newkey [newkey2 ...]**
> See copymap
> 
> **copytmap key newkey [newkey2 ...]**
> See copymap
> 
I notice the `See copymap ` pointer but i dont find anything on it yet.. what does this command exacly mean?

Sorry for create an issue for this question maybe will be a good idea to have a place for ranger users questions and guides!

toonn:
Fixes #1678


Vifon:
Do you think we should do the same thing with `:cunmap` and its family of commands?

toonn:
Yes, I'll add it to the PR tomorrow.

toonn:
Hmm, something I thought of. Maybe we should include all the commands in the `=item` heading, something like:
```
=item map [I<keys> ...]
=item cmap [I<keys> ...]
=item pmap [I<keys> ...]

Body of the explanation.
```
Not sure how that works with pod formatting?

toonn:
@Vifon, I added my multiple section header suggestion. What do you think? Alternatives would be dropping the `key newkey...` from the variations, shortening the variants to `cmap/pmap/tmap` or dropping these section headers altogether like it was before.

Vifon:
On second thought, I don't think that's valid POD-wise, though if it formats correctly, I'll allow it.

toonn:
It seems valid syntax to me. It just gives us some empty sections whose headers happen to be formatted one right under the other. My goal is to make it really clear when people are searching that those sections apply to all the variants.

PR #1685, opened 2019-09-02
Update colorscheme creation method

parmort:
Continuation of PR #1229. Sorry that this is a year later.
For those who don't know, this is going to make defining colorschemes easier.

With the proposed method, it goes from a (no offense) dirty interface to a cleaner one, imo:

```python
from ranger.colorschemes.default import Default, color

# Must be `Scheme` because of how ranger detects colorschemes
class Scheme(Default):
    colors = {
        'browser': {
            'selected': color(bg=red)
        }
    }
```

#### ISSUE TYPE

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
Instead of overriding the entire use method directly for even the smallest changes, like changing the color of something, one can add keys to the `colors` dictionary, which makes life a lot easier. So as not to break the existing schemes, I modified the default scheme (which looks the same) and left the original colorscheme class alone, which @toonn suggested in #1229.

#### MOTIVATION AND CONTEXT
I went to create my own colorscheme and found a poorly documented (for a newcomer) mess of if statements and contexts. I just wanted to switch around a few colors.

### DIFF
--- a/ranger/colorschemes/default.py
+++ b/ranger/colorschemes/default.py
@@ -9,23 +9,23 @@ from ranger.gui.color import (
     normal, bold, reverse, dim, BRIGHT,
     default_colors,
 )
-from ranger.core.actions import Actions
-
 
 def color(fg=-1, bg=-1, attr=0):
     return (fg, bg, attr)
 
-def getkey(d, key):
+
+def getkey(dictionary, key):
     keys = key.split('.')
     code = ''
     for i in dir():
-        if eval(i) == d:
+        if eval(i) == dictionary:  # pylint: disable=eval-used
             code += i
 
     for k in keys:
         code += '["' + k + '"]'
 
-    return eval(code)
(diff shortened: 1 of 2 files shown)


toonn:
I don't understand why you're opening a new PR about this? What's different between this and #1229 and why did you PR it to ranger/colorschemes like I asked?

parmort:
I created a new PR is because the old one doesn't track the commits I'm making in my fork. I also started from scratch this time as well.

Correct me if I'm wrong, but it seems that this change would also get buried in the colorscheme repo because it is not advertised as well as it could be, it is a stagnant repo that hasn't had any new PRs in over a year, the issues that are there are about the ones packaged with ranger and not the ones included in the repo, and what I'm making isn't a colorscheme per se, but more of a library that by default implements the current default colorscheme.

If you do think that it is a good fit in the colorscheme repo instead, I'll revert the changes I made here, update the documentation, and open a PR over there.

PR #1687, opened 2019-09-03
Add GitHub Actions ci

toonn:
Run our python tests and shellcheck in seperate workflows only when
relevant files are changed, respectively python files or `scope.sh`.

This tests with one extra version of python, 3.7.

GitHub's version of shellcheck seems to be missing a few flags. So we
download the latest stable version, 0.7.0 currently.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,22 @@
+name: Shellcheck scope.sh
+
+on: 
+  push:
+    paths:
+      - '.github/workflows/shellcheck.yml'
+      - 'ranger/data/scope.sh'
+
+jobs:
+  test_shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: Install newer shellcheck (0.7.0 rather than 0.4.6)
+      run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,22 @@
+name: Shellcheck scope.sh
+
+on: 
+  push:
+    paths:
+      - '.github/workflows/shellcheck.yml'
+      - 'ranger/data/scope.sh'
```
Vifon: Let's prepend a slash to this path, so that we only match `ranger/data/scope.sh` and not `anything/ranger/data/scope.sh`. Mostly for the sake of correctness.

Vifon: That is, assuming it works like gitignore. If it doesn't, ignore my comment. :)

toonn: So, the GH Actions seem to use [this function](https://golang.org/pkg/path/#Match) for matching.

From its docs:
  > Match requires pattern to match all of name, not just a substring.

If I read this right that means it won't match anything but `ranger/data/scope.sh` so I think prefixing `/` wouldn't work anymore.

I just force-pushed a test to verify that nothing gets triggered, neither `ghcactions/ranger/data/scope.sh` nor `ghcactions-ranger/data/scope.sh` triggered the shellcheck action.

PR #872, opened 2017-05-22
Image displayer improvements

joe-rice:
Added ImageDisplayer registry, automatic selection and a new ImageDisplayer.

#### ISSUE TYPE
- Improvement/feature implementation

#### RUNTIME ENVIRONMENT
- Operating system and version: Arch Linux
- Terminal emulator and version: st 0.7
- Python version: 3.6.1
- Ranger version/commit: 1.8.1/10d7174613e9865b065c52e57af8b0799cd106b0
- Locale: en_us

#### CHECKLIST
- [X] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [X] All changes follow the code style **[REQUIRED]**
- [X] All new and existing tests pass **[REQUIRED]**
- [ ] Changes require config files to be updated
    - [ ] Config files have been updated
- [X] Changes require documentation to be updated
    - [ ] Documentation has been updated
- [ ] Changes require tests to be updated
    - [ ] Tests have been updated

#### DESCRIPTION
- commit b7af7b6fcd32e60141e64cb1162e8d1de74a45c6
I've added a registry for ImageDisplayers, to manage a growing number of sub-classes and to allow users to add their own sub-classes. This commit provides a register function and a metaclass for ImageDisplayer. Each ImageDisplayer can simply set METHOD_NAME to the desired key.
- commit e4005d10ba28a6a27644dad325619364d089229a
The user can now choose "set preview_method auto" to choose an ImageDisplayer that will run in the given environment. Simple heuristics are used, such as searching the parent process tree (optionally with psutils) for the currently used terminal. An arbitrary test ordering was used: iterm2 > w3m > urxvt.
- commit b05c6858ae016d90e4004f6b766d08c8c0b2a856
The user can now "set preview_method mpv" to use mpv as their ImageDisplayer. This uses mpv 0.25's socket IPC mechanism. The new "auto" order is: iterm2 > w3m > mpv > urxvt. More work on this will need to be done to configure mpv to not resize the window on media change.

#### MOTIVATION AND CONTEXT
To ease the management of many ImageDisplayers and allow users to easily implement their own. Also, to utilize a popular media viewer (mpv) that behaves differently from current ImageDisplayers.

Link to relevant issues:
MPV provides an alternative that may act as a satisfactory workaround for these issues: #864 #820 #856 #467

#### TESTING
What tests have been run?
I have run the included tests and have been using the mpv method myself.

How does the changes affect other areas of the codebase?
I don't think it will.

PR #1691, opened 2019-09-11
added registry for ImageDisplayer classes

joe-rice:
Added ImageDisplayer registry

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Terminal emulator and version: st 0.7
- Python version: 3.7.4
- Ranger version/commit: ranger-master v1.9.2-319-g3b8a3252/86281229606e147870b365d7aff445e394883674
- Locale: en_us

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
- commit 3b8a32522ebb392a29759ec814947db048592f2e
I've added a registry for ImageDisplayers, to manage a growing number of sub-classes and to allow users to add their own sub-classes. Each ImageDisplayer can register with a key by using the register_image_displayer class decorator. These classes can then be used by setting preview_images_method accordingly in their rc.conf.

For example, in commands.py:
```
from ranger.ext.img_display import register_image_displayer, ImageDisplayer

@register_image_displayer('my-custom-image-displayer')
class MyImageDisplayer(ImageDisplayer):
    def draw(self, path, start_x, start_y, width, height):
        import subprocess
        subprocess.check_call(['sxiv', path])
```

And in rc.conf:
```
set preview_images_method my-custom-image-displayer
```

#### MOTIVATION AND CONTEXT
To ease the management of many ImageDisplayers and allow users to easily implement their own.

Link to relevant issues:
Splits the registry portion of #872. Apologies for the asinine delay and thanks to @Vifon for splitting off the MPV portion of that old PR.

#### TESTING
What tests have been run?
I have run the included tests and have been using the registry myself with a few additional image preview methods (mostly various MPV configurations). Before submitting, I also re-ran it with w3m on urxvt.

How does the changes affect other areas of the codebase?
I don't think it will. However it should allow image displayers to be externalized.

### DIFF
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -22,7 +22,7 @@ from ranger.container.tags import Tags, TagsDummy
 from ranger.gui.ui import UI
 from ranger.container.bookmarks import Bookmarks
 from ranger.core.runner import Runner
-from ranger.ext import img_display
+from ranger.ext.img_display import get_image_displayer
 from ranger.core.metadata import MetadataManager
 from ranger.ext.rifle import Rifle
 from ranger.container.directory import Directory
@@ -100,7 +100,7 @@ class FM(Actions,  # pylint: disable=too-many-instance-attributes
         def set_image_displayer():
             if self.image_displayer:
                 self.image_displayer.quit()
-            self.image_displayer = self._get_image_displayer()
+            self.image_displayer = get_image_displayer(self.settings.preview_images_method)
         set_image_displayer()
         self.settings.signal_bind('setopt.preview_images_method', set_image_displayer,
                                   priority=settings.SIGNAL_PRIORITY_AFTER_SYNC)
@@ -222,11 +222,6 @@ class FM(Actions,  # pylint: disable=too-many-instance-attributes
         for entry in logutils.QUEUE:
             for line in entry.splitlines():
                 yield line
-
-    def _get_image_displayer(self):
-        registry = img_display.IMAGE_DISPLAYER_REGISTRY
-        cls = registry[self.settings.preview_images_method]
-        return cls()
 
     def _get_thisfile(self):
         return self.thistab.thisfile
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -22,11 +22,7 @@ from ranger.container.tags import Tags, TagsDummy
 from ranger.gui.ui import UI
 from ranger.container.bookmarks import Bookmarks
 from ranger.core.runner import Runner
-from ranger.ext.img_display import (W3MImageDisplayer, ITerm2ImageDisplayer,
-                                    TerminologyImageDisplayer,
-                                    URXVTImageDisplayer, URXVTImageFSDisplayer,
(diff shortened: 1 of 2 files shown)


toonn:
This looks pretty cool. I have still to test this. Would registering with multiple names be as simple as using the decorator multiple times, .e.g., to register `urxvt-full` as `urxvt-background` in addition or both `urxvt` and `rxvt-unicode`?

joe-rice:
@toonn Yea, stacking the decorator will associate each key with the same class.

ranger/core/fm.py
```diff
@@ -227,22 +223,10 @@ def get_log():
...
-        elif self.settings.preview_images_method == "urxvt-full":
-            return URXVTImageFSDisplayer()
-        elif self.settings.preview_images_method == "kitty":
-            return KittyImageDisplayer()
-        elif self.settings.preview_images_method == "ueberzug":
-            return UeberzugImageDisplayer()
-        return ImageDisplayer()
+    def _get_image_displayer(self):
+        registry = img_display.IMAGE_DISPLAYER_REGISTRY
+        cls = registry[self.settings.preview_images_method]
```
toonn: `cls` is so generic, let's go with `imageDisplayer` or something.

ranger/core/fm.py
```diff
@@ -22,11 +22,7 @@
 from ranger.gui.ui import UI
 from ranger.container.bookmarks import Bookmarks
 from ranger.core.runner import Runner
-from ranger.ext.img_display import (W3MImageDisplayer, ITerm2ImageDisplayer,
-                                    TerminologyImageDisplayer,
-                                    URXVTImageDisplayer, URXVTImageFSDisplayer,
-                                    KittyImageDisplayer, UeberzugImageDisplayer,
-                                    ImageDisplayer)
+from ranger.ext import img_display
```
toonn: Since we only use the registry, can we import this as `from ranger.ext.img_display import IMAGE_DISPLAYER_REGISTRY as imgRegistry` or something?

ranger/ext/img_display.py
```diff
@@ -72,6 +73,24 @@ class ImgDisplayUnsupportedException(Exception):
...
+    raise ImgDisplayUnsupportedException
+
+
+IMAGE_DISPLAYER_REGISTRY = defaultdict(fallback_image_displayer)
+
+
+def register_image_displayer(nickname=None):
+    """Register an ImageDisplayer by nickname if available."""
+
+    def decorator(cls):
```
toonn: `imgDisplayerClass` or similar variation.

ranger/core/fm.py
joe-rice: Sure. I can do that. Would it be at all preferable to just expose a `ranger.ext.img_display.get_image_displayer` and replace the uses of `FM._get_image_displayer` with it?

joe-rice: Or go full ham and replace image_displayer with a data descriptor?

toonn: Sounds like a good suggestion, re the first. Re the second, depends on what exactly you call a data descriptor. I'm not well-versed in OOP design patterns.

Issue #1704, opened 2019-10-01
No commit on status bar


davoudarsalani:
ranger version: ranger 1.9.2
Python version: 3.7.4 (default, Jul 16 2019, 07:12:58) [GCC 9.1.0]
Locale: en_US.UTF-8

The commit sometimes does not show up on the status bar. It seems to happen when the commit is too long.

![1398-07-09-22-39-52-ss](https://user-images.githubusercontent.com/50229325/65992662-8215ab00-e49c-11e9-9a23-4e97598cc3ae.png)

Vifon:
It makes sense considering how the status bar works but it'd also make sense to truncate the commit message if it's really long. I'll take a look into it.

PR #1706, opened 2019-10-01
Add a configurable max VCS message length

Vifon:
Previously it was hardcoded as 50, let's make it easy to change for the user.

Improves upon #1705.
Related to #1704.

### DIFF
--- a/doc/ranger.pod
+++ b/doc/ranger.pod
@@ -1170,6 +1170,10 @@ Sets the state for the version control backend. The possible values are:
  local      display only local state.
  enabled    display both, local and remote state. May be slow for hg and bzr.
 
+=item vcs_msg_length [int]
+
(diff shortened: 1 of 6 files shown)


doc/ranger.pod
```diff
@@ -1170,6 +1170,10 @@ Sets the state for the version control backend. The possible values are:
  local      display only local state.
  enabled    display both, local and remote state. May be slow for hg and bzr.
 
+=item vcs_msg_length [int]
+
+Truncate the long commit messages to this length when shown in the statusbar.
```
toonn: Maybe mention that only the first line is shown?

Reads a bit meh to me. How about, "Length to truncate first line of commit messages to when shown in the statusbar. Defaults to 50."?

Vifon: Fair enough.

PR #1707, opened 2019-10-01
Check current dir exists when pasting

### LINKED ISSUE #1660: Moving a file to a renamed directory causes file deletion
#### Runtime Environment
ranger version: ranger 1.9.2
Python version: 3.7.4 (default, Jul 16 2019, 07:12:58) [GCC 9.1.0]
Locale: en_US.UTF-8

#### Current Behavior

Ranger deletes a file that you move and fails to create the file in the new directory.

0) mkdir /tmp/foo
1) touch /tmp/bla 
2) touch /tmp/foo/bla2
3) ranger /tmp/foo
4) in ranger: :tab_new /tmp
5) in ranger in tab (/tmp) select file /tmp/bla
6) in ranger in tab (/tmp): dd (cut selection)
7) in ranger switch to tab /tmp/foo
8) in a terminal: mv /tmp/foo /tmp/foo2
9) in ranger: pp

Result: the file /tmp/bla is deleted, and it's not in directory /tmp/foo2. 

#### Expected Behavior

Either an error that the directory is no longer exsting, but actually better would be to move /tmp/bla to /tmp/foo2

#### Context

I renamed a directory while ranger still had an open tab.

toonn:
I changed the default `dest` for paste to the value that's used if
`None` were passed to the `CopyLoader`, i.e., `self.thistab.path`,
because the current directory was bypassing the `isdir()` check
otherwise.

Fixes #1660

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1591,19 +1591,19 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 link(source_path,
                      next_available_filename(target_path))
 
-    def paste(self, overwrite=False, append=False, dest=None):
+    def paste(self, overwrite=False, append=False, dest=self.thistab.path):
         """:paste
 
         Paste the selected items into the current directory or to dest
         if provided.
         """
-        if dest is None or isdir(dest):
+        if isdir(dest):
             loadable = CopyLoader(self.copy_buffer, self.do_cut, overwrite,
                                   dest)
             self.loader.add(loadable, append=append)
             self.do_cut = False
         else:
-            self.notify('Failed to paste. The given path is invalid.', bad=True)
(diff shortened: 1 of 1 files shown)


toonn:
@Ablesius, thanks for catching this!

toonn:
Of course, today I can't reproduce it... Does it reproduce reliably for you @Ablesius?

Ablesius:
This still happens on the current master; on the PR branch (rebased against master) it works now as expected! \o/

toonn:
This is weird though. I don't think anything relevant to this changed. @Zaeph, can you test the rebased PR too? It looks to me like a bug due to a race condition or something. The directory not being unlinked yet at the time of the `isdir` check but then being unlinked before the `copyLoader`'s created and therefore the target path becomes a rename rather than a move to a destination.

PR #1709, opened 2019-10-01
Deprecate c/p/tunmap in favor of unc/p/tmap

### LINKED ISSUE #1698: Rename cunmap, punmap, tunmap
These names break the pattern of expectation set by `copycmap` and friends, which simply prefix `cmap` and variations. So you might reasonably expect `uncmap` rather than `cunmap`.

My suggestion, either we rename these or the `copymap` variants, taking them out of the docs but not actually out of the code to avoid breaking backwards compatibility.

Imo, I'd be more likely to think/type `un<tab>` to find the right mode to unmap for than `p<tab>` to find the right action I want to perform for the pager mode.

toonn:
`cunmap` definitely won't be missed because it was buggy. I'd expect the other two to be even less missed.

I considered writing a decorator for deprecation because the code is repetitive but imo it should be removed in the *next* version after 1.9.3 so it shouldn't matter much.

Fixes #1698

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1321,28 +1321,62 @@ class unmap(Command):
             self.fm.ui.keymaps.unbind(self.context, arg)
 
 
-class cunmap(unmap):
-    """:cunmap <keys> [<keys2>, ...]
+class uncmap(unmap):
+    """:uncmap <keys> [<keys2>, ...]
 
     Remove the given "console" mappings
     """
     context = 'console'
 
+class cunmap(uncmap):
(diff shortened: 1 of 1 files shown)


toonn:
Forgot to update docs.

toonn:
Done, no idea how it got there. I didn't touch the `map_` command.

toonn:
Gonna be force amending that latest commit a couple times probably. Bad whitespaces are popping out of the woodwork left and right.
I hate how janky python tooling generally is tbh. Why did pylint not catch these before?

Issue #1699, opened 2019-09-22
[Enhancement] Put items into trash


Hocuri:
#### Runtime Environment

Operating system and version: Manjaro
ranger version: ranger-master 1.9.2
Python version: 3.7.4 (default, Jul 16 2019, 07:12:58) [GCC 9.1.0]
Locale: en_US.UTF-8

#### Current Behavior

On press of F8, items are deleted after confirmation

#### Expected Behavior

On press of F8, put item into trash without asking

#### Context
I am much faster with Ranger than with other file managers but one feature it lacks is trash support, which is very sad because this makes using Ranger a little dangerous (because I could accidentally delete a file) and annoying (because I have to confirm deletion, when I delete a folder even twice)

#### Possible Solutions

Integrate trash-cli

Vifon:
I think it could be resolved as a user's configuration:

```
map <F8> shell trash-rm %s
```

I don't think including it in the defaults would be a good idea, because of the added dependency among other reasons. But I am open to discussing it.

toonn:
I guess there wouldn't be harm in adding a rule for `trash-rm` to `rifle.conf`, not that that would be incredibly comfortable to use. But could be used in a mapping. And if we use a "delete" label maybe we can even have `:delete` use `rifle delete/trash/whatever` by default?

davoudarsalani:
I use this map to use a script I have made to send selected items to trash directory:
map <S-delete> shell $HOME/scripts/trash %s

thiswillbeyourgithub:
Hi, sorry to reopen this but I spend quite some time fine tuning a command to trash and untrash files. As it turns out my version on ubuntu was wayy outdated (1.8 something) so contained no mention of the trash function.

Anyway my keybinding is still okay for me so I'll keep using it, but I thought some people might benefit from these poorly written lines :

```
map DD shell mv -i %s /home/$USER/.local/share/Trash/files/ && echo "trashed %s        from_dir $(pwd)" >> /home/$USER/.config/ranger/trash_history 

map uD shell fileToRestore=$(cat /home/$USER/.config/ranger/trash_history | grep "^trashed" | tail -n 1 | cut -c10- | awk -F"from_dir " '{print $1}' | rev | cut -c10- | rev) && mv -i /home/$USER/.local/share/Trash/files/$fileToRestore $fileToRestore && echo "         untrashed $(pwd)/$fileToRestore" >> /home/$USER/.config/ranger/trash_history && unset $fileToRestore
```

[...]

PR #1710, opened 2019-10-01
Add a trash-cli support to the default rifle.conf

Vifon:
`trash-cli` is usually used to remove file using a traditional "Trash bin" pattern with a possible undo and all these fancy whistles. Since some of our users would like such a support (#1699), let's add it.


ranger/config/rifle.conf
```diff
@@ -264,5 +264,14 @@ label wallpaper, number 14, mime ^image, has feh, X = feh --bg-fill "$1"
...
+######################################################################
+# The actions below are left so low down in this file on purpose, so #
+# they are never triggered accidentally.                             #
+######################################################################
+
+# Execute a file as program/script.
 mime application/x-executable = "$1"
+
+# Move the file to trash using trash-cli.
+has trash-put = trash-put -- "$@"
```
toonn: I'd also add `mv "$@" "$XDG_DATA_DIR"/ranger-trash` (or whatever xdg dir is approriate).

Vifon: I don't think we need a fallback here. `rm` is perfectly fine if use doesn't want/need `trash-cli`.

PR #1341, opened 2018-10-15
Do not include "ask" in open_with options (fixes #1047)

GermainZ:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch
- Terminal emulator and version: urxvt v9.22
- Python version: 3.7.0
- Ranger version/commit: ranger-master 1.9.2
- Locale: en_US.UTF-8

#### CHECKLIST

- [X] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [X] All changes follow the code style **[REQUIRED]**
- [X] All new and existing tests pass **[REQUIRED]**
- [ ] Changes require config files to be updated
    - [ ] Config files have been updated
- [ ] Changes require documentation to be updated
    - [ ] Documentation has been updated
- [ ] Changes require tests to be updated
    - [ ] Tests have been updated

#### DESCRIPTION

Do not include "ask" in open_with options, since that option does nothing there and is redundant.

#### MOTIVATION AND CONTEXT

https://github.com/ranger/ranger/issues/1047

#### TESTING

Tested with Python 2 and 3.

PR #1711, opened 2019-10-01
Hotfix for an off-by-one in rifle

Vifon:
After #1341 the action IDs displayed by the UI were inconsistent with the actual numbers used by rifle.  It was caused by the `ask` action being skipped entirely (including incrementing the ID) but only in the UI.

This is a hotfix that makes the hidden `ask` action to still take up a number (usually (always?) 0) and thus keeping the numbering consistent.  Bad side effect: sometimes the counting starts with 1 instead of 0 but for a time being let's choose a small inconsistency over a BIG one.

### DIFF
--- a/ranger/ext/rifle.py
+++ b/ranger/ext/rifle.py
@@ -293,6 +293,9 @@ class Rifle(object):  # pylint: disable=too-many-instance-attributes
             self._app_flags = ''
             self._app_label = None
             if skip_ask and cmd == ASK_COMMAND:
+                # TODO(vifon): Fix properly, see
+                # https://github.com/ranger/ranger/pull/1341#issuecomment-537264495
+                count += 1
                 continue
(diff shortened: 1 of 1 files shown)


toonn:
Just recording here that I couldn't reproduce the problem this would fix. So I don't forget.

Vifon:
@toonn Try on an empty file while running `./ranger.py -c` from the repo. Press <kbd>r</kbd> and then select "1", supposedly a pager. It should launch an editor which is marked as "0" while there actually is also a hidden "ask" option before them, making the numbers all wrong.

toonn:
Crap on a cracker, you're right. Tried `0` for editor and it opened the pager at `1` : (
Testing the PR and merging.

toonn:
I opened #1756 so we will not forget to *actually* fix this ; )

PR #1714, opened 2019-10-02
Document :trash command

toonn:
Shame on us, we didn't add any docs.


Vifon:
I don't think this description does `:trash` justice. Can you add a concise description of what actually happens (i.e. call `trash-cli` or use our fallback)? And actually mention that trashing means moving to a trash bin instead of deleting, it's not that obvious.

Vifon:
Rebase onto master and we should be good to go.

toonn:
Rebased

Issue #1668, opened 2019-08-20
Date and Time Formats


cpchan:
Right now the date and time formats are hard coded in Ranger. However, IMHO, I think  it would be nicer to change it to respect locale settings.

toonn:
Could you please specify where in the UI you've noticed these? It'll make it much easier for people to fix.

cpchan:
In the statusbar and in any build in status line that shows date and/or time. Right now it is using iso date and 24 hours clock. I am in North America (Canada) where our date and time format is mm-dd-yyyy and 12 hour clock.

bigbigtor:
I've taken some time to look at this issue, and it seems like the cleanest solution would be to leverage the locale's appropriate date and time formats defined on strftime.
According to the [docs](https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior), there's two options.

- `'%c'` which will output a more human readable format (e.g. _Mon Sep 30 07:06:05 2013_), but IMO is too verbose.

- `'%x %X'` which is more similar to the current format (e.g. _02/23/2019 03:33:31 PM_), but it differs in the date order, and it includes the seconds.

What are your thoughts on this?

toonn:
+1 for '%x %X'!

PR #1715, opened 2019-10-02
Use locale appropriate date and time format

bigbigtor:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Terminal emulator and version: Termite v15
- Python version: 2.7.16 / 3.7.4
- Ranger version/commit: 
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

The formats for dates/times show on the status bar and file details have been changed to use the locale appropriate version (%x for date, %X for time).

#### MOTIVATION AND CONTEXT

This changes will make the UI more user-friendly as suggested in #1668 

#### TESTING

I have tested the output with the following locales to check it behaves correctly with different formats and scripts:
 - ar_SA.UTF-8
 - en_CA.UTF-8
 - en_US.UTF-8
 - es_ES.UTF-8
 - zh_CN.UTF-8


toonn:
Apologies, we already had something similar to this in the works in #1619. It'd be great if you could PR a version of these changes as I described to @cibinmathew's branch. So, splitting the settings and then using the settings in the code. Oh, and make it three settings, one for time, date and one for the combination.

bigbigtor:
I think this covers all the changes you asked for. The default values for the settings won't work until #1629 gets merged. Just to be clear, you want me to PR these changes into @cibinmathew's master branch?

doc/ranger.pod
```diff
@@ -1136,6 +1147,11 @@ Put the status bar at the top of the window?
 
 Abbreviate $HOME with ~ in the titlebar (first line) of ranger?
 
+=item time_format [string]
+
+The time format to be display in the GUI. It defaults to "%X" which is locale's
```
toonn: "displayed", s/GUI/interface, ranger has no GUI. "the locale"

toonn: "for all valid values" -> "for valid format strings"

doc/ranger.pod
```diff
@@ -883,6 +883,17 @@ Ask for a confirmation when running the "delete" command?  Valid values are
 "always" (default), "never", "multiple". With "multiple", ranger will ask only
 if you delete multiple files at once.
 
+=item date_format [string]
+
+The date format to be display in the GUI. It defaults to "%x" which is locale's
```
toonn: "displayed", s/GUI/interface, "the locale"
"for all valid values" -> "for valid format strings"

doc/ranger.pod
```diff
@@ -883,6 +883,17 @@ Ask for a confirmation when running the "delete" command?  Valid values are
...
+appropriate date format. See python's strftime docs for all valid values.
+
+=item date_plus_time_format [string]
+
+The combined date and time format to be display in the GUI. It defaults to "%x %X"
```
toonn:
```
Format used when both date and time are displayed, as in the statusbar. It defaults to "date_format time_format". See python's I<strftime> docs for valid format strings.
```

ranger/config/rc.conf
```diff
@@ -163,6 +163,15 @@ set draw_progress_bar_in_status_bar true
 # Both combines the two.
 set draw_borders none
 
+# The date format used in the GUI
```
toonn: No mentioning GUI.

ranger/config/rc.conf
```diff
@@ -163,6 +163,15 @@ set draw_progress_bar_in_status_bar true
...
+set date_format %x
+
+# The time format used in the GUI
+set time_format %X
+
+# The combined date and time format used in the GUI
+set date_plus_time_format %x %X
```
toonn: `#set date_and_time_format %x %X`
Comment this out and implement defaulting to `date_format time_format` in the code. I don't think we can easily refer to other settings when setting something.

ranger/container/settings.py
```diff
@@ -36,6 +36,8 @@
     'colorscheme': str,
     'column_ratios': (tuple, list),
     'confirm_on_delete': str,
+    'date_format': str,
+    'date_plus_time_format': str,
```
toonn: Let's rename to `date_and_time_format`, plus kinda sounds like it might be a timezone thing.

ranger/core/linemode.py
```diff
@@ -118,7 +119,8 @@ def filetitle(self, fobj, metadata):
     def infostring(self, fobj, metadata):
         if fobj.stat is None:
             return '?'
-        return datetime.fromtimestamp(fobj.stat.st_mtime).strftime("%x %X")
+        fmt = SettingsAware.settings.date_format
```
toonn: Should be `date_and_time_format` afaics?

ranger/core/linemode.py
```diff
@@ -130,8 +132,9 @@ def filetitle(self, fobj, metadata):
     def infostring(self, fobj, metadata):
         if fobj.stat is None:
             return '?'
+        fmt = SettingsAware.settings.date_format
```
toonn: Same thing, `date_and_time_format`.

ranger/gui/widgets/statusbar.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, win, column=None):
         self.column = column
         self.settings.signal_bind('setopt.display_size_in_status_bar',
                                   self.request_redraw, weak=True)
+        self.timeformat = self.settings.date_plus_time_format
```
toonn: I *think* this'd better be above the redraw request signal.

toonn (changes requested):
Couple things I'd like to see changed.

bigbigtor:
That should cover the last request. I hope that was the right place to set the default value for the combined format.

toonn (dismissed):
LGTM, I'll test this later today before merging.

toonn:
Hmm, my statusbar isn't adhering to the locales.

bigbigtor:
Is it just the statusbar, or does it affect the file details as well?
In my case I had to run
`$ unset LANG`
`$ source /etc/profile.d/locale.sh`
after setting the new locale so it got applied immediately without the need of login out.
Could that be the issue?

toonn:
I'm testing with `LC_ALL`, shouldn't that work?

bigbigtor:
> I'm testing with `LC_ALL`, shouldn't that work?

I didn't test it using `LC_ALL`, but it should work, let me check what's happening.

bigbigtor:
It works fine for me, so I'm not sure what could be the cause.
lc_all

toonn:
I'm running into this:
```
Error in line `set date_format %x':
  'NoneType' object has no attribute 'thisfile'
Error in line `set time_format %X':
  'NoneType' object has no attribute 'thisfile'
```
Because whenever `%` occurs in `rc.conf` we try to resolve macros but some attributes we need for that aren't available yet.

bigbigtor:
> I don't understand why you're not running into the same errors though.

This branch needs #1629's changes before it can work, so I was testing it with those changes patched into my local repo. Sorry if that wasn't clear.

And about the format strings, would this be enough?
1. Add the extra `%` to the defaults
2. Specify in the docs that the `%` needs doubling up

toonn:
Yep, sounds good!

doc/ranger.pod
```diff
@@ -887,12 +887,16 @@ if you delete multiple files at once.
 
 The date format to be displayed in the interface. It defaults to "%x" which is the locale
 appropriate date format. See python's I<strftime> docs for valid format strings.
+To avoid ranger trying to resolve the format strings as macros, they need to be
```
toonn: Maybe we should merge these sections to avoid duplication? Like I did for the map variations. Also, maybe a more thorough explanation's in order? Explain that doubling up the `%%`'s is a good habit because *some* formatting characters need the escaping, like `%d`.

toonn (commented):
What do you think about my docs suggestion?

doc/ranger.pod
bigbigtor: Do you mean adding a new topic under the *CONCEPTS* section in **doc/ranger.pod**?

toonn: No, I meant empty sections so you only have to do the explanation once. Look at `map/cmap/tmap/...` to see what I mean. The rendered result puts the headers together so it looks like the explanation applies to all the commands or settings in this case.

ranger/config/rc.conf
```diff
@@ -163,6 +163,15 @@ set draw_progress_bar_in_status_bar true
...
 set draw_borders none
 
+# The date format used in the interface
+set date_format %%x
+
+# The time format used in the interface
+set time_format %%X
+
+# The combined date and time format used in the interface
+#set date_and_time_format %x %X
```
toonn: Don't forget to do so here in case people uncomment the setting.

ranger/container/settings.py
```diff
@@ -199,6 +202,9 @@ def _get_default(self, name):
                 value = self.fm.confpath('scope.sh')
                 if not os.path.exists(value):
                     value = self.fm.relpath('data/scope.sh')
+        elif name == 'date_and_time_format':
```
toonn: Let's add defaults for both `time_format` and `date_format` too.

toonn (commented):
Couple more things.

ranger/container/settings.py
bigbigtor: Should I comment out the defaults in **ranger/config/rc.conf** as well?

toonn: Nah, those don't hurt.

bigbigtor: I added the default values, but they still need a way to ensure them being loaded before date_and_time_format. Is there a way to do this currently? Or should I work on some sort on setting dependency solution?

toonn: I'm not sure the `%`s should still be escaped at this point. Did you test with some format string that has something that conflicts with a macro, `%d` or something?

doc/ranger.pod
```diff
@@ -885,18 +885,21 @@ if you delete multiple files at once.
...
+The formats to be used in the interface when date, time,
+or both date and time (as in the statusbar) are displayed.
+
+The defaults are "%%x" for date_format, "%%X" for time format,
+and "date_format time_format" for date_and_time_format.
 See python's I<strftime> docs for valid format strings.
+
 To avoid ranger trying to resolve the format strings as macros, they need to be
 escaped by doubling up the "%" symbols. e.g. B<%x %H:%M> must be set as B<%%x %%H:%%M>.
+This is also a good habit since those strings could be interpreted as python's
```
toonn: I don't think this makes sense. That would only matter if we used them as format strings but I don't think we do? What I was getting at is that `%x` doesn't actually need the doubling up, because it's not a valid macro, though it doesn't hurt either, but `%%d` *does* because it *is* a valid macro.

toonn (commented):
Hmm, sounds a bit gnarly. What order are settings loaded now? Order they happen to be in in a dictionary?

doc/ranger.pod
bigbigtor: > What order are settings loaded now?

They're sorted by name so `date_and_time_format` gets loaded first, even though `date_format` is declared just above it.

> I'm not sure the %s should still be escaped at this point. Did you test with some format string that has something that conflicts with a macro, %d or something?

I think I got stuck with the loading order and forgot to test those :/
I did now and they're not working when escaped, even with `%d`, so I'll have to unescape them.

> I don't think this makes sense. That would only matter if we used them as format strings but I don't think we do?

Because I'm using the _format()_ function to build the `date_and_time_format` default value and you previously mentioned them as _formatting characters_, I got the two concepts mixed-up.
Sorry about that, I'll fix it.

bigbigtor:
I fixed the loading order problem, turned out to be easier than I originally thought.

toonn:
Hmm, testing this to merge. Changing `time_format` seems to work. `date_format` isn't used anywhere so can't be tested. Changing `date_and_time_format` doesn't seem to work.

bigbigtor:
> Hmm, testing this to merge. Changing `time_format` seems to work. `date_format` isn't used anywhere so can't be tested. Changing `date_and_time_format` doesn't seem to work.

Let me merge master into the branch, fix the conflicts and check everything again, it's been a long time.

bigbigtor:
After merging upstream, it seems to be working for me, could you please check?

toonn:
Now it works but only if I configure it in `rc.conf`. Still doesn't if I set it while ranger's running.

bigbigtor:
So after a bit of debugging, what I see is the settings are actually set, but the UI needs refreshing.
One solution I can think of, is creating a list of settings that would trigger this UI refresh after being set.
So when any of those settings is set, a signal is emitted requesting the redraw.
This could be easily leveraged by any potential setting that behaves similarly in the future.
What do you think?

toonn:
Doesn't this already exist? I'm fairly certain the draw_borders setting causes refreshes. But I"m not sure it's enough because manually causing redraws (using `C-l`) didn't suffice.

bigbigtor:
This takes care of the status bar, only file details are pending.

toonn:
I'm understanding correctly this isn't ready for review yet, right?

bigbigtor:
> I'm understanding correctly this isn't ready for review yet, right?

Not yet, let’s hope I can work on it soon.

PR #1722, opened 2019-10-14
Add support for more apps in rifle default config

swalladge:
#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Adds support for more filetypes/apps in the rifle default config. These are things I use in my own config that I felt could be useful to others too. Found them while merging new changes from the upstream rifle.conf into mine ( https://github.com/swalladge/dotfiles/commit/e4b62ed6c277a94ae28cef3f54d1cf7f6fb73e65 ). Summary of changes:

- audacity
- firefox dev edition
- scribus projects
- sc (cli spreadsheet editor)
- krita image editor
- fix bug where wine is erroneously displayed if not installed

#### MOTIVATION AND CONTEXT

These are simple config additions to support more filetypes out-of-the-box for rifle. Simply convenience.

#### TESTING

Does not affect the codebase. The added rules in rifle.conf should be inspected to ensure they don't conflict or clobber existing rules.

### DIFF
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -94,7 +94,7 @@ ext 1                         = man "$1"
 ext s[wmf]c, has zsnes, X     = zsnes "$1"
 ext s[wmf]c, has snes9x-gtk,X = snes9x-gtk "$1"
 ext nes, has fceux, X         = fceux "$1"
-ext exe                       = wine "$1"
+ext exe, has wine             = wine "$1"
 name ^[mM]akefile$            = make
 
 #--------------------------------------------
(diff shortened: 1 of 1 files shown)


ranger/config/rifle.conf
```diff
@@ -70,6 +70,7 @@ ext x?html?, has chromium-browser, X, flag f = chromium-browser -- "$@"
 ext x?html?, has chromium,         X, flag f = chromium -- "$@"
 ext x?html?, has google-chrome,    X, flag f = google-chrome -- "$@"
 ext x?html?, has opera,            X, flag f = opera -- "$@"
+ext x?html?, has firefox-developer-edition, X, flag f = firefox-developer-edition -- "$@"
```
toonn: I know we usually go by rarity. But I'm not sure betas and dev editions should follow the same rule. Opinions, @Vifon?

ranger/config/rifle.conf
```diff
@@ -129,6 +130,8 @@ mime ^video,       has mplayer,  X, flag f = mplayer -fs -- "$@"
 mime ^video|audio, has vlc,      X, flag f = vlc -- "$@"
 mime ^video|audio, has totem,    X, flag f = totem -- "$@"
 mime ^video|audio, has totem,    X, flag f = totem --fullscreen -- "$@"
+mime ^audio|ogg$,  has audacity, X, flag f = audacity -- "$@"
```
toonn: Do we need the explicit `ogg`?

swalladge: Probably not. Perhaps this was to include ogx (mime application/ogg instead of audio/ogg)?  I would say better to remove it though and just use `^audio` there.

Vifon:
A bit offtopic but if `^audio|ogg$` even *works*, it makes me concerned about the other rules. What I mean, there are two interpretations: `^(audio|ogg)$` or `(^audio)|(ogg$)`. Most of our rules assume the former (for example `^video|audio` looks like it expects it). After a very brief testing, it seems indeed the regexp anchors (`^$`) get interpreted after the alternative (`|`), so the latter case. It seems we have a bit of fixing to do, as half of the patterns are matching too much.

ranger/config/rifle.conf
Vifon: I'd keep it out of the default rifle.conf. Users can and should add it manually in such cases.

toonn: I'm not really opposed with having it in the default config. Do you think it'd be fine at the end of the list?

swalladge:
@Vifon True, but I think it's only the `^video|audio` pattern that is affected. I don't see any other rules that use the `|` regex operator.

The `^audio|ogg$` pattern was originally added in https://github.com/ranger/ranger/pull/105 , which I can only assume was meant to catch `application/ogg` (.ogx files). I don't see anywhere that mentions support for ogx in audacity, mpv, mplayer... Perhaps it can be safely dropped and replaced with `^audio` everywhere.

ranger/config/rifle.conf
swalladge: Fair enough, I can see why you wouldn't want to clutter the default with loads of test/dev editions of software :+1:

Vifon: @swalladge, if you'll remove this line, I'll be happy to merge the PR. :)

toonn:
@swalladge, @Vifon, I think a rebase is also in order since we changed the order, don't wanna accidentally lose that in the merge.

Vifon:
How exactly would we lose it?

toonn:
Sometimes diffs suck, we've lost changes before. I'm just saying a rebase would avoid having to pay attention when merging.

swalladge:
@Vifon @toonn I removed the firefox dev edition line, rebased, and squashed. :)

swalladge:
@vifon are you still interested in merging this PR?  :)  If so, I'm happy to revisit, rebase, and retest.

toonn (approved):
LGTM : )
Apologies for the oversight, we'd as good as given our approval.

toonn:
Feel free to ping us in the irc channel or something if we neglect to reply here : )

swalladge:
@toonn thanks!  

> Feel free to ping us in the irc channel 

Yep for sure. :smile: To be honest, I totally forgot about this PR.

PR #1725, opened 2019-10-15
adding nautilus as command to open current dir in ui file manager,

dumbPy:
Adding `:nautilus` as command to open the current directory in nautilus file manager with fallback to `xdg-open`

#### ISSUE TYPE
- Improvement/feature implementation

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 18.04
- Terminal emulator and version: kitty
ranger version: ranger 1.9.2
Python version: 3.6.8 (default, Oct  7 2019, 12:59:55) [GCC 8.3.0]
Locale: en_IN.ISO8859-1

#### CHECKLIST

- [X] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [X] All changes follow the code style **[REQUIRED]**
- [X] All new and existing tests pass **[REQUIRED]**
- [X] Changes require config files to be updated
    - [X] Config files have been updated
- [ ] Changes require documentation to be updated
    - [ ] Documentation has been updated
- [ ] Changes require tests to be updated
    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT
some use cases need a gui file manager like for grad and dropping files `browse` buttons directly.   
A simple `:nautilus` command to open file manager reduces the hustle.

#### TESTING
`make test` fails for `test_pylint` for already existing code, but has nothing to do with my commits.
Doesn't affect other areas of codebase.

PR #1726, opened 2019-10-16
open current directory in X filemanager with xdg-open, binding=gf

dumbPy:
Open current Directory in X file manager using xdg-open.
binding is `gf` for (go to filemanager)

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 18.04
- Terminal emulator and version: kitty
ranger version: ranger 1.9.2
Python version: 3.6.8 (default, Oct  7 2019, 12:59:55) [GCC 8.3.0]
Locale: en_IN.ISO8859-1

#### CHECKLIST

    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT
some use cases need a X file manager like for grad and dropping files `browse` buttons directly.   
A simple `gf` binding (go to filemanager) to open X file manager reduces the hustle.

#### TESTING
`make test` fails for `test_pylint` for already existing code, but has nothing to do with my commits.
Doesn't affect other areas of codebase.


dumbPy:
Also, this addition is based on @toonn 's suggestion in #1725

toonn:
For drag and drop [there's also dragon](https://github.com/ranger/ranger/wiki/Drag-and-Drop), let's include that.

dumbPy:
`<C-o>` seems available. Ctrl-o to open in external filemanager, seems fair.

`xdg-open` takes  care of opening in system's default/preinstalled file manager in most distributions. which is why I haven't mentioned Nautilus.
For dragon, I will play around a bit and add it too.

For TUIs, their respective users may open PRs  since I have already made a separate section for opening current Dir in X (will change to Other) file manager

toonn:
Sounds fair. Let's not include TUIs yet. Except for ranger, ofc : ) Put it at the bottom of the list though. Let's not use `xdg-open` for this, people might want to use other file managers than the system default sometimes. Just try to order them by reverse popularity. So Nautilus should probably end up at the bottom just above ranger.

dumbPy:
In rc.conf, We can access current directory path as `fm.thisdir.path`, but I am not able to cast it as an extra argument to `:open_with filemanager`. From the already existing rules in rifle.conf, `"$@"` is used for all selected files and "$[n]" is used for nth selected file, there's no way to define cwd.
Let me know if there is a way to cast the value of `fm.thisdir.path` to `:open_with filemanager` or else `$1/../` seems the only way.   
To make is more understandable, I can rename the rule to, `cwd_in_other_filemanager`

toonn:
Huh, you're right. I mistakenly thought `open_with` was a shallower wrapper around `execute_file` than it is. So let's implement a custom command to `self.fm.execute_file(currentdir, app=filemanager)`.

dumbPy:
I think it would be more generic to just give access to Current Directory in rifle.conf  
Like replacing say `$#` in making command with `fm.thisdir.path` would fix it.  
We already have access to `fm.thisfile` with `$1` and all selected files with `$@`, adding something like `$#` for `fm.thisdir` makes sense, also it's only because you don't want the non-standard `$1/../` in the rifle.conf

toonn:
No, you're misunderstanding rifle's `$@` etc. My problem with your approach is that it's a weird rifle rule. You have to pass a file *in* a directory to trigger the rule on the directory. That's silly and makes it hard to trigger the rule on empty directories.

dumbPy:
@toonn  See if the this works for you.
I have added ability to pass file[s]/dir object to open_with argument, saperated by `--`. Tried a few combinations, and it works as expected. Also, without `--`, it works as same as before i.e., uses `[f for f in self.fm.thistab.get_selection()]`.

running `:open_with open -- ./` opens the current directory with xdg-open, and sould similarly work for macOS is suppose. running `:open_with open` would fall back to opening current file selection[s].
running `:open_with open -- regexstr` would match `regexstr` with basenames of the files in current dir and open them. though xdg-open doesn't support multiple files opening, `vim` does. so `:open_with vim -- .*.py` would open all files in the current dir in vim in separate buffers.

dumbPy:
What remains now is how to make `:open_with open -- ./` or something similar, say `:open_with editor -- .*.py` into a binding.
I tried quite a few things but failed to deduce how rc.conf works for multi word strings.
Need help

toonn:
@dumbPy, looking at this with fresh eyes it seems like your original goal can be achieved rather easily like this: `map gf shell rifle %d`. Because we added generic openers. You could easily add rules for other file managers, change the command to specify a label or other rifle features.

dumbPy:
it would still require adding 
```
mime ^inode/directory, has xdg-open = xdg-open "$@"
mime ^inode/directory, has open = open "$@"
```
to the rifle.conf [tested for xdg-open, hoping it would be same for mac's `open` command].

Now that I can add this, there is no need to handle `./` in open_with. What do you think of regex though?
Ofcourse even that can be done with chaining of commands i.e., chaining regex based marking then calling `open_with` on them and then unmarking  the files again, in a single biniding. But still I think it would good to have regex based filtering in open_with itself

toonn:
Imo, `open_with` should stay as simple as possible.

Adding those mime constraints can only make the rules match *less* not more. So it can't possibly be necessary.

dumbPy:
What mime constraints are you taking about? If this is in regard to #1762 , `rifle /path/to/some/dir` doesn't work and throws`Unknown file type: inode/directory`, which is why I am adding inode/directory as a mime type explicitly.   
Same for your suggestion i.e., from ranger `:shell rifle %d` throws the same error.

Issue #1654, opened 2019-08-04
How can I disable the top header and bottom status bar?


sdushantha:
#### Current Behavior
Not able to disblae the top header and bottom status bar

#### Expected Behavior
Being able to disable both, top header and bottom status bar.

#### Context

I want to be able to disable the top header and the bottom status bar in ranger to a get miminal look.

aRkedos:
I started experimenting with a check whether the "disable_titlebar" option is true or false (implemented the check during ui initialization and added the option in settings.py and rc.conf) and it works.
If true the titlebar does not get shown if not it will be displayed like normal. Still trying to figure out how you can do that during runtime.

Would that be the correct approach? Could I move on from there?

toonn:
It sounds ok. I'd flip the setting though, double negatives are unnecessarily confusing.

aRkedos:
Cool. So far it only works in startup when the rc.conf gehts parsed. When I use the toggle_option / set_option command nothing happens.

First I thought it is because the ui needs to be redrawn if one of the commands is issued but that is not the case since it already gets redrawn as I saw by using the viewmode option.

I think I miss one piece of the puzzle. I'll keep trying :)

asoymos:
Well, in the mean time I altered the color scheme to display the title and status bar black:
1) Copy /usr/share/doc/ranger/config/colorschemes to ~.config/ranger/colorschemes
2) I edited default.py
`elif context.in_titlebar:`                                                                                             
`if context.hostname:`                                                                                                                                                                   
            `elif context.directory:`                                                                                             
                `fg = 16`
`elif context.in_statusbar:`                                                                                              
            `if context.permissions:`                                                                                             
                `if context.good:`                                                                                                
                    `fg = 16`     
            `if context.marked:`                                                                                                  
                `attr |= bold | reverse`                                                                                            
            `else:`     
                `fg = 16`

toonn:
@asoymos, have you tried running the patch? Testing and reviewing by others can speed up merging the PR.

asoymos:
@toonn, no idea how to do that, but if explained can try

alexaandru:
@asoymos maybe I can help. I see an open PR about this, https://github.com/ranger/ranger/pull/1764 so I will assume @toonn was referring to that. Here is how I tried it:

```Bash
git clone https://github.com/aRkedos/ranger.git
cd ranger
git switch status-bar 
make
# you need to edit your config, see below
[...]

PR #1732, opened 2019-10-19
Adding options for enabling or disabling the status- / titlebar

aRkedos:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: MANJARO Linux 5.2.21-1
- Terminal emulator and version: xterm
- Python version: 3.7.4
- Ranger version/commit: 1.9.2 ranger-master 
- Locale: de_DE.UTF8

#### CHECKLIST

- [X] All changes follow the code style **[REQUIRED]** <---- not entirely sure though
- [X] All new and existing tests pass [REQUIRED] <---- yes 
    - [ ] Tests have been updated

#### DESCRIPTION
I added two options to disable / enable the status and the titlebar as one user asked for.
Manpage was also updated with the new options.

#### MOTIVATION AND CONTEXT

Changes are not required, they simply add more options for customization for users
see: https://github.com/ranger/ranger/issues/1654

#### TESTING

make test_py has been run.

Tests ran with a 9.83 / 10 rating. Only issues the tests raised were on unchanged areas of the codebase. 

Additionaly I actively used Ranger with these changes. Everything ran fine on my System.

As far as I can tell these changes have no impact on other parts of the codebase.

### DIFF
--- a/ranger/data/mime.types
+++ b/ranger/data/mime.types
@@ -17,6 +17,8 @@ audio/wavpack           wv wvc
 audio/webm              weba
 audio/x-ape             ape
 
+image/webp              webp
+
 video/mkv               mkv
 video/webm              webm
 video/flash             flv


doc/rifle.1
```diff
@@ -1,4 +1,4 @@
-.\" Automatically generated by Pod::Man 4.10 (Pod::Simple 3.35)
+.\" Automatically generated by Pod::Man 4.11 (Pod::Simple 3.35)
```
toonn: Remove this file from your commits please. We only want it committed *if* it's changed.

ranger/container/settings.py
```diff
@@ -100,6 +100,8 @@
     'w3m_offset': int,
     'wrap_scroll': bool,
     'xterm_alt_key': bool,
+    'show_titlebar': str,
```
toonn: These should be `bool` values, not `str`.

ranger/gui/ui.py
```diff
@@ -397,6 +402,7 @@ def draw(self):
     def finalize(self):
         """Finalize every object in container and refresh the window"""
         DisplayableContainer.finalize(self)
+
```
toonn: Seems like a weird whitespace to add.

ranger/gui/ui.py
```diff
@@ -339,7 +344,7 @@ def redraw(self):
             self.console.focused = False
             self.console.visible = False
             self.status.visible = True
-
+
```
toonn: Remove this trailing whitespace.

ranger/gui/ui.py
```diff
@@ -485,6 +491,33 @@ def get_pager(self):
     def _get_viewmode(self):
         return self._viewmode
 
+    def _set_titlebar(self, value):
```
toonn: Can't you work with a default value here? `... , value=True):` And then simplify the body to test on a boolean rather than these `ALLOWED_STATUS` strings?

ranger/gui/ui.py
```diff
@@ -485,6 +491,33 @@ def get_pager(self):
...
+            value = self.ALLOWED_STATUS[0]
+        if value in self.ALLOWED_STATUS:
+            if value == self.ALLOWED_STATUS[0]:
+                self.add_child(self.titlebar)
+            elif value == self.ALLOWED_STATUS[1]:
+                self.remove_child(self.titlebar)
+        
+        self.redraw_window()
+
+    def _set_statusbar(self, value):
```
toonn: This is almost the exact same code, just refactor both methods into a generic method you can pass `self.titlebar` to.

doc/ranger.pod
```diff
@@ -1090,6 +1090,14 @@ directories are displayed at once. A value of 0 turns off this feature.
 
 Always show the terminal cursor?
 
+=item show_titlebar [str]
+
+Enable / disable the titlebar. Alowed values are "true" or "false"
```
toonn: We don't really mention a list of allowed values for `[bool]` options.

ranger/config/rc.conf
```diff
@@ -27,6 +27,10 @@
 set viewmode miller
 #set viewmode multipane
 
+#enable / disable the status or the titlebar. Allowed values are 'true' or 'false'
```
toonn: Make this a proper sentence like the other comments. Drop the allowed values. Maybe phrase it as a question: "Show the statusbar?" Make a comment for each don't group these settings.

ranger/gui/ui.py
aRkedos: Expected that feedback. I'll do :)

ranger/gui/ui.py
aRkedos: will do.

ranger/gui/ui.py
aRkedos: Same like the white space before.

ranger/gui/ui.py
aRkedos: happend by accident, will remove it

ranger/container/settings.py
aRkedos: I did that first. But then I looked at other options which also used 'false' / 'true' as `str`. And when I used a `boolean` you always had to write True / False with a capital letter in the beginning which I think from a user perspective is not too practical.

But I'll change it :)

ranger/gui/ui.py
toonn: Not sure what you mean? Blank lines should be completely blank. The diff shows spaces.

aRkedos: I meant it happened by accident and that I'll remove it :) like the one in line 405 ;)

aRkedos:
I basically did all the changes you requested. All Tests are now positive and the whitespace is gone, 
I refactored the methods, so they test against a `bool` and the code should be cleaner now.

Anyway what I couldn't manage to merge those methods into a single generic method. I think one of the challenges are how instances of Signal work. If you bind a method / function to a Signal the signal basically becomes the parameter of that said function.

`def function(self, value):` which is bound to a signal gets the Signal-instance which is created upon using `signal_bind()` passed as an argument so `value` basically represents that Signal-instance.

So adding an additional parameter to that function (which represents either the status or the titlebar object won't work. And that is why something like 

`def _toggle_bars(self, bar): def bar(value): if..... add_child(bar)......return bar`
won't work either because you have to bind the signal to _toggle_bars but than bar would again represent an instance of Signal instead of the passed bar object.

At least that is how I understand `signal_bind` for the time being.

aRkedos:
Okay managed to merge those functions.

ranger/container/settings.py
```diff
@@ -117,6 +119,8 @@
     'vcs_backend_hg': ['disabled', 'local', 'enabled'],
     'vcs_backend_svn': ['disabled', 'local', 'enabled'],
     'viewmode': ['miller', 'multipane'],
+    'show_titlebar': [False, True],
```
toonn: Remove these, they're redundant for boolean options.

ranger/core/filter_stack.py
```diff
@@ -149,3 +149,8 @@ def __str__(self):
 
     def decompose(self):
         return [self.subfilter]
+
+
+class TagFilter(BaseFilter):
```
toonn: Why are we touching `filter_stack`?

ranger/gui/ui.py
```diff
@@ -302,10 +302,12 @@ def setup(self):
         self.taskview.visible = False
         self.add_child(self.taskview)
 
-        # Create the status bar
+        # Create the status bar whether the setting is "True" or "False"
```
toonn: This means it doesn't matter whether the setting is true or false.

ranger/gui/ui.py
```diff
@@ -286,10 +284,12 @@ def setup(self):
         from ranger.gui.widgets.taskview import TaskView
         from ranger.gui.widgets.pager import Pager
 
-        # Create a titlebar
+        # Create a titlebar depending whether the setting is "True" or "False"
```
toonn: Simplify to "Draw titlebar if set." Same for the other comment.

ranger/core/filter_stack.py
aRkedos: Because the Travis CI check stated that there was only 1 empty line after the `def __str__()` and the check failed. So I added one more line.

ranger/gui/ui.py
```diff
@@ -1,5 +1,5 @@
 # This file is part of ranger, the console file manager.
-# License: GNU GPL version 3, see the file "AUTHORS" for details.
+# License: GNU GPL version 3, see the file "AUTHORS" for detailset
```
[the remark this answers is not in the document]
aRkedos: Oh man am I clumsy :/

ranger/gui/ui.py
aRkedos: Done

ranger/gui/ui.py
aRkedos: Done

ranger/container/settings.py
aRkedos: Done.

aRkedos:
added a fix for #1739

toonn:
@aRkedos, please roll back that fix. It needs to go in its own PR.

aRkedos:
Oh okay I'll make another fork then

toonn:
@aRkedos, no need to make another fork. Don't think github even lets you? Just make another branch.

aRkedos:
Yeah you are right. You can only fork once. i created a new branche and reset that branch to ranger/master. Worked l, thanks for the advice and patience.

toonn:
Why did you close this?

hyh1750522171:
11

Issue #1739, opened 2019-10-24
Tmux window gets not renamed even if option is set


aRkedos:
#### Runtime Environment

- Operating system and version: Linux Manjaro 5.2.21
- Terminal emulator and version: urxrvt
- Python version: Python 3.7.4
- Ranger version/commit: Ranger-Master **1.9.20**
- Locale: de.DE_UTF.8

#### Current Behavior
If using Tmux, the window in which Ranger is started does not get renamed even if the option is set.

#### Expected Behavior
Tmux window should be renamed if option is set. And should again rename itself when Ranger closes.

#### Possible Solutions

I've found the issue in ui.py. In ui.py the command `tmux show-window-option -v automatic-rename` gets invoked via the `check_output` method which was imported from `ranger.ext.spawn`. This returns the output of `stdout` of that command. Nothing more (see `gui/ui.py` **lines 116 - 128*). This output is stored in the variable `self._tmux_automatic_rename.`

In order to rename the window, ranger must issue the command `tmux rename-window ranger`. 

Later on in order to restore the automatic-rename behaviour of tmux Ranger issues the command `tmux set-window-option automatic-rename` with `self._tmux_automatic_rename as parameter` (gui/ui.py lines 178 - 189). This does not get really issued because `if self._tmux_automatic_rename` is not a boolean since the `stdout` of `tmux show-window-option -v automatic-rename` is `on` or `off`

Solution
If `tmux show-window-option -v automatic-rename` returns `off` set it to `on` instead to make sure tmux gets renamed after closing
Add a check_output which issues `tmux rename-window ranger` to actually rename the thing.
remove the else-block in `gui/ui.py` (starting at line 185) since it is no longer needed with the aforementioned changes.
Maybe even the `sys.stdout.write` call can be removed (see line 127) could be removed (but not entirely sure which purpose it serves)

#### Steps to reproduce
1. Simply start Ranger in a Tmux session

aRkedos:
added it to the still open PR #1732

lonelykid:
I think this should be better documented. Tmux has ``allow-rename`` and ``automatic-rename``. The former is what ranger needs, and ranger would set the latter to off.

toonn:
Ranger emits a warning if you enable the setting but allow-rename is set to off.
What would you expect to be documented better?

lonelykid:
[...]

PR #1740, opened 2019-10-25
applying proposed tmux renaming fix

aRkedos:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Manjaro Linux 5.2.21-1
- Terminal emulator and version: urxvt
- Python version: Python 3.7.4
- Ranger version/commit: 1.9.20 Master
- Locale: de_DE-UTF.8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Applying the proposed Tmux renaming window fix from issue #1739.

#### MOTIVATION AND CONTEXT
Change fixes a problem that the current Tmux window does not get rename even if the option is set.

#### TESTING
Tests have been run. Very small changes to the codebase.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -188,6 +188,10 @@ set update_title false
 # Set the tmux window-name to "ranger"?
 set update_tmux_title true
 
+# Set the title to "ranger" in the screen program?
+
+set update_screen_title true
+
 # Shorten the title if it gets long?  The number defines how many
 # directories are displayed at once, 0 turns off this feature.
 set shorten_title 3
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -185,12 +185,8 @@ set display_tags_in_all_columns true
 # Set a title for the window? Updates both `WM_NAME` and `WM_ICON_NAME`
 set update_title false
 
-# Set the tmux window-name to "ranger"?
+# Set the tmux/screen window-name to "ranger"?
 set update_tmux_title true
-
-# Set the title to "ranger" in the screen program?
-
-set update_screen_title true
 
 # Shorten the title if it gets long?  The number defines how many
 # directories are displayed at once, 0 turns off this feature.
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -122,7 +122,9 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
             if self.settings.update_tmux_title and 'TMUX' in os.environ:
                 try:
                     self._tmux_automatic_rename = check_output(
-                        ['tmux', 'show-window-options', '-v', 'automatic-rename']).strip()
+                        ['tmux', 'show-window-option', '-v', 'automatic-rename']).strip()
+                    if self._tmux_automatic_rename == 'off':
+                        self._tmux_automatic_rename = 'on'
                 except CalledProcessError:
                     self._tmux_automatic_rename = None
 
@@ -130,6 +132,7 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
         self.is_on = True
 
         if self.settings.update_tmux_title and 'TMUX' in os.environ:
+            check_output(['tmux', 'rename-window', 'Ranger'])
             sys.stdout.write("\033kranger\033\\")
             sys.stdout.flush()
 
@@ -186,11 +189,6 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
                 try:
                     check_output(['tmux', 'set-window-option',
                                   'automatic-rename', self._tmux_automatic_rename])
-                except CalledProcessError:
-                    pass
-            else:
-                try:
-                    check_output(['tmux', 'set-window-option', '-u', 'automatic-rename'])
                 except CalledProcessError:
                     pass
 
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -73,7 +73,7 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
         self.multiplexer = None
         self._draw_title = None
         self._tmux_automatic_rename = None
-        self._tmux_allow_rename = None
+        self._tmux_title = None
         self._screen_title = None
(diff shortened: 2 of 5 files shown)


ranger/gui/ui.py
```diff
@@ -116,14 +116,17 @@ def initialize(self):
             if self.settings.update_tmux_title and 'TMUX' in os.environ:
                 try:
                     self._tmux_automatic_rename = check_output(
-                        ['tmux', 'show-window-options', '-v', 'automatic-rename']).strip()
+                        ['tmux', 'show-window-option', '-v', 'automatic-rename']).strip()
```
aRkedos: that lef out `s` is on purpose. Command is actually `show-window-option`, `tmux show-window-options` produces no output.

toonn: `man tmux` suggests the official command *is* "show-window-option**s**", both work for me but I prefer that one. It doesn't return a result for me unless the `-g` flag is passed though. I think you have to pass a target window otherwise. Though I get no output even if I provide a target, not even when I ask for all options rather than just `automatic-rename`.

ranger/gui/ui.py
```diff
@@ -182,11 +185,6 @@ def destroy(self):
                                   'automatic-rename', self._tmux_automatic_rename])
                 except CalledProcessError:
                     pass
-            else:
```
toonn: I don't understand why you'd remove restoring the option to unset? Is the option unset automatically when we issue `rename-window`?

ranger/gui/ui.py
```diff
@@ -116,14 +116,17 @@ def initialize(self):
...
+                        self._tmux_automatic_rename = 'on'
                 except CalledProcessError:
                     self._tmux_automatic_rename = None
 
         self.update_size()
         self.is_on = True
 
         if self.settings.update_tmux_title and 'TMUX' in os.environ:
+            check_output(['tmux', 'rename-window', 'Ranger'])
             sys.stdout.write("\033kranger\033\\")
```
toonn: Turns out this was broken in tmux 2.7, cf. tmux/tmux#1322.
This is pretty sad because I think it's cross compatible with screen. Let's try to stay screen compatible.

toonn: I guess setting `allow-rename` first would have the control sequence work again. But that's dirty. We should just tell the user in `rc.conf` that they have to allow renaming and then emit a warning if `update_tmux_title` is set but `allow-rename` is not.

ranger/gui/ui.py
aRkedos: As far as I understand it is. When I start tmux `automatic-rename` is `on` by default

When I issue a `rename-window` it turns to `off`

aRkedos:
Wrote a method for handle terminal multiplexers, added settings for screen as well and updated the man page.

ranger/config/rc.conf
```diff
@@ -185,6 +185,10 @@ set update_title false
 # Set the title to "ranger" in the tmux program?
 set update_tmux_title true
 
+# Set the title to "ranger" in the GNU Screen program?
+
```
toonn: Too much whitespace.

ranger/gui/ui.py
```diff
@@ -87,7 +90,6 @@ def initialize(self):
         self.win.leaveok(0)
         self.win.keypad(1)
         self.load_mode = False
-
```
toonn: Too little whitespace?

ranger/gui/ui.py
```diff
@@ -177,6 +182,10 @@ def destroy(self):
             del self.__dict__['vcsthread']
         DisplayableContainer.destroy(self)
 
+<<<<<<< HEAD
```
toonn: Unresolved merge conflict.

aRkedos: should be resolved now

ranger/gui/ui.py
aRkedos: fixed

ranger/gui/ui.py
```diff
@@ -174,19 +183,8 @@ def destroy(self):
...
-                    check_output(['tmux', 'set-window-option',
-                                  'automatic-rename', self._tmux_automatic_rename])
-                except CalledProcessError:
-                    pass
-            else:
-                try:
-                    check_output(['tmux', 'set-window-option', '-u', 'automatic-rename'])
-                except CalledProcessError:
-                    pass
+        if self.multiplexer is not None:
```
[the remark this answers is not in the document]
aRkedos: I thought it would be better so the method does not get called in the first place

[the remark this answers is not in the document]
toonn: I prefer the check to be done where it matters. This spreads around the checks for negligible performance gain and makes it seem to others looking at the code like it's part of the magic incantation.

ranger/gui/ui.py
```diff
@@ -474,6 +460,56 @@ def throbber(self, string='.', remove=False):
...
+                else:
+                    try:
+                        self._tmux_automatic_rename = check_output(
+                            ['tmux', 'show-window-option', '-v', 'automatic-rename']).strip()
+                        if self._tmux_automatic_rename == 'off':
+                            self._tmux_automatic_rename = 'on'
+                    except CalledProcessError:
+                        self._tmux_automatic_rename = None
+
+        if multiplexer is not None and not restore_required:
```
toonn: Are you interpreting `restore_required` as whether to restore the window name *right now*?

If so this should be seperate functions.

ranger/gui/ui.py
```diff
@@ -474,6 +460,56 @@ def throbber(self, string='.', remove=False):
...
+                if multiplexer == 'TMUX':
+                    check_output(['tmux', 'set-window-option',
+                                  'automatic-rename', self._tmux_automatic_rename])
+                elif multiplexer == 'screen':
+                    check_output(['screen', '-X', 'title', self._screen_title])
+            except CalledProcessError:
+                pass
+
+        # Stores the screen window name before renaming it
+        # gives out a warning if $TERM is not "screen"
```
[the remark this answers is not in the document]
toonn: Whoops, `screen-256color`.

Issue #1513, opened 2019-03-31
Ranger crash after resizing a terminal window in multipane mode


nQ8hzkZw:
#### Steps to reproduce
1. Switch to multipane viewmode `~`.
2. Open 3 or 4 tabs ALT+2, ALT+3, ALT+4.
3. Resize (vertically and horizontally) terminal window to a smaller size.

#### Traceback

```
ranger version: ranger-master 1.9.2
Python version: 3.6.7 (default, Oct 22 2018, 11:32:17) [GCC 8.2.0]
Locale: en_US.UTF-8
Current file: '/media/ramdisk/ranger-master/doc'

Traceback (most recent call last):
  File "/media/ramdisk/ranger-master/ranger/gui/curses_shortcuts.py", line 37, in addstr
    self.win.addstr(*args)
_curses.error: addwstr() returned ERR

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/media/ramdisk/ranger-master/ranger/core/main.py", line 190, in main
    fm.loop()
  File "/media/ramdisk/ranger-master/ranger/core/fm.py", line 398, in loop
    ui.redraw()
  File "/media/ramdisk/ranger-master/ranger/gui/ui.py", line 341, in redraw
    self.draw()
  File "/media/ramdisk/ranger-master/ranger/gui/ui.py", line 368, in draw
    DisplayableContainer.draw(self)
  File "/media/ramdisk/ranger-master/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/media/ramdisk/ranger-master/ranger/gui/widgets/view_base.py", line 43, in draw
    DisplayableContainer.draw(self)
  File "/media/ramdisk/ranger-master/ranger/gui/displayable.py", line 256, in draw
    displayable.draw()
  File "/media/ramdisk/ranger-master/ranger/gui/widgets/browsercolumn.py", line 186, in draw
    self._draw_directory()
  File "/media/ramdisk/ranger-master/ranger/gui/widgets/browsercolumn.py", line 323, in _draw_directory
    self.execute_curses_batch(line, drawn.display_data[key])
  File "/media/ramdisk/ranger-master/ranger/gui/widgets/browsercolumn.py", line 122, in execute_curses_batch
    self.addstr(text, attr)
  File "/media/ramdisk/ranger-master/ranger/gui/curses_shortcuts.py", line 42, in addstr
    self.win.move(y, x)
_curses.error: wmove() returned ERR

[...]

Issue #846, opened 2017-04-03
Support draw_borders true in split pane (MC Style) view


zyklotomic:
#### ISSUE TYPE

- Bug report [More of enhancement, not critical] 

#### RUNTIME ENVIRONMENT

- Operating system and version: Gentoo Linux
- Terminal emulator and version: urxvt
- Python version: 3.4.5
- Ranger version/commit: Latest

#### EXPECTED BEHAVIOR
Pressing ~ which splits the pane to show multiple tabs should draw a grid as well as defined 
by the option in rc.conf `set draw_borders true`.

#### CURRENT BEHAVIOR

Does not draw borders in split plane MC mode. 

#### POSSIBLE SOLUTIONS

Draw borders in split pane mode.

#### STEPS TO REPRODUCE

0. Make sure draw_borders has been set to true
1. Open new tabs
2. `~` 
3.  See that no borders are drawn

PR #1742, opened 2019-10-27
Settings draw_borders for view_multipane

zyklotomic:
Implement draw_borders='both','separators','outline' settings for view_multipane

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: nixos 19.03
- Terminal emulator and version: st
- Python version: 3.7.4
- Locale: en_US.UTF-8

#### CHECKLIST

Rest are N/A
    - [ ] Tests have been updated

When I run the makefile on the master branch of the ranger/ranger repository, the tests fail regardless. I am not sure if this is a problem with the master branch. Pylint passes on the only file that I changed: `ranger/gui/widgets/view_multipane.py`.

#### DESCRIPTION

Self-explanatory in the images/videos 

#### MOTIVATION AND CONTEXT

https://github.com/ranger/ranger/issues/846 
Slightly less than 1000 days later!

#### TESTING

pylint on the file that I have changed, `ranger/gui/view_multipane.py`
There is a bug that causes ranger to crash when resizing when in multipane view, but this
is a known issue. https://github.com/ranger/ranger/issues/1513

No affect, changes only in view_multipane

#### IMAGES / VIDEOS

draw_settings=both
2019-10-26-230204_955x341_scrot
draw_settings=separators
2019-10-26-230253_953x344_scrot
hints properly draw over the borders
2019-10-26-233411_1193x376_scrot


zyklotomic:
A slight explanation for the latest commit. I thought that a "floating" border on the currently focused column in `view_multipane` would be visually helpful. I've attached a screenshot on what it looks like. 

There is an implementation problem however. In `settings.py`, `draw_borders` can only take in a string. The new `draw_borders="floating"` option only pertains to `view_multipane` and not `view_miller`. Any thoughts?
2019-12-06-190223_985x357_scrot

toonn:
This looks good to me so far but I haven't had time to properly review it and we're nearing release so unfortunately I'm slipping this. Will try to got back to it in a timely manner!

zyklotomic:
No worries, thanks for the affirmation!

toonn:
This works ok but it's missing some stuff. Like the handling of the draw_borders setting changing. Can you go through view_miller again and copy that behavior more closely? Also, change `floating` to `active-pane` or something.

zyklotomic:
This should handle the draw_borders refreshing upon the setting changing (via `set draw_borders=` inside ranger I assume).
I checked alongside view_miller and don't see anything else that is missing after this fix.

However, how should the border style settings be specified? `active-pane` is specific to view_miller only, a user might also prefer different border styles for different views.

toonn:
Yeah, you do have a point. Let's add a setting, `draw_borders_multipane` which defaults to `None`. `active-pane` isn't valid for `draw_borders`. In multipane viewmode you use this new setting but fall back to `draw_borders` if it's `None`.

zyklotomic:
I hope I didn't miss anything with adding `draw_borders_multipane` as a new setting.

Also oops, I realize I confused multipane and miller in my previous comment, but you caught on anyways.

toonn (changes requested):
You implemented defaulting to the `draw_borders` setting using `or` but `draw_borders_multipane` is set to `"none"` by default, which is a non-empty string and therefore truth-y in python.

zyklotomic:
> Yeah, you do have a point. Let's add a setting, `draw_borders_multipane` which defaults to `None`. `active-pane` isn't valid for `draw_borders`. In multipane viewmode you use this new setting but fall back to `draw_borders` if it's `None`.

I misunderstood the fundamental meaning of `None` here. I thought it referred to the None type in Python and I did not realize ranger behavior was to use the first element in the list of allowed values as the default. Would using an explicit None in the allowed values be better?

ranger/container/settings.py
```diff
@@ -107,6 +108,8 @@
     'cd_tab_case': ['sensitive', 'insensitive', 'smart'],
     'confirm_on_delete': ['multiple', 'always', 'never'],
     'draw_borders': ['none', 'both', 'outline', 'separators'],
+    'draw_borders_multipane': ['none', 'both', 'outline',
```
zyklotomic, line 111:
```suggestion
    'draw_borders_multipane': [None, 'none', 'both', 'outline',
```
Like this?

zyklotomic (commented):
My proposed change

ranger/container/settings.py
toonn: First off, my comment did assume `None` was a valid setting. But I didn't mean you should copy my words verbatim. I'm not going to say "`None` or the equivalent value" all the time, just need to get my point across.

I'm not sure we should use `None`, settings should be case-insensitive and I'm not 100% on how `set draw_borders none` would be handled if we assume only `None` means no borders. If it works fine then I'm ok with it. But do not forget that "false" is an expected value too for compat reasons.

toonn: Distinguishing between no setting at all and being explicitly set to "none" does complicate matters a bit.
Let's also drop the cleverness of `or` and make it explicit with `if setting is None` or something, truthiness is such an ugly implicit thing in python.

zyklotomic: Apologies if that came off that way. I meant that as in that was totally 100% my fault for misunderstanding you.

`None` would only be the set value if the user never explicitly specifies the setting for `set draw_borders_multipane` and to instead fallback onto `set draw_borders`, it would be impossible to manually set `None`, only `'none'` for no borders.

I'll check through the logic and make sure `"false"` is properly handled too. Hopefully ASAP.

toonn: No offense was experienced. Just wanted to clarify.
No rush either, take your time.

toonn (approved):
LGTM! 🎉

toonn:
Thank you for your patience on this PR. Great first contribution! 👍

zyklotomic:
Yay! The most fulfilling part is realizing that I reported this missing feature back in April 2017, almost forgot that.

toonn:
Multipane viewmode was originally added as an april fool's, hence the missing features : )

PR #1743, opened 2019-10-27
Refactored examples/bash_automatic_cd.sh

dmitmel:
Refactored `examples/bash_automatic_cd.sh`

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Terminal emulator and version: kitty 0.14.6
- Python version: 3.7.4 `(default, Oct  4 2019, 06:57:26) [GCC 9.2.0]`
- Ranger version/commit: `ranger-master v1.9.2-379-gc915dff4`
- Locale: `en_US.UTF-8`

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
I basically replaced deprecated/useless syntax with newer, removed unnecessary spawned processes and removed arguments to `mktemp` when creating a temporary file. The script remains compatible with Zsh. **Note that this PR breaks compatibility with POSIX sh.** Although by the name of the file I concluded that this is not needed anyway.

I also applied the same logic to the shell script embedded in `ranger.py`, however that one I made POSIX-compliant, though this requires spawning a `cat` process to read the file. I can workaround that using the `read` builtin or drop POSIX sh support at all, let me know if I should do that.

#### IMAGES / VIDEOS


examples/bash_automatic_cd.sh
```diff
@@ -7,14 +7,14 @@
...
 # original directory.
 
-function ranger-cd {
-    tempfile="$(mktemp -t tmp.XXXXXX)"
-    ranger --choosedir="$tempfile" "${@:-$(pwd)}"
-    test -f "$tempfile" &&
-    if [ "$(cat -- "$tempfile")" != "$(echo -n `pwd`)" ]; then
-        cd -- "$(cat "$tempfile")"
+ranger-cd() {
+    local temp_file chosen_dir
```
toonn: Is this basically the only thing we're giving up POSIX compliance for?

examples/bash_automatic_cd.sh
```diff
@@ -7,14 +7,14 @@
...
+    temp_file="$(mktemp)"
```
toonn: I'd rather change the pattern to `mktemp -t ranger-cd.XXXXXX` or something. I've had platforms like darwin complain when the X's aren't part of the pattern when called from a script even though it works fine on the command line.

examples/bash_automatic_cd.sh
```diff
@@ -7,14 +7,14 @@
...
+    ranger --choosedir="$temp_file" -- "${@:-$PWD}"
+    if chosen_dir="$(<"$temp_file")" && [[ -n "$chosen_dir" ]]; then
```
toonn: This doesn't test whether we're already in the dir.

examples/bash_automatic_cd.sh
```diff
@@ -7,14 +7,14 @@
...
+        cd -- "$chosen_dir"
```
toonn: POSIX says nothing about `--` for `cd` so let's drop it.

examples/bash_automatic_cd.sh
dmitmel: The only thing in this script that is not POSIX-compliant is use of the `$(<"$filename")` construct which is not available in POSIX sh. I can revert back to `cat` then.

examples/bash_automatic_cd.sh
dmitmel: Is this required? Does `cd` give any errors? Tested on my machine with `sh`, `bash` and `zsh` and none of them give any errors when `cd`ing into `$PWD`.

examples/bash_automatic_cd.sh
dmitmel: 1. This was here before I edited this file.
2. Yes, [specification](https://pubs.opengroup.org/onlinepubs/009695399/utilities/cd.html) doesn't specify `--`, although this can lead to bugs when a directory is named `-L` or `-P`. Is this small bug negligible for POSIX-compliance?

examples/bash_automatic_cd.sh
dmitmel: done

examples/bash_automatic_cd.sh
toonn: It's not about errors it's about adding a useless entry to the history. Some people actually use `cd -` to get to the previous dir.

dmitmel: Okay, give me a minute.

examples/bash_automatic_cd.sh
toonn: Hmm, I'm not sure, let's lean towards the status quo then.

examples/bash_automatic_cd.sh
toonn: `local` is definitely not POSIX and neither is `[[ ]]`.

dmitmel: Functions with a dash in their name are neither. Do you want full POSIX compliance?

toonn: Preferably yes. But it's a lot more important in `ranger.py` of course.

examples/bash_automatic_cd.sh
dmitmel: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html#tag_12_02

> <dt><b>Guideline&nbsp;10:</b></dt>
> <dd>The argument <b>--</b> should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the <tt>'-'</tt> character. The <b>--</b> argument should not be used as an option or as an operand.</dd>

As far as I can tell, support for `--` is specified in this document.

dmitmel:
Done. `ranger.py` and `examples/bash_automatic_cd.sh` are now fully POSIX-compliant. Should I squash all commits in one because the history is a bit messy?

examples/bash_automatic_cd.sh
toonn: Good enough for me.

examples/bash_automatic_cd.sh
```diff
@@ -7,14 +7,13 @@
 # To undo the effect of this function, you can type "cd -" to return to the
 # original directory.
 
-function ranger-cd {
-    tempfile="$(mktemp -t tmp.XXXXXX)"
-    ranger --choosedir="$tempfile" "${@:-$(pwd)}"
-    test -f "$tempfile" &&
-    if [ "$(cat -- "$tempfile")" != "$(echo -n `pwd`)" ]; then
-        cd -- "$(cat "$tempfile")"
+ranger_cd() {
```
toonn: Dropping the `function` keyword is POSIX?

examples/bash_automatic_cd.sh
```diff
@@ -7,14 +7,13 @@
...
+    temp_file="$(mktemp -t "ranger_cd.XXXXXXXXXX")"
```
toonn: This change I like. The rest leaves me pretty ambivalent. @Vifon, could use some opinions.

examples/bash_automatic_cd.sh
dmitmel: Yes. Check the output of `shellcheck` with and without it.

dmitmel: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_05

ranger.py
```diff
@@ -9,17 +9,18 @@
...
-if [ "$(cat -- "$tempfile")" != "$(echo -n `pwd`)" ]; then
-    cd "$(cat "$tempfile")"
+if [ -n "$1" ]; then
+    shift
 fi
-rm -f -- "$tempfile"
-return $returnvalue
+"$ranger" --choosedir="$temp_file" -- "${@:-$PWD}"
+return_value="$?"
+if chosen_dir="$(cat -- "$temp_file")" && [ -n "$chosen_dir" ] && [ "$chosen_dir" != "$PWD" ]; then
```
Vifon: Again pwd.

examples/bash_automatic_cd.sh
```diff
@@ -7,14 +7,13 @@
...
+    ranger --choosedir="$temp_file" -- "${@:-$PWD}"
```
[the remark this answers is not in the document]
dmitmel: > we don't need any nesting which `$(pwd)` eases

Could you please explain this in more detail?

> I seem to recall some cases where `$PWD` isn't defined.

For example? `$PWD` is defined by POSIX.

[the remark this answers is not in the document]
Vifon: > Could you please explain this in more detail?

I meant that if we'd need something like `$(cat $(echo somefile))` (I know, a silly example), `$()` is much easier to nest and so often preferred. Since we only need a single level of command substitution, let's go with the older syntax.

Sorry for making it seem more complex than it is. :)

> For example? $PWD is defined by POSIX.

Sadly, I don't remember the details but I remember explicitly preferring `` `pwd` `` in some code I wrote some time ago.

On a second thought, considering I really doubt it all would run on a minimal shell like the original sh and would need bash/zsh instead, it should be fine.

Vifon:
@toonn 

> I'm not convinced of the merits of changing this code just for the sake of newer less-compatible syntax. I'd be fine with renaming the file automatic_cd.sh. Some BSDs have default shells different from bash let's not make their lives harder just because.

Can you confirm our code works on these shells in the first place?

toonn:
@Vifon, the consideration is for people who're using ranger on BSDs already. Remember my effort to port `scope.sh` to POSIX?

examples/bash_automatic_cd.sh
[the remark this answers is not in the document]
Vifon: I guess it'll be ok too and it'll hardly be a loss of any compat considering we were already using this syntax before.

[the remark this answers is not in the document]
Vifon: To clarify: I mean that I'm fine with `$()` too.

toonn:
@Vifon, I'm talking about the opposite. That more POSIX is never a bad thing. You seemed to be saying "Does it even work there now?" as an argument for sticking to dirty dirty bashisms.

dmitmel:
Is there still something that has to be changed in order for this PR to be accepted?

toonn:
Not really, you could rename the file so it doesn't suggest only bash compatibility, maybe add a shebang, if that makes sense for a function definition. Oh, and update the comments to say shell rather than bash.

Only thing blocking it rn is some testing.

dmitmel:
@toonn Done. I've also edited `examples/bash_subshell_notice.sh` since it is already compatible with POSIX sh.

dmitmel:
Whoops, totally forgot about that :) Fixed now.

toonn:
Looks a lot better : )

examples/shell_automatic_cd.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
```
toonn: Let's remove the shebang and mention this script is supposed to be sourced. You can't define a function by running a script anyway.

examples/shell_automatic_cd.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Compatible with ranger 1.4.2 through 1.7.*
```
toonn: Guess this range should be expanded?

examples/shell_subshell_notice.sh
```diff
@@ -1,7 +1,9 @@
+#!/bin/sh
```
toonn: Same thing about the sourcing.

examples/shell_subshell_notice.sh
```diff
@@ -1,7 +1,9 @@
...
+
 # Compatible with ranger 1.5.3 through 1.7.*
```
toonn: Same thing about the range.

examples/shell_automatic_cd.sh
dmitmel: Done.

toonn: Could you add explanation that these files are supposed to be sourced from .bashrc/.zshrc/etc.?

dmitmel: Sure, just did that.

toonn: Great, but it's still missing from `shell_subshell_notice.sh`.

dmitmel: That file already contains the following line:

```
# Add this line to your shell startup file (.bashrc, .zshrc etc) for it to work.
```

toonn: Yeah but it'd be smarter to source the file from the repo in your `.bashrc`, that way you'd get updates if we ever change it.

dmitmel: Done.

BTW, just recalled why I added a shebang to these two files: I did that to specify the type of shell for `shellcheck`.

toonn: Let's do that this way instead, because we don't want to suggest running the scripts:

dmitmel: Wow, didn't know that `shellcheck` had directives (especially this one)! Thanks a lot!

hut:
Hi Dmytro, thank you for the PR. I have a follow-up question: Do you remember why you added the `--` argument in `ranger --choosedir="$temp_file" -- "${@:-$PWD}"`? It prevents people from passing options when using this feature, and `source ranger ranger --help` will think that `--help` is a path.

dmitmel:
Yeah, that's right, I added `--` so that the current working directory path is always treated as a path, even if it starts with dashes. Of course, now that you mention that it doesn't work with options, I have a possible solution: add `--` only when there are no arguments to the function (i.e. `$# == 0`). Then the end user has a choice of adding `--` or not, and the default case never breaks. Honestly I don't know what is the best way to express this in a shell script, but my current idea is the following:
```sh
  if [ "$#" -eq 0 ]; then
    ranger --choosedir="$temp_file" -- "$PWD"
  else
    ranger --choosedir="$temp_file" "$@"
  fi
```

toonn:
Can't you simply do `${@:-"-- $PWD"}`? I'm ok with having people deal with escaping if they pass explicit arguments.

hut:
Thanks for the explanation, @dmitmel :)

@toonn, that sounds like a good solution. I'll go ahead and fix it.

dmitmel:
@toonn Hey, does that thing (`${@:-"-- $PWD"}`) work on bash and zsh? Because I had trouble (syntax errors) when trying it.

hut:
Actually, `$PWD` seems to be always an absolute path, therefore always starting with a `/`. I think we can simply remove the `--`, would you agree?

dmitmel:
Just chcked, the construction suggested by @toonn does indeed work (I had troubles with a different variant, `${@:"-- $PWD"}` because I was stupid and there is no `${variable:default-value}` syntax). It is also POSIX-compatible (checked with `dash`). I propose that we play safe in this regard.

toonn:
@hut, don't think the path being absolute matters: `/this/is/a/valid/absolute/path/with/an --option=launchMissiles /though/admittedly/silly`

toonn:
Btw, the only reason I included the double quotes is to make sure the `--` wasn't parsed as part of the `:-` I'm not sure we need them and since the variable substitution is already in quotes they may actually hinder readability.

dmitmel:
Won't `"-- $PWD"` get parsed as a single argument though?

toonn:
I'm not sure, shell interpretation is too complicated to know by head : )
I'm sure we could split it up somehow, like doing two substitutions `"${@:--- }""${@:-$PWD}"`.
I only wanted to share the idea with my original comment.

dmitmel:
Oh God, this is getting really hacky really fast :sweat_smile:. I suggest we go with by original idea (an `if` statement).

toonn:
An if statement is more confusing imo, especially with the limitations of POSIX test(1). So I'd prefer making use of the variable substitution that is being done anyway.

hut:
I don't think there's a need to get (more) hacky. This works for me in bash even when I'm in a folder called "`/home/hut/x --help`":

`"$ranger" --choosedir="$temp_file" "${@:-$PWD}"`

The double quotes around the `${@...}` ensure that the individual parts of `$@` are properly escaped.

Actually, even a simple `"$ranger" --choosedir="$temp_file" "$@"` works, but that would pass an empty string to ranger as a first argument, and relies on the undocumented (or undefined?) behavior that an empty string as an argument is treated like "$PWD", so we probably shouldn't do that.

toonn:
No, that's documented behavior. And IMO we *should* make sure it works in a directory that happens to contain something looking like an option. Does "$@" escape all the arguments individually? And did you test it in a directory that looks like an option and trips up ranger if `--` is not provided?

PR #1746, opened 2019-11-03
port automatic_cd to fish, fix #960

### LINKED ISSUE #960: Add fish_automatic_cd to examples
ranger's /usr/share/doc/ranger/examples already has bash_automatic_cd.sh

Could we add a fish_automatic_cd.fish? Here's an example

    function ranger-cd                                                               
    
      set tempfile '/tmp/chosendir'                                                  
      /usr/bin/ranger --choosedir=$tempfile (pwd)                                    
    
      if test -f $tempfile                                                           
          if [ (cat $tempfile) != (pwd) ]                                            
             cd (cat $tempfile)                                                       
          end                                                                        
      end                                                                            
    
      rm -f $tempfile                                                                
    
    end                                                                              
    
    function fish_user_key_bindings                                                  
        bind \co 'ranger-cd ; fish_prompt'                                           
    end

Taken from https://stackoverflow.com/questions/18807813/how-to-port-ranger-cd-function-to-fish-shell

In my own config I also add bind -M insert to the user_key_bindings as I use vi mode.

artemisart:
Simple alias to port the automatic_cd example to fish shell, fix #960 

#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

### DIFF
--- a/examples/fish_automatic_cd.fish
+++ b/examples/fish_automatic_cd.fish
@@ -1,5 +1,15 @@
+# Automatically change the directory in fish after closing ranger
+#
+# This is a fish alias to automatically change the directory to the last visited
+# one after ranger quits.
+# To undo the effect of this function, you can type "cd -" to return to the
+# original directory.
+#
 # Note: funcsave save the alias in fish's config files, you do not need to copy
 # this file anywhere, just execute it once
 
 alias ranger-cd 'ranger --choosedir=$HOME/.rangerdir; set RANGERDIR (cat $HOME/.rangerdir); cd $RANGERDIR'
 funcsave ranger-cd
+
+# To bind Ctrl-O to ranger-cd, save this in `~/.config/fish/config.fish`:
+bind \co 'ranger-cd ; fish_prompt'
--- a/examples/fish_automatic_cd.fish
+++ b/examples/fish_automatic_cd.fish
@@ -1,0 +1,5 @@
+# Note: funcsave save the alias in fish's config files, you do not need to copy
+# this file anywhere, just execute it once
+
+alias ranger-cd 'ranger --choosedir=$HOME/.rangerdir; set RANGERDIR (cat $HOME/.rangerdir); cd $RANGERDIR'
+funcsave ranger-cd
--- a/examples/fish_automatic_cd.fish
+++ b/examples/fish_automatic_cd.fish
@@ -8,8 +8,14 @@
 # Note: funcsave save the alias in fish's config files, you do not need to copy
 # this file anywhere, just execute it once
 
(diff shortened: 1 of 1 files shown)


toonn:
Hmm, considering #1743 is making this script posix, is this still necessary?

artemisart:
`source ranger` and the `shell_automatic_cd.sh` script do not work with fish.

toonn:
Ah, yes, ofc. Both ranger and the auto_cd scripts need to be sourced rather than executed and that means they'd need to be fish-compatible syntax.

examples/fish_automatic_cd.fish
```diff
@@ -0,0 +1,15 @@
...
+#
+# This is a fish alias to automatically change the directory to the last visited
+# one after ranger quits.
+# To undo the effect of this function, you can type "cd -" to return to the
+# original directory.
+#
+# Note: funcsave save the alias in fish's config files, you do not need to copy
+# this file anywhere, just execute it once
+
+alias ranger-cd 'ranger --choosedir=$HOME/.rangerdir; set RANGERDIR (cat $HOME/.rangerdir); cd $RANGERDIR'
```
Vifon: A fully fledged function would be more justified here. It seems there is no cleanup of the rangerdir file either.

examples/fish_automatic_cd.fish
```diff
@@ -0,0 +1,15 @@
...
+funcsave ranger-cd
+
+# To bind Ctrl-O to ranger-cd, save this in `~/.config/fish/config.fish`:
+bind \co 'ranger-cd ; fish_prompt'
```
Vifon: Instead of `fish_prompt`, `commandline -f repaint` would be more fitting here. Additionally consider calling it from `ranger-cd` and not from the binding.

examples/fish_automatic_cd.fish
toonn: While I agree from a syntax pov, note that `alias` is just an alias for `function` basically : )

Vifon: I know.  All the more reason to use the actual function syntax.

toonn:
Could you add a `fish_subshell_notice.fish`? Would be nice to have a fish version of all the scripts.

examples/fish_automatic_cd.fish
artemisart: :+1: I changed to a function and added the cleanup.

examples/fish_automatic_cd.fish
artemisart: Indeed that's much better

toonn:
Great contribution 👍

PR #1647, opened 2019-07-30
Add image/webp to known mimetypes

callahad:
#### ISSUE TYPE

- Improvement/feature implementation

#### RUNTIME ENVIRONMENT

- Operating system and version: Linux (NixOS unstable)
- Ranger version/commit: 1.9.2

#### CHECKLIST

- [X] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [X] All changes follow the code style **[REQUIRED]**
- [X] All new and existing tests pass **[REQUIRED]**

#### DESCRIPTION

Ranger now identifies WebP images as having the MIME type of `image/webp`.

#### MOTIVATION AND CONTEXT

WebP is new enough that Python's `mimetypes.guess_type` did not identify them, which meant the `.image` property on `FileSystemObject` was incorrectly set to `False`. This breaks functionality like the `open_all_images` setting.

#### TESTING

Ranger now identifies `.webp` files as images, and `open_all_images` with sxiv works as expected on directories full of WebP images.

PR #1757, opened 2019-11-17
Refine custom mime types for rifle

toonn:
Some changes in the interest of ergonomics.

### DIFF
--- a/ranger/data/mime.types
+++ b/ranger/data/mime.types
@@ -8,24 +8,35 @@
 #  Mimetypes are used for colorschemes and the builtin filetype detection
 #  to execute files with the right program.
 #
+#  You can find the official list of Media Types assigned by IANA here,
+#  https://www.iana.org/assignments/media-types/media-types.xhtml
+#  We deviate from these in certain cases when formats lack an official type
+#  or the type is too generic, application/* for a video format for example.
+#  In such cases we try to adhere to what file(1) returns.
+#
 ###############################################################################
 
-audio/flac              flac
+audio/ape               ape   # [1]
+audio/flac              flac  # [1]
 audio/musepack          mpc mpp mp+
-audio/ogg               oga ogg spx opus
+audio/ogg               oga ogg spx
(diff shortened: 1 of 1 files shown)


ranger/data/mime.types
```diff
@@ -8,24 +8,30 @@
...
 
-audio/flac              flac
+application/javascript  js
+
 audio/musepack          mpc mpp mp+
 audio/ogg               oga ogg spx opus
 audio/wavpack           wv wvc
 audio/webm              weba
 audio/x-ape             ape
+audio/x-flac            flac
```
Vifon: Do we need this line? The Python `mimetypes` module already returns it for the FLAC files. If it didn't, I'd go with `audio/flac` which `file(1)` returns.

toonn: I guess we don't need it. But both `mimetypes` and `file(1)` return `audio/x-flac` for me.

Actually, I'd say we err on the side of caution. #1647 wouldn't be necessary either according the same guidelines but a user *did* experience an issue.

toonn: Seems like NixOS's purity prevents the mimetypes module from loading system-specific files:
```
[nix-shell:~]$ python
Python 2.7.16 (default, Mar  2 2019, 18:34:01) 
[GCC 7.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from mimetypes import guess_type
>>> guess_type('E.mkv')
(None, None)
>>> guess_type('E.flac')
(None, None)
>>> guess_type('E.webp')
(None, None)
>>> guess_type('E.webm')
('video/webm', None)
>>> guess_type('E.weba')
(None, None)
>>> 
```

So I think we should add *a* type for all of these, I don't really care whether we go `flac` or `x-flac`.

Vifon: Isn't it a bug that should be fixed on NixOS's side?

toonn: Not really imo. These aren't official mime types so the OS doesn't really have any business distributing them.

Issue #1620, opened 2019-07-10
Confirmation before any action


jcguu95:
I appreciate the option `confirm_on_delete`, but also wonder if it is possible to let ranger to ask for confirmation for other actions?

Recently, I combined `ranger` and `trash-cli` for a safer workflow. It would be nice if even before moving files to trash `ranger` will ask for a confirmation. Thank you!

toonn:
This is actually a good idea. I seem to remember this coming up a while back too.

Vifon:
Doesn't `:console SOMETHING` already serve as a poor man's confirmation?

toonn:
Noting that a `:confirm` command, which prompts for `Execute <arg>? [Y/n]` would be nice.

catzybluphish:
@vifon @toonn I'd like to use `console` as a simple confirmation, but if paired with `shell`, macros aren't shell-escaped. E.g. `console shell mv %s /tmp` doesn't automatically escape characters. `shell mv %s /tmp` does though. Is there a quick fix for this?

toonn:
@catzybluphish, only double-escaping so the expansion is delayed. I have a fix in mind but having trouble getting it out of there : /
`console shell mv %%s /tmp`

PR #1765, opened 2019-11-28
Added a confirm command.

aRkedos:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Manjaro Linux 5.3.12-1 
- Terminal emulator and version: Termite v15
- Python version: 3.7.4
- Ranger version/commit: Ranger 1.9.2 4ff19f6
- Locale: de_DE.UTF8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
This PR adds a Command which can be used for any Ranger command and asks the user for confirmation if the command shall be executed.

#### MOTIVATION AND CONTEXT
User requested to possibilty to be asked for confirmation before any command in Issue #1620

#### TESTING
Make test has been run.
So far I tested the following commands:
- delete (if confirmation is a activated ranger correctly asks for conformation twice) 
- cd
- set
- chain with different commands. Worked, although chaining delete with cd can be risky.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -2003,43 +2003,35 @@ class confirm(Command):
 
     def execute(self):
         from functools import partial
-        selection_cmd = [
-            'bulkrename', 'chmod', 'delete', 'grep',
-            'meta', 'narrow', 'stage', 'trash', 'unstage'
-        ]
 
-        # Maps the arguments of a command to said command.
-        command_argument_mapping = {}
-        for index in range(1, len(self.args)):
-            if self.args[index] in self.fm.commands.commands:
-                command_argument_mapping[self.args[index]] = ''
-            else:
-                command_argument_mapping[self.args[index-1]] = self.args[index]
-
-        if self.fm.thistab.get_selection():
-            selection = ''
-            for fobj in self.fm.thistab.get_selection():
-                selection += fobj.relative_path + ' '
-
-        # Make sure the selection only gets shown if the command has no
-        # arguments.
-        for argument in range(1, len(self.args)):
-            if self.arg(argument) in self.fm.commands.commands:
-                if self.arg(argument) in selection_cmd:
-                    if command_argument_mapping[self.arg(argument)] == '':
-                        self.args.insert(argument+1, selection)
-
-        # Finally build the command to be shown.
         command = ''
-        for element in range(1, len(self.args)):
-            command += self.args[element] + ' '
+        if len(self.args) > 1 and self.arg(1) != 'chain':
+            for index in range(1, len(self.args)):
+                if index != len(self.args):
+                    command += self.args[index] + ' '
+            self.fm.ui.console.ask(
+                "Confirm execution of: %s (y/N)" % ''.join(command),
+                partial(self._question_callback, command),
+                ('n', 'N', 'y', 'Y'),
+            )
+        else:
+            arguments = self.rest(2).split(';')
+            for item in range(0, len(arguments)):
+                arguments[item] = arguments[item].lstrip()
+            self._chaining(len(arguments) - 1, arguments)
 
-        self.fm.ui.console.ask(
-            "Confirm execution of: %s (y/N)" % ''.join(command),
-            partial(self._question_callback),
-            ('n', 'N', 'y', 'Y'),
-        )
-
-    def _question_callback(self, answer):
+    def _question_callback(self, command, answer):
         if answer == 'y' or answer == 'Y':
-            self.fm.execute_console(self.arg(1) + ' ' + self.rest(2))
+            self.fm.execute_console(command)
+
+    def _chaining(self, recursion_count, commands):
+        from functools import partial
+        if recursion_count >= 0:
+            command = commands[recursion_count]
+            self.fm.ui.console.ask(
+                "Confirm execution of: %s (y/N)" % ''.join(command),
+                partial(self._question_callback, command),
+                ('n', 'N', 'y', 'Y'),
+            )
+
+            self._chaining(recursion_count - 1, commands)
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -2000,38 +2000,33 @@ class confirm(Command):
     given command.
     """
 
-
     def execute(self):
-        from functools import partial
-
         command = ''
         if len(self.args) > 1 and self.arg(1) != 'chain':
             for index in range(1, len(self.args)):
                 if index != len(self.args):
                     command += self.args[index] + ' '
-            self.fm.ui.console.ask(
-                "Confirm execution of: %s (y/N)" % ''.join(command),
-                partial(self._question_callback, command),
-                ('n', 'N', 'y', 'Y'),
-            )
+            self._ask_for_confirmation(command)
         else:
-            arguments = self.rest(2).split(';')
-            for item in range(0, len(arguments)):
-                arguments[item] = arguments[item].lstrip()
+            arguments = [arg.lstrip() for arg in self.rest(2).split(';')]
+            for index, item in enumerate(arguments):
+                arguments[index] = item
             self._chaining(len(arguments) - 1, arguments)
 
     def _question_callback(self, command, answer):
         if answer == 'y' or answer == 'Y':
             self.fm.execute_console(command)
 
-    def _chaining(self, recursion_count, commands):
+    def _ask_for_confirmation(self, command):
         from functools import partial
-        if recursion_count >= 0:
-            command = commands[recursion_count]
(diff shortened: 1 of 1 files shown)


ranger/config/commands.py
```diff
@@ -738,8 +738,8 @@ def is_directory_with_files(path):
             files = [f.relative_path for f in self.fm.thistab.get_selection()]
             many_files = (cwd.marked_items or is_directory_with_files(tfile.path))
 
-        confirm = self.fm.settings.confirm_on_delete
-        if confirm != 'never' and (confirm != 'multiple' or many_files):
+        confirm_setting = self.fm.settings.confirm_on_delete
+        if confirm_setting != 'never' and (confirm_setting != 'multiple' or many_files):
```
aRkedos: These changes were necessary due to potential namespace conflict since the new class is also named "confirm".

ranger/config/commands.py
```diff
@@ -1959,3 +1959,25 @@ def tab(self, tabnum):
...
+    """:confirm <command_to_be_confirmed> + <potential command arguments>
+
+        Can be used with any command. Ranger will ask if it should execute the
+        given command.
+    """
+
+    def execute(self):
+
+        from functools import partial
+        command = self.arg(1)
```
toonn: I don't like this as much. We should show the entire command *at least*. Ideally the command with macros resolved.

toonn (changes requested):
One issue, the rest looks good to me.

ranger/config/commands.py
aRkedos: Okay two questions here:
1. What do you mean by macros? I have done some searching through the codebse and on the Wiki here on Github, are macros just key bindings of commands? So if this is correct, you want to show which command exactly gets executed?

2. Shall I also show selection of files if you operate on commands that use a selection of files (e.g: delete or open_with)?

toonn: Yes, selection's a good idea too. Though I think you should only show that if there's no arguments specified. In other words, always show just exactly what files it'll operate on.

aRkedos:
Okay this is the version we discussed in IRC. It works. Chained commands will be split up and called recursively.

ranger/config/commands.py
```diff
@@ -1959,3 +1959,46 @@ def tab(self, tabnum):
...
+            for index in range(1, len(self.args)):
+                if index != len(self.args):
+                    command += self.args[index] + ' '
+            self.fm.ui.console.ask(
+                "Confirm execution of: %s (y/N)" % ''.join(command),
+                partial(self._question_callback, command),
+                ('n', 'N', 'y', 'Y'),
+            )
+        else:
+            arguments = self.rest(2).split(';')
```
toonn: Let's use a list comprehension here instead.

ranger/config/commands.py
```diff
@@ -1959,3 +1959,46 @@ def tab(self, tabnum):
...
+                arguments[index] = item.strip()
+            self._chaining(len(arguments) - 1, arguments)
+
+    def _question_callback(self, command, answer):
+        if answer == 'y' or answer == 'Y':
+            self.fm.execute_console(command)
+
+    def _chaining(self, recursion_count, commands):
+        from functools import partial
+        if recursion_count >= 0:
```
toonn: Why do we need this count?

ranger/config/commands.py
```diff
@@ -1959,3 +1959,46 @@ def tab(self, tabnum):
...
+            command = commands[recursion_count]
+            self.fm.ui.console.ask(
```
toonn: Why is this duplicated here?

toonn (changes requested):
Starting to look better. Not quite happy yet though.

ranger/config/commands.py
aRkedos: Well I could work with len() of the list, or delete a command from the list after every recursion and check if the len() is still greater 0

ranger/config/commands.py
aRkedos: You mean the module import? Or do you mean I could use the commands[recursion_count] directly in the message?

Well without the module import the partial function is not part of the namespace in the chaining function and would not be executed.

ranger/config/commands.py
toonn: Aren't you already splitting the list on `;`s?

ranger/config/commands.py
toonn: I mean the entire ask part.

aRkedos: Well if I do it recursively I would have to call on that function again and again I guess?

ranger/config/commands.py
aRkedos: Yes but each command is one element of a list . I used the counter to determine how long the recursion should go on. When the counter reaches zero, the first element of the commands list will be called (commands will be called from last to first, I could reverse that order if preferred), on the next iteration the counter will be -1 and the recursion stops. To me it seemed pretty straight forward.

ranger/config/commands.py
aRkedos: I already tried that, I did something like `arguments = [arg.strip(';') for arg in self.args]`. But this separates arguments of a command from the command itsself, e.g. If you do `:confirm chain delete; echo hi` you would geld a list with `['confirm', 'chain', 'delete', 'echo', 'hi']`. 
So what benefit would a list comprehension have?

Please don't get me wrong I will try my best to make this PR fit the needs of ranger, just curious about the reasons in order to learn something :)

aRkedos: Okay figured it out. Just had to iterate over self.rest(2).split(';') in the comprehension. Sometimes I fail to see the obvious. 
:)

ranger/config/commands.py
toonn: What does recursion have to do with duplication of code?

aRkedos: Nothing directly, but I need to call the ask part in the non-recursive part and in the recursive part. Well I could pack that into another function which gets called where needed.

aRkedos: Okay packed it into a helper function and thus avoiding duplicated code and the duplicated module import.

toonn: Can't you just stow it in the non-recursive part? And then "recurse" (to a depth of 1) even for the simple case?

toonn: A helper might be fine too. Go with whatever seems clearest.

ranger/config/commands.py
```diff
@@ -1990,3 +1990,34 @@ def make_safe_path(dst):
...
+            self.fm.notify("Nothing to confirm.")
+        elif self.arg(1) == 'chain':
+            # TODO: This doesn't work for chained chain commands.
+            for command in [arg.lstrip() for arg in self.rest(2).split(';')]:
+                self._ask_for_confirmation(command)
+        else:
+            self._ask_for_confirmation(self.rest(1))
+
+    def _question_callback(self, command, answer):
+        if answer in ('y', 'Y'):
```
Vifon: Let's use `answer.lower()` here, this way there will be no need to enumerate every variant by hand.

ranger/config/commands.py
```diff
@@ -1990,3 +1990,34 @@ def make_safe_path(dst):
...
+    Can be used with any command. Ranger will ask if it should execute the
+    given command.
+    """
+
+    def execute(self):
+        if len(self.args) <= 1:
+            self.fm.notify("Nothing to confirm.")
+        elif self.arg(1) == 'chain':
+            # TODO: This doesn't work for chained chain commands.
+            for command in [arg.lstrip() for arg in self.rest(2).split(';')]:
```
Vifon: Will this ask for each `:chain` subcommand separately? I think it would make more sense to ask once for the whole chain. What was the rationale behind this decision?

vifon (changes requested):

ranger/config/commands.py
toonn: I think this stems from a misunderstanding of something I said. If we do listing of the files a command operates on then we'd have to recurse for chain. But the PR doesn't do that yet so all it would do is make the commands slightly more readable (because they're shorter) when confirming at the cost of having to confirm each command in a chain. But in addition the way the chain is split doesn't work, cf. the TODO I added.

ranger/config/commands.py
```diff
@@ -1959,3 +1959,55 @@ def tab(self, tabnum):
             in sorted(self.modes.keys())
             if mode
         )
+
+
+class confirm(Command):
+    """:confirm <command_to_be_confirmed> + <potential command arguments>
+
+        Can be used with any command. Ranger will ask if it should execute the
```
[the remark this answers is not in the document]
toonn: I could've sworn I caught this.

PR #1791, opened 2019-12-27
Drop the update_icon_title setting

toonn:
I forgot to remove the setting from `rc.conf` when I merged the behavior
into the `update_title` setting.


toonn:
LGTM : >

PR #1799, opened 2019-12-30
cd: fix parsing relative paths with '../' to absolute

wilecoyote2015:
#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
This pull request implements parsing the absolute path constructed from `self.fm.thisdir.path` and `dest_exp` in `_tab_args()` using `os.path.abspath()`. 
This is necessary because otherwise, if the user types relative navigation tokens in `cd`, like `../myDir` or `./myDir`, tab completion will not work, contrary to the usual shell's tab completion.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -169,7 +169,7 @@ class cd(Command):
                 dest_exp = tail
         else:
             dest_exp = ''
-        return (start, dest_exp, os.path.join(self.fm.thisdir.path, dest_exp),
+        return (start, dest_exp, os.path.abspath(os.path.join(self.fm.thisdir.path, dest_exp)),
                 dest.endswith(os.path.sep))
 
     @staticmethod


toonn:
Lets use `normpath` instead. Could you describe the problem better? I noticed some funkiness regarding `..` when I reviewed the last change but I didn't wanna bother understanding, guess it's time I bothered.

wilecoyote2015:
Okay, I've updated the PR with normpath.

The problem is that tab fuzzy tab completion does not work with paths containing dots. For example, when the current directory in ranger is `/usr`, `cd ../e` will only work if `cd_tab_fuzzy false`.
I've just noticed that I've forgotten to mention that the fix is mainly related to fuzzy matching. I'll add that.

I must confess that haven't had the time to understand (and debug) the code of the cd operator completely. 
At first glance, it seems that in `_tab_fuzzy()`, the `..` is parsed as a `token`. Then, in `_tab_fuzzy_match()`, this can of course not be matched with a directory name and therefore the whole fuzzy matching chain fails.

Properly handling traversing previous-level directories using `..` in the present fuzzy matching algorithm seems relatively complicated to me. Hence, within the scope of this pull request, it is proposed to feed the `_tab_fuzzy()` with a normalized absolute path directly.

Plase note that fuzzy matching will still not work if `..` is present as another than the first place in the path. For example, `e/../p/`. However, I struggle to find a realistic scenario where that would be a reasonable use case. 
I'd thus assume that with the proposed change, at least the usual and most probable use case of using `..` at the beginning of a cd pattern for fuzzy tab completion is now working.

I hope that I haven't missed any conceptional aspect regarding the intention of the tab matching algorithm. Please let me know what you think :)

toonn:
Does this change behavior when tab completing without `cd_tab_fuzzy`?

Are you sure `e/../p` isn't normalized to `p`? That's what `normpath` should do.

wilecoyote2015:
> Does this change behavior when tab completing without cd_tab_fuzzy?
AFAIK not, because in `_tab_normal()`, the absolute path is only used to provide its basename for `os.walk()`, so that it should not make a difference whether the path is already normalized or not.

> Are you sure e/../p isn't normalized to p? That's what normpath should do.
You're right, it should get normalized to '/p`. I didn't see that.

toonn:
Not `/p`, just `p`.

wilecoyote2015:
Of course, my bad.

PR #1802, opened 2020-01-03
Add example - Fish alias using ranger as file opener

MatejKastak:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch linux 5.4.6-arch3-1
options: perl,xft,styles,combining,blink,iso14755,unicode3,encodings=eu+vn+jp+jp-ext+kr+zh+zh-ext,fade,transparent,tint,pixbuf,XIM,frills,selectionscrolling,wheel,slipwheel,cursorBlink,pointerBlank,scrollbars=plain+rxvt+NeXT+xterm
- Python version: Python 3.8.1
- Ranger version/commit: ranger 1.9.2
- Locale: 

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Add new example of using fish alias as file opener.

#### MOTIVATION AND CONTEXT

I am using ranger as a file opener and I thought that others might find it also useful.
`urxvt -e fish -c "ranger-open"` will open a new terminal with the ranger and wait for the user to select a file. If you select a file it will close the selector window and open the file.

### DIFF
--- a/examples/fish_automatic_open.fish
+++ b/examples/fish_automatic_open.fish
@@ -12,7 +12,7 @@ function ranger-open
 	set ranger_bin (which ranger)
 	$ranger_bin --choosefile=$dir $argv
 	echo (cat $dir)
-	nohup xdg-open (cat $dir) &
+	nohup xdg-open (cat $dir) > /dev/null &
 	rm $dir
 end
 funcsave ranger-open
--- a/examples/fish_automatic_open.fish
+++ b/examples/fish_automatic_open.fish
@@ -1,0 +1,18 @@
+# Automatically open the file with xdg-open after selecting it
+#
+# This is a fish alias to automatically open selected the file and then exit
+# the ranger. It ilustrates how to use ranger as a file opener. For example you
+# can use 'urxvt -e fish -c "ranger-open"'.
+#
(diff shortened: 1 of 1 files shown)


vifon:
I know you closed it, but what's the benefit of having it as a fish function? It doesn't seem to utilize the fact it's running in the same shell and your scenario closes fish as soon as you close ranger. It could be a standalone script just as well, which would also make it more portable and easier to run.

MatejKastak:
Hello, thanks for the feedback :)
I am using this as a file opener, that spawns ranger and lets me select a file and then opens it (closing the file selector dialog).

> but what's the benefit of having it as a fish function?
- for short scripts like this, I prefer having them defined as functions (subjective)
    - they are easier to version
    - fish has a little description
    - easier to install

> It could be a standalone script just as well, which would also make it more portable and easier to run.
- you are 100% right, the portability is not good
- and there is no benefit to running in the same shell

What do you think we should do?
1) I can reopen the PR and rewrite it POSIX shell - if you think this is of any benefit to other users
2) We can keep this PR closed

toonn:
If it were posix it could probably go into the examples/ directory. Otherwise the wiki might still be suitable.

PR #1809, opened 2020-01-15
fix the wrong length of previewing ansi text when setting wrap_plaint…

kevinhwang91:
When setting wrap_plaint_previews, calculate the wrong length of the text because of including the ansi char.
 
issue:

after patch:

### DIFF
--- a/ranger/gui/ansi.py
+++ b/ranger/gui/ansi.py
@@ -8,9 +8,8 @@ from __future__ import (absolute_import, division, print_function)
 
 import re
 
-from ranger.ext.widestring import WideString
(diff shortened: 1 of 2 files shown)


kevinhwang91:
@toonn 
Fix some bug what I consider:
1. Disable wrap lines in pager;
2. Reset max scroll at wrap_plaintext_previews mode;

I found the current code couldn't support to preview wrapping line for the last char of east asian(char may be missing at the end of the wrapping line).
I think `WideString` inside `ranger/gui/ansi.py` is a bad design, so I remove it.

If some wrong what I did, please correct me.

PR #1838, opened 2020-02-12
Allow equal mtime for cached previews

### LINKED ISSUE #1837: Support for symbolic links in the image cache directory
### Current Behavior
In the current version of ranger an image in the cache directory of ranger is outdated if it's last modified timestamp is smaller than **or equal** to the last modified timestamp of the file it's a preview of.  
See: 
https://github.com/ranger/ranger/blob/3f8e7c14103a6570b0e55fbcf84242c86f42a7cb/ranger/core/actions.py#L1127

#### Expected Behavior / Context

I just wanted to ask whether the the `>` could be changed to `>=` as it would allow to use a work around to prevent the execution of the scope.sh script.  
(Creating a symbolic link for actual images would prevent the execution of the scope.sh script and therefore save the time required to execute this script)
(Another possibility to support symbolic links would be to extend the 'last modified' clause with os.path.islink)

toonn:
This allows the use of symlinks to prevent copying files to the cache.

Fixes #1837

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -37,8 +37,6 @@ from ranger.container.settings import ALLOWED_SETTINGS, ALLOWED_VALUES
 
 MACRO_FAIL = "<\x01\x01MACRO_HAS_NO_VALUE\x01\01>"
 
-PY3 = version_info[0] >= 3
-
 LOG = getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


toonn:
I think we should forego checking the mtime and just include mtime in the hash, and the inode and the device ID.
What do you think, @vifon?

vifon:
Yes, it makes sense to "unify" the checking logic this way into a single hash.

Issue #738, opened 2016-12-16
Show copy speed in status line


SammysHP:
Show copy speed in status line

d3vr:
This feature is definitely something I would love to see implemented in ranger.

sebastianst:
any progress here?

toonn:
Don't think anyone's started work on this. I don't consider this a high priority feature because I recommend against doing considerable file transfers with ranger. Tools like rsync are more suitable.

toonn:
How would it be hard to write the rsync command? It's a command like any other, you get used to it through using it.

You only have to come up with the command once though, then you can map it in ranger so you can "paste with rsync" rather than ranger. Something along the lines of `map pr shell rsync -a --info=progress2 %c %d` would probably do. It'll rsync all the yanked/cut files to the current directory, note that cut files won't be removed from their original location, to do that you need either two mappings or to implement a custom command in `commands.py`.

Note that I have nothing against a useful progress indication. The issue's been open for 2 years though and no one has stepped forward to work on it. So either you have to use a workaround like I described or step forward yourself and submit a PR. If you need some help with the latter, visit us in #ranger@freenode, we'd be glad to help a new contributor : )

moneroexamples:
Copying speed is one of the main reasons I still use `mc`.

toonn:
@moneroexamples, and the rsync keybinding I listed before isn't a decent workaround? It has the advantage of being robust since rsync's tried and tested and working today : )

moneroexamples:
the `pr` blocks entire ranger process. I wonder if it would be possible to add the status into regular copy progress status, so that you can still user ranger (as you can do now when copying/moving files), and see the speed in the progress bar?

toonn:
You can use the fork flag but then you don't see the progress and you can use the `t` flag to run the rsync in a new terminal (caveats do apply).

sebastianst:
It still is just much nicer, cleaner and less error prone if ranger supports it out of the box and as an internal process. Especially because of this situation that an external command would either block ranger or run in a new terminal where you have to switch to back and forth.

toonn:
I'm just offering workarounds since no one's working on this.

alscaldas:
[...]

PR #1840, opened 2020-02-15
Separate progress tracking; show copy speed in TaskView

mik13ST:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch Linux, kernel 5.4.15-arch1-1
- Terminal emulator and version: termite v15
- Python version: 3.8.1
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

A bunch of users wished to be able to know the copy/move speed of a task. The only statistic tracked was progress percent and it was done in place. The percent tracking was separated into a ProgressTracker class which allowed clean implementation of copy/move speed calculation. 

#### MOTIVATION AND CONTEXT

#738
Although the issue specifically mentions status line, I believe this satisfies the request. 

#### TESTING

Testing was done manually on a single drive (move was immediate). I used one complex directory with many small files (`node_modules`) and a second directory containing one big (~600MB) binary file. I duplicated one of the directories each time to test the progress indication. During the copying I switched to TaskView (W key) and observed the progress. I tested both of these task running at the same time. Switching does work, but the speed is not calculated correctly when the task is paused. Testing was only done on ext4 partitions.

Directory loading (which also has progress) was also tested. Running a high disk IO operation (`dd if=/dev/zero of=./test1.img bs=1G count=10 oflag=dsync`) allowed me to simulate a slow IO storage. This was crucial for testing of directory loading. 

The one slight problem is the copy speed is being displayed for directory loading as well and the units happen to be in bytes.

The impact on CPU utilization or actual copy speed was not tested in any way.

### DIFF
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -19,6 +19,7 @@ except ImportError:
     HAVE_CHARDET = False
 
 from ranger.core.shared import FileManagerAware
+from ranger.core.progress_tracker import ProgressTracker
 from ranger.ext.safe_path import get_safe_path
 from ranger.ext.signals import SignalDispatcher
 from ranger.ext.human_readable import human_readable
@@ -28,10 +29,10 @@ class Loadable(object):
     paused = False
     progressbar_supported = False
 
-    def __init__(self, gen, descr):
+    def __init__(self, gen, descr, progress=ProgressTracker(0)):
         self.load_generator = gen
         self.description = descr
-        self.percent = 0
+        self.progress = progress
 
     def get_description(self):
         return self.description
@@ -60,10 +61,10 @@ class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instan
         self.original_path = dest if dest is not None else self.fm.thistab.path
         self.overwrite = overwrite
         self.make_safe_path = make_safe_path
-        self.percent = 0
+        self.progress = ProgressTracker(self._calculate_size(1))
         if self.copy_buffer:
             self.one_file = self.copy_buffer[0]
-        Loadable.__init__(self, self.generate(), 'Calculating size...')
+        Loadable.__init__(self, self.generate(), 'Calculating size...', self.progress)
 
     def _calculate_size(self, step):
         from os.path import join
@@ -89,10 +90,7 @@ class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instan
 
         from ranger.ext import shutil_generatorized as shutil_g
         # TODO: Don't calculate size when renaming (needs detection)
-        bytes_per_tick = shutil_g.BLOCK_SIZE
-        size = max(1, self._calculate_size(bytes_per_tick))
         size_str = " (" + human_readable(self._calculate_size(1)) + ")"
-        done = 0
         if self.do_cut:
             self.original_copy_buffer.clear()
             if len(self.copy_buffer) == 1:
@@ -110,12 +108,13 @@ class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instan
                         self.fm.tags.tags[new_path] = tag
                         self.fm.tags.dump()
                 n = 0
+                last_n = 0
                 for n in shutil_g.move(src=fobj.path, dst=self.original_path,
                                        overwrite=self.overwrite,
                                        make_safe_path=self.make_safe_path):
-                    self.percent = ((done + n) / size) * 100.
+                    self.progress.step(n - last_n)
+                    last_n = n
                     yield
-                done += n
         else:
             if len(self.copy_buffer) == 1:
                 self.description = "copying: " + self.one_file.path + size_str
@@ -124,6 +123,7 @@ class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instan
             for fobj in self.copy_buffer:
                 if os.path.isdir(fobj.path) and not os.path.islink(fobj.path):
                     n = 0
+                    last_n = 0
                     for n in shutil_g.copytree(
                             src=fobj.path,
                             dst=os.path.join(self.original_path, fobj.basename),
@@ -131,17 +131,18 @@ class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instan
                             overwrite=self.overwrite,
                             make_safe_path=self.make_safe_path,
                     ):
-                        self.percent = ((done + n) / size) * 100.
+                        self.progress.step(n - last_n)
+                        last_n = n
                         yield
-                    done += n
                 else:
                     n = 0
+                    last_n = 0
                     for n in shutil_g.copy2(fobj.path, self.original_path,
                                             symlinks=True, overwrite=self.overwrite,
                                             make_safe_path=self.make_safe_path):
-                        self.percent = ((done + n) / size) * 100.
+                        self.progress.step(n - last_n)
+                        last_n = n
                         yield
-                    done += n
         cwd = self.fm.get_directory(self.original_path)
         cwd.load_content()
 
@@ -427,7 +428,7 @@ class Loader(FileManagerAware):
             except Exception as ex:  # pylint: disable=broad-except
                 self.fm.notify(
                     'Loader work process failed: {0} (Percent: {1})'.format(
-                        item.description, item.percent),
+                        item.description, item.progress.percent),
                     bad=True,
                     exception=ex,
                 )
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -19,7 +19,7 @@ except ImportError:
     HAVE_CHARDET = False
 
 from ranger.core.shared import FileManagerAware
-from ranger.core.progress_tracker import ProgressTracker
+from ranger.core.progress_tracker import FileProgressTracker
 from ranger.ext.safe_path import get_safe_path
 from ranger.ext.signals import SignalDispatcher
 from ranger.ext.human_readable import human_readable
@@ -29,7 +29,7 @@ class Loadable(object):
     paused = False
     progressbar_supported = False
 
-    def __init__(self, gen, descr, progress=ProgressTracker(0)):
+    def __init__(self, gen, descr, progress=FileProgressTracker(0)):
         self.load_generator = gen
         self.description = descr
         self.progress = progress
@@ -61,7 +61,7 @@ class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instan
         self.original_path = dest if dest is not None else self.fm.thistab.path
         self.overwrite = overwrite
         self.make_safe_path = make_safe_path
-        self.progress = ProgressTracker(self._calculate_size(1))
+        self.progress = FileProgressTracker(self._calculate_size(1))
         if self.copy_buffer:
             self.one_file = self.copy_buffer[0]
         Loadable.__init__(self, self.generate(), 'Calculating size...', self.progress)
--- a/ranger/core/progress_tracker.py
+++ b/ranger/core/progress_tracker.py
@@ -25,3 +25,16 @@ class ProgressTracker(object):
         if self.total_items == 0:
             return 0
         return (self.done / self.total_items) * 100
+
+    def get_status_text(self):
+        raise NotImplementedError('Please Implement this method')
+
+
+class DirectoryProgressTracker(ProgressTracker):
+    def get_status_text(self):
+        return '{:3.2f}%'.format(self.percent)
+
+
+class FileProgressTracker(ProgressTracker):
+    def get_status_text(self):
+        return '{:3.2f}%'.format(self.percent)
--- a/ranger/core/progress_tracker.py
+++ b/ranger/core/progress_tracker.py
@@ -4,11 +4,12 @@
 """
 A Progress Tracker of any Loadable that manages all the progress
 related statistics.
-
-Currently, only percent reporting is implemented.
 """
 
 from __future__ import division
+
+import time
+from ranger.ext.human_readable import human_readable
 
 
 class ProgressTracker(object):
@@ -16,6 +17,7 @@ class ProgressTracker(object):
     def __init__(self, total_items=0):
         self.total_items = total_items
         self.done = 0
+        self.start_time = time.time()
 
     def step(self, stepsize=1):
         self.done += stepsize
@@ -25,6 +27,12 @@ class ProgressTracker(object):
         if self.total_items == 0:
             return 0
         return (self.done / self.total_items) * 100
+
+    @property
+    def average_rate(self):
+        time_diff = time.time() - self.start_time
+        rate = self.done / time_diff
+        return rate
 
     def get_status_text(self):
         raise NotImplementedError('Please Implement this method')
@@ -37,4 +45,5 @@ class DirectoryProgressTracker(ProgressTracker):
 
 class FileProgressTracker(ProgressTracker):
     def get_status_text(self):
-        return '{:3.2f}%'.format(self.percent)
+        average_rate = human_readable(self.average_rate, '') + '/s'
+        return '{:3.2f}% {:>7}'.format(self.percent, average_rate)
--- a/ranger/core/progress_tracker.py
+++ b/ranger/core/progress_tracker.py
@@ -1,0 +1,27 @@
+# This file is part of ranger, the console file manager.
+# License: GNU GPL version 3, see the file "AUTHORS" for details.
+
+"""
+A Progress Tracker of any Loadable that manages all the progress
+related statistics.
+
+Currently, only percent reporting is implemented.
+"""
+
+from __future__ import division
+
+
+class ProgressTracker(object):
+
+    def __init__(self, total_items=0):
+        self.total_items = total_items
+        self.done = 0
+
+    def step(self, stepsize=1):
+        self.done += stepsize
+
+    @property
+    def percent(self):
+        if self.total_items == 0:
+            return 0
+        return (self.done / self.total_items) * 100
--- a/ranger/gui/widgets/taskview.py
+++ b/ranger/gui/widgets/taskview.py
@@ -53,9 +53,9 @@ class TaskView(Widget, Accumulator):
                         clr.append('selected')
 
                     descr = obj.get_description()
-                    if obj.progressbar_supported and obj.percent >= 0 and obj.percent <= 100:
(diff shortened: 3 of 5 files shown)


ranger/core/loader.py
```diff
@@ -60,10 +61,10 @@ def __init__(self, copy_buffer, do_cut=False, overwrite=False, dest=None,
         self.original_path = dest if dest is not None else self.fm.thistab.path
         self.overwrite = overwrite
         self.make_safe_path = make_safe_path
-        self.percent = 0
+        self.progress = ProgressTracker(self._calculate_size(1))
```
mik13ST: The build is failing because the `size` variable is unused. This lead me to this which could use the `size` variable. Or maybe remove the `size` variable and use its definition here.

While looking at the `size` variable definition, I realized I could use the trick with the `max(1, x)` to avoid possible division by zero.

ranger/core/progress_tracker.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @property
+    def percent(self):
+        if self.total_items == 0:
+            return 0
+        return (self.done / self.total_items) * 100
+
+    @property
+    def average_rate(self):
+        time_diff = time.time() - self.start_time
```
FichteFoll: Just something to think about: If the progress has been paused (for some reason), the average will be wrong. Furthermore, averaging over the entire period produces results that apply to the process as a whole, but for the most part these are time intervals that humans don't parse well or are even interested in. The current speed calculated with a sliding window, for example, is much more useful for speed calculation because it considers recency and values it more (or only) while abandoning old and outdated information.

To put it into actual suggestions: You could either create a reference point on every `step` call with both the progress at the time and the timestamp, or you create a fifo queue of limited size that you push discrete rates since the last update to and then calculate the average with those rates only.

Not saying you should definitely do that here or now, but something I wanted to mention nonetheless.

ranger/core/progress_tracker.py
```diff
@@ -0,0 +1,33 @@
...
+        self.total_items = total_items
+
+    def step(self, stepsize=1):
+        self.done += stepsize
+
+    @property
+    def percent(self):
+        if self.total_items == 0:
+            return 0
+        return (self.done / self.total_items) * 100
```
FichteFoll: If total and done are both integers, Python 2 will perform integer division here. To prevent that, multiply with `1.0` before the division.

ranger/core/progress_tracker.py
```diff
@@ -0,0 +1,33 @@
...
+
+import time
+
+class ProgressTracker(object):
+    def __init__(self, total_items=0):
+        self.total_items = total_items
+        self.done = 0
+        self.start_time = time.time()
+
+    def set_total_items(self, total_items):
```
FichteFoll: You may want to consider allowing to set `total_items` explicitly (since it's a "public" attribute) instead of having this boilerplatey setter. Many attributes in the standard library also operate like this.

ranger/gui/widgets/taskview.py
```diff
@@ -53,9 +54,11 @@ def draw(self):
...
 
                     descr = obj.get_description()
-                    if obj.progressbar_supported and obj.percent >= 0 and obj.percent <= 100:
-                        self.addstr(y, 0, "%3.2f%% - %s" % (obj.percent, descr), self.wid)
-                        wid = int((self.wid / 100) * obj.percent)
+                    if obj.progressbar_supported and 0 <= obj.progress.percent <= 100:
+                        rate = human_readable(obj.progress.average_rate, '') + "/s"
+                        status = '{:3.2f}% {:>7} - {:s}'.format(obj.progress.percent, rate, descr)
+                        self.addstr(y, 0, status, self.wid)
+                        wid = int((self.wid / 100) * obj.progress.percent)
```
FichteFoll: Not your change, but when reading `wid` I think of `w_id`, or more precisely "window ID" and not `width`. That's a quite unfortunate shortening, but it's also used in the class attribute.

ranger/core/progress_tracker.py
toonn: No, please use `from __future__ import division` instead.

FichteFoll: Oh, yes, please do that. I haven't used future imports in years.

ranger/gui/widgets/taskview.py
mik13ST: Ok, I added that to my TODO list. I might rename this before I try expanding the progress to multiple rows (if I ever justify doing that).

ranger/core/progress_tracker.py
mik13ST: Using FIFO buffer to implement sliding window speed calculation sounds good to me, added to my TODO list.

ranger/core/loader.py
mik13ST: The `size` variable was removed. Using the `max(1, x)` trick in ProgressTracker didn't seem to provide any benefits over the `if total_items = 0` condition.

mik13ST:
> The one slight problem is the copy speed is being displayed for directory loading as well and the units happen to be in bytes.

I thought about this and came up with some possible solutions:
1. Create a `FileOnlyProgressTracker` class which is based on (inherits from) the general `ProgressTracker` and overloads the `average_rate` property function.
2. Let `ProgressTracker` know what unit the progress is using (bytes, files). It could be a constructor parameter.
3. Add units conditionally in the GUI (if obj.is_directory then use "files" as unit).
4. Use a formatting function which would be passed as an argument to the constructor and then used in the `average_rate` property function.

Any other ideas?

I like option 2, although I am unhappy with making the `ProgressTracker` less general.

toonn:
I think 2 is fine. It's not *less* general if it works for more cases.

mik13ST:
Ok, 2 it is. The other problem is using the `human_readable()` to format the progress rate. This function should not engage for files progress (not possible with option 2).

The best solution I see here is to create a `human_readable_rate(rate, units)` which would decide using the units whether and how to format the rate. The units could then be a constant (defined in the `ProgressTracker`.

Is this acceptable? Any other ideas? Is option 2 still the best way to do this?

FichteFoll:
What about having a function for the base progress tracker that is responsible for the human readable representation of the whole status and that has information about both the unit (bytes/files) and how it is supposed to be represented (prefixes or not) in a single overloaded function for `FileTracker` and `DirectoryTracker` subclasses?

toonn:
I'm fine with subclasses too. Maybe we can simply not use the `human_readable` function for directories?

Otoh, a design where the units progress tracking shows are configurable would be cool. For users transferring a directory with huge files a MB/s view would probably be more useful than 0 files/sec or "5 files transferred" being show for ages.

mik13ST:
> I'm fine with subclasses too. Maybe we can simply not use the human_readable function for directories?

This is similar to the option 3, having the login in the GUI. I am afraid this would become obsolete the second somebody decides to implement more progress stats (e.g. "done: 3/10 files")

> Otoh, a design where the units progress tracking shows are configurable would be cool. For users transferring a directory with huge files a MB/s view would probably be more useful than 0 files/sec or "5 files transferred" being show for ages.

To clarify what directory progress I am talking about: it is the progress of listing a directory (when a directory preview is requested). It comes from `container/directory.py`. IMO it makes sense to use files as unit for it. When copying a directory, the units will be bytes (both `copy2` and `copytree` from `ext/shutil_generatorized.py` yield bytes).

Ok, I will try the two subclasses approach. There should be a `get_status_text()` which will be overloaded by both classes (`DirectoryProgressTracker` and `FileProgressTracker`). This allows for improving the file copy/move progress without impacting the directory listing progress or to use different formatting (units, human readability) for each of them. Improvements of the directory listing progress could be done later.

ranger/core/loader.py
```diff
@@ -29,7 +29,7 @@ class Loadable(object):
     paused = False
     progressbar_supported = False
 
-    def __init__(self, gen, descr, progress=ProgressTracker(0)):
+    def __init__(self, gen, descr, progress=FileProgressTracker(0)):
```
FichteFoll: Mutable default values in function calls are not intended in most cases because they will be shared for all invocations the default value is used. I'm confident you want to use `=None` here and check for that later.

Besides, imo that `0` can be removed since it's the default anyway.

ranger/core/progress_tracker.py
```diff
@@ -37,4 +45,5 @@ def get_status_text(self):
 
 class FileProgressTracker(ProgressTracker):
     def get_status_text(self):
-        return '{:3.2f}%'.format(self.percent)
+        average_rate = human_readable(self.average_rate, '') + '/s'
+        return '{:3.2f}% {:>7}'.format(self.percent, average_rate)
```
FichteFoll:
```suggestion
        rate = human_readable(self.average_rate, '')
        return '{:3.2f}% {:>5}/s'.format(self.percent, rate)
```

Are 5 characters even enough?

ranger/core/loader.py
mik13ST: This indeed isn't necessary, default value replaced with `=None`.

ranger/core/progress_tracker.py
mik13ST: I don't like your suggestion, because in it the contents of `rate` variable happen to have incorrect units if taken out of context (e.g. `60MB`). I'd like to have the status items either preformatted completely on a separate row (this case) or formatted in the template string (e.g. percent). This would be a hybrid between the two.

I will rename the variable to `rate` no problem. I might eventually rename it back to `average_rate` when I decide to implement current rate so there will be two rates to distinguish between.

>Are 5 characters even enough?

It should be, the `human_readable()` seems to never produce more than 4 digits + separator + unit(1 character). Since empty separator is used, this turns out to be 5 characters. See the function docstring for example outputs. Anyways, I am unhappy with having the character count tightly coupled to the `human_readable()` output format. The `.format()` should not know how many characters does each statistic need. However, I have no other solution to prevent jumping of the whole text when the rate changes (e.g. from `40M/s` to `39.9M/s`). Maybe fixed lengths for all of them?

FichteFoll: > I'd like to have the status items either preformatted completely on a separate row (this case) or formatted in the template string (e.g. percent). This would be a hybrid between the two.

I don't understand that argument. On one hand, you want the unit in the final format string, and on the other you don't.

> The `.format()` should not know how many characters does each statistic need.

But that's precisely what `.format` is for. I don't see a reasonbale way to do it differently.

> 4 digits + separator + unit(1 character)

I see, that makes sense. I didn't check the docstring or source for `human_readable`, so I wasn't sure about its result. Since the rate is given in bytes, afaik, that should also be part of the unit and become `40MB/s`.

mik13ST: > I don't understand that argument. On one hand, you want the unit in the final format string, and on the other you don't.

I am sorry for confusing explanation. Alright, ignore the two categories (preformatted completely, formatted in the template string), I realized they are not the actual argument. In your suggestion: if I would pause the code before executing the `format()`, the immediate value would lack the `/s` and therefore it wouldn't be a valid rate representation. It would be a representation of data size. If the variable was named something like `data_per_second` then the variable name *would* actually correspond to its content and I would be fine with it. I hope its clear now.

> But that's precisely what .format is for. I don't see a reasonbale way to do it differently, unless human_readable gets a parameter for column width or number precision.

Should I refactor `human_readable()` then? I don't think I want to go that far right now though. 

> Since the rate is given in bytes, afaik, that should also be part of the unit and become 40MB/s.

That's another reason to refactor the function. How should this be kept compatible with the tree view (`gui/widgets/browsercolumn.py`) which (for some reason) uses single character file size units ('K', 'M', 'G' etc.)? Something like a `abbreviate_units` function parameter?

FichteFoll: >  It would be a representation of data size.

Yes, that's a reason I can get behind.

> Should I refactor `human_readable()` then?

Definitely not my decision, but I would refrain from doing so. I don't see much value in that, unless there are different places where it would be useful. I believe having the column width in the using format string is fine.

> That's another reason to refactor the function.

Eh, no, you could just add the "B" for display in the format string The `human_readable` function doesn't need to know the unit to do its job, you can just append it afterwards, if desired.

toonn: @FichteFoll, what do you mean by "implicit format spec numbering?" The empty `{}`'s for positional formatting arguments work perfectly well in python 2.7.

FichteFoll: Yes, that is what I meant. I definitely remember that not working for me in 2.6, so maybe it was added for 2.7. However, since ranger still supports 2.6, this is still relevant.

Unfortunately, I don't have a 2.6 interpreter at hand to verify.

toonn: Ah, yep, you're right. Nvm, me, I'm just over here doing nothing of interest...

mik13ST: > Yes, that's a reason I can get behind. Ideally we'd put the human_readable call in the same line as .format, but that might be a bit too long.

Ok, I will use your suggestion but with a different variable name.

> Eh, no, you could just add the "B" for display in the format string The human_readable function doesn't need to know the unit to do its job, you can just append it afterwards, if desired.

That would work for all but bytes. In case the size is below 1K the `human_readable()` adds a `B` as a unit. This would result in `BB/s` in these cases. I don't see a way to correct the units without refactoring the function or creating a wrapper(s) for it. I will investigate this in this PR thread but separated from this discussion.

> However, since ranger still supports 2.6, this is still relevant.

Alright, will be testing all changes in Python 2.6 from now on. I will probably use a Vagrant box, because I don't want to mess up my system by downgrading packages. This box looks promising: https://app.vagrantup.com/pcbadger/boxes/centos6-64-python2-6.

toonn: Checking docs don't mention "Since 2.7 or 3.x" is usually enough, and running `make test` with a python 2, 2.7 is fine. Python 3 doesn't support the pylint version we're still using.
I'm working on updating to newer pylint but it involves manually implementing some checkers to prevent python 2 incompatibilities from sneaking in.

mik13ST:
The use of `human_readable()` to format copy speed is problematic. The units this function adds are `B`, `K`, `M` etc. which corresponds to how common UNIX utilities format bytes when requested usually by `-h` (e.g. `ls -h`, `df -h`). This makes sense until `/s` is added to the units to make the value represent data transfer (copy speed in this case). The units become: `B/s`, `K/s`, `M/s`.

Possible solutions:
1. Add a `B` to the end which would turn most units into what we want: `BB`, `KB`, `MB`. But `BB` is not a valid unit.
2. Conditionally add the `B` character in the `human_readable()`. There could be an argument called e.g. `abbreviate_units` used to switch this behavior.
3. Create a wrapper around `human_readable()` that would add the `B` character conditionally ("if it is missing then add it"?). It could be named e.g. `human_readable_...`. That's difficult to name which means it probably is not the correct way to do it.
4. Create a `human_readable_speed()` which would return the correct units of data transfer: `B/s`, `MB/`, `KB/s`. It would be internally calling the `human_readable()`.

Any other ideas? Is this actually a problem worth doing the changes?

There is a [comment](https://github.com/ranger/ranger/blob/bc47bc24fe45aaa4b8f3502a5d11187962692bd1/ranger/ext/human_readable.py#L29) in the `human_readable()` warning about possible performance impact. Should I test the performance when doing the changes? It should be trivial since (with little modification) the function can be used outside ranger.

toonn:
Factor out the common part of `human_readable` maybe call it [`engineering_form`](https://en.wikipedia.org/wiki/Engineering_notation) or something, I think that's the closest conceptually? I imagine it'd return not just the number but a tuple with `(number, power_of_ten)`.

Then have `human_readable` and `human_readable_rate` call that function and add the appropriate units.

mik13ST:
Sounds good, will split into private `engineering_form` (or something) and two public methods that format it based on the exponent.

The `human_readable()` still uses the old string templates (e.g. [here](https://github.com/ranger/ranger/blob/bc47bc24fe45aaa4b8f3502a5d11187962692bd1/ranger/ext/human_readable.py#L35)). Should I first refactor that to use the now preferred format function? Should I do this refactoring in a separate branch/PR?

toonn:
Sure, note that implicit format specs aren't allowed though, so no empty {}'s.

mik13ST:
The research conducted in #1891 revealed that the old %-formatting is the fastest of possible python 2.6 compatible solutions. The `format` function will not be used here unless %-formatting becomes deprecated.

mik13ST:
> Factor out the common part of `human_readable` maybe call it [`engineering_form`](https://en.wikipedia.org/wiki/Engineering_notation) or something, I think that's the closest conceptually? I imagine it'd return not just the number but a tuple with `(number, power_of_ten)`.
> 
> Then have `human_readable` and `human_readable_rate` call that function and add the appropriate units.

The refactoring attempts done in #1891 failed because adding any more logic to the function causes performance degradation. Let's backtrack to previous solutions:

>1. ~Add a `B` to the end which would turn most units into what we want: `BB`, `KB`, `MB`. But `BB` is not a valid unit.~

Not a solution.
>2. ~Conditionally add the `B` character in the `human_readable()`. There could be an argument called e.g. `abbreviate_units` used to switch this behavior.~

Nope, that would have a performance impact on `human_readable()` which could potentially manifest somewhere else in Ranger. 
>3. Create a wrapper around `human_readable()` that would add the `B` character conditionally ("if it is missing then add it"?). It could be named e.g. `human_readable_...`. That's difficult to name which means it probably is not the correct way to do it.

Not ideal, but could be a possible solution without a lot of code duplication. Although I suggest it might as well add the `/s` to the end of the string. With the `/s` at the end the name of the method is obvious (`human_readable_rate` or `human_readable_speed`).
>4. ~Create a `human_readable_speed()` which would return the correct units of data transfer: `B/s`, `MB/`, `KB/s`. It would be internally calling the `human_readable()`.~

Too vague.

New proposal: 

5. Duplicate code of `human_readable()` as `human_readable_rate()` (or `_speed`) and change the hardcoded units in the newly duplicated code. This will have identical performance as `human_readable()` with the cost of having duplicated code. The code of `human_readable()` has been touched about 5 times in the last 10 years so the increased maintenance might be negligible.

Which of the remaining solutions (3. and 5.) is preferred? Are there any other solutions that I should explore?

toonn:
You still haven't tried what I'm suggesting though?

mik13ST:
> You still haven't tried what I'm suggesting though?

Oh, you are right. Please take a look at https://github.com/ranger/ranger/pull/1891#issuecomment-614304819 then.

mik13ST:
No longer using ranger and #1891 was abandoned.

PR #1850, opened 2020-02-22
Fix trash location

### LINKED ISSUE #1848: ranger/config/rifle.conf: trash using wrong directory
#### Context

https://github.com/ranger/ranger/blob/870435dd9bec311d0866eb4a801ff174366ed0a9/ranger/config/rifle.conf#L284

#### Possible Solutions

XDG_DATA_DIR should be XDG_DATA_HOME as is also specified in the man page of ranger.

The standard fallback if that environment variable is not defined is usually ``$HOME/.local/share``.  Ranger already uses ``$HOME/.local/share/ranger`` for data files. Wouldn't it be better to use that for the trash as well?

My suggestion would be:

```
diff --git a/ranger/config/rifle.conf b/ranger/config/rifle.conf
index 86f53fd1..60b4b22d 100644
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -281,4 +281,4 @@ mime application/x-executable = "$1"

 # Move the file to trash using trash-cli.
-label trash = mkdir -p -- ${XDG_DATA_DIR:-$HOME/.ranger}/ranger-trash; mv -- "$@" ${XDG_DATA_DIR:-$HOME/.ranger}/ranger-trash
+label trash = mkdir -p -- ${XDG_DATA_HOME:-$HOME/.local/share}/ranger/trash; mv -- "$@" ${XDG_DATA_HOME:-$HOME/.local/share}/ranger/trash

```

emberfade:
#### ISSUE TYPE

- Improvement

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Change the fallback trash location to use the correct environment variable and
ranger's default data directory. Update the man page to reflect the change.

#### MOTIVATION AND CONTEXT

Fixes #1848

#### TESTING

I can't get the tests to run properly. I only manually changed `rifle.conf` and `ranger.pod` and ran `make man`.

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -133,7 +133,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.9.3" "2019-12-31" "ranger manual"
+.TH RANGER 1 "ranger-1.9.3" "2020-02-22" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
 .\" way too many mistakes in technical documents.
 .if n .ad l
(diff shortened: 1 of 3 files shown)


emberfade:
I have one more suggestion for the `mv` command. It will currently overwrite if the same filename already exists in the trash folder. I could add `mv --backup=numbered` so that the command isn't destructive like delete.

toonn:
Great first contribution! 🎉 
Keep 'em coming : )

emberfade:
Thank you very much.

> The `--backup` idea is good but it's not portable. MacOS for example doesn't have it.

Oh right, I forgot about portability.

Issue #1805, opened 2020-01-07
Ranger Crashes on with 'not finding file or directory'


Dovermore:
Runtime Environment

- Operating system and version: Raspbian GNU/Linux 10 (buster)
- Terminal emulator and version: Alacritty
- Python version: python3.7.3
- Ranger version/commit: 1.9.3
- Locale: Australia

#### Current Behavior
Running pypl `ranger-fm` give the following error:
```
ranger version: ranger 1.9.3
Python version: 3.7.3 (default, Apr  3 2019, 05:39:12) [GCC 8.2.0]
Locale: en_US.UTF-8

Traceback (most recent call last):
  File "/home/pi/.local/lib/python3.7/site-packages/ranger/core/main.py", line 171, in main
    fm.initialize()
  File "/home/pi/.local/lib/python3.7/site-packages/ranger/core/fm.py", line 132, in initialize
    self.ui.initialize()
  File "/home/pi/.local/lib/python3.7/site-packages/ranger/gui/ui.py", line 127, in initialize
    self.handle_multiplexer()
  File "/home/pi/.local/lib/python3.7/site-packages/ranger/gui/ui.py", line 500, in handle_multiplexer
    ['screen', '-Q', 'title']).strip()
  File "/home/pi/.local/lib/python3.7/site-packages/ranger/ext/spawn.py", line 35, in check_output
    process = Popen(popenargs, stderr=fd_devnull, **kwargs)
  File "/usr/lib/python3.7/subprocess.py", line 775, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.7/subprocess.py", line 1522, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'screen': 'screen'

ranger crashed. Please report this traceback at:
https://github.com/ranger/ranger/issu
```

Cloning then run `ranger.py` give the following behavior (python2)
```
ranger version: ranger-master v1.9.3-1-g3965e3ee
Python version: 2.7.16 (default, Apr  6 2019, 01:42:57) [GCC 8.2.0]
Locale: None.None

Traceback (most recent call last):
  File "/home/pi/scratch/ranger/ranger/core/main.py", line 171, in main
    fm.initialize()
  File "/home/pi/scratch/ranger/ranger/core/fm.py", line 132, in initialize
    self.ui.initialize()
[...]

PR #1852, opened 2020-02-23
Fix `FileNotFoundError` on startup

tkukurin:
Fixes 'FileNotFound' if `screen` is not available on the system.

#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch
- Terminal emulator and version: Fish 3.0.2
- Python version: 3.8.1
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Change `screen` to be invoked via a shell so it doesn't raise unexpected error type. 

#### MOTIVATION AND CONTEXT
https://github.com/ranger/ranger/issues/1805

### DIFF
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -497,7 +497,7 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
                 # gives out a warning if $TERM is not "screen"
                 try:
                     self._screen_title = check_output(
-                        ['screen', '-Q', 'title']).strip()
+                        ['screen', '-Q', 'title'], shell=True).strip()
                 except CalledProcessError:
                     self._screen_title = None
 
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -9,6 +9,7 @@ import threading
 import curses
 from subprocess import CalledProcessError
 
+from ranger.ext.get_executables import get_executables
 from ranger.ext.keybinding_parser import KeyBuffer, KeyMaps, ALT_KEY
 from ranger.ext.lazy_property import lazy_property
 from ranger.ext.signals import Signal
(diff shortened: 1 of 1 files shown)


tkukurin:
Yeah, I think the 2nd easiest option is just catching `FileNotFound`.

I went with this option because it requires fewer changes and appears 
semantically more meaningful.

According to my `timeit` tests catching an exception is faster, but I feel
`FileNotFound` shouldn't be treated the same way as `CalledProcessError`
so maybe it'd require some additional logic.

Another alternative is checking whether `screen` is installed via `shutil.which`;
it only exists since 3.3, but the
[code](https://hg.python.org/cpython/file/6860263c05b3/Lib/shutil.py#l1068)
appears relatively straightforward.
The problem is it won't always work (e.g. `alias screen=my_real_screen`), and 
also searching the path may be really slow only for catching a few edge cases.

WDYT? I'm fine with either of the first two options.

toonn:
I guess this'd only run once so as long as the slowdown's not measured in hundreds of milliseconds it's probably fine.

toonn:
I took a different approach because the rest of that code wasn't doing what it should be doing. It'd be great if you could test it to verify it's no longer resulting in errors.

ranger/gui/ui.py
```diff
@@ -73,8 +84,8 @@ def __init__(self, env=None, fm=None):  # pylint: disable=super-init-not-called
         self.multiplexer = None
         self._draw_title = None
         self._tmux_automatic_rename = None
-        self._tmux_title = None
-        self._screen_title = None
+        self._multiplexer_title = None
+        self._multiplexer_title = None
```
Vifon: No need to set it twice now that it's a single variable.

ranger/gui/ui.py
```diff
@@ -469,58 +480,64 @@ def throbber(self, string='.', remove=False):
...
-                except CalledProcessError:
-                    self._screen_title = None
+            except CalledProcessError:
+                self.fm.notify("Couldn't access previous multiplexer window"
+                               " name, won't be able to restore.",
+                               bad=False)
+            if not self._multiplexer_title:
+                self._multiplexer_title = os.environ.get(
+                    "SHELL",
+                    "shell").split("/")[-1]
```
Vifon: Consider using `os.path.basename` instead of `str.split()[-1]`.

PR #1871, opened 2020-03-05
Fix crashes when deleting to trash (#1798)

### LINKED ISSUE #1798: Ranger crashes when deleting to Trash
OS: Linuxmint 19.3
Python: 3.6.9
Ranger: v1.9.2
trash-cli: 0.12.9.14-2.1

When deleting a highlighted file to trash, I get the following error:
"object has no attribute 'path'"

When deleting a (one) selected file to trash, I am asked to confirm delete, and after inputing 'y' ranger crashes.
----------------------------------------------------------------------
ranger version: ranger-master
Python version: 3.6.9 (default, Nov  7 2019, 10:44:02) [GCC 8.3.0]
Locale: en_GB.UTF-8
Current file: '/home/standard_user/file.txt'

  File "/usr/local/lib/python3.6/dist-packages/ranger/core/main.py", line 201, in main
    fm.loop()
  File "/usr/local/lib/python3.6/dist-packages/ranger/core/fm.py", line 383, in loop
    ui.handle_input()
  File "/usr/local/lib/python3.6/dist-packages/ranger/gui/ui.py", line 279, in handle_input
    self.handle_key(key)
  File "/usr/local/lib/python3.6/dist-packages/ranger/gui/ui.py", line 209, in handle_key
    elif not DisplayableContainer.press(self, key):
  File "/usr/local/lib/python3.6/dist-packages/ranger/gui/displayable.py", line 271, in press
    focused_obj.press(key)
  File "/usr/local/lib/python3.6/dist-packages/ranger/gui/widgets/console.py", line 188, in press
    self.type_key(key)
  File "/usr/local/lib/python3.6/dist-packages/ranger/gui/widgets/console.py", line 214, in type_key
    self._answer_question(answer)
  File "/usr/local/lib/python3.6/dist-packages/ranger/gui/widgets/console.py", line 197, in _answer_question
    callback(answer)
  File "/usr/local/lib/python3.6/dist-packages/ranger/config/commands.py", line 757, in _question_callback
    self.fm.execute_file(files, label='trash')
  File "/usr/local/lib/python3.6/dist-packages/ranger/core/actions.py", line 458, in execute_file
    filenames = [f.path for f in files]
  File "/usr/local/lib/python3.6/dist-packages/ranger/core/actions.py", line 458, in <listcomp>
    filenames = [f.path for f in files]
AttributeError: 'str' object has no attribute 'path'

------------------------------------------------------------

    Thank you for contributing to ranger by opening this issue.
    Please check through this list, so you can be as helpful as possible:

    1. Was this issue already reported?  Please do a quick search.
    2. Maybe the problem is solved in the current master branch already?
       Simply clone ranger's git repository and run ./ranger.py to find out.
    3. Provide all the relevant information, as outlined in this template.
       Feel free to remove any sections you don't need.
-->

#### Runtime Environment

- Python version: 
- Ranger version/commit: 
- Locale: 

#### Steps to reproduce
1. 
2. 
3. 

#### Traceback

```

```

5hir0kur0:
#### ISSUE TYPE

- Bug fix: This fixes #1798.

#### RUNTIME ENVIRONMENT

- Operating system and version: `Arch Linux, kernel 5.4.23-1-lts, last full pacman update at: 2020-03-05T08:20:22`
 ```
rxvt-unicode (urxvt) v9.22 - released: 2016-01-23
options: perl,xft,styles,combining,blink,iso14755,unicode3,encodings=eu+vn+jp+jp-ext+kr+zh+zh-ext,fade,transparent,tint,XIM,frills,selectionscrolling,wheel,slipwheel,cursorBlink,pointerBlank,scrollbars=plain+rxvt+NeXT+xterm
```
- Python version: `Python 3.8.1`
- Ranger version/commit: `v1.9.3-18-g081e7315` (output of `git describe HEAD`)
- Locale: `en_US.UTF-8`

#### CHECKLIST

  (`make test_py` outputs is the same before and after the change)
  (`make test_py` output is the same before and after the change)
    - [ ] Tests have been updated

#### DESCRIPTION

The `execute()` method of the trash command (in `ranger/config/commands.py`)
used to pass a list of file paths (as strings) to `fm.execute_file()`.
The documentation of the `execute_file()` method states that the 'files'
parameter must not be a list of strings:
>     [...]
>     files: a list of file objects (not strings!)
>     [...]
So I changed `files` to be a list of `File` objects and that seems to fix the issue.

*Note: I'm not sure if I need to pass any additional parameters to the `File` constructor.
Passing just the path seemed to work fine in my tests.*
(I'm talking about [this](https://github.com/5hir0kur0/ranger/commit/f65e6f08bcf63b7d915d2a2e98d1f6f891cd30de#diff-ac22d5c05a75c2e025c82cdfcbf695b1R732) line: `files = [File(name) for name in file_names]`).

#### MOTIVATION AND CONTEXT

Relevant issue: #1798

#### TESTING

I ran `make test_py` as I only changed a python file.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -722,14 +722,15 @@ class trash(Command):
     def execute(self):
         import shlex
         from functools import partial
-        from ranger.container.file import File
 
         def is_directory_with_files(path):
             return os.path.isdir(path) and not os.path.islink(path) and len(os.listdir(path)) > 0
 
         if self.rest(1):
             file_names = shlex.split(self.rest(1))
-            files = [File(name) for name in file_names]
+            files = self.paths_to_filesystem_objects(file_names)
+            if files is None:
+                return
             many_files = (len(files) > 1 or is_directory_with_files(files[0].path))
         else:
             cwd = self.fm.thisdir
@@ -753,6 +754,43 @@ class trash(Command):
         else:
             # no need for a confirmation, just delete
             self.fm.execute_file(files, label='trash')
+
+    @staticmethod
+    def group_paths_by_dirname(paths):
+        """
+        Groups the paths into a dictionary with their dirnames as keys and a set of
+        basenames as entries.
+        """
+        groups = dict()
+        for path in paths:
+            abspath = os.path.abspath(os.path.expanduser(path))
+            dirname, basename = os.path.split(abspath)
+            groups.setdefault(dirname, set()).add(basename)
+        return groups
+
+    def paths_to_filesystem_objects(self, paths):
+        """
+        Find FileSystemObjects corresponding to the paths if they are already in
+        memory and load those that are not.
+        """
+        result = []
+        # Grouping the files by dirname avoids the potentially quadratic running time of doing
+        # a linear search in the directory for each entry name that is supposed to be deleted.
+        groups = trash.group_paths_by_dirname(paths)
+        for dirname, basenames in groups.items():
+            directory = self.fm.get_directory(dirname)
+            directory.load_content_if_outdated()
+            for entry in directory.files_all:
+                if entry.basename in basenames:
+                    result.append(entry)
+                    basenames.remove(entry.basename)
+            if basenames != set():
+                # Abort the operation with an error message if there are entries
+                # that weren't found.
+                names = ', '.join(basenames)
+                self.fm.notify('Error: No such file or directory: {}'.format(names), bad=True)
+                return None
+        return result
 
     def tab(self, tabnum):
         return self._tab_directory_content()
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -728,7 +728,7 @@ class trash(Command):
 
         if self.rest(1):
             file_names = shlex.split(self.rest(1))
-            files = self.paths_to_filesystem_objects(file_names)
+            files = self.fm.get_filesystem_objects(file_names)
             if files is None:
                 return
             many_files = (len(files) > 1 or is_directory_with_files(files[0].path))
@@ -754,43 +754,6 @@ class trash(Command):
         else:
             # no need for a confirmation, just delete
             self._trash_files_catch_arg_list_error(files)
-
-    @staticmethod
-    def group_paths_by_dirname(paths):
-        """
-        Groups the paths into a dictionary with their dirnames as keys and a set of
-        basenames as entries.
-        """
-        groups = dict()
-        for path in paths:
-            abspath = os.path.abspath(os.path.expanduser(path))
-            dirname, basename = os.path.split(abspath)
-            groups.setdefault(dirname, set()).add(basename)
-        return groups
-
-    def paths_to_filesystem_objects(self, paths):
-        """
-        Find FileSystemObjects corresponding to the paths if they are already in
-        memory and load those that are not.
-        """
-        result = []
-        # Grouping the files by dirname avoids the potentially quadratic running time of doing
-        # a linear search in the directory for each entry name that is supposed to be deleted.
-        groups = trash.group_paths_by_dirname(paths)
-        for dirname, basenames in groups.items():
-            directory = self.fm.get_directory(dirname)
-            directory.load_content_if_outdated()
-            for entry in directory.files_all:
-                if entry.basename in basenames:
-                    result.append(entry)
-                    basenames.remove(entry.basename)
-            if basenames != set():
-                # Abort the operation with an error message if there are entries
-                # that weren't found.
-                names = ', '.join(basenames)
-                self.fm.notify('Error: No such file or directory: {}'.format(names), bad=True)
-                return None
-        return result
 
     def tab(self, tabnum):
         return self._tab_directory_content()
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -722,13 +722,15 @@ class trash(Command):
     def execute(self):
         import shlex
         from functools import partial
+        from ranger.container.file import File
 
         def is_directory_with_files(path):
             return os.path.isdir(path) and not os.path.islink(path) and len(os.listdir(path)) > 0
 
         if self.rest(1):
-            files = shlex.split(self.rest(1))
-            many_files = (len(files) > 1 or is_directory_with_files(files[0]))
+            file_names = shlex.split(self.rest(1))
+            files = [File(name) for name in file_names]
+            many_files = (len(files) > 1 or is_directory_with_files(files[0].path))
         else:
             cwd = self.fm.thisdir
             tfile = self.fm.thisfile
@@ -736,14 +738,15 @@ class trash(Command):
                 self.fm.notify("Error: no file selected for deletion!", bad=True)
                 return
 
+            files = self.fm.thistab.get_selection()
             # relative_path used for a user-friendly output in the confirmation.
-            files = [f.relative_path for f in self.fm.thistab.get_selection()]
+            file_names = [f.relative_path for f in files]
             many_files = (cwd.marked_items or is_directory_with_files(tfile.path))
 
         confirm = self.fm.settings.confirm_on_delete
         if confirm != 'never' and (confirm != 'multiple' or many_files):
             self.fm.ui.console.ask(
-                "Confirm deletion of: %s (y/N)" % ', '.join(files),
+                "Confirm deletion of: %s (y/N)" % ', '.join(file_names),
                 partial(self._question_callback, files),
                 ('n', 'N', 'y', 'Y'),
             )
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -753,7 +753,7 @@ class trash(Command):
             )
         else:
             # no need for a confirmation, just delete
-            self.fm.execute_file(files, label='trash')
+            self._trash_files_catch_arg_list_error(files)
 
     @staticmethod
     def group_paths_by_dirname(paths):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by toonn** [comment]
I think we need to use the actual `File` objects ranger has in memory, not create new ones. They're not singletons. If we create new ones the interface will have outdated information in memory.

**Comment by 5hir0kur0** [comment]
@toonn So you mean I should do something like this?
```python
    def find_existing_files(self, file_names):
        '''
        Find file objects corresponding to the file names that are already in
        memory and load those that are not.
        '''
        result = []
        for name in file_names:
            path = os.path.abspath(os.path.expanduser(name))
            if os.path.isdir(path):
                result.append(self.fm.get_directory(path))
            else:
                dirname = os.path.dirname(path)
                directory = self.fm.get_directory(dirname)
                directory.load_content_if_outdated()
                desired_file = trash.find_file_by_path(directory.files_all, path)
                if desired_file is not None:
                    result.append(desired_file)
                    continue
                # Exit the operation with an error message if the file could
                # not be found.
                self.fm.notify('Error: path does not exist: ' + name, bad=True)
                return None
        return result

    @staticmethod
    def find_file_by_path(files, path):
        '''
        Search a list of File objects for a file matching path.
        Assumes that path is absolute.
        '''
        if files is None:
            return None
        for file_object in files:
            if file_object.path == path:
                return file_object
        return None
```

Sorry if I misunderstood you; I am not very familiar with ranger's code base.

The problem with my implementation above is that it uses linear search to find files.
For directories, I found the `fm.get_directory()` method which seems to use a dictionary of loaded directories internally and thus should be quite fast.

But I didn't find a similar method for files, so I just do linear search in the corresponding directory, which could be a performance issue when deleting lots of files.

toonn:
Something like that, yeah. I don't think the linear search can be avoided, the files are sorted but the order depends on the sorting method the user has set. Maybe you could group files by directory, turn them into a set and then go through the directory linearly *once*, checking the set for membership and remembering those file objects?

5hir0kur0:
That's a good idea! I've implemented it.
____
There is another problem though: If you trash lots of files you run into the argument list length limitation of the operating system and this actually crashes ranger.
I'm not sure if this is easily fixable though. I think it is this line in `rifle.conf`:
```
```
If there are lots of files, the argument list gets too long, so it should actually be split up into multiple calls to `trash-put` (similar to the way the command line program `xargs` handles this case).

While I don't really think this limitation itself is a problem, the fact that it crashes ranger is (IMO).
This can be fixed by wrapping the call to `execute_file()` in a try-except like so:
```python
try:
    self.fm.execute_file(files, label='trash')
except OSError as e:
    self.fm.notify(str(e), bad=True)
```
*However, I am not sure if this is the right place to put it. Maybe it would be better to have the `fm.execute_file()` method itself handle this case?*
____
Here are the steps to reproduce the crash and the traceback for reference:
Steps:
```
:cd /tmp
:mkdir temp
:cd temp
:shell touch empty_file{1..10000}
:mark_files all=True val=True
:trash %s
(confirm prompt by pressing y)
```
Result: Ranger crashes with the following traceback:
```
ranger version: ranger-master v1.9.3-20-gefa32996
Python version: 3.8.1 (default, Jan 22 2020, 06:38:00) [GCC 9.2.0]
Locale: en_US.UTF-8
Current file: '/tmp/temp/empty_file1'

  File "/home/me/code/ranger/ranger/core/actions.py", line 465, in execute_file
    return execute()
  File "/home/me/code/ranger/ranger/core/actions.py", line 463, in execute
  File "/home/me/code/ranger/ranger/ext/rifle.py", line 442, in execute
    process = Popen(cmd, env=self.hook_environment(os.environ))
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1702, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 7] Argument list too long: '/bin/sh'

  File "/home/me/code/ranger/ranger/core/main.py", line 203, in main
    fm.loop()
  File "/home/me/code/ranger/ranger/core/fm.py", line 383, in loop
    ui.handle_input()
  File "/home/me/code/ranger/ranger/gui/ui.py", line 276, in handle_input
    self.handle_key(key)
  File "/home/me/code/ranger/ranger/gui/ui.py", line 206, in handle_key
    elif not DisplayableContainer.press(self, key):
  File "/home/me/code/ranger/ranger/gui/displayable.py", line 275, in press
    focused_obj.press(key)
  File "/home/me/code/ranger/ranger/gui/widgets/console.py", line 188, in press
    self.type_key(key)
  File "/home/me/code/ranger/ranger/gui/widgets/console.py", line 214, in type_key
    self._answer_question(answer)
  File "/home/me/code/ranger/ranger/gui/widgets/console.py", line 197, in _answer_question
    callback(answer)
  File "/home/me/code/ranger/ranger/config/commands.py", line 800, in _question_callback
    self.fm.execute_file(files, label='trash')
  File "/home/me/code/ranger/ranger/core/actions.py", line 473, in execute_file
    return execute()
  File "/home/me/code/ranger/ranger/core/actions.py", line 463, in execute
  File "/home/me/code/ranger/ranger/ext/rifle.py", line 442, in execute
    process = Popen(cmd, env=self.hook_environment(os.environ))
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1702, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 7] Argument list too long: '/bin/sh'
```

ranger/config/commands.py
```diff
@@ -727,30 +727,71 @@ def is_directory_with_files(path):
...
+        basenames as entries.
+        """
+        groups = dict()
+        for path in paths:
+            abspath = os.path.abspath(os.path.expanduser(path))
+            dirname, basename = os.path.split(abspath)
+            groups.setdefault(dirname, set()).add(basename)
+        return groups
+
+    def paths_to_filesystem_objects(self, paths):
```
toonn: Maybe this should go next to `get_directory`? Or in `ranger.ext`?

5hir0kur0: OK, I moved it into `fm.py`. I'm not sure if it's really important enough to be moved out of the `trash` class though. (Since no one seems to have needed this kind of method up to now.)

5hir0kur0:
I also "fixed" the crash I [mentioned](#issuecomment-595832917) above by displaying an error message instead of crashing. 
It does not actually move the files to trash in this case because I don't know how to get rifle to do that (see above).

kevinhwang91:
I hate the redraw when using trash-put, and never use it happily. So I write a simple command plugin to trash file.
https://github.com/kevinhwang91/dotfiles/blob/master/ranger/plugins/trash_put.py

kevinhwang91:
Not only `trash` but also `delete` will crash. I make a hack patch for solving this.
```diff
diff --git a/ranger/gui/widgets/statusbar.py b/ranger/gui/widgets/statusbar.py
index fd44613e..245beb62 100644
--- a/ranger/gui/widgets/statusbar.py
+++ b/ranger/gui/widgets/statusbar.py
@@ -192,11 +192,24 @@ class StatusBar(Widget):  # pylint: disable=too-many-instance-attributes
         directory = target if target.is_directory else \
             target.fm.get_directory(os.path.dirname(target.path))
         if directory.vcs and directory.vcs.track:
-            if directory.vcs.rootvcs.branch:
-                vcsinfo = '({0:s}: {1:s})'.format(
-                    directory.vcs.rootvcs.repotype, directory.vcs.rootvcs.branch)
-            else:
-                vcsinfo = '({0:s})'.format(directory.vcs.rootvcs.repotype)
+            try:
+                if directory.vcs.rootvcs.branch:
+                    vcsinfo = '({0:s}: {1:s})'.format(
+                        directory.vcs.rootvcs.repotype, directory.vcs.rootvcs.branch)
+                else:
+                    vcsinfo = '({0:s})'.format(directory.vcs.rootvcs.repotype)
+            except TypeError:
+                root_path = directory.vcs.rootvcs.path
+                for path, fdir in self.fm.directories.items():
+                    if not fdir.files:
+                        continue
+                    for fobj in fdir.files:
+                        if path.startswith(root_path):
+                            if fobj.is_directory and fobj.vcs:
+                                del fobj.vcs
+                            fobj.vcsstatus = None
+                return
+
             left.add_space()
             left.add(vcsinfo, 'vcsinfo')
```

PopeRigby:
Any progress on this?

toonn:
Reviews and testing reports welcome.

PopeRigby:
It seems to have fixed the issue for me on Arch Linux.

ranger/config/commands.py
```diff
@@ -727,36 +727,54 @@ def is_directory_with_files(path):
...
+            files = self.fm.thistab.get_selection()
             # relative_path used for a user-friendly output in the confirmation.
-            files = [f.relative_path for f in self.fm.thistab.get_selection()]
+            file_names = [f.relative_path for f in files]
             many_files = (cwd.marked_items or is_directory_with_files(tfile.path))
 
         confirm = self.fm.settings.confirm_on_delete
         if confirm != 'never' and (confirm != 'multiple' or many_files):
             self.fm.ui.console.ask(
-                "Confirm deletion of: %s (y/N)" % ', '.join(files),
```
toonn, line 746: Switch to py3 formatting syntax here.

ranger/core/fm.py
```diff
@@ -329,6 +329,43 @@ def get_directory(self, path, **dir_kwargs):
...
+            directory.load_content_if_outdated()
+            for entry in directory.files_all:
+                if entry.basename in basenames:
+                    result.append(entry)
+                    basenames.remove(entry.basename)
+            if basenames != set():
+                # Abort the operation with an error message if there are entries
+                # that weren't found.
+                names = ', '.join(basenames)
+                self.fm.notify('Error: No such file or directory: {}'.format(names), bad=True)
```
toonn, line 365: Py2 doesn't support implicit format references.

toonn (changes requested):
Two small fixes needed but I'll do those myself when testing.

ranger/config/commands.py
toonn: Hmm, there's an open PR taking care of commands.py formatting strings.

toonn:
Thank you for the contribution! I'm sorry it took so long to see it through, especially because the code quality was so high!

da-the-dev:
Has this been fixed? I'm on ranger 1.9.3 and i get the exact same error message

humanplayer2:
I changed the keybinding in `rc.conf` inspired by [ArchWiKi](https://wiki.archlinux.org/title/ranger#Move_to_trash):

    map dT shell gio trash %s

PR #1876, opened 2020-03-10
Move man pages to support MANPATH

peey:
Make a subdirectory man1 and move man pages ranger.1 and rifle.1 to it
so that adding "doc" directory to "MANPATH" works.

I also tried to make a `man8` directory instead, but it wouldn't work https://unix.stackexchange.com/a/3587/39009

See https://askubuntu.com/a/244810/370026, which suggests that you need to make a man1 subdirectory.

After making these changes, my MANPATH works.

This is useful when you haven't installed ranger, and have just downloaded ranger repo and aliased `ranger` to full path of `ranger.py`


toonn:
Are you sure you don't want to run `make install`? It should install the man pages for you, as well as ranger and rifle of course.

peey:
Yep. I think having applications installed portably is pretty standard.

(Also, I'm not sure but it might be the the only route a user can take when they don't have root access)

Should I make additional changes in makefile to update the paths so installs don't break?

toonn:
Current plan:
  1. Drop the generated man pages from the repo, they get in the way when merging.
  2. Change the `man` target in the Makefile to write the generated files to a `man1` subdir.
  3. Have the `install` target depend on the `man` target, requiring perl for people installing from the repo but not when installing from a tarball.
  4. Evaluate whether anything else needs changing to accommodate the above.

peey:
Thanks a lot for outlining this. 

I'll try to take a stab at this the next week. Please also feel free to go ahead with it yourself if you think it'll be easier and you have time.

Issue #1256, opened 2018-08-13
Filenames containing combining characters cause misaligned display


craigbarnes:
#### Runtime Environment

- Operating system and version: Arch Linux
- Terminal emulator and version: Various; urxvt 9.22, xfce4-terminal 0.8.7.4, etc.
ranger version: ranger-master 1.9.1
Python version: 3.7.0 (default, Jul 15 2018, 10:44:58) [GCC 8.1.1 20180531]
Locale: en_GB.UTF-8

#### Current Behavior

Filenames containing combining or other non-spacing characters caused the file size column to become unaligned (shift more and more to the left, depending on the number of combining characters)

#### Expected Behavior

File size text aligned to the right edge of the column, as ASCII-only filenames are.

#### Possible Solutions

Use `wcwidth(3)` or similar to compute the number of terminal columns a filename occupies.

#### Steps to reproduce
1. `touch ฮั่ฮั่ฮั่ฮั่ฮั่ฮั่ฮั่ฮั่ฮั่ฮั่ฮั่ฮั่ฮั่`
2. `ranger`
3. Observe that the file size text is shifted too far left, due to miscalculated display width

toonn:
Can confirm on the latest master. Thanks for the report!

Fujiwara-Chika:
![image](https://user-images.githubusercontent.com/44722787/147486564-1a28ad02-57fd-442b-9969-b0d7034117d8.png)

Still not resolve?

toonn:
No, that's why the issue is still open. The available fix isn't universal and I don't want to complicate matters while introducing more such subtle bugs.

jquast:
I might be able to help with this, it looks like @qrvd and I have already discussed it lately, I have been reviewing projects for proper wcswidth/grapheme integration this week, this looks like a popular tool so let's make it CJK / emoji friendly :)

PR #1882, opened 2020-03-18
Fix string (character) width retrieving method

crazymug:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 18.04.2
- Terminal emulator and version: GNOME Terminal 3.28.2
- Python version: 
3.8.2 (default, Mar 15 2020, 13:26:29) [GCC 7.4.0] 
2.7.17 (default, Nov  7 2019, 10:07:09) [GCC 7.4.0]
- Ranger version/commit: ranger-master v1.9.3-28-g9bac5190
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Current implementation of string (char) width determination is not correct while working with some non-ASCII strings (e.g. #1256).

There's good and pretty small library that implements two POSIX functions, which are not present in Python's standard lib: wcwidth and wcswidth https://pypi.org/project/wcwidth/. This library gets the most actual version of Unicode code tables from http://www.unicode.org during installation/updates.

I propose here a variant of using this library. Another way is although possible - include this library's code base entirely with respect to maintainers will.

#### MOTIVATION AND CONTEXT
#1256

### DIFF
--- a/ranger/ext/widestring.py
+++ b/ranger/ext/widestring.py
@@ -9,7 +9,7 @@ from unicodedata import east_asian_width
 
 try:
     from wcwidth import wcwidth, wcswidth
-    WCWIDTH_AVAILABLE = True 
+    WCWIDTH_AVAILABLE = True
 except ImportError:
     WCWIDTH_AVAILABLE = False
 
@@ -21,22 +21,22 @@ WIDE = 2
 WIDE_SYMBOLS = set('WF')
 
 
-def uwid(string):	
-    """Return the width of a string"""	
-    if not PY3:	
-        string = string.decode('utf-8', 'ignore')	
+def uwid(string):
+    """Return the width of a string"""
+    if not PY3:
+        string = string.decode('utf-8', 'ignore')
     if WCWIDTH_AVAILABLE:
         return wcswidth(string)
-    return sum(utf_char_width(c) for c in string)	
+    return sum(utf_char_width(c) for c in string)
 
 
-def utf_char_width(string):	
-    """Return the width of a single character"""	
+def utf_char_width(string):
+    """Return the width of a single character"""
     if WCWIDTH_AVAILABLE:
         return wcwidth(string)
     else:
-        if east_asian_width(string) in WIDE_SYMBOLS:	
-            return WIDE	
+        if east_asian_width(string) in WIDE_SYMBOLS:
+            return WIDE
         return NARROW
 
 
@@ -178,7 +178,7 @@ class WideString(object):  # pylint: disable=too-few-public-methods
                 string = self.string.decode('utf-8', 'ignore')
             return wcswidth(string)
         else:
-            return len(self.chars) 
+            return len(self.chars)
 
 
 if __name__ == '__main__':
--- a/ranger/ext/widestring.py
+++ b/ranger/ext/widestring.py
@@ -9,13 +9,21 @@ from unicodedata import east_asian_width
 
 try:
     from wcwidth import wcwidth, wcswidth
+
+    def uwid(string):
+        """Return the width of a string"""
+        return wcswidth(string)
+
+    def utf_char_width(char):
+        """Return the width of a single character"""
+        return wcwidth(char)
 except ImportError:
-    def wcswidth(string):
+    def uwid(string):
         if not PY3:
             string = string.decode('utf-8', 'ignore')
         return sum(utf_char_width(c) for c in string)
 
-    def wcwidth(char):
+    def utf_char_width(char):
         if east_asian_width(char) in WIDE_SYMBOLS:
             return WIDE
         return NARROW
@@ -26,16 +34,6 @@ ASCIIONLY = set(chr(c) for c in range(1, 128))
 NARROW = 1
 WIDE = 2
 WIDE_SYMBOLS = set('WF')
-
-
-def uwid(string):
-    """Return the width of a string"""
-    return wcswidth(string)
-
-
-def utf_char_width(char):
-    """Return the width of a single character"""
-    return wcwidth(char)
 
 
 def string_to_charlist(string):
--- a/ranger/ext/widestring.py
+++ b/ranger/ext/widestring.py
@@ -10,7 +10,15 @@ from unicodedata import east_asian_width
 try:
     from wcwidth import wcwidth, wcswidth
 except ImportError:
-    pass
+    def wcswidth(string):
+        if not PY3:
+            string = string.decode('utf-8', 'ignore')
+        return sum(utf_char_width(c) for c in string)
+
+    def wcwidth(char):
+        if east_asian_width(char) in WIDE_SYMBOLS:
+            return WIDE
+        return NARROW
 
 
 PY3 = sys.version_info[0] >= 3
@@ -22,22 +30,12 @@ WIDE_SYMBOLS = set('WF')
 
 def uwid(string):
     """Return the width of a string"""
-    if not PY3:
-        string = string.decode('utf-8', 'ignore')
-    try:
-        return wcswidth(string)
-    except NameError:
-        return sum(utf_char_width(c) for c in string)
+    return wcswidth(string)
 
 
 def utf_char_width(char):
     """Return the width of a single character"""
-    try:
-        return wcwidth(char)
-    except NameError:
-        if east_asian_width(char) in WIDE_SYMBOLS:
-            return WIDE
-        return NARROW
+    return wcwidth(char)
 
 
 def string_to_charlist(string):
@@ -171,12 +169,7 @@ class WideString(object):  # pylint: disable=too-few-public-methods
         >>> len(WideString("モヒカン"))
         8
         """
-        try:
-            if not PY3:
-                return wcswidth(self.string.decode('utf-8', 'ignore'))
-            return wcswidth(self.string)
-        except NameError:
-            return len(self.chars)
+        return uwid(self.string)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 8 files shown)


toonn (commented):
I don't understand why you would fix `ranger.ext.widestring` and then remove its uses?
We have a very strict policy on external dependencies in that there are none allowed.
How about updating `ranger.ext.widestring` to defer to `wcwidth` *iff* it's available?

toonn (changes requested):
Rather than work with flags just define a width function or two with a differing implementation depending on whether the module is available. I.e., in the simple case it's just import and alias (or us the `wcwidth/wcswidth` names and in the other case it's two function definitions.

toonn (changes requested):
Almost, import and define these functions once at the top then simply use them in the various locations.

toonn (changes requested):
Not quite, the point is to get all the exception crap out of the functions. Makes the control flow so much clearer. `try import wcwidth; except IE: def wcwidth(): whatever`

toonn (changes requested):
Ok, almost there. Move the uwid and utf_char_width definitions into the except branch and alias wcwidth and wcswidth in the try branch.

Are you sure string_to_charlist doesn't need any updating?

toonn (commented):
Looks good, I'll have to get around to testing this soon.

ranger/ext/widestring.py
```diff
@@ -5,7 +5,32 @@
...
+    def utf_char_width(char):
+        """Return the width of a single character"""
+        return wcwidth(char)
+except ImportError:
+    from unicodedata import east_asian_width
+
+    def uwid(string):
+        if not PY3:
+            string = string.decode('utf-8', 'ignore')
+        return sum(utf_char_width(c) for c in string)
```
jquast, line 27: May I suggest,
```
         return sum(max(0, utf_char_width(c)) for c in string)
```

To account for possible return value of -1 for non-printable characters

jquast (approved):
👍

qrvd:
We could finish off this PR with a few more changes. Many terminals don't adhere strictly to the Unicode standard, so `wcwidth` won't always match the actual character width.

To fix this I think we could manually measure the character width. I would suggest we use `curses` to print a given string and see how many columns it really occupies. Here's an example:

```python
s = "এবং লোকজন অনেক অতিথিপরায়ণ। বাংলাদেশের ইতিহাস এবং সংস্কৃতি অত্যন্ত সমৃদ্ধ।"
import curses
curses.initscr()
w = curses.newwin(1, 1023, 0, 0)
w.addstr(0, 0, s)
w.refresh()
curses.nocbreak()
curses.echo()
curses.endwin()
_, width = w.getyx()
print(s)
print('A' * width)
```

The printed rows of As should visually line up exactly with the printed variable `s`, in which case this method works in your terminal.

I'm getting accurate results with this approach in several terminals. But it doesn't work in a handful of terminals, like `kitty`, and`mlterm`, which will instead display text of width `wcwidth.wcswidth(s)`. Maybe we should try to detect these with the `TERM` environment variable and use `wcwidth` in those cases...?

jquast:
@qrvd that's exactly the work I did with my python wcwidth library and tool ucs-detect, for example, results for Bengali language in mlterm: https://ucs-detect.readthedocs.io/sw_results/mlterm.html#bengali

I hope to publish some updated results and charts soon

qrvd:
@jquast Any plans for `wcwidth` to measure text widths directly with `curses`? I think it'd be really helpful when using non-Unicode-compliant terminals. For example Tilix doesn't have ZWJ support so some Emoji sequences take up more columns than `wcwidth` but Curses is giving me the correct value in those cases.

But I understand if it's kinda out of scope/excessive for `wcwidth` to initialize Curses, draw a string, and retrieve the cursor position in order to determine the width.

jquast:
I'm not against adding a new API function in wcwidth, like, `detect_tty_unicode_version(timeout=1)`, but I am against misleading people into thinking it is the solution :(

python "wcwidth" accepts a "UNICODE_VERSION" environment value or variable unicode_version, and [the hope was that we could detect just that variable,](https://www.jeffquast.com/post/terminal_wcwidth_solution/) and maybe that could be a new optional function in wcwidth to call.

However, after writing ucs-detect and gathering data, the problem is  complicated, many (most!) terminmals  support different versions of unicode depending on the character, for example ghostty is unicode version 15.0 for wide characters and version 17 for zero-width, and so I no longer document or suggest it, a more detailed issue https://github.com/jquast/wcwidth/issues/104 suggests a more complex solution

Related, I do hope to extend wcwidth for new function that makes a best-effort to measure strings containing terminal sequences, eg. "hello\x1b[0m" should be length 5 https://github.com/jquast/wcwidth/issues/92

jquast:
Oh and maybe I misunderstand, i think you made an edit to the comment, but regarding,

> Any plans for wcwidth to measure text widths directly with curses?

No, not in wcwidth, sorry. This technique is way too slow to be practical for all purposes, like in an editor or something.

But if somebody really wishes to have it, a terminal-accurate measurement, so long as they don't mind that the terminal will be printing it onto the screen, and some care has to be taken for very long strings that wrap the margin, in any case I could reduce it to a reusable public API of ucs-detect, but probably not wcwidth, it's pretty much what you already wrote (but without using curses) https://github.com/jquast/ucs-detect/blob/2576e8803577e4281bdecab0194532346125a7b4/ucs_detect/measure.py#L196-L215

To be clear, the "getyx" function is sending a hidden terminal sequence, "Where is your cursor?" and interpreting the automatic reply from the terminal. So this technique only works with a live terminal attached, and a place to print text. Curses is not aware of the printable width, just telling you where your cursor went to..

edit: from https://www.jeffquast.com/post/terminal_2/

> Applications may write hidden messages that change the state of your terminal, but they may also request your terminal emulator to write hidden messages in return!

> Let's try one, Report Cursor Position:
> ```
> $ printf "\x1b[6n"; read input
> $ set | grep ^input
> input=$'\E[38;1R'
> ```

qrvd:
Yes I made a few edits to elaborate what I meant, sorry about the misunderstanding.

Some points in favor of the curses method is that we can apparently create a "virtual window" to draw any string to, and it wouldn't be visible to the user at all. And from what I can tell such "virtual windows" needn't be as big as the terminal, they can be up to 1023 columns wide (I'm stilll uncertain about the limit). Hence that was my approach in the proof-of-concept code in the earlier message.

> To be clear, the "getyx" function is sending a hidden terminal sequence, "Where is your cursor?" and interpreting the automatic reply from the terminal. So this technique only works with a live terminal attached, and a place to print text. Curses is not aware of the printable width, just telling you where your cursor went to..

I'm still unsure why `curses` can tell where the cursor is with the more difficult Unicode strings. After briefly browsing the source code of ncurses I didn't see any such escape sequence involved*, instead it seems to track and update the cursor position internally in the call to `addstr`.

Another thing I don't understand is, if terminal escape sequence was used to read the position, why does this same code *overestimate* the text width in a terminal like `kitty` for example? (try it with the string `এবং লোকজন অনেক অতিথিপরায়ণ। বাংলাদেশের ইতিহাস এবং সংস্কৃতি অত্যন্ত সমৃদ্ধ।`) This needs more investigation but it seems that kitty is purposefully exposing to programs a larger width than it actually uses to display the text, maybe in order to not break old applications that don't support Unicode.

> So this technique only works with a live terminal attached, and a place to print text.

I had a look at the code you just linked and the ucs-detect measurement indeed should work just as well as curses. For users with no terminal atatched, maybe it would interest you to try measuring in a ["pseudo-terminal"](https://docs.python.org/3/library/pty.html)? (I haven't tried this method though)

jquast:
Ah sorry that I didn’t review the source of getxy, I assumed because it got correct results it worked this way, I will say that it is very likely it “accidentally” gets the correct position sometimes, like that a zero-wide and full-wide can cancel each other out and both measure the correct final result 2, and, to encourage you to examine the “udhr” data files in ucs-detect and make more extended experiments of your own.

As for pty, I make extended use of them in a lot of my code so I can say for certain that it will not work, as it cannot negotiate or “interact” with automatic sequences, as a pseudo terminal it would be your job as controlling the “master” side to look for and conditionally react to terminal sequences found in output, that’s exactly how “pyte” library works, all terminals use pty actually, that they respond to the sequences in output is what makes them capable

Also to suggest that curses is very dumb about terminal capabilities. It looks no further than TERM environment variable, the “notcurses” library works very hard to do what curses doesn’t, and, like ucs-detect to interrogate and understand the difference of kitty and mlterm,, for example,m that legacy curses doesn’t do

-- 
  Jeff Quast
  ***@***.***

qrvd:
Thanks for the clarification on pty. I'll have a look at ucs-detect when I get back to this. 

And notcurses sounds really good, but I can't make the call on switching this project over to it. I think pragmatically we might end up settling for whichever solution tells us how much the cursor moved by, even if the particular terminal used doesn't join codepoints correctly etc. since the reason we are doing all these calculations is to have our UI line up regardless of the text shown (broken or not). The maintainer seemed to prefer solutions that don't add dependencies, and thankfully we already use curses. But as I said, not my call exactly. notcurses is another option on the table.

jquast:
I'll look into improving on this PR this week, I hope to help, it requires a bit more study, I think maybe WideString.\_\_getslice\_\_ can plug-in to [wcwidth.clip](https://wcwidth.readthedocs.io/en/latest/intro.html#clip) for example, and also acquire better awareness of more complex like emoji and flags, I will try to include demonstrating examples and test cases I have made many lately

toonn:
It's great to see interest in this! It's such a deceptively complex issue.

PR #1891, opened 2020-03-28
Refactor human_readable() from template strings to format function

mik13ST:
While adding a copy speed and other stats to the TaskView in #1840, the code of the `human_readable()` was about to be refactored when I saw the old string formatting syntax. This syntax should be avoided in favor of the `.format()`. This does the refactoring.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: Arch Linux, kernel 5.4.15-arch1-1
- Terminal emulator and version: termite v15
- Python version: 3.8.1
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Refactoring this should be done carefully because there apparently were some performance issues with this function. See 7bc8b3f. Also see [this](https://github.com/ranger/ranger/blob/master/ranger/ext/human_readable.py#L29) comment.

The conclusion is the performance of the `human_readable()` is not critical and not noticeable on a RaspberryPi 1 even if it had 20 times worse performance. See the research in https://github.com/ranger/ranger/pull/1891#issuecomment-622968812.

#### TESTING

 I created simple tests myself (the first commit of the chain) and these are the results (the units are seconds):

Performance was tested using 8b64be4877446af630665884a906a0dd24a655de.
| method | asb diff | rel diff | avg | avg (pypy2) | run1 | run2 | run3 | run4 | run5 | run6 | run7 | run8 | run9 | run10 |
| - | - | - | - | - | - | - | - | - | - | - | - | - | - | - |
| original | 0 | 0% | 6.60 | 2.40 | 6.30 | 6.41 | 8.32 | 6.45 | 6.38 | 6.44 | 6.43 | 6.43 | 6.43 | 6.45 | 
| `format()` python 2.6 incompatible | +1.42 | +21% | 8.02 |   | 7.96 | 7.99 | 7.92 | 7.94 | 8.01 | 7.99 | 7.98 | 8.06 | 8.29 | 8.06 | 
| `format()` | +1.72 | +26% | 8.32 |  3.35 | 8.27 | 8.24 | 8.36 | 8.25 | 8.24 | 8.32 | 8.27 | 8.53 | 8.30 | 8.45 | 
| `format()` without separator | +0.62 | +9% | 7.22 |   | 7.23 | 7.17 | 7.36 | 7.15 | 7.22 | 7.18 | 7.21 | 7.22 | 7.22 | 7.32 | 
| custom float format (see snippet) |  |  |  | 2.91  |  |  |  |  |  |  |  |  |  |  | 
| external [1] python2 incompatible? | +3.25 | +49% | 9.85 |   | 9.86 | 9.85 | 9.87 | 9.80 | 9.90 |  |  |  |  |  | 
| external [2] | a lot more | a lot more | 21.24 | 9.86  | 21.28 | 21.21 | 21.23 |  |  |  |  |  |  |  | 
| external [3] | +2.60 | +39% | 9.20 | 3.24  | 9.00 | 8.94 | 8.98 | 9.20 | 9.20 | 9.06 | 10.28 | 9.20 | 9.12 | 9.15 | 
| `humanize` library (python 2.x incompatible) | a lot more | a lot more | about 25 |   | 28.08 | 25.21 | 25.74 |  |  |  |  |  |  |  | 
| external [4] skips decimals | +1.54 | +23% | 8.14(WTH? how is this possible?) | 0.89  | 8.45 | 8.33 | 8.68 | 8.27 | 8.35 |  |  |  |  |  | 
| `hurry.filesize` library | +2.63 | +39% | 9.23 |   | 9.19 | 9.44 | 9.07 |  |  |  |  |  |  |  | 
| external [5] | +1.68 | +25% | 8.28 | 3.15 | 8.21 | 8.22 | 8.41 | 8.33 | 8.25 |  |  |  |  |  | 

Ranger itself was not tested yet and the functionality test I created is failing with current implementation. Let's not worry about that now, I will eventually fix the issues. The performance impact is what should be solved first.

#### Custom float formatting and string array concat using `join` ####
```
if byte < 2**20:
  # return '{0:.3f}{1:s}K'.format((byte / 2**10), separator)
  return ''.join([str(int(((byte / 2**10) * 100) + 0.5) / 100.0), separator, "K"])
```

#### Useful resources for optimizing speed ####
- https://wiki.python.org/moin/PythonSpeed/PerformanceTips#String_Concatenation
- https://github.com/rkern/line_profiler
- Somebody having a similar string formatting performance problems, algouth ab bit old thread. https://stackoverflow.com/a/2638807/5951116
- [1] Mathematical take on formatting the file size. https://stackoverflow.com/a/25613067/5951116
- [2] Another Mathematical take on formatting the file size. https://stackoverflow.com/a/10171475/5951116
- [3] Yet another mathematical approach. https://stackoverflow.com/a/1094933/5951116
- [4] Using recursion. https://stackoverflow.com/a/43750422/5951116
- [5] Simple and perspective yet relatively fast. https://stackoverflow.com/a/27094657/5951116

### DIFF
--- a/ranger/ext/human_readable.py
+++ b/ranger/ext/human_readable.py
@@ -8,7 +8,7 @@ from datetime import datetime
 from ranger.core.shared import SettingsAware
 
 
-def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-statements
+def human_readable(byte_count, separator=' '):  # pylint: disable=too-many-return-statements
     """Convert a large number of bytes to an easily readable format.
 
     >>> human_readable(54)
@@ -22,19 +22,19 @@ def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-stat
     units = ['B', 'K', 'M', 'G', 'T', 'P']
 
     # handle automatically_count_files false
-    if byte is None:
+    if byte_count is None:
         return ''
 
     if SettingsAware.settings.size_in_bytes:
-        return format(byte, 'n')  # 'n' = locale-aware separator.
+        return format(byte_count, 'n')  # 'n' = locale-aware separator.
 
-    [formatted_number, unit_index] = _preformat_bytes(byte)
+    [formatted_number, unit_index] = _preformat_number(byte_count)
     if unit_index is None:
         return formatted_number
     return '{0:s}{1:s}{2:s}'.format(formatted_number, separator, units[unit_index])
 
 
-def _preformat_bytes(byte):  # pylint: disable=too-many-return-statements
+def _preformat_number(n):  # pylint: disable=too-many-return-statements
     # I know this can be written much shorter, but this long version
     # performs much better than what I had before.  If you attempt to
     # shorten this code, take performance into consideration.
@@ -50,30 +50,30 @@ def _preformat_bytes(byte):  # pylint: disable=too-many-return-statements
     # was not perceptible. I think we did our best to take
     # performance into consideration.
 
-    if byte <= 0:
+    if n <= 0:
         return ('0', None)
-    if byte < 2**10:
-        return ('{0:d}'.format(byte), 0)
-    if byte < 2**10 * 999:
-        return ('{0:.3g}'.format(byte / 2**10), 1)
-    if byte < 2**20:
-        return ('{0:.4g}'.format(byte / 2**10), 1)
-    if byte < 2**20 * 999:
-        return ('{0:.3g}'.format(byte / 2**20), 2)
-    if byte < 2**30:
-        return ('{0:.4g}'.format(byte / 2**20), 2)
-    if byte < 2**30 * 999:
-        return ('{0:.3g}'.format(byte / 2**30), 3)
-    if byte < 2**40:
-        return ('{0:.4g}'.format(byte / 2**30), 3)
-    if byte < 2**40 * 999:
-        return ('{0:.3g}'.format(byte / 2**40), 4)
-    if byte < 2**50:
-        return ('{0:.4g}'.format(byte / 2**40), 4)
-    if byte < 2**50 * 999:
-        return ('{0:.3g}'.format(byte / 2**50), 5)
-    if byte < 2**60:
-        return ('{0:.4g}'.format(byte / 2**50), 5)
+    if n < 2**10:
+        return ('{0:d}'.format(n), 0)
+    if n < 2**10 * 999:
+        return ('{0:.3g}'.format(n / 2**10), 1)
+    if n < 2**20:
+        return ('{0:.4g}'.format(n / 2**10), 1)
+    if n < 2**20 * 999:
+        return ('{0:.3g}'.format(n / 2**20), 2)
+    if n < 2**30:
+        return ('{0:.4g}'.format(n / 2**20), 2)
+    if n < 2**30 * 999:
+        return ('{0:.3g}'.format(n / 2**30), 3)
+    if n < 2**40:
+        return ('{0:.4g}'.format(n / 2**30), 3)
+    if n < 2**40 * 999:
+        return ('{0:.3g}'.format(n / 2**40), 4)
+    if n < 2**50:
+        return ('{0:.4g}'.format(n / 2**40), 4)
+    if n < 2**50 * 999:
+        return ('{0:.3g}'.format(n / 2**50), 5)
+    if n < 2**60:
+        return ('{0:.4g}'.format(n / 2**50), 5)
     return ('>9000', None)
 
 
--- a/ranger/ext/human_readable.py
+++ b/ranger/ext/human_readable.py
@@ -29,30 +29,42 @@ def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-stat
     # I know this can be written much shorter, but this long version
     # performs much better than what I had before.  If you attempt to
     # shorten this code, take performance into consideration.
+    #
+    # Revisit of 2020: In order to implement file copy speed with
+    # correct units, this had to be refactored. While we are at it,
+    # let's use format functions. It now has about 2 times worse
+    # performance than the original solution using %-formatting. The
+    # justification here is this function doesn't account for much
+    # execution time in the overall rendering time of Ranger. The
+    # function was benchmarked and profiled. Ranger as a whole was
+    # tested on a weak hardware and a user experience degradation
+    # was not perceptible. I think we did our best to take
+    # performance into consideration.
+
     if byte <= 0:
         return '0'
     if byte < 2**10:
-        return '%d%sB' % (byte, separator)
+        return '{0:d}{1:s}B'.format(byte, separator)
     if byte < 2**10 * 999:
-        return '%.3g%sK' % ((byte / 2**10), separator)
+        return '{0:.3g}{1:s}K'.format((byte / 2**10), separator)
     if byte < 2**20:
-        return '%.4g%sK' % ((byte / 2**10), separator)
+        return '{0:.4g}{1:s}K'.format((byte / 2**10), separator)
     if byte < 2**20 * 999:
-        return '%.3g%sM' % ((byte / 2**20), separator)
+        return '{0:.3g}{1:s}M'.format((byte / 2**20), separator)
     if byte < 2**30:
-        return '%.4g%sM' % ((byte / 2**20), separator)
+        return '{0:.4g}{1:s}M'.format((byte / 2**20), separator)
     if byte < 2**30 * 999:
-        return '%.3g%sG' % ((byte / 2**30), separator)
+        return '{0:.3g}{1:s}G'.format((byte / 2**30), separator)
     if byte < 2**40:
-        return '%.4g%sG' % ((byte / 2**30), separator)
+        return '{0:.4g}{1:s}G'.format((byte / 2**30), separator)
     if byte < 2**40 * 999:
-        return '%.3g%sT' % ((byte / 2**40), separator)
+        return '{0:.3g}{1:s}T'.format((byte / 2**40), separator)
     if byte < 2**50:
-        return '%.4g%sT' % ((byte / 2**40), separator)
+        return '{0:.4g}{1:s}T'.format((byte / 2**40), separator)
     if byte < 2**50 * 999:
-        return '%.3g%sP' % ((byte / 2**50), separator)
+        return '{0:.3g}{1:s}P'.format((byte / 2**50), separator)
     if byte < 2**60:
-        return '%.4g%sP' % ((byte / 2**50), separator)
+        return '{0:.4g}{1:s}P'.format((byte / 2**50), separator)
     return '>9000'
 
 
--- a/ranger/ext/human_readable.py
+++ b/ranger/ext/human_readable.py
@@ -19,6 +19,8 @@ def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-stat
     '1023 M'
     """
 
+    units = ['B', 'K', 'M', 'G', 'T', 'P']
+
     # handle automatically_count_files false
     if byte is None:
         return ''
@@ -26,13 +28,20 @@ def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-stat
     if SettingsAware.settings.size_in_bytes:
         return format(byte, 'n')  # 'n' = locale-aware separator.
 
+    [formatted_number, unit_index] = _preformat_bytes(byte)
+    if unit_index is None:
+        return formatted_number
+    return '{0:s}{1:s}{2:s}'.format(formatted_number, separator, units[unit_index])
+
+
+def _preformat_bytes(byte):  # pylint: disable=too-many-return-statements
     # I know this can be written much shorter, but this long version
     # performs much better than what I had before.  If you attempt to
     # shorten this code, take performance into consideration.
     #
     # Revisit of 2020: In order to implement file copy speed with
     # correct units, this had to be refactored. While we are at it,
-    # let's use format functions. It now has about 2 times worse
+    # let's use format functions. It now has about 3 times worse
     # performance than the original solution using %-formatting. The
     # justification here is this function doesn't account for much
     # execution time in the overall rendering time of Ranger. The
@@ -42,30 +51,30 @@ def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-stat
     # performance into consideration.
 
     if byte <= 0:
-        return '0'
+        return ('0', None)
     if byte < 2**10:
-        return '{0:d}{1:s}B'.format(byte, separator)
+        return ('{0:d}'.format(byte), 0)
     if byte < 2**10 * 999:
-        return '{0:.3g}{1:s}K'.format((byte / 2**10), separator)
+        return ('{0:.3g}'.format(byte / 2**10), 1)
     if byte < 2**20:
-        return '{0:.4g}{1:s}K'.format((byte / 2**10), separator)
+        return ('{0:.4g}'.format(byte / 2**10), 1)
     if byte < 2**20 * 999:
-        return '{0:.3g}{1:s}M'.format((byte / 2**20), separator)
+        return ('{0:.3g}'.format(byte / 2**20), 2)
     if byte < 2**30:
-        return '{0:.4g}{1:s}M'.format((byte / 2**20), separator)
+        return ('{0:.4g}'.format(byte / 2**20), 2)
     if byte < 2**30 * 999:
-        return '{0:.3g}{1:s}G'.format((byte / 2**30), separator)
+        return ('{0:.3g}'.format(byte / 2**30), 3)
     if byte < 2**40:
-        return '{0:.4g}{1:s}G'.format((byte / 2**30), separator)
+        return ('{0:.4g}'.format(byte / 2**30), 3)
     if byte < 2**40 * 999:
-        return '{0:.3g}{1:s}T'.format((byte / 2**40), separator)
+        return ('{0:.3g}'.format(byte / 2**40), 4)
     if byte < 2**50:
-        return '{0:.4g}{1:s}T'.format((byte / 2**40), separator)
+        return ('{0:.4g}'.format(byte / 2**40), 4)
     if byte < 2**50 * 999:
-        return '{0:.3g}{1:s}P'.format((byte / 2**50), separator)
+        return ('{0:.3g}'.format(byte / 2**50), 5)
     if byte < 2**60:
-        return '{0:.4g}{1:s}P'.format((byte / 2**50), separator)
-    return '>9000'
+        return ('{0:.4g}'.format(byte / 2**50), 5)
+    return ('>9000', None)
 
 
 def human_readable_time(timestamp):


mik13ST:
Is this performance impact acceptable? How to make it more efficient? I can use a [line profiler](https://github.com/rkern/line_profiler) to get the times spent on execution of each line if that is helpful here.

toonn:
You can not use implicit format specs, i.e., you'd *have* to use `{0}{1}`, rather than `{}{}`. Check if the performance difference is the same if you do this.

Thank you for doing this benchmarking. 1.5 s on a function that's called fairly often isn't very inspiring : / I wonder if maybe we can come up with a cleverer algorithm?

mik13ST:
> You can not use implicit format specs, i.e., you'd have to use {0}{1}, rather than {}{}.

Oh, the implicit format spec means Python 2.6 need to have these fields numbered. OK, that's easy.

> Check if the performance difference is the same if you do this.

Yes, there is an impact. The performance is even worse, but we to deal with it because Python 2.6 support is mandatory.

I will try replacing things like `10**20` with constants and see whether it has an effect on performance. The language interpreter might just be caching these multiplications.

mik13ST:
Replacing exponential numbers with constants did not change anything. 

I decided to share the profiling results so anybody can get an idea of what should be optimized. Any tips are greatly appreciated.

```
Timer unit: 1e-06 s

Total time: 58.6983 s
File: /home/mik/src/ranger/ranger/ext/human_readable.py
Function: human_readable at line 11

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    11                                           @profile
    12                                           def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-statements
    13  10000000      9502279      1.0     16.2      if byte <= 0:
    14         1            1      1.0      0.0          return '0'
    15   9999999      8155892      0.8     13.9      if byte < 2**10:
    16      1023         1486      1.5      0.0          return '{0:d}{1:s}B'.format(byte, separator)
    17   9998976      7856534      0.8     13.4      if byte < 2**10 * 999:
    18   1021952      1940458      1.9      3.3          return '{0:.2f}{1:s}K'.format((byte / 2**10), separator)
    19   8977024      6970607      0.8     11.9      if byte < 2**20:
    20     25600        48266      1.9      0.1          return '{0:.3f}{1:s}K'.format((byte / 2**10), separator)
    21   8951424      7037689      0.8     12.0      if byte < 2**20 * 999:
    22   8951424     17185102      1.9     29.3          return '{0:.2f}{1:s}M'.format((byte / 2**20), separator)
    23                                               if byte < 2**30:
    24                                                   return '{0:.3f}{1:s}M'.format((byte / 2**20), separator)
    25                                               if byte < 2**30 * 999:
    26                                                   return '{0:.2f}{1:s}G'.format((byte / 2**30), separator)
    27                                               if byte < 2**40:
    28                                                   return '{0:.3f}{1:s}G'.format((byte / 2**30), separator)
    29                                               if byte < 2**40 * 999:
    30                                                   return '{0:.2f}{1:s}T'.format((byte / 2**40), separator)
    31                                               if byte < 2**50:
    32                                                   return '{0:.3f}{1:s}T'.format((byte / 2**40), separator)
    33                                               if byte < 2**50 * 999:
    34                                                   return '{0:.2f}{1:s}P'.format((byte / 2**50), separator)
    35                                               if byte < 2**60:
    36                                                   return '{0:.3f}{1:s}P'.format((byte / 2**50), separator)
    37                                               return '>9000'

```

toonn:
I think you'd need to compare with the old syntax? It looks like most of the time is spent checking the conditions, not sure that can be sped up.

mik13ST:
What is the stance toward Cython? This indeed is a time critical module. Is any compiling viable?

Simply compiling the Python code without any changes should make a significant difference (35% for a certain code snippet). The compiling can be done in a typical way (using `setup.py`) or [dynamically in code](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiling-with-pyximport)(not advised for production) or even [inline](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiling-with-pyximport) witch caching. This would require installing the Cython library of course. I haven't checked Python 2.6 compatibility. I have no way of checking OSX compatibility.

Should I explore this more or is it straight unacceptable? I haven't been able to make it work in about half an hour, but I am sure any solution will be significantly faster (especially if Cython type notations are used).

toonn:
I'd rather stick with the old syntax than resort to Cython tbh. You did make me curious about the performance in PyPy though.

mik13ST:
> I'd rather stick with the old syntax than resort to Cython tbh.

In order to finish the implementation of the progress copy speed (#1840), I will have to refactor this using mathematical form anyways so sticking to the old syntax is not compatible with the changes required in #1840. 

There is an alternative solution to #1840: don't call the `human_readable()` so often. Refreshing the progress stats once or twice per second should be enough. But still, this does not mean we can neglect the performance of `human_readable()` (judging by the comment near the function).

> You did make me curious about the performance in PyPy though.

I just ran the tests using `pypy3` and they finish in about 4 seconds (mathematical) or 5 seconds (if statements). I might update the stats table just for fun. If changing the interpretor isn't a problem then this could be a solution.

toonn:
Mathematical form?

I actually meant pypy2, pypy3 isn't at the same level yet. It'd be useful to see a comparison between the old and the new format syntax.

mik13ST:
> Mathematical form?

I mean a method suitable for splitting into `human_readable_size()` and `human_readable_rate()` as discussed in #1840. You mentioned creating something like `engineering_form()` in https://github.com/ranger/ranger/pull/1840#issuecomment-595136585. I believe that isn't possible while keeping the efficient multiple if statements syntax.

> I actually meant pypy2, pypy3 isn't at the same level yet. It'd be useful to see a comparison between the old and the new format syntax.

Comparisons updated with PyPy (python2) results. Original is still the fastest. The second best is no longer the format function, I found a solution which is comparable with the format function while actually being really easy to split into rate and size specific implementations. See [[5]](https://stackoverflow.com/a/27094657/5951116) for details.

toonn:
Why is it hard to change `human_readable` to return a tuple with a number or string and a character or equivalent power? The other two functions then just add different units basically. The original function does completely different rounding from all the other implementations I tested. We should stick with it.

%-formatting being "deprecated" basically seems like FUD at this point, they started this before 2.7, we're now on 3.8 and it's still not gone, even from the stdlib afaik. So given it's ~30% faster I see no good reason to replace it with the newer syntax until it actually stops working. Other places in the codebase mostly use the newer syntax anyway so it's no big deal if we have to update.

The reason I recommended engineering form is I assumed the stdlib would have C calls to do this. Turns out python's `decimal` module has a `to_eng_form` but it's basically completely useless because it follows a terrible spec.

mik13ST:
> %-formatting being "deprecated" basically seems like FUD at this point, they started this before 2.7, we're now on 3.8 and it's still not gone, even from the stdlib afaik. So given it's ~30% faster I see no good reason to replace it with the newer syntax until it actually stops working. Other places in the codebase mostly use the newer syntax anyway so it's no big deal if we have to update.

Alright, abandoning the commits. Should I update the comment above the function with the newly discovered information (%-formatting is ~30% faster than `format` function)? The comment would prevent anyone attempting this refactoring in the future (unless it's necessary due to deprecation).

> Why is it hard to change human_readable to return a tuple with a number or string and a character or equivalent power? The other two functions then just add different units basically.

I was having difficulties with understanding the current implementation. Now I tried to summarize what I know and now I understand it almost completely. Thanks.

toonn:
Yes, feel free to add as much detail to the comment as you think is useful. Even though the result is basically the status quo it's still valuable work! Now we *know* rather than taking hut's word for it : )

mik13ST:
Should I rename and reuse this PR or create a new one for the comment update? The branch name 
from which it was merged from (mik13ST:human-readable-refactor) cannot be renamed without creating a new PR, as far as I know. Can I reference this PR in the commit message?

toonn:
Yes, do reference the PR from the commit message. And imo, the branch name is perfectly applicable. You're still going to refactor the function to extract the formatting parts, no?

mik13ST:
Comment updated, see 02397c1.

I tried to separate out the `engineering_form` but that still takes some 15 seconds to complete test. I used the profiling tool to cut some of the time down, but this is as far as I got. Here are the profiling results:
```
Timer unit: 1e-06 s

Total time: 89.2225 s
File: /home/mik/src/ranger/ranger/ext/human_readable.py
Function: engineering_form at line 11

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    11                                           @profile
    12                                           def engineering_form(number):
    13   9999999      4053270      0.4      4.5      power_of_ten = 0
    14  66996993     33296936      0.5     37.3      while number > 1000 or power_of_ten % 3 != 0:
    15  56996994     25168579      0.4     28.2          number /= 10
    16  56996994     22983169      0.4     25.8          power_of_ten += 1
    17   9999999      3720536      0.4      4.2      return (number, power_of_ten)

Total time: 231.267 s
File: /home/mik/src/ranger/ranger/ext/human_readable.py
Function: human_readable at line 19

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    19                                           @profile
    20                                           def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-statements
    21  10000000      7765262      0.8      3.4      if byte is None:
    22                                                   return ''
    23                                           
    24  10000000      7150488      0.7      3.1      if byte <= 0:
    25         1            1      1.0      0.0          return '0'
    26   9999999    176333944     17.6     76.2      n,power_of_ten = engineering_form(byte)
    27   9999999      6860671      0.7      3.0      if power_of_ten == 0:
    28      1000          942      0.9      0.0          return '%d%sB' % (n, separator)
    29   9998999      6541487      0.7      2.8      if power_of_ten < 3:
    30                                                   return '%d%sB' % (n, separator)
    31   9998999      6467799      0.6      2.8      if power_of_ten < 6:
    32    999000      1386210      1.4      0.6          return '%.4g%sK' % (n, separator)
    33   8999999      5948618      0.7      2.6      if power_of_ten < 9:
    34   8999999     12811545      1.4      5.5          return '%.4g%sM' % (n, separator)
    35                                               if power_of_ten < 12:
    36                                                   return '%.4g%sG' % (n, separator)
    37                                               # TODO continue to higher numbers
    38                                               return '>9000'
```

The `decimal.to_eng_string()` indeed cannot be used, since it accepts only nice numbers like 1000000. It just does not work for uneven numbers like 1234567. I also tried formatting the number to a scientific form using `'%.2E' % number` and then parsing the number and exponent from the string using a regular expression. This took about 30 seconds, most of the time was spent on `re.compile(...)` calls.

What about putting the units (`B`, `K`, `M`) into a list passed as an argument and then create a wrapper (or two - for size and rate) which would call it with the correct units as an argument?

toonn:
You still went with the "math form". Just take `human_readable`, rip out the units and the corresponding formats, return a tuple with the number string and the unit power, then write two functions that format the end result, mapping the unit number to a unit string (which could be done by using it as an index of a list of units). That shouldn't affect timing much, if at all?

mik13ST:
Seems like it still is slower that the original solution. In the current version it takes about 10 seconds. I first implemented it using a dictionary of units but that was really slow (about 16 seconds). It looked somewhat like this:

```
def human_readable_size(byte, separator=' '):
    units = {
        10: 'B',
        20: 'K',
        30: 'M',
        40: 'G',
        50: 'T',
        60: 'P'
    }

    number_string,power_of_two = human_readable(byte)
    if power_of_two == None:
        return number_string
    return '%s%s%s' % (number_string, separator, units[power_of_two])
```

I also tried to use a list (array) here while the indexes were directly returned by the called function (avoiding any numerical conversion like dividing by ten and subtracting one), but it performed exactly the same as the current solution (many if statements). It was faster than dict thought, so at least some lessons learned.

Any suggestions?

The profiling results are attached as files because I am too lazy to strip all the code comments out and then wait 5 minutes again for the profiling to complete. Also it would probably be quite cumbersome to read it here in a code block since it's quite long.

[profiling_results_of_split.txt](https://github.com/ranger/ranger/files/4423270/profiling_results_of_split.txt)
[profiling_results_of_original.txt](https://github.com/ranger/ranger/files/4423271/profiling_results_of_original.txt)

mik13ST:
Couldn't we call a shell command to format the number for us? There is a GNU coreutils `numfmt` to do the formatting, which, I believe, is basically the same formatting all the GNU/Linux tools use (`df -h`, `ls -h`). The `human_readable()` appears to mimic just this exact behavior (the units are the same). We might just call it instead of reinventing the wheel. It would however have to be checked on Mac OS. The invocation would have to have better or equal performance of course.

https://www.gnu.org/software/coreutils/manual/html_node/numfmt-invocation.html

The invocation could be something like:
```
numfmt --to=si 432432432432
> 433G
```

For rate (or other derived units) a `--suffix` could be used:
```
numfmt --to=si --suffix="/s" 432432432432
> 433G/s
```

Shell commands are already being used somewhere else so that might not be a problem. See for example the `__init__.py`:
```
git_describe = subprocess.Popen(['git', 'describe'],
```

Could we rely on GNU coreutils being available on all systems? How common is it?

mik13ST:
Oh, nevermind. Running commands like this is stupendously slow (test takes >30 minutes)... at least with ensuring some cross platform compatibility. Excuse my lack of knowledge and experience, I have never tried to call shell commands in a performance critical application.

```
    myCmd = os.popen("/usr/bin/numfmt --to=si %d" % byte).read()
    return myCmd.strip()
```

mik13ST:
Oh, I misinterpreted some aspects of your suggestion, again. Sorry.

Anyways, the performance is still notably worse than the original, the test now takes about 9.2 seconds compared to the 6.6 seconds of the original solution. See 0c99b6e.

Did I miss anything? Any ideas on how to improve this solution?

[profiling_results_of_0c99b6e.txt](https://github.com/ranger/ranger/files/4483946/profiling_results_of_0c99b6e.txt)

toonn:
Maybe we should just duplicate the code : /

mik13ST:
I am not happy with the duplication either but why do we care about the execution time so much? Two times worse execution time could still be OK if it accounts for only a negligible portion of time spent rendering the application. In other words, I think we should also measure the execution time relative to other parts of the application. Why did @hut refactor the function in 7bc8b3fc32b44a8db8bfb321423a1bb7718350ab? What was his motivation?

I did some initial experiments trying to find out when the execution time of the function starts noticeably affecting user experience. I tested this by adding this to the beginning of the function:

```
def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-statements
    """Convert a large number of bytes to an easily readable format.

    >>> human_readable(54)
    '54 B'
    >>> human_readable(1500)
    '1.46 K'
    >>> human_readable(2 ** 20 * 1023)
    '1023 M'
    """
    import math
    import random
    for x in range(0, 10**1):
        math.sqrt(random.random())
```
I incremented the exponent to prolong the execution time.

| extra calls to `math.sqrt()` per one function call | exponent | speed test duration (s) | felt ranger performance degradation |
|-|-|-|-|
| 0 (reference) |  | 7 | no |
| 1 | 0 | 17 | no |
| 10 | 1 | 32 | no |
| 100 | 2 | 193 | no |
| 1000 | 3 |  | a bit, the loading text appears for a brief moment, otherwise it's hard to spot a difference unless there are hundreds of files in a folder |
| 1000 | 4 |  | yes, launching ranger takes a bit longer, it's still usable |
| 10000 | 5 |  | yes |

This is however subjective and relative to hardware performance. Should I bother to compare the user experience of ranger with the your suggestion (0c99b6e) applied on a RaspberryPi 1 Model B (considering it to be the weakest device one would ever use ranger on)? I could record a video of it (running ranger on the RPi over SSH) so anyone can judge.

I could also do a profiling of the all the functions that call `human_readable()` so we know what does  `human_readable()` account for in a bigger scope. Should I?

Don't worry about TaskView rendering, I would like to limit the refresh of the stats to only a few times a second, eventually.

toonn:
Tbh, If 2 orders of magnitude don't make a perceptible difference I think it's probably fine. You've put quite enough time into this already.

mik13ST:
My time spent is irrelevant to certain extent, I am enjoying the fact there is no time budget on this project (unlike at work) so I can make this as perfect as I want (and leave when I want, because there is no deadline).

Ok, so I will convert to the format function as originally intended, update the comment to include a justification for the changes, test the performance on the RaspberryPi 1 and test the performance of separated units (0c99b6e) on the RaspberryPi (while using format functions). If all performance tests find no perceptible performance degradation, then I will remove "DO NOT MERGE" and this PR will be ready for review. Sounds good?

mik13ST:
There we go. Even though the performance is almost three times worse (test now takes about 16 seconds), I think I see no performance degradation even on the RaspberryPi.

Here is a screen capture of the performance test on the RaspberryPi: https://youtu.be/X2GTRnjYJjs.

I left the "add performance tests" commit (8b64be4) in so it can be utilized during code review. A maintainer can remove that commit from the chain when no longer needed (and if it still exists at that time).

mik13ST:
Tests now pass, the problem was this.

```
************* Module ranger.ext.human_readable
C: 31, 7: Comparison to None should be 'expr is None' (singleton-comparison)
E: 54,16: Too many arguments for format string (too-many-format-args)
```

The changes were tested in Python 2.6 and it appears to work (tested using a [CentOS with Python 2.6 Vagrant box](https://app.vagrantup.com/pcbadger/boxes/centos6-64-python2-6)).

ranger/ext/human_readable.py
```diff
@@ -19,42 +18,61 @@ def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-stat
...
 
+    units = ['B', 'K', 'M', 'G', 'T', 'P']
+
     # handle automatically_count_files false
     if byte is None:
         return ''
 
-    if SettingsAware.settings.size_in_bytes:
-        return format(byte, 'n')  # 'n' = locale-aware separator.
+    # if SettingsAware.settings.size_in_bytes:
```
toonn, line 27: Looks like you no longer take this setting into account?

ranger/ext/human_readable.py
```diff
@@ -19,42 +18,61 @@ def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-stat
...
+    #     return format(byte, 'n')  # 'n' = locale-aware separator.
+
+    [formatted_number, unit_index] = _preformat_bytes(byte)
+    if unit_index is None:
+        return formatted_number
+    return '{0:s}{1:s}{2:s}'.format(formatted_number, separator, units[unit_index])
```
toonn, line 33: I don't understand the purpose of separator very well is it what separates the number from the unit?

ranger/ext/human_readable.py
```diff
@@ -19,42 +18,61 @@ def human_readable(byte, separator=' '):  # pylint: disable=too-many-return-stat
...
 
+def _preformat_bytes(byte):  # pylint: disable=too-many-return-statements
```
toonn, line 35: Maybe we should rename `byte` everywhere? It's a bit confusing cause it's not a byte we're storing.

ranger/ext/human_readable.py
mik13ST: Yes, you understand it correctly. It typically is single white space or an empty string.

ranger/ext/human_readable.py
mik13ST: This made the module easily testable. This is part of the first commit which I didn't intended to be merged. I you want to keep the performance tests of the first commit in the repo, then I could figure out a way around this.

mik13ST: I gave it about half an hour and I cannot figure out hot to make my performance tests work with SettingsAware. So I will simply drop the first commit. I don't think my performance tests are that useful and they have served their purpose. I could just place them to the suggested directory without being actually executable without this modification (commenting out SettingsAware), but I don't like that.

ranger/ext/human_readable.py
mik13ST: That's a good point. Renamed in 322960b.

toonn:
I never got around to merging this : /
I'm going to close it in favor of #2802.

Apologies, @mik13ST, your detailed work on this was still really appreciated! ❤️ ❤️

mik13ST:
It's alright. I am no longer using Ranger so I had no plans to finish my PRs anyways. You made sure to use my research if possible so it's fine by me. I think there is another PR by me that was blocked by this that can be abandoned too.

sz-hankus:
Thank you for the research! It was very valuable.

PR #1912, opened 2020-04-09
modified file: ranger/ext/img_display.py

mayoumin:
#### ISSUE TYPE

- Bug fix

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -189,7 +189,6 @@ class W3MImageDisplayer(ImageDisplayer, FileManagerAware):
         self.process.stdin.write(input_gen)
         self.process.stdin.flush()
         self.process.stdout.readline()
-        self.is_initialized = False
 
     def clear(self, start_x, start_y, width, height):
         if not self.is_initialized or self.process.poll() is not None:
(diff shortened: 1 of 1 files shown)


toonn:
Could you provide some more detail as to what this fixes?

mayoumin:
### Detail:
System Version: Arch Linux 5.5.6
Python Version: Python 2.7.17
```
diff --git a/ranger/ext/img_display.py b/ranger/ext/img_display.py
index ffaa4c07..98736664 100644
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -189,7 +189,6 @@ class W3MImageDisplayer(ImageDisplayer, FileManagerAware):
         self.process.stdin.write(input_gen)
         self.process.stdin.flush()
         self.process.stdout.readline()
-        self.quit()
         self.is_initialized = False
 
     def clear(self, start_x, start_y, width, height):
@@ -216,6 +215,7 @@ class W3MImageDisplayer(ImageDisplayer, FileManagerAware):
             raise
         self.process.stdin.flush()
         self.process.stdout.readline()
+        self.quit()
 
     def _generate_w3m_input(self, path, start_x, start_y, max_width, max_height):
         """Prepare the input string for w3mimgpreview

```

toonn:
That's not detail. Explain, the problem you experienced and why the fix works. Maybe point to an existing open issue.

mayoumin:
When using the w3m image preview function under xterm, the image cannot be previewed. After modifying the code in this way, you can preview the image success.
```
!Sorry, English is not good. :-(
```

PR #1930, opened 2020-04-22
Fixing freezing ranger when press Ctrl-s

mrgoodvin:
Disable usage of XON/XOFF on termianl

#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5+deb10u2 (2019-08-08) x86_64 GNU/Linux
- Terminal emulator and version: any (practical used and checked: urxvt, lxterminal)
- Python version: Python 2.7.16
- Ranger qversion/commit: any (practical used: 1.9.2-4)
- Locale: ru_UA.UTF-8

#### CHECKLIST

- [] Changes require config files to be updated
    - [] Config files have been updated
- [] Changes require documentation to be updated
    - [] Documentation has been updated
- [] Changes require tests to be updated
    - [] Tests have been updated

#### DESCRIPTION

add one row code, not very deeply

#### MOTIVATION AND CONTEXT

normal usage when start ranger by terminal and out comands DE (on hotkey for example)

#### TESTING

start ranger by it comands and press Ctrl-s and usage next

#### IMAGES / VIDEOS

no any influence

### DIFF
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -15,8 +15,9 @@ import tempfile
 
 from ranger import VERSION
 
-
 LOG = getLogger(__name__)
+
+os.system("stty -ixon")
 
 VERSION_MSG = [
     'ranger version: {0}'.format(VERSION),
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -15,6 +15,7 @@ import tempfile
 
 from ranger import VERSION
 
+
 LOG = getLogger(__name__)
 
 os.system("stty -ixon")


toonn:
I'm not sure we should touch this. The user decides whether they want Software Flow Control. Ranger has no business altering that configuration.

When in doubt, defer to vim. Vim doesn't disable this by default so imo we shouldn't either.

mrgoodvin:
This value is on time work ranger, is true? Maybe need to method wrapper on suches terminal-opetations (what need in SFC), but disable any hot key not good idea. If user want to do it him need call wrapper method. No matter vim is it or emacs. I think that vim is doing it mayby. I do not know exactly, im sorry. Vim do not make Ctrl-s, realy? Thanks

toonn:
Whether the user has Software Flow Control enabled or not is up to them. I don't see a good reason for ranger to interfere with that. In a terminal some key sequences are interpreted as control characters, that's a limitation we have to accept.

mrgoodvin:
I can not accept :). Sorry. What prevents using this option when it is really necessary. Therefore, I am changing this.
And I start the ranger with hot keys, and many do it. And what should we do if we really don't need Software Flow Control (XON / XOFF) exactly?
I can handle it, but how about the rest?

mrgoodvin:
It is better to have a plug and not need it than to need and not have it.
So in this situation :)

mrgoodvin:
Do you know, how check a debug flag in pyton module? I would like to add disabling Software Flow Control only when debugging. In pycharm, this fix breaks the connection  in the debug mode.

toonn:
Ranger does not mess with the user's terminal configuration. You could easily wrap the ranger command with a script that runs `stty ixof` before starting ranger. Or you could simply configure your terminal to always disable SFC.

PR #1931, opened 2020-04-23
Fix qualificators %s in shell

mrgoodvin:
Fails of multiple fetch of files for commands in system shell with space in names.
./
*file 1
*file 2
Folder 1
console shell mv %s %f
trouble ==> $ console shell mv file 1 file 2 Folder 1
    fix ==> $ console shell mv 'file 1' 'file 2' 'Folder 1'

#### ISSUE TYPE

- Improving

#### RUNTIME ENVIRONMENT

xfce 4.12

1.9.2.-4
- Operating system and version: 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5+deb10u2 (2019-08-08) x86_64 GNU/Linux
- Terminal emulator and version: lxterminal
- Python version: Python 2.7.16
- Ranger version/commit: 1.9.2.-4
- Locale: ru_UA.UTF-8

#### CHECKLIST

- [X ] Changes require documentation to be updated
    - [ ] Tests have been updated

#### DESCRIPTION
 wraping of names into ''

#### MOTIVATION AND CONTEXT

correct handling of system shell operating (mv, cp...)

fixing

#### TESTING

copyng and moving files and directories in system shell

probably no any

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -310,9 +310,9 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             macros['f'] = MACRO_FAIL
 
         if self.fm.thistab.get_selection:
-            macros['p'] = [os.path.join(self.fm.thisdir.path, fl.relative_path)
+            macros['p'] = ["'" + os.path.join(self.fm.thisdir.path, fl.relative_path) + "'"
                            for fl in self.fm.thistab.get_selection()]
-            macros['s'] = [fl.relative_path for fl in self.fm.thistab.get_selection()]
+            macros['s'] = ["'" + fl.relative_path + "'" for fl in self.fm.thistab.get_selection()]
         else:
             macros['p'] = MACRO_FAIL
             macros['s'] = MACRO_FAIL
@@ -345,9 +345,9 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             i = str(i)
             macros[i + 'd'] = tabdir.path
             if tabdir.get_selection():
-                macros[i + 'p'] = [os.path.join(tabdir.path, fl.relative_path)
+                macros[i + 'p'] = ["'" + os.path.join(tabdir.path, fl.relative_path) + "'"
                                    for fl in tabdir.get_selection()]
-                macros[i + 's'] = [fl.path for fl in tabdir.get_selection()]
+                macros[i + 's'] = ["'" + fl.path + "'" for fl in tabdir.get_selection()]
             else:
                 macros[i + 'p'] = MACRO_FAIL
                 macros[i + 's'] = MACRO_FAIL
@@ -379,9 +379,9 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             else:
                 macros['F'] = MACRO_FAIL
             if next_tab_dir.get_selection():
-                macros['P'] = [os.path.join(next_tab.path, fl.path)
+                macros['P'] = ["'" + os.path.join(next_tab.path, fl.path) + "'"
                                for fl in next_tab.get_selection()]
-                macros['S'] = [fl.path for fl in next_tab.get_selection()]
+                macros['S'] = ["'" + fl.path + "'" for fl in next_tab.get_selection()]
             else:
                 macros['P'] = MACRO_FAIL
                 macros['S'] = MACRO_FAIL
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -15,8 +15,9 @@ import tempfile
 
 from ranger import VERSION
 
-
 LOG = getLogger(__name__)
+
+os.system("stty -ixon")
 
 VERSION_MSG = [
     'ranger version: {0}'.format(VERSION),
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -15,6 +15,7 @@ import tempfile
 
 from ranger import VERSION
 
+
 LOG = getLogger(__name__)
 
 os.system("stty -ixon")


toonn:
This function already provides functionality for escaping macro values: https://github.com/ranger/ranger/blob/d5ab70df4780de1aa9759913761248c38553c0df/ranger/core/actions.py#L278

Which ranger commands are you noticing problems with?

mrgoodvin:
> This function already provides functionality for escaping macro values:
> 
> https://github.com/ranger/ranger/blob/d5ab70df4780de1aa9759913761248c38553c0df/ranger/core/actions.py#L278
> 
> Which ranger commands are you noticing problems with?
I will see it latter. Thanks 
OK. Try to copy two files contained ' ' (space) in home, please: **shell cp -iv %s ~**, where selection files are **file 1** and **file 2**

toonn:
I can't reproduce what you're claiming.
I create `file 1` and `file 2` with the spaces. I create a directory `~/test`.
I open ranger and mark `file 1` and `file 2`.
Then I execute `:shell mv %s ~/test`.
The files end up in the `~/test` directory without errors of any kind.

mrgoodvin:
How do you separate files in the your system shell? Ok. I will look at myself again.

mrgoodvin:
No. It no work in my terminal. Maybe ru.utf-8. I must use "file 1" 'file 2' or My\ Drawings/ as example only. I think that is local influence on result of terminal. For cirillic DE it st actualy. It exactly. Done!

toonn:
But if it works on my end that means `shell_quote` *is* being called, so if it doesn't work that function would need fixing.
Including `'`'s around the filenames isn't good enough `This filenames contains a '.txt` is a valid filename and wouldn't be escaped properly if you just include single quotes around it.

mrgoodvin:
Yes? It is thus. Ok. Adding in comment my fix Ctrl+s and marked "Only Cyrillic users uncomment". And doc updating with it. If attribute has be found, addin insted %spece some-any  :). Yes it is problem, I did think about too :). Delimiter need be space is true. ' in name of file and directory is valid. I checked it. But only we need use ' aroud files and directories. In my patch done it. Or we can around with ' files alweys or if have attributing of fix Ctrl+s. I see it thus. Two way one true :)

toonn:
> ' in name of file and directory is valid. I checked it. But only we need use ' aroud files and directories.

Could you rephrase this?
As I understand it you're saying two thing:
  1. `'` (apostrophe) is indeed a valid character in file and directory names.
  2. We only need to surround file and directory names with `'`'s.

I agree with 1., all characters are allowed in file names except for `/` and NUL.
I don't understand 2., putting apostrophes around a file name isn't enough to escape it. Let me give an example:

Filename: `file with ' and spaces.txt`
If you surround this name with apostrophes, you get: `'file with ' and spaces.txt'`
This is interpreted as three names: `file with`, `and` and `spaces.txt'`

toonn:
Is the escaping of file names still wrong when you run `ranger --clean`?
Run `ranger --clean` and try selecting some files with spaces in their name, then run `:shell mv %s ~/somewhere`.

mrgoodvin:
1. I Agree to.
2. Why? Enough if add ' when files to be on process to adiing in list. What diding patch.
Certainly, if to forming string on shell when all are mixed, will be that you say about. Yes. Adding ' in process to adde in list when we know name of file.

mrgoodvin:
> Is the escaping of file names still wrong when you run `ranger --clean`?

No, do not copy

toonn:
Ah, I think what's happening is you're adding `console` to the command which causes the macros to be evaluated outside of a "shell" context.

If you either drop the "console" or escape the macros `%%s`, it'll work.

This cannot be the default for the `console` command because `console touch %s`, should result in `:touch file 1`, not `:touch 'file 1'` which would create a new file with two apostrophes in the name.

mrgoodvin:
Yes, it is true. I want to show  operation for eddit by my :) I try with out console operating. But.. may system console corectly handling touch 'file 1'. And create file with name of 1 when touch file 1

mrgoodvin:
Proble in that what conactinate happen in file of config and mixing. A handling you do after, in shell. One way or only one :), to process identification of files to do on adding. You may to create function addin file name in list.
Proble in that what conactination happen on config file and mix. A handling you do after in shell where you do not controll nothing. One way or only one :), to processing identification files on adding in list. You may to create function addin file name in list. I see thus it.

mrgoodvin:
> If you either drop the "console" or escape the macros `%%s`, it'll work.

no, do not working

mrgoodvin:
Ok. I am staring to cripting...

toonn:
I'm talking about ranger's `:touch` which you'd be more likely to use from a ranger key binding. That's just an example, there's endless possibilities.

I understand what you want now though.
I think we should add a flag to the `:console` command that enable escaping of macros. Maybe it should even become the default. Existing mappings that shouldn't get escaping can then use a `--no-shell-escape` flag or something.

mrgoodvin:
Ok. No

> I understand what you want now though.
> I think we should add a flag to the `:console` command that enable escaping of macros. Maybe it should even become the default. Existing mappings that shouldn't get escaping can then use a `--no-shell-escape` flag or something.

_Agree, we can go this way.

mrgoodvin:
Tell me, please, which symbols need to shield for your file name that to be used without arounds quotation mark?
These in my system shell:
en: ,[];'`<>?{}:"$^&
ua, ru: ,"№;:?

toonn:
You shouldn't have to care. We use a python stdlib function to do the escaping. If there's a bug in that function you should report it to the python project.

mrgoodvin:
I am understanding You. I see one small problem. I show it as example.
Have one file with next name:
"file '1' "name "" "
At now ranger do not handling it correct. Excluding that commands which using calls in 'os', for example 'rename' in 'core.actions.py'. It is correct name in OS, but ranger do not know it.
How we can do not handle  any file name at all. We can translate it name to system shell, and a system shell need to handle a file name.
But before transfer it in system shell we need to translate this name in to next view:
command \"file\ \'1\'\ \"name\ \"\"\ \"
We can pass more files to system shell without delimiters, simply as string.
We do not around file names and directories at all.
I'm thinking how do it, and a need more information for analyse.

toonn:
Hmm, I touched that filename `:touch "file '1' "name "" "1'` which created the file just fine.
I then ran `:console -p7 rename '%f'` this opened the console like this (where ^ indicates the position of the cursor) `:rename ^'"file '1' "name "" "1''`.
Pressing CR/enter move the file just fine `"file '1' "name "" "1' -> '"file '1' "name "" "1''`.

If I run `:eval fm.open_console("rename '%f'", position=7)` I get the following `:rename ^'%f'` (^ is the location of the cursor). No exceptions.

You don't have to worry about this issue anymore. I have two fixes in mind, one of which is elegant in my opinion, `console` would just detect whether macros need shell escaping or not.

mrgoodvin:
Well, maybe I broke something in my local version. I will then look at myself.

mrgoodvin:
I update idea for it on my settings for ranger (it's last). It seems to me more universal, but probably the most difficult to implement.

hut:
Thanks for your work on this, @mrgoodvin!

I reviewed the PR (sorry for the delay!) and it looks to me like it's not the right way to solve this issue. As @toonn mentioned, it should be solved in the `console` command, not as a workaround in the macro substitution function.

I will close this PR for now.

PR #1934, opened 2020-04-23
Task. Translate kb from en in ogther(ua ru)

mrgoodvin:
TASK to keyboard translation key bindings
Inputs key bindings commands only Latin with any language-type of keyboards

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

no matter

no matter
- Operating system and version: no matter
- Terminal emulator and version: no matter
- Python version: 
- Ranger version/commit: 
- Locale: ALL

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

!This text mayby changes or|and updated. Do watch for it desctiption fo to be competent in this task
2020-04-29::4

DEVELOPMENT TASK
REORGANISATION

1. planning option setting-value 'en-keys' and 'other_keys' as string: For example see code
2. planed create pyton module with name keyboards.py
   • It containe a code to handling of mapped pressed key onto Latin ascii.
   • (not actual) If need to correct file name in shell operation, it prepares base to do it
   ‣ This module will be imported in ui.py
   ‣ This module wiil be check existing 'en-keys' and 'other_keys' in rc.conf
   ‣ This module maybe adding source code "os.system("stty -ixon")" ... Now it is unknown

Base ending result: Translate input key bindings for hot keys commands (only for hot-key commands and only) without swith keyboard for any language KB

EXAMPLE: user pressed hot keys ss(find) to find directory name "Фильмы" (ru) without it
----------------------------------------------------------------------
user action1; user controll curren KB. If curren KB not EN, user selected KB=EN
user action2; pressed-> ss
user action3; user selected KB=RU
user action4; input-> Фильмы
user action5(1); if need command next, user selected KB=EN

Next: User can do it on KB=EN | KB=RU | KB=AU selected himself languages KB of marke uppered

EXAMPLE: user pressed hot keys ss(find) to find directory name "Фильмы" (ru) with it
----------------------------------------------------------------------
user action1; pressed-> ss
user action2; user controll curren KB. If curren KB not RU -> user selected KB=RU
user action3; input-> Фильмы

User of Ukraine often use three languge keyboards: EN RU UA
It will be useful too else languages. Example for Germany user where use umlauts. And it like...
It making any file.conf universl. It will be work in else users of different countries, for it enough remap keys bindings fillint two strings in your rc.conf.

en_keys - keys on English (Latin)
other_keys - keys of user mapping on en (now is 'ru' in it)

other_keys for default contain "" and|or is comment. User need userself input these string
en_keys for default contain alphabetic latin [a-zA-Z] and|or is comment.
For enable supprot userself keyboards, need uncomment it and fill it

mapping next symbols:
`abcdefghijklmnopqrstuvwxyz,./[];''`ABCDEFGHIJKLMNOPQRSTUVWXYZ<>?{}:"~@#$^&`

other_keys mayby contain two and more KB for one letter
----------------------------------------------------------------------
Example key 'r' for RU and UA (ruusina and ukrain 'к' and more else for ukrain and russian are identical)
en_keys = "rR"
other = "кК"

Example key 's' for RU and UA
en_keys = "ssSS"
other = "ыіЫІ"

Example key 's' for RU only
en_keys = "sS"
other = "ыЫ"

Example key 's' for UA only
en_keys = "sS"
other = "іІ"

String-values 'en_keys' and 'other-keys' will be existing in single existens instance. User maybe to use it, for adding suport userself keys with userself languages. See code it (Russian and Ukraine mapping to English)

Example for Cyrillic:
en ="abcdefghijklmnopqrsstuvwxyz,./[]];''``" \
    'ABCDEFGHIJKLMNOPQRSSTUVWXYZ<>?{}}:""~~' \
    '@#$^&'
other ="фисвуапршолдьтщзйкыіегмцчнябю.хъїжэєё'" \
    'ФИСВУАПРШОЛДЬТЩЗЙКЫІЕГМЦЧНЯБЮ,ХЪЇЖЭЄЁʼ'	\
ru_mm ='"№;%:?'

Not translated symbols
----------------------------------------------------------------------
There are special keyboard layout keys/symbols for certain languages that are not included in the “other_keys”, which like “+” do not participate in the translation (ru,ua exactly). They pass as is.
For example: '=' or '+' (Shift-=) for russian, ukrain and english is identical.

Duplication|Conflict of some up-digital keyboard shortcuts in Cyrillic languages (for converting only)
----------------------------------------------------------------------
Usually users of languages from the Cyrillic group do not use the S-6 key and similar at all. For the reason that they are duplicated on an alphabetic keyboard. There is two way solve it small problem. See down next.
Theory:
For example RU: S-6 (EN) duplicate S -; (RU). Whay that is problem? This is that because pressed it hotkey (S-6) for these languages give two different symbols, where one from it is be in bouth languages — it is ':' for this hotkey.
```
|----+-----+----+----+-------------------------------------------|
| KB | KEY | CH | IN | DESCRIPTION                               |
|----+-----+----+----+-------------------------------------------|
| EN | S-6 | ^  | :  | has intersection S-; at alphabetic block  |
| RU | S-6 | :  |    | 'Ж' at on pressed S-; at alphabetic block |
|----+-----+----+----+-------------------------------------------|
| EN | S-4 | $  | ;  | has intersection  ;  at alphabetic block  |
| RU | S-4 | ;  |    |                                           |
|----+-----+----+----+-------------------------------------------|
| EN | S-2 | @  | "  | has intersection S-' at alphabetic block  |
| RU | S-2 | "  |    | 'Э' at on pressed S-' at alphabetic block |
|----+-----+----+----+-------------------------------------------|
```

The user expects the same command to press S-6 on EN and on RU (UA). But he gets two different command on press it keybinding. Or comands works only in EN. The user is in shock. No, advanced users know this fact, and do not use these keyboard shortcuts at all. But ...

Possible
1 - do not use these hotkey at all
2 - merge these hotkeys on bouth keyboards
For to marge hotkeys used strings in en_key and other_keys
other_keys_mm = '"№;:?'

If used mergin method (MM) then applied hotkey on key bindings of EN only, keybindings of other KB will be ignored.
If used different commands method (DCM) then applied keybindings all keyboards.

MM: Indeed even it maybe bring in of user in a shock. Now two hotkeys are works at equals - the one command.

DCM: One hotkey -> two commands
MM: Two hotkeys -> one command

Public custom generation string 'en' and 'other' ranger user
----------------------------------------------------------------------
It can implement on web-site, where will be select userneeded language kayboards. It allow easy update data base and public access for anybody.
• multiple choice kayboard layouts
• public access for anybody
• ability selecting of translate method DCM/MM (for two of many selected languages if ability)

General information
----------------------------------------------------------------------
For hieroglyphic-languages it need to be remade appreciably. It demand hard deep thinking in a process of Unicode

#### MOTIVATION AND CONTEXT

useful

useful

#### TESTING

Simple TEST

Marginal changes and margin in the fature

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -321,6 +321,18 @@ set nested_ranger_warning true
 # setlocal path=~/downloads sort mtime
 
 # ===================================================================
+# == Command description
+# ===================================================================
+# You can add description on command or group command use next syntax
+# map KEY [chain echo YOUR_NAME_GROUP_COMMAND::[YOUR_NAME_COMMAND];] commands
+# where commands is/are comand1[; command2][; ...commandN]
+
+# ===================================================================
+# == Uses multiple selected files %s, %p, %S, %P
+# ===================================================================
+# Need use these qualificators without including in '' for correct handling of shell
+
+# ===================================================================
 # == Command Aliases in the Console
 # ===================================================================
 
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1295,8 +1295,10 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             return self.tab_open(narg)
         assert isinstance(offset, int)
         tablist = self.get_tab_list()
-        current_index = tablist.index(self.current_tab)
-        newtab = tablist[(current_index + offset) % len(tablist)]
+        tab_count = len(tablist)
+        go_index = tablist.index(self.current_tab) + offset if abs(
+            offset) <= tab_count else 0 if offset <= 0 else tab_count - 1
+        newtab = tablist[go_index % tab_count]
         if newtab != self.current_tab:
             self.tab_open(newtab)
         return None
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -277,7 +277,7 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
             elif keys[0] == 27:
                 keys[0] = ALT_KEY
                 tkeys = key_translate(keys[1:])
-                if tkeys: keys = keys[0:1] + tkeys + keys[2:]
+                if tkeys: keys = keys[0:1] + tkeys
             if self.settings.xterm_alt_key:
                 if len(keys) == 2 and keys[1] in range(127, 256):
                     if keys[0] == 195:
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -69,8 +69,9 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
     is_on = False
     termsize = None
 
-    ru ="фисвуапршолдьтщзйкыіегмцчнябю.хъїжэєё'" +'ФИСВУАПРШОЛДЬТЩЗЙКЫІЕГМЦЧНЯБЮ,ХЪЇЖЭЄЁʼ' + '"№;:?'
-    en ="abcdefghijklmnopqrsstuvwxyz,./[]];''``" +'ABCDEFGHIJKLMNOPQRSSTUVWXYZ<>?{}}:""~~' + '@#$^&'
+    ru = "фисвуапршолдьтщзйкыіегмцчнябю.хъїжэєё'" + 'ФИСВУАПРШОЛДЬТЩЗЙКЫІЕГМЦЧНЯБЮ,ХЪЇЖЭЄЁʼ'
+    ru_mm = '"№;%:?'
+    en = "abcdefghijklmnopqrsstuvwxyz,./[]];''``" + 'ABCDEFGHIJKLMNOPQRSSTUVWXYZ<>?{}}:""~~' + '@#$%^&'
 
     def __init__(self, env=None, fm=None):  # pylint: disable=super-init-not-called
         self.keybuffer = KeyBuffer()
@@ -259,8 +260,8 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
                     self.update_size()
                 else:
                     if not self.fm.input_is_blocked():
-                        if chr(key) in self.ru:  # translate S-[0-9], etc from ua or ru into en as simple key
-                            key = ord(self.en[self.ru.index(chr(key))])
+                        if chr(key) in self.ru_mm:  # translate S-[0-9], etc from ua or ru into en as simple key
+                            key = ord(self.en[len(self.ru)+self.ru_mm.index(chr(key))])
                         self.handle_key(key)
             elif key == -1 and not os.isatty(sys.stdin.fileno()):
                 self.fm.exit()  # STDIN has been closed
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -69,6 +69,9 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
     is_on = False
     termsize = None
 
+    ru ="фисвуапршолдьтщзйкыіегмцчнябю.хъїжэєё'" +'ФИСВУАПРШОЛДЬТЩЗЙКЫІЕГМЦЧНЯБЮ,ХЪЇЖЭЄЁʼ'
+    en ="abcdefghijklmnopqrsstuvwxyz,./[]];''``" +'ABCDEFGHIJKLMNOPQRSSTUVWXYZ<>?{}}:""~~'
+
     def __init__(self, env=None, fm=None):  # pylint: disable=super-init-not-called
         self.keybuffer = KeyBuffer()
         self.keymaps = KeyMaps(self.keybuffer)
@@ -235,35 +238,17 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
         for key in keys:
             self.handle_key(key)
 
-    def handle_input(self):  # pylint: disable=too-many-branches
-        key = self.win.getch()
-        if key == curses.KEY_ENTER:
-            key = ord('\n')
-        if key == 27 or (key >= 128 and key < 256):
-            # Handle special keys like ALT+X or unicode here:
-            keys = [key]
-            previous_load_mode = self.load_mode
-            self.set_load_mode(True)
-            for _ in range(4):
-                getkey = self.win.getch()
-                if getkey != -1:
-                    keys.append(getkey)
-            if len(keys) == 1:
-                keys.append(-1)
-            elif keys[0] == 27:
-                keys[0] = ALT_KEY
-            if self.settings.xterm_alt_key:
-                if len(keys) == 2 and keys[1] in range(127, 256):
-                    if keys[0] == 195:
-                        keys = [ALT_KEY, keys[1] - 64]
-                    elif keys[0] == 194:
-                        keys = [ALT_KEY, keys[1] - 128]
-            self.handle_keys(*keys)
-            self.set_load_mode(previous_load_mode)
-            if self.settings.flushinput and not self.console.visible:
-                curses.flushinp()
-        else:
-            # Handle simple key presses, CTRL+X, etc here:
+    def handle_input(self):
+        def key_translate(keys):  # translate key from from ua or ru into en
+            if not self.console.visible or self.console.question_queue:
+                ch = b''.join([bytes.fromhex('{0:X}'.format(i)) for i in keys]).decode('utf-8')
+                if not ch: return []
+                if ch in self.ru:
+                    i = self.ru.index(ch)
+                    ch = self.en[i]
+                    return [ord(ch)]
+
+        def simple_key(key):  # Handle simple key presses, CTRL+X, etc here:
             if key >= 0:
                 if self.settings.flushinput and not self.console.visible:
                     curses.flushinp()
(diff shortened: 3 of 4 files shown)


mrgoodvin:
Join to work, please!

toonn:
What are you trying to solve here? Is this supposed to make ranger's key bindings work with a russian layout? Why no just create a localized version of `rc.conf` where the mappings use the cyrillic alphabet?

mrgoodvin:
Because we use two languages, and sometimes three - ru, ua, en

toonn:
Oh, so it's to have both the latin alphabet and cyrillic alphabet characters as bindings?
I think a nicer solution would be to make some new files `rc_ru.conf` for example that only contain `copymap b Б` (Just an example I don't read cyrillic so I don't know which letters'd correspond.) and similar commands.

Then anyone who wants to make use of them can source those language specific files from `rc.conf` to get the bindings working with both alphabets.

mrgoodvin:
No, it is no way true. I may addin self key-binding file, You will be surprised :). Many hot-keys, and include changes hot keys for userself it will be big trouble :). No, not good, maybe. May be to create py with handling other language and dinamicaly attach him by availability Ctrl+s attributa. It py checking having other_keys and en_keys in rc.conf(they option and they is can absent). But how attach py and how use it.... selected function in condition availability Ctrl+s? Hm....
Or.... Maybe simple refer to them from views.py :) ? You will maybe do not like it... I will understand You.

mrgoodvin:
> Oh, so it's to have both the latin alphabet and cyrillic alphabet characters as bindings?

It posibile, yes. But we don't do thus. It not universally.

toonn:
Are you asking whether a contribution of an `rc_ru.conf` file would be accepted? It probably would, maybe in a separate repository.

I'm not sure what you mean by "replays function adding ' in file if other=ru."

Any `SettingsAware` object should have access to the settings (`rc.conf`).

mrgoodvin:
You can see on my hot keys conf https://files.dp.ua/TZMeawWV

> I'm not sure what you mean by "replays function adding ' in file if other=ru."

Mistake. I means replacer-function (wraper)

toonn:
I can't view your link because of my adblocker. You can include the text in your comment.

So this is supposed to translate characters when using macros? But if you change the characters from `Мир.txt` to `Mir.txt` the commands won't work, they'll affect the wrong file?

mrgoodvin:
> I can't view your link because of my adblocker. You can include the text in your comment.

https://github.com/mrgoodvin/ranger_settings

> So this is supposed to translate characters when using macros? But if you change the characters from `Мир.txt` to `Mir.txt` the commands won't work, they'll affect the wrong file?

I will description it more in task, than wiil be more information. It script translate pressed keys only. If I change `Мир.txt` to `Mir.txt`... Of course is different valid names. I do not understend question.

toonn:
Can you give an example? How would it affect a keybinding like `+ux` for example. (Or some other bindings if there's a better example.)

mrgoodvin:
I try find or make that binding and test it. Now tell to next. Must wiil be all coll :) because i do not mapping 0-1 end else similar symbols. Rempping only next abcdefghijklmnopqrsstuvwxyz,./[]];''``
+ will be pass for translating
Why i do not use 0-9 + - and similar, because mayby duplication on Shift key. Your right, that symbols do not recommend to user at all. Even in without it improvement :)

toonn:
Ok, so why not add `copymap : <S-6>`? (Or the symbol shifted 6 inputs, of course.)

mrgoodvin:
It makes no sense to convert this because a different result on different KB. And in Russian it is also used since the analogue <S -;> in English KB.
For good, we need to use only letters. They do not intersect on any keyboard layouts UTF8 under the condition of use utf8.
User can add anything to the settings file. Ultimately, it is the user's prerogative to reassign hotkeys.
But it trouble are all where use Cyrillic: vim, emacs and for ranger to will be too

mrgoodvin:
Tell me, please, which symbols need to shield for file name that to be used without  arounds quotation mark?
These in my system shell:
en: ,[];'`<>?{}:"$^&
ua, ru: ,"№;:?

mrgoodvin:
_update description about problem for Cyrillic-users for your information. But seem we solved it. I am checked only S-6. But i do not use similar hotkeys :).

toonn:
Ok, I think I understand what you're trying to do and I appreciate the effort.
But this is a really complicated system full of surprises. Does RU "a" mean the same thing as EN "a" or is there a different corresponding character, is a very confusing question to have to ask yourself.

Because of the complication and fragility this will not be accepted into the ranger codebase.
You're welcome to refactor this as a plugin however. I could imagine users with your exact problem to find this really useful.

What we *would* accept in a PR is localized key binding files. There's two options here, either you translate the entire `rc.conf` into an `rc_ua.conf` for example, including comments. This would be useful for people who only/mainly use the Ukrainian keyboard layout. The other option is to have a `ua_maps.conf` file that only contains `copymap` commands, users of both an English and a Ukrainian keyboard layout could then source that file from `rc.conf` in addition.

Example `rc_ua.conf` (I used google translate so the translation might be terrible.):
```
# Показати приховані файли? Ви можете перемкнути це, ввівши "яє"
set show_hidden false
.
.
.
map І shell $SHELL
.
.
.
map і  console shell%space
.
.
.
map яє    set show_hidden!
```

Example `ua_maps.conf`:
```
copymap s і
copymap S І
```

mrgoodvin:
> Does RU "a" mean the same thing as EN "a" or is there a different corresponding character, is a very confusing question to have to ask yourself.

No, RU "a" and EN "a" are different symbols as symbol "ON PROGRAM EXECUTION" pressed key on different KB. In my conf use ascii for `map as`, Yes. Do not translate it in any language. My conf file same as your conf file. I typed keysbinding in my conf on EN only. Differences only in words such as `Видео, Документы, Музыка, Шаблоны` what in english mean `Video, Documents, Music, Templates`.

> Because of the complication and fragility this will not be accepted into the ranger codebase.
You're welcome to refactor this as a plugin however. I could imagine users with your exact problem to find this really useful.

Yes, please. I will only be glad.
The code for converting hot files is not fragile, my file is yes (for now, but I'm improving it).
Please try my code (last commit) on your implementation. You will see and understand everything. He will not change this code for you :)
.... unless of course you put the Russian or Ukrainian layout.

Well, I will give you an example with whom it will be useful for me to talk on this topic.
Russia, Ukraine, Belarus, Poland, Serbia, Slovenia, Greece, Germany, Yugoslavia, possibly Sweden and Switzerland, Croatia. They all have problems with this. And they redefine the same keyboard shortcut twice if they use English and their native language. This is if they use two layouts. If three, then three times :). I use three because I live in Ukraine.

> What we would accept in a PR is localized key binding files. There's two options here, either you translate the entire rc.conf into an rc_ua.conf for example, including comments. This would be useful for people who only/mainly use the Ukrainian keyboard layout. The other option is to have a ua_maps.conf file that only contains copymap commands, users of both an English and a Ukrainian keyboard layout could then source that file from rc.conf in addition.

And there is a third way. Why don't you upset him?
This is to make the ranger universal.

> Example rc_ua.conf (I used google translate so the translation might be terrible.):

Please, do not do that. Just use it as a native settings file for yourself. Let me fix it a little bit of it, but you can use the main team at home right now. Problems only with file paths.

But you will not be able to check my code for correctness. You will not be involved. To check, you need a Russian or Ukrainian layout.

Sorry, I see that I’m doing a lot of imprint. I need to set the English check and enable the default check in my emacs.

toonn:
Does the "a" in the Ukrainian alphabet have a different UTF-8 encoding than the "a" in the latin alphabet?

mrgoodvin:
> Does the "a" in the Ukrainian alphabet have a different UTF-8 encoding than the "a" in the latin alphabet?

Yes.

toonn:
And what's the UTF-8 encoding of Ukrainian "a"?

mrgoodvin:
_do not remember it. In code been used symbols, that is more comfortably. You can see it on [https://en.wikipedia.org/wiki/A_(Cyrillic)](url)

toonn:
And how do you ensure the right unicode value is used if you put the character in the string? What if someone else opens the file to make further changes and their encoding isn't set to utf-8, are you sure it'll be preserved? How about utf-8 normalization?

mrgoodvin:
Your are right __('your right' was been used me earlier is wrong, i was mean other sense. It is that :) ). Only utf-8 must be used on user PC. But now is it a standard in linux, yes? If it not thas, a code not will be worked or maybe work not correct. Ok. And if do not use utf-8 normalization, what maybe problems without a normalization? I seem that symbol is determined of language-KB. A string ru and en are been filling by us.
I think what may be problems with different type of same-equals KB (when '.' and ','  with (Shift) are placed at an up of numerical part instead of alphabetical part). I will test it on my PC.

mrgoodvin:
S-6 maybe worked, but it create problem in console, instead EN(") apply EN(@). And i do know how its identify in console.

mrgoodvin:
Ok, it worked but nedd to improvement|reorganization. I than close branch, Ok? If I to do it, I create new branch improvement.

toonn:
To be clear this will *not* be accepted into the ranger codebase. So new branches and new PRs aren't useful.

I recommended implementing this as a *plugin*.

mrgoodvin:
No, please, it is can not be a plugin. Throw away this code altogether then implement the idea as clean PR.

mrgoodvin:
Thanks for job!

toonn:
You're on your own. The api for plugins isn't well documented. Have a look at other existing plugins.

mrgoodvin:
Thanks.

toonn:
Good luck! And be sure to add your plugin to the ranger wiki once it works : )

mrgoodvin:
Of course : ). Good luck!

mrgoodvin:
_It was been implemented as plagin. See plugins for ranger._

toonn:
I'd remove the extra page and just add explanation to the entry in the plugin list tbh. More visible that way. Also consider adding a short example, that helps make explanations a lot easier to understand.

mrgoodvin:
> I'd remove the extra page and just add explanation to the entry in the plugin list tbh. More visible that way. Also consider adding a short example, that helps make explanations a lot easier to understand.

Thanks. Done.

Issue #1946, opened 2020-05-04
Memory leak. Ranger takes all my RAM


ThePonyCoder:
#### Runtime Environment

- Operating system and version: NixOS 20.03
- Terminal emulator and version: kitty 0.16.0 
- Python version: 3.7.6 
- Ranger version/commit: 1.9.3 
- Locale: en_US.UTF-8

#### Current Behavior
When trying to preview djvu file ranger uses all of my memory
Maybe memory leak
I don't know, maybe it is problem in only this file

#### Expected Behavior
Don't preview/raise exeption

#### Steps to reproduce
1. Download this djvu file https://dropmefiles.com/qaa3V
2. Try to preview it with ranger

#### Traceback
no traceback

toonn:
Does `djvutxt` work fine on the file? Or, if you don't have that installed, `exiftool`?

firmart:
Try the solution depicted here #1815

toonn:
Oh, thanks for noticing the similarity, firmart.

ThePonyCoder:
I found fix. I've added this to `scope.sh` to function `handle_image():`
```
## DjVu
image/vnd.djvu)
    exit 0;;
```

ThePonyCoder:
I found fix. I've added this to `scope.sh` to function `handle_image():`
```
## DjVu
image/vnd.djvu)
    exit 0;;
```

Issue #1815, opened 2020-01-23
RAM filled fastly while `ranger` tried to identify `djvu` "orientation".


firmart:
#### Runtime Environment

- Operating system and version: Ubuntu 19.04 x86_64
- Terminal emulator and version: XTerm(330) and termit 3.0.0
- ranger version: ranger 1.9.2
- Python version: 3.7.3 (default, Oct  7 2019, 12:56:13) [GCC 8.3.0]
- Locale: en_US.UTF-8

#### Current Behavior

I was on a 5MB `.djvu`  and my 16GB of RAM was fulled within 3 seconds. More specifically, I've seen in `htop` (before my computer gets frozen) that the command `identify -format '%[EXIF:Orientation]\n' --  file.djvu` was run (which comes from `scope.sh`). I guess there is an issue when `ranger` is trying to identify the orientation of the `djvu` file (MIME: `image/vnd.djvu`). The problem probably comes from `djvu` mime which is categorized as `image`.

toonn:
Is this reproducible with other djvu files? (Preferably sample djvu files.)

firmart:
I have tried to create a one-page `.djvu` file, but nothing happens with it. However, with `.djvu` of 100+ pages, the issue is reproduced (didn't try intermediates). The attached sample `.djvu` file is created by the following `sample.tex` file :  
```latex
\documentclass{article}
\usepackage{lipsum}
\begin{document}
\Huge
\lipsum[1-150]
\end{document}
```
with the command `pdflatex sample.tex && pdf2djvu sample.pdf > sample.djvu`.
[sample-djvu.zip (500KB)](https://github.com/ranger/ranger/files/4109593/sample-djvu.zip)

firmart:
An easy way to solve the problem is to select `image` mime type only supported by `identify` instead to quantify universally all `image` type which includes `.djvu`.

That's it, instead of this,

https://github.com/ranger/ranger/blob/3965e3eeafac8ce44670cfdc25ccc7b5997a3366/ranger/data/scope.sh#L142-L145
do this for instance:
```bash
 image/png | image/jpeg) 
     local orientation 
     orientation="$( identify -format '%[EXIF:Orientation]\n' -- "${FILE_PATH}" )" 
     ## If orientation data is present and the image actually
```

toonn:
Ok, so we've narrowed it down to multiple versions of ranger with multiple versions of imagemagick on multiple distros with multiple file types...

[...]

PR #1947, opened 2020-05-04
Fix memory allocation issue with `identify`.

firmart:
Use `ulimit` to bound `identify` in
- space: 100MB + filesize 
- time: 3 seconds

#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 19.04 x86_64
- Terminal emulator and version: XTerm(330) and termit 3.0.0
- ranger version: ranger 1.9.2
- Python version: 3.7.3 (default, Oct 7 2019, 12:56:13) [GCC 8.3.0]
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
There is an excessive memory allocation issue when `ranger` is using `identify` in `scope.sh`  
to query the orientation of the file of mimetype `image/*`.  (See #1815 #1929 #1946). 

I suggest to bound `identify` resources both in time and space (100MB + filesize and 3 seconds) and recover this limitation later. Moreover, since the mimetype of `djvu`  is categorized under `image`, I suggest to exclude it and only match `jpeg` and `png`. 

#### MOTIVATION AND CONTEXT

Bug: #1815 #1929 #1946 

#### TESTING

- No python code modified (all tests passed).
- `.png`, `.jpeg`, `.pdf` are still correctly displayed. `.djvu` is exclude.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -139,9 +139,17 @@ handle_image() {
         #           && exit 6 || exit 1;;
 
         ## Image
-        image/*)
+        image/jpeg|image/png)
+            local filesize=$(($(stat -c%s "${FILE_PATH}")/1024)) # filesize in kb
+            local limit=$(ulimit -v) # address size limit in kb
+            local timeout=$(ulimit -t) # cpu time limit in sec
+            # set address size limit to 100MB + filesize
+            ulimit -v $((100*1024 + $filesize))
+            ulimit -t 3 # limit cpu time to 3 sec
             local orientation
             orientation="$( identify -format '%[EXIF:Orientation]\n' -- "${FILE_PATH}" )"
+            ulimit -v $limit # restore address size afterward
+            ulimit -v $timeout # restore cpu time
             ## If orientation data is present and the image actually
             ## needs rotating ("1" means no rotation)...
             if [[ -n "$orientation" && "$orientation" != 1 ]]; then


ThePonyCoder:
It looks like your fix doesn't work on my system, I've changed scope.sh in my `~/.config/ranger` and ranger still takes all my ram.

firmart:
By "all tests passed" I mean those of `pytest`. `make test` which uses `pylint` gives me runtime error even If I didn't change a python loc.

ThePonyCoder:
Your changes don't work on my machine. It doesn't preview images. Now I see raw text instead of preview

firmart:
@ThePonyCoder What's the mime-type of the images you test on? My change excludes those other than `image/png` and `image/jpeg`.

ThePonyCoder:
@firmart yea, I've tried jpeg, jpg, png and others. I don't know, maybe something wrong in nixos

toonn:
Restricting the mime types is not an option imo. Djvu *can* be used perfectly well to store images. That also happens to be its main use, albeit in the form of scanned documents, hence the mime type.

But there's tons of other image formats people want to preview, raw for example or webp, but even common ones like gif and tiff that you're excluding.

firmart:
You're right, excluding in fact is not necessary because the process dies naturally if it exceeds the limit. However, it will respawn if the file is still selected. 
The `ulimit` should fix #1929 anyway. I'll continue trying tomorrow with djvu containing images and without.

firmart:
@toonn  I ended up with a solution which lets me preview `djvu` files (for the first time!).  The first solution is abolished, but I messed up when I tried to squash them. Consequently, this doesn't fix #1929 anymore, but I guess it may be the expected behaviour of `identify` when it deals with 30 MB `png`.

glyh:
Hope this will be merged. On my computer ranger call so many identify instances that eat up all of my memory.

toonn:
Wow, almost 2 years : s This one definitely slipped through the cracks. I'll be testing it later today and merging.

@glyh, could you report on #1929 instead of here? In either case this PR won't help you unless the source of the problem is DjVu files.

elliotwesoff:
I'm getting this issue with djvu files. Is this going to be merged, or can I fix it locally somehow?

toonn:
I forgot about this and resigned myself to the same fix, basically, in #2743. Now I'm thinking maybe we need to allow fall-through to the general image case again because DjVu *is* a valid image format. Testing of both version of the solution is welcome though. I don't know why this one relies on `convert` rather than simply output a TIFF from `ddjvu` for example. Maybe some image preview methods aren't compatible with TIFFs?

PR #1966, opened 2020-05-16
User friendly defaults

eMPee584:
Current defaults could be made more user-friendly.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -131,7 +131,7 @@ set iterm2_font_width 8
 set iterm2_font_height 11
 
 # Use a unicode "..." character to mark cut-off filenames?
-set unicode_ellipsis false
+set unicode_ellipsis true
 
 # BIDI support - try to properly display file names in RTL languages (Hebrew, Arabic).
 # Requires the python-bidi pip package
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -170,7 +170,7 @@ set draw_progress_bar_in_status_bar true
 set draw_borders none
 
 # Display the directory name in tabs?
-set dirname_in_tabs false
+set dirname_in_tabs true
 
 # Enable the mouse support?
 set mouse_enabled true
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -534,7 +534,7 @@ map cm search_next order=mtime
 map ca search_next order=atime
 
 # Tabs
-map <C-n>     tab_new
+map <C-n>     tab_new .
 map <C-w>     tab_close
 map <TAB>     tab_move 1
 map <S-TAB>   tab_move -1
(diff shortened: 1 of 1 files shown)


ranger/config/rc.conf
```diff
@@ -167,7 +170,7 @@ set draw_progress_bar_in_status_bar true
 set draw_borders none
 
 # Display the directory name in tabs?
-set dirname_in_tabs false
+set dirname_in_tabs true
```
toonn, line 173: Don't feel very strongly about this but, again, you give up some info (path of the current directory) in narrow terminals.

ranger/config/rc.conf
```diff
@@ -531,7 +534,7 @@ map cm search_next order=mtime
 map ca search_next order=atime
 
 # Tabs
-map <C-n>     tab_new
+map <C-n>     tab_new .
```
toonn, line 537: This is a noop afaict?

ranger/config/rc.conf
```diff
@@ -36,6 +36,9 @@ set hidden_filter ^\.|\.(?:pyc|pyo|bak|swp)$|^lost\+found$|^__(py)?cache__$
 # Show hidden files? You can toggle this by typing 'zh'
 set show_hidden false
 
+# Display size and modification time for each object
+default_linemode sizemtime
```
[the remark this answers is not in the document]
eMPee584: No I usually don't. Within tmux, my ranger panes sometimes get such small dimensions but I routinely zoom them to full screen.
I'm somewhat sceptical an 80x25 term is what to optimize defaults for in 2020..

ranger/config/rc.conf
```diff
@@ -128,7 +131,7 @@ set iterm2_font_width 8
 set iterm2_font_height 11
 
 # Use a unicode "..." character to mark cut-off filenames?
-set unicode_ellipsis false
+set unicode_ellipsis true
```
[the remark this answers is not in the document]
eMPee584: How could I; there is always going to be someone using an even more arcane system stuck in the distant past of computing.
Optimize for maximum compatibility or good comprehensibility, that is the question.
I think the meaning of `…` as placeholder for elided text is much more obvious than `~` .. String width is the same.

ranger/config/rc.conf
eMPee584: For very narrow terminals and long paths, that is true. In most other cases, better usability through exposing at least a hint to each tab's location.

ranger/config/rc.conf
eMPee584: No; `tab_new` by default opens each tab in the home directory. Which I found quite annoying..

toonn: That's simply false as far as I can ascertain. I'm not making this up, I tested it back then and I tested it again now. `C-n` opens a new tab in whatever directory you're in in ranger.

ranger/config/rc.conf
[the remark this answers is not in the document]
toonn: That's a cool feature request.

ranger/config/rc.conf
eMPee584: Yes, you were absolutely right.. shame on me back then.. `ranger --clean` ftw 😅

PR #1975, opened 2020-05-25
add 'hide' alias for scout

hektr:
i've added hide alias for scout command. 
It does the same as `:filter`, but with inverted results of matching. 
`:filter` shows everything, that matches regex and `:hide` hides everything, that matches regex.


hektr:
Sorry, I don't know, why there are 2 commits, when I changed only 1 file :) I'm not good in git !

toonn:
Next time just PR from your feature branch.

Issue #1939, opened 2020-04-29
Implement moving to first and last tab


toonn:
#1932 is a good feature suggestion.

PR #1977, opened 2020-05-25
Switching between first/last tab

hektr:
Switching between first/last tab 

I tried to do a quest from here #1939 ! I don't know, whether I've done this correct or not. But at first glance it works.

I added keyword argument to tab_move function to make that function use offset as tab index(we don't need offset for going to the first/last tab, but we need index -- zero or -1).

So, we can now write at rc.conf something like:

`map gz tab_move -1 use_offset_as_index=yes`
and it will use -1 not as offset, but as index of tab. So, it switch to the last tab.

`ga` is used for switching to the first tab and `gz` for last. It's easy to remember, because `a` is the first letter of alphabet, and `z` is the last. And it's easy to type :)

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -25,7 +25,6 @@
 #     multipane: Midnight-commander like multipane view showing all tabs next
 #                to each other
 set viewmode miller
-#set viewmode multipane
 
 # How many columns are there, and what are their relative widths?
 set column_ratios 1,3,4
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -539,8 +539,8 @@ map <A-Right> tab_move 1
 map <A-Left>  tab_move -1
 map gt        tab_move 1
 map gT        tab_move -1
-map ga        tab_move 0 useOffsetAsTabIndex=yes
-map gz        tab_move -1 useOffsetAsTabIndex=yes
+map ga        tab_move 0 use_offset_as_index=yes
+map gz        tab_move -1 use_offset_as_index=yes
 map gn        tab_new
 map gc        tab_close
 map uq        tab_restore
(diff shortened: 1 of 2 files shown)


hektr:
Sory, first time it showed errors, because of keyword argument name(it was useOffsetAsTabIndex). Tests don't like varibales, writed on camelStyle.

mrgoodvin:
Thanks. This version is better, imho. Might it be worth making the parameter / argument name shorter? (maybe — 'is_index'?). Who would like to remember such a long word combination?

hektr:
> Thanks. This version is better, imho. Might it be worth making the parameter / argument name shorter? (maybe — 'is_index'?). Who would like to remember such a long word combination?

yes, you're right. It has one more disadvantage. I'll change it a liitle.

toonn:
Hmm, looking more closely at the code I'm not sure we need this. If you specify a quantifier this already happens, for example `1gt` moves to the first tab.

hektr:
> Hmm, looking more closely at the code I'm not sure we need this. If you specify a quantifier this already happens, for example `1gt` moves to the first tab.

heh, yes, you are right. I don't know all possibilities of ranger :) May be it's not needed if you can use quantifier with gt

mrgoodvin:
I did not tested. And yes. It is a pity that it is. In terms of performance, the code was faster.

toonn:
The quantifier approach does some signaling to make sure the tab is loaded properly which the `tab_move` function neglects otherwise. Something to keep in mind.

mrgoodvin:
> The quantifier approach does some signaling to make sure the tab is loaded properly which the `tab_move` function neglects otherwise. Something to keep in mind.

Thanks.

> If you specify a quantifier this already happens, for example `1gt` moves to the first tab.

This is not completely true. If the tab has been closed or moved, specifying the `1gt` quantifier will create a new tab.

toonn:
You're right, it moves to the tab with the index you provided or opens a new one there if none exists. And 1 really means "tab named 1" not "first tab on the list."

PR #1982, opened 2020-05-31
Fix line length computation when there are ANSI escape codes

ipkiss42:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: `Manjaro Linux 20.0.1`
- Terminal emulator and version: `xfce4-terminal 0.8.9.2 (Xfce 4.14)`
- Ranger version: `ranger-master v1.9.3-24-g1654128f`
- Python version: `3.8.2 (default, Apr  8 2020, 14:31:25) [GCC 9.3.0]`
- Locale: `en_US.UTF-8`

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION

This PR fixes the computation of the "displayable" line length in the preview, properly ignoring the ANSI escape codes, instead of using `len(line)`.

#### MOTIVATION AND CONTEXT

Without this fix, when `wrap_plaintext_previews` is set to `true` and syntax highlighting in the preview is enabled (e.g. after installing `highlight` of `pygmentize`), then the line wrapping is done before the full width of the pager, resulting in shorter previews than possible and in spurious empty lines, not present in the previewed file.

The root cause is that the algorithm to wrap lines splits them based on the pager width, but the line length is computed "naively" using `len(line)`. Invisible ANSI escape codes used for highlighting are counted, making the line seem longer than it really is. So the line is wrapped earlier that needed. For short enough lines this can introduce extra empty lines.

The fix is thus simply to calculate and use the display length of the line in the wrapping algorithm.

#### TESTING

I ran `make tests_py`, but there is no automated test relevant to this part of the code.

I did manual testing both with and without highlighter installed, and both with and without line wrapping. The bug disappeared... and I didn't notice new ones :).

### DIFF
--- a/ranger/gui/widgets/pager.py
+++ b/ranger/gui/widgets/pager.py
@@ -246,9 +246,10 @@ class Pager(Widget):  # pylint: disable=too-many-instance-attributes
         while True:
             try:
                 line = self._get_line(i).expandtabs(4)
+                line_len = ansi.char_len(line) if self.markup == 'ansi' else len(line)
                 for part in ((0,) if not
                              self.fm.settings.wrap_plaintext_previews else
-                             range(max(1, ((len(line) - 1) // self.wid) + 1))):
+                             range(max(1, ((line_len - 1) // self.wid) + 1))):
                     shift = part * self.wid
                     if self.markup == 'ansi':
                         line_bit = (ansi.char_slice(line, startx + shift,
--- a/ranger/gui/widgets/pager.py
+++ b/ranger/gui/widgets/pager.py
@@ -246,7 +246,7 @@ class Pager(Widget):  # pylint: disable=too-many-instance-attributes
         while True:
             try:
                 line = self._get_line(i).expandtabs(4)
-                line_len = ansi.char_len(line) if self.markup == 'ansi' else len(line)
+                line_len = ansi.char_len(line)
                 for part in ((0,) if not
                              self.fm.settings.wrap_plaintext_previews else
                              range(max(1, ((line_len - 1) // self.wid) + 1))):


ranger/gui/widgets/pager.py
```diff
@@ -246,9 +246,10 @@ def _generate_lines(self, starty, startx):
         while True:
             try:
                 line = self._get_line(i).expandtabs(4)
+                line_len = ansi.char_len(line) if self.markup == 'ansi' else len(line)
```
toonn, line 249: Doesn't `ansi.char_len` always give the right answer? I don't think we need the fallback to `len`?

toonn (changes requested):
Thanks for the fix!

ranger/gui/widgets/pager.py
ipkiss42: It probably gives the right answer, but I simply followed the pattern from a few lines below, where the char slice is taken :
https://github.com/ranger/ranger/blob/1654128f955b320354eea98a0bb6970a159ab16f/ranger/gui/widgets/pager.py#L253-L259


I assumed it was done for performance reasons, but I'm not sure.

Would you like me to always use `ansi.char_len()`?

ipkiss42: I took that as a yes :)
Done.

ipkiss42: @toonn Is there anything I can do to help get this PR merged?

toonn: Only waiting for me (or someone else on the team *gasp*) to get a round Tuit : /

PR #1992, opened 2020-06-13
Apply refactored function name to bind command on example script

jj-143:
#### ISSUE TYPE

- Breaking changes

#### CHECKLIST

#### DESCRIPTION

Function name was refactored from `ranger-cd` to `ranger_cd` in #1743, but it hadn't applied to the bind command at the bottom.

#### OTHER ISSUE
Maybe I should create an Issue for it? Anyway, in the comments,
```
# directory to the last visited one after ranger quits. Either put it into your
# .zshrc/.bashrc/etc or source this file from your shell configuration.
```
This implies sourcing it would also work in zsh and other shells with key binding feature out of the box, but it's not:
The function's compatible with zsh but I found out `bind` is bash's built-in command,  
`bindkey -s "\C-o" "ranger_cd\C-m"` works on zsh.

### DIFF
--- a/examples/shell_automatic_cd.sh
+++ b/examples/shell_automatic_cd.sh
@@ -19,5 +19,9 @@ ranger_cd() {
     rm -f -- "$temp_file"
 }
 
-# This binds Ctrl-O to ranger_cd:
-bind '"\C-o":"ranger_cd\C-m"'
+# This binds Ctrl-O to ranger_cd on bash or zsh:
+if [ -n "${ZSH_VERSION:+x}" ]; then
+    bindkey -s "\C-o" "ranger_cd\C-m"
+else
+    bind '"\C-o":"ranger_cd\C-m"'
+fi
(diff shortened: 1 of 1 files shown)


toonn:
Oh, good catch. Let's actually change back. `-` is a lot more common as a character in commands.
Feel free to refactor for zsh compatibility, I think fish's builtin corresponds to bash in this case.
Also update the ranger versions at the top of the file maybe.

jj-143:
Hmmm, I read #1743 and I thought this script was meant to be POSIX compatible? `-` are not specified in POSIX's "valid names" - so it's not guaranteed - whereas `_` is specified. I know the whole script cannot be POSIX compatible due to the keybinding command, but still, at least the function part should be remain that way?

Done with zsh support, but not sure about the ranger version; `--choosedir` was indeed introduced in `1.4.2`.

toonn:
Looks good, I just meant bumping the latest supported version, 1.7.x is pretty old.

jj-143:
> # Compatible with ranger 1.4.2 through 1.9.*
It's already 1.9.* on master branch :)

PR #1994, opened 2020-06-14
dirname_in_tabs setting can be changed from ranger

hektr:
dirname_in_tabs setting can be changed from ranger using `zt`

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -593,6 +593,7 @@ map zs    set sort_case_insensitive!
 map zu    set autoupdate_cumulative_size!
 map zv    set use_preview_script!
 map zf    console filter%space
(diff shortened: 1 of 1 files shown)


toonn:
This is a fine key binding if you do this often but I think most people will prefer either having it on or having it off so I don't see a good reason to provide this by default. Feel free to try to change my mind though.

hektr:
@toonn, yes, you are right. But this option is popular. A lot of people will try to enable it. It's more easy and fast to do using zt than by editing config file, especially for new users.

toonn:
Sure but it's not permanent. Key bindings to toggle settings are more for things you might often want to change on an open ranger instance.

hektr:
@toonn, thanks :) in this case, you are right

PR #1999, opened 2020-06-16
:touch create missing directories

### LINKED ISSUE #1998: Create a command to create dirs and files
This is a feature suggestion.

We have mkdir which creates directories and touch that creates files but we have no way to recursively create the directories if they don't exist.

For example doing :command a/b/c/d would create dirs a/b/c and then the file d inside it

toonn:
`:mkdir` creates directories recursively like `mkdir -p`.
`:touch` failed when parent directories didn't exist, for convenience
and consistency it now creates missing directories recursively.

Fixes #1998

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -928,10 +928,14 @@ class touch(Command):
     """
 
     def execute(self):
-        from os.path import join, expanduser, lexists
+        from os.path import join, expanduser, lexists, dirname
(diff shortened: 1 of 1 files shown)


vifon:
I'm not sure it's a good idea, at least not when it's called `:touch`. It's definitely no longer just a `touch(1)` replacement. I get the intent but it should be a separate command or -flag for `:touch`.

PR #2001, opened 2020-06-18
Gracefully fallback when file(1) is not available to identify mime-type

soraxas:
#### ISSUE TYPE

- Improvement
When I was running `ranger` in an embedded device (a NAS) with ssh, I was trying to edit something and it keeps refusing to open some files. After checking the log, I found the following:

```sh
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/core/actions.py", line 271, in execute_console
    cmd.execute()
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/config/commands.py", line 1144, in execute
    self.fm.execute_file([File(listpath)], app='editor')
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/core/actions.py", line 465, in execute_file
    return execute()
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/core/actions.py", line 463, in execute
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/ext/rifle.py", line 332, in execute
    for count, cmd, lbl, flgs in self.list_commands(files, mimetype):
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/ext/rifle.py", line 303, in list_commands
    if not self._eval_condition(test, files, None):
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/ext/rifle.py", line 187, in _eval_condition
    return self._eval_condition2(condition, files, label)
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/ext/rifle.py", line 217, in _eval_condition2
    return bool(re.search(argument, self.get_mimetype(files[0])))
  File "/volume1/main/home/soraxas/git-repo/ranger/ranger/ext/rifle.py", line 261, in get_mimetype
    process = Popen(["file", "--mime-type", "-Lb", fname], stdout=PIPE, stderr=PIPE)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

It turns out it's because of this line
```python
process = Popen(["file", "--mime-type", "-Lb", fname], stdout=PIPE, stderr=PIPE)
```
where my NAS does not has `file` installed (i.e. `which file` exit with error). I saw that the code had the try/catch block around the Popen `mimetype` line, but not with `file`. I guess it was assumed that `file` will always be available on a system, but I think that was not true on NAS device.

#### DESCRIPTION

After extending the try/catch block around the whole **Popen** and adding the fallback as plaintext (I think it's a very reasonable fallback and is the best that you can do without `file` and `mimetype`), I was able to edit files and use `:bulkrename` with vim without any issues.

#### CHECKLIST

### DIFF
--- a/ranger/ext/rifle.py
+++ b/ranger/ext/rifle.py
@@ -279,7 +279,10 @@ class Rifle(object):  # pylint: disable=too-many-instance-attributes
         elif function == 'path':
             return bool(re.search(argument, os.path.abspath(files[0])))
         elif function == 'mime':
-            return bool(re.search(argument, self.get_mimetype(files[0])))
+            mimetype = self.get_mimetype(files[0])
+            if mimetype is None:
+                return False
+            return bool(re.search(argument, mimetype))
         elif function == 'has':
             if argument.startswith("$"):
                 if argument[1:] in os.environ:
@@ -322,11 +325,15 @@ class Rifle(object):  # pylint: disable=too-many-instance-attributes
         self._mimetype, _ = mimetypes.guess_type(fname)
 
         if not self._mimetype:
-            with Popen23(
-                ["file", "--mime-type", "-Lb", fname], stdout=PIPE, stderr=PIPE
-            ) as process:
-                mimetype, _ = process.communicate()
-            self._mimetype = mimetype.decode(ENCODING).strip()
+            try:
+                with Popen23(
+                    ["file", "--mime-type", "-Lb", fname], stdout=PIPE, stderr=PIPE
+                ) as process:
+                    mimetype, _ = process.communicate()
+                self._mimetype = mimetype.decode(ENCODING).strip()
+            except OSError:
+                self._mimetype = None
+                self.hook_logger("file(1) is not available to determine mime-type")
             if self._mimetype == 'application/octet-stream':
                 try:
                     with Popen23(


toonn:
Hmm, good catch. I think we shouldn't fall back to `text/plain` though, opening images and videos with a text editor is no fun. We should log a warning about `file(1)` not being available, though I'm not sure rifle has access to ranger's logging : /
I think an empty string or maybe even None is a better fallback. Mime-based rules won't fire but there's still extension-based rules which could be meaningful. Maybe we should also add default editor and pager rules for the .txt extension lower down in rifle.conf.

soraxas:
Thanks @toonn for your message.

We cannot directly return `None`, because the regex here
https://github.com/ranger/ranger/blob/1654128f955b320354eea98a0bb6970a159ab16f/ranger/ext/rifle.py#L217
excepts string, or else it would result in
```python
TypeError: expected string or bytes-like object
```

Empty string could work, perhaps I can switch the fallback `"text/plain"` to empty string `""`? (I'm not familar with the code base, do you think would an empty mime string would breaks anything else?)

toonn:
I don't think so. Actually, I'm leaning toward None. And then change the code not to test the mimetype at all if it's None. That way we won't have a weird match if someone manages to match empty strings with their regex.

soraxas:
My apology for the delayed response; had forgotten about it. The rebased push should address the issue

toonn (approved):
LGTM

PR #2013, opened 2020-06-23
Implement passing of flags for :bulkrename

### LINKED ISSUE #2002: Bulkrename confirmations
#### Current Behavior
After confirming the changes and exiting vim for the second time, you are greeted with "Press ENTER to continue" message.

 

#### Expected Behavior
It would be nice if there were a way to stop this message from appearing at all. Also the "Review changes" message could be made opt-out as well.

toonn:
The flags a bulkrename command passed to the command runner were fixed
to `w`, which means "wait for an enter keypress." Now it's possible to
pass whichever flags you want. Passing any flag overrides the default so
if you still want the confirmation you need to explicitly pass `w`.

Fixes #2002

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1124,6 +1124,12 @@ class bulkrename(Command):
     After you close it, it will be executed.
     """
 
+    def __init__(self, *args, **kwargs):
+        super(bulkrename, self).__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


toonn:
@kohane27, that's still the default behavior. You need to pass other flags if you want different behavior. But we do tend to recommend vidir anyway so if the alternative works better for you then stick to it.

PR #2017, opened 2020-06-28
added w3mimgdisplay render location fix for tmux panes

Gnostiphage:
Added some slight awareness to the ImageDisplayer for w3m so that images would be rendered in the appropriate tmux pane.

#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch Linux (kernel 5.7.6-arch1-1)
- Terminal emulator and version: rxvt-unicode-pixbuf, 9.22-2, running tmux 3.1_b-1
- Python version: 3.8.3-1
- Ranger version/commit: 1.9.3/4bb83b3
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

The _generate_w3m_input() and clear() methods of the W3MImageDisplayer class would choose a point in the terminal to draw the image, which works unless you're using ranger within tmux and tmux has been split into panes, and ranger isn't running in the top-left pane. I've edited those methods to call tmux if the TMUX environment variable is set, to grab where the current pane starts, and add those values to start_x and start_y.

#### MOTIVATION AND CONTEXT

I believe many people use tmux, and many people like the image previews in ranger. Making them work better together increases usability.

This solves an image-rendering issue.

#### TESTING

"make test" has been run, which returned that two of my lines were too long (which appears to be relatively acceptable? let me know if that needs to change)

It doesn't affect other areas of the codebase.

#### IMAGES / VIDEOS

Before
After

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -198,9 +198,24 @@ class W3MImageDisplayer(ImageDisplayer, FileManagerAware):
 
         fontw, fonth = self._get_font_dimensions()
 
+        # change the offset for panes if tmux is running
+        tmux_offset_x = 0
+        tmux_offset_y = 0
+        if 'TMUX' in os.environ:
+            tmux_cmd = "tmux display -pt \"${TMUX_PANE:?}\" '#{pane_left} #{pane_top} #{status-position}'"
+            tmux_offset_proc = Popen(tmux_cmd, shell=True, stdout=PIPE)
+            tmux_offset = tmux_offset_proc.stdout.read().decode().split()
+            tmux_offset_x = int(tmux_offset[0])
+            tmux_offset_y = int(tmux_offset[1])
+            if tmux_offset[2] == "top":
+                tmux_offset_y += 1
+
+        start_x = int((start_x - 0.2) * fontw) + (tmux_offset_x * fontw)
+        start_y = (start_y * fonth) + (tmux_offset_y * fonth)
+
         cmd = "6;{x};{y};{w};{h}\n4;\n3;\n".format(
-            x=int((start_x - 0.2) * fontw),
-            y=start_y * fonth,
+            x=start_x,
+            y=start_y,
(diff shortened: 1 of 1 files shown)


toonn:
CI needs to pass for all PRs, so yeah you need to fix those. (Currently testing requires pylint<2 : /, working on this as we speak.)

This sounds like an AWESOME contribution! 🎉 🥇 💯 

Could you take a look at the other methods to see whether a similar fix is necessary/possible?

Gnostiphage:
I've already looked, and because tmux almost acts like its own little terminal, the only image-display method that works consistently is w3m regardless of the underlying terminal (the direct urxvt method sort of works, unless you have a transparency script running too), and in W3MImageDisplayer, the only methods that really need editing are clear and _generate_w3m_input (I considered draw, but draw calls _generate_w3m_input, and the latter was simpler). Sorry about the lack of pylint<2, I'm new to python development (just took a SANS course on it a month or so ago, it covered 2-to-3 and linting, but none of the methods to check those).

toonn:
Don't worry about the linting, I'll fix it when merging.

The iterm2 method kinda works in tmux, just not for large files. Could you take a look at that?

Gnostiphage:
I'll see what I can do, but because I don't have a mac I won't be able to test anything (plus I'm not sure if there'd be anything wonky related to the subprocess class or tmux on mac). At best it'll just be theoretical.

Gnostiphage:
So the iterm2 method of encoding image formats is completely different. I'm not sure if the problem you're looking at has anything to do with where an image gets rendered on the terminal (not that that seems like it can be fixed, it doesn't allow for starting coordinates like with w3m), but if you're saying it's mostly to do with image size, I'd imagine it's a limit on the length of the base64-encoded string, which would probably mean to successfully render you'd need to first create a reduced-size copy of whatever image. That's a bit beyond my abilities, sorry.

PR #2018, opened 2020-06-28
Typo in the command binding

Aster89:
The shell function is called `ranger_cd`, but the `bind` command has a typo, `ranger-cd`.


hut:
Thanks @Aster89 for the fix! It seems that somewhere along the way, this has already been fixed in https://github.com/ranger/ranger/commit/e5a06fa2180fd3ba2f86172be8fddd469425749b . Closing this PR.

Issue #2026, opened 2020-07-01
command touch to auto select the file


ri-aje:
#### Runtime Environment

- Operating system and version: 
- Terminal emulator and version: 
- Python version: 
- Ranger version/commit: 
ranger version: ranger-master v1.9.3-27-gd17e532a
Python version: 2.7.17 (default, Apr 15 2020, 17:20:14) [GCC 7.5.0]
Locale: None.Non
- Locale: 

#### Current Behavior
the file is made and left amid the rest of the files/folders in the current dir.

#### Expected Behavior
to select/highlight the newly made (or existing) file.

#### Context

I prefer touch to auto select the file after making it, currently I have to make a second step to move the cursor to the file, which is inconvenient.

#### Possible Solutions
it seems to be fairly easy to implement with `self.fm.select_file(fname)`, but my attempt [here](https://github.com/ri-aje/ranger/blob/d17e532a06e5353927ee58b52857052b2ec256f0/ranger/config/commands.py#L938) doesn't work. I am looking for some pointers. thanks.

if the file already exists, then `select_file` will highlight the file. if the file doesn't exist, touch will make it, but `select_file` won't highlight it. I am not sure why the second scenario doesn't work. same behavior with `mkdir`.

#### Steps to reproduce
1. `:touch test`
2. notice test is created, but is not highlighted. would like it be auto selected.
3. 

#### Traceback

```

```

toonn:
#1200 already tries to implement this but it's far harder than you think for hysterical raisins.

ri-aje:
thanks. I didn't find those in my history digging. looks like a known issue long time ago. given the fact that they were open since two years ago, is it the case that they are unlikely to be fixed? even for the most simple case where the file is created directly in the current dir? I can see it become more complicated when multiple subdirs are involved in the creation, but I can live without auto selecting those deep files.

toonn:
I intend to finish the work but it involves changing ranger internals, which involves being confident in the changes, which involves thoroughly testing the changes, which involves mocking up a ranger instance...

It's a bit of a rabbit hole. If you feel like working on this come join us in #ranger @freenode.net, we're a friendly bunch : )

ri-aje:
gotcha. I just got to know ranger, so I will start small, such as making improvements like what I was after in this issue. I might have some idea to improve `bulkrename`, I will share when I get it.

kevinhwang91:
@ri-aje If you are interested in this feature, you can try my patch.

[...]

PR #1200, opened 2018-06-09
:touch and :mkdir focus their argument

toonn:
:touch now behaves more like the unix command in case the file already
exists, the notification is now info-level instead of error.
Both :touch and :mkdir now focus the new/existing file/directory.

Fix #1028

#1036 was based on my master branch which was muy inconvenient.

PR #2038, opened 2020-07-08
After touch and mkdir, automatically select generated file or dir

kevinhwang91:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: ArchLinux
- Terminal emulator and version: Alacritty
- Python version:  3.8
- Locale: 

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Use directory.load_content(schedule=False) force update content.

#### MOTIVATION AND CONTEXT

Improve https://github.com/ranger/ranger/issues/2026 and https://github.com/ranger/ranger/issues/1028
directory.files_all will not update when creating new file or dir because
directory.load_content without parameter schedule=False is asynchronous

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -914,8 +914,10 @@ class mkdir(Command):
         dirname = join(self.fm.thisdir.path, expanduser(self.rest(1)))
         if not lexists(dirname):
             makedirs(dirname)
+            self.fm.get_directory(os.path.dirname(dirname)).load_content(schedule=False)
         else:
-            self.fm.notify("file/directory exists!", bad=True)
+            self.fm.notify("file/directory exists!")
+        self.fm.select_file(dirname)
 
     def tab(self, tabnum):
         return self._tab_directory_content()
@@ -933,8 +935,11 @@ class touch(Command):
         fname = join(self.fm.thisdir.path, expanduser(self.rest(1)))
         if not lexists(fname):
             open(fname, 'a').close()
+            self.fm.get_directory(os.path.dirname(fname)).load_content(schedule=False)
         else:
-            self.fm.notify("file/directory exists!", bad=True)
+            self.fm.notify("file/directory exists!")
+            os.utime(fname)
+        self.fm.select_file(fname)
 
     def tab(self, tabnum):
         return self._tab_directory_content()


kevinhwang91:
master branch crash at preview caused by sha512, there's log, so I only test under v1.9.3.

```log
ranger version: ranger-master
Python version: 3.8.3 (default, May 17 2020, 18:15:42) [GCC 10.1.0]
Locale: en_US.UTF-8
Current file: '/home/kevin/WorkSpaces/build/ranger/AUTHORS'

  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/core/main.py", line 203, in main
    fm.loop()
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/core/fm.py", line 376, in loop
    ui.redraw()
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/gui/ui.py", line 343, in redraw
    self.draw()
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/gui/ui.py", line 370, in draw
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/gui/displayable.py", line 257, in draw
    displayable.draw()
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/gui/widgets/view_miller.py", line 100, in draw
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/gui/displayable.py", line 257, in draw
    displayable.draw()
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/gui/widgets/browsercolumn.py", line 187, in draw
    self._draw_file()
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/gui/widgets/browsercolumn.py", line 206, in _draw_file
    path = self.target.get_preview_source(self.wid, self.hei)
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/container/file.py", line 96, in get_preview_source
    return self.fm.get_preview(self, width, height)
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/core/actions.py", line 1096, in get_preview
    cacheimg = os.path.join(ranger.args.cachedir, self.sha512_encode(path))
  File "/home/kevin/.local/lib/python3.8/site-packages/ranger/core/actions.py", line 1024, in sha512_encode
    sha = sha512(stat_.st_dev)
TypeError: object supporting the buffer API required

```

toonn:
Because of my work on #1200, I'm very skeptical this will work. Please go through everything in my earlier commits and painstakingly address all my concerns.

Yep, I broke master. Have a PR ready to fix it but not satisfied yet.

kevinhwang91:
I have seen https://github.com/ranger/ranger/pull/1200 before I intend to improve this feature, this patch works fine to me.

toonn:
Then I suspect you didn't test it thoroughly enough.

vejkse:
I tested this, and all cases in https://github.com/ranger/ranger/pull/1200#issuecomment-395973824 work, **as long as no `..`, `.` or symlinks are involved**.  In those cases, either `get_directory` doesn’t return the right `Directory` [**CORRECTION: actually, `get_directory` already does `abspath` — the problem is only:**] or `select_file` can’t go to to the touched/made item because the path is not the one recorded.

To make @kevinhwang91 solution work in all cases (as far as I can tell we always end up with the touched/made item selected), even when relative paths or symlinks are involved, it suffices to add `os.path.abspath` when joining `self.fm.thisdir.path` to the argument.

Here’s the plugin I used to test this: there are two changes wrt to the initial commit here:
- I added `abspath` as explained above.
- I resolved conflicts.

```python
import os.path


class mkdir(Command):
    """:mkdir <dirname>

    Creates a directory with the name <dirname>.
    """

    def execute(self):
        from os import makedirs

        # Canonicalize the path so that it is recognized by
        # select_file.
        dirname = os.path.abspath(join(self.fm.thisdir.path, expanduser(self.rest(1))))
        if not lexists(dirname):
            makedirs(dirname)
            self.fm.get_directory(os.path.dirname(dirname)).load_content(schedule=False)
        else:
            self.fm.notify("file/directory exists!")
        self.fm.select_file(dirname)

    def tab(self, tabnum):
        return self._tab_directory_content()


class touch(Command):
    """:touch <fname>

    Creates a file with the name <fname>.
    """

    def execute(self):
        from os.path import join, expanduser, lexists, dirname
        from os import makedirs

        # Canonicalize the path so that it is recognized by
        # select_file.
        fname = os.path.abspath(join(self.fm.thisdir.path, expanduser(self.rest(1))))
        dirname = dirname(fname)
        if not lexists(fname):
            if not lexists(dirname):
                makedirs(dirname)
            with open(fname, 'a', encoding="utf-8"):
                pass  # Just create the file
            self.fm.get_directory(dirname).load_content(schedule=False)
        else:
            self.fm.notify("file/directory exists!")
            os.utime(fname)
        self.fm.select_file(fname)

    def tab(self, tabnum):
        return self._tab_directory_content()
```

There remains (at least) one problem on the other hand: when touching/creating a directory in a new subdirectory of a new subdirectory … (any number of times), since the closest existing “ancestor” directory does not receive `load_content(schedule=False)`, when navigating back to it from where one arrived, we don’t end up on the new subdirectory but on the previously selected item or the first item of that ancestor.  To fix that, one could
- compute that closest existing “ancestor” directory and made it `load_content(schedule=False)`;
- recursively run `load_content(schedule=False)` on the ancestors, to avoid having to determine the closest existing one.

vejkse:
There could be a second remaining problem, with symlinks.  Let’s consider the case of
- two directories `/tmp/a` and `/tmp/b`
- a symlink `/tmp/a/b` to `/tmp/b`.

If one navigates inside `/tmp/a/b`, we get different behaviours:
1. `:shell touch ../c` creates a file `/tmp/c` (the symlink is resolved).
2. `:touch ../c` with the current version of the code creates a file `/tmp/c`.
3. `:touch ../c` with the original pull request above creates a file `/tmp/c` (the symlink is resolved) but we jump to `/tmp/a` (the symlink is not resolved).
4. `:touch ../c` with my proposed modification with `abspath` creates a file `/tmp/a/c` and we move to it inside `/tmp/a` (the symlink is not resolved).
5. `:touch ../c` with the same modification using `realpath` instead creates a file `/tmp/c` and we move to it (the symlink is resolved).

So using `realpath` would be better in the sense that it respects the behaviour of shell `touch` and of current `:touch`.

On the other hand, with `realpath`, if one now run `:touch d` inside `/tmp/a/b`, the file is created in `/tmp/b` and one is unexpectedly moved to it as seen in `/tmp/b`, not through the symlink in `/tmp/a/b` where one was the instant before.  Of course, it’s the same directory actually, but the view is different, so that the parent and sibling directories will be different when navigating inside ranger.

I prefer the behaviour with `abspath`, mainly because I don’t want to unexpectedly jump to the resolved place in the directory hierarchy when touching a file in a directory reached through symlinks.  And I actually find it more intuitive that `../c` is created in the parent directory as seen from ranger, even though it is not the correct parent directory with symlinks resolved.

vejkse:
For the first problem (when navigating back to the closest existing ancestor we don’t arrive on the right item), I think the problem is simply that the directory on which we need to run `load_content(schedule=False)` is not the parent directory but that closest already existing ancestor.

So it should be something like this awkward and maybe dangerous (in an ideal world, `makedirs` would return the path to that ancestor):
```python
import os.path

class mkdir(Command):
    """:mkdir <dirname>

    Creates a directory with the name <dirname>.
    """

    def execute(self):
        from os import makedirs

        # Canonicalize the path so that it is recognized by select_file.
        dirname = os.path.abspath(join(self.fm.thisdir.path, expanduser(self.rest(1))))
        if not lexists(dirname):
            last_existing_ancestor = os.path.dirname(dirname)
            while not lexists(last_existing_ancestor):  # risk of infinite loop?
                last_existing_ancestor = os.path.dirname(last_existing_ancestor)
            makedirs(dirname)
            self.fm.get_directory(last_existing_ancestor).load_content(schedule=False)
        else:
            self.fm.notify("file/directory exists!")
        self.fm.select_file(dirname)

    def tab(self, tabnum):
        return self._tab_directory_content()


class touch(Command):
    """:touch <fname>

    Creates a file with the name <fname>.
    """

    def execute(self):
        from os.path import join, expanduser, lexists, dirname
        from os import makedirs

        # Canonicalize the path so that it is recognized by select_file.
        fname = os.path.abspath(join(self.fm.thisdir.path, expanduser(self.rest(1))))
        dirname = dirname(fname)
        if not lexists(fname):
            last_existing_ancestor = dirname
            while not lexists(last_existing_ancestor):  # risk of infinite loop?
                last_existing_ancestor = os.path.dirname(last_existing_ancestor)
            if not lexists(dirname):
                makedirs(dirname)
            with open(fname, 'a', encoding="utf-8"):
                pass  # Just create the file
            self.fm.get_directory(last_existing_ancestor).load_content(schedule=False)
        else:
            self.fm.notify("file/directory exists!")
            os.utime(fname)
        self.fm.select_file(fname)

    def tab(self, tabnum):
        return self._tab_directory_content()
```

vejkse:
Actually, the obvious solution of just doing `self.fm.select_file(fname)` — which first `cd`s into the directory and then `move_to_obj`s to the file inside that directory — didn’t work for 3 reasons:
1. The `Directory` wasn’t up to date and the new path wasn’t available for `move_to_obj` to jump to it.  This is solved by @kevinhwang91’s use of `load_content`.
2. `move_to_obj` doesn’t work with paths with `..` in them.  This is fixed by using `abspath` or `realpath`.
3. `mkdir` and `open` normalizes relative paths as `realpath` does (resolving symlinks), while `cd` uses `normpath`, that behaves as `abspath` does on most systems (not resolving symlinks).  This is fixed by normalizing one way or the other before doing anything.

For symlinks, I looked at other commands behaviours:
- `cd` does not resolve symlinks (unless explicitly used with `-r`);
- `edit`, `relink` and `rename` resolve symlinks.

Issue #1734, opened 2019-10-20
Cannot move files with console command


catzybluphish:
#### Runtime Environment

- Operating system and version: Arch
- Terminal emulator and version: urxvt

#### Context

I'm trying to make my own trash binding where I have `map dD console shell mv %s /tmp` to move files into /tmp. I've added `console` because I wanted a confirmation (although not as good as `delete`), but now I can't delete files with spaces in their names because the input line at the bottom doesn't automatically add quotation marks. So I've changed it to `map dD console shell mv "%s" /tmp` but now it can't delete multiple selected files because the quotes make them seem like a single file. Is there a bind where I can move multiple selected files at once with spacees in names followed by a prompt (or Enter as with `console)?`

toonn:
Could you check whether latest master escapes properly? (Btw, master has trash support using rifle, maybe check that out after confirming the bug.)

catzybluphish:
> Could you check whether latest master escapes properly? (Btw, master has trash support using rifle, maybe check that out after confirming the bug.)

The latest master still has the issue, and I unfortunately don't know any python. It's not a bug; more so a feature request for ranger to automatically add quotation marks around multiple of each selected file in the bottom input bar when moving through `console shell`. Or allow a command that prompts the user before executing a shell command like how `delete` does.

toonn:
It's a bug. Filenames should be escaped. Does tab-completing `%s` escape them properly?

catzybluphish:
Tab-completion does escape properly.

chu4ng:
May be we should add option for the trash path in rc.conf, not hardcode it in riffle?

hut:
Sorry for the late reply.

I believe the correct key binding is:

`map dD console shell mv %%s /tmp`

You have 3 nested commands here, `map`, `console`, and `shell`, which get executed in that order. The macro substitution (from %s to the file names) needs to happen when `shell` is executed, not `map` or `console`.

Fortunately, the `map` command is smart enough to not do any macro substitution. (commands.py class `map_` has `resolve_macros = False`)

However, the `console` command is not. So you need to escape the macro for it by using `%%s` instead of `%s`.

This is an easy trap to fall into. I make the same mistake in some of my own key bindings. This leads to the question: Should we add `resolve_macros = False` for `:console` as well?

Anybody see any downsides? @toonn?

hut:
One obvious downside is that those who copied the config before this fix will have the following line in their bindings:

```
map @  console -p6 shell  %%s
[...]

PR #2040, opened 2020-07-08
fix macros quoting in console command

crazymug:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

Operating system and version: Ubuntu 18.04.2
Terminal emulator and version: GNOME Terminal 3.28.2
Python version:
3.8.2 (default, Mar 15 2020, 13:26:29) [GCC 7.4.0]
2.7.17 (default, Apr 15 2020, 17:20:14) [GCC 7.5.0]
Ranger version/commit: ranger-master v1.9.3-24-g1654128f
Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
When invoking command through 'console' command, macroses are not escaped with quotes. Thus, substitution of file names containing spaces or other should-be-escaped chars becomes wrong. This happens because invoked command class instance, which handles escaping control var state (escape_macros_for_shell) is not created by the time of substitution. To fix this behavior i've added escape_macros_for_shell variable to 'console' initialization. 'console' __init__ adds shell-like macros escaping if given command need it (e.g. there are three commands that set macros escaping: shell, delete, trash).

#### MOTIVATION AND CONTEXT
#1734

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -830,7 +830,14 @@ class console(Command):
     """:console <command>
 
     Open the console with the given command.
+    Add shell-like macros escaping if given command need it.
     """
+    def __init__(self, *args, **kwargs):
+        super(console, self).__init__(*args, **kwargs)
+
+        command_class = self.fm.commands.get_command(self.arg(1))
+        if command_class:
+            self.escape_macros_for_shell = command_class.escape_macros_for_shell
 
     def execute(self):
         position = None


toonn:
Look at #1930 for more info. This solution isn't good enough.

crazymug:
> Look at #1930 for more info. This solution isn't good enough.

Just tried to fix the bug with minimum amount of rewriting already working things.
Looked at #1930 - i'm not interfering with terminal configuration.
Will appreciate if you'll give more details on which solution will be better. Thx.

Do i correctly understand - good solution is to delegate somehow quoting to subprocess.Popen() and eliminate usage of ranger/ext/shell_escape.py?

toonn:
No, what I want to do is have console inherit the escaping behavior from the command it's running.

Issue #2052, opened 2020-07-24
Set default column_ratios to Golden Cut (screenshot included)


desmap:
not a bug but a very nice-to-have, just `set column_ratios 100,161,259` and look:

![image](https://user-images.githubusercontent.com/43666255/88366956-db7ffb00-cd8a-11ea-98fc-46a9651ea589.png)

PR #2053, opened 2020-07-24
Set column_rations to Golden Cut

desmap:
Check https://github.com/ranger/ranger/issues/2052 for full background and multiple screenshots and comparisons to now.

It's a about a much more welcoming UX using Golden Cut ratios instead the current setup.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- irrelevant

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

It's a small change but creates a big impact making ranger a more polished impression.

#### MOTIVATION AND CONTEXT

It's not a must but it's an easy change and low hanging fruits which give ranger a more mature appearance.

https://github.com/ranger/ranger/issues/2052

#### TESTING
no test are required

##### How it looks now

image

##### Golden Cut ratio to the column left

`set column_ratios 100,161,259`

image

##### Golden Cut ratio to all columns left
kind of a recursive Golden Cut:  `set column_ratios 100,161,420`

image

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -27,7 +27,7 @@
 set viewmode miller
 
 # How many columns are there, and what are their relative widths?
-set column_ratios 1,3,4
+set column_ratios 100,161,420
 
 # Which files should be hidden? (regular expression)
 set hidden_filter ^\.|\.(?:pyc|pyo|bak|swp)$|^lost\+found$|^__(py)?cache__$


toonn:
Could you show what this looks like on a file with and without a preview? And let's call this golden ratio.

desmap:
Yes, right, golden ratio, IDK why I wrote golden cut all the time. However, here you find it with a file open, hope you like it! :)

image

desmap:
Here the screenshot for a file without a preview which looks strange but is the exact same as with the default ratio of 1,3,4. Is this a bug or a special setting?

image

toonn:
I don't hate the idea but it only really works in wide terminals. For the default of 80 columns there's very little space left for filenames. Most terminals default to 80 columns so I think that's what we should set the default for. We could add a note in the comment about the golden ratio being nice for people who prefer wider terminals though.

A cool feature would be if you could specify limits, like "20<1.6<50," which would only use the ratio if the result is over 20 columns and under 50 columns, capping it at either end when the result would under/overflow.

desmap:
> A cool feature would be if you could specify limits, like "20<1.6<50,"

Kind of CSS' media queries? Yeah that would be cool!

> For the default of 80 columns there's very little space left for filenames

I agree and used `/` full of short dir name for my prio examples, so yeah. What about the less aggressive non-recursive golden ratio: `set column_ratios 100,161,259`

Then, it would look like this on 80 chars and I used now dirs with a bit longer dir names:

image

Is this better? 

Again the recursive golden ratio from before on 80 chars for comparison:

image

toonn:
Consensus seems to be against making this a new default. I still think a comment would be useful. Prompting people with wide terminals to use different ratios. Either than or using different linemodes that put more useful information in the center column.

eMPee584:
I actually like a setting of `1,6,5` (so a wide center column, usually on fullscreen terminal with f.e. 159x38 chars) but vote for close.. nice idea though 😉

ranger-miller-ratio_2025-10-05_12-17-34

PR #2068, opened 2020-08-09
Add escaped path and name yank

dumbPy:
Add ability to yank path `yP` and filename `yN` with <s>whitespace escaped</s> shell escaping.  
<s>using `re.sub` so that the escapable chars can be expanded if required. Currently escapes `<space>` char</s>  
escaped paths and names are `<space>` joined unlike non-escaped names and paths that are `\n` joined.
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 18.04
- Terminal emulator and version: Kitty v0.14.2
- Python version: 3.7
- Ranger version/commit: 1.9.3 (master branch)
- Locale: 

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

`yp` and `yn` yank the path and name of the selection to clipboard.  
In order to yank files and use them in shell, you need to manually escapes the whitespaces chars.  
`yP` and `yN` can now yank path and name and escape the whitespace

#### TESTING

Doesn't affect the existing bindings/working.  
`Make` passed all the tests  
installed with `pip install ./` (because anaconda) and tested the added functionality. works as expected


toonn:
Look for `shell_escape` in the codebase for less fragile escaping. Great PR idea!

dumbPy:
Added `shell_escaped_path` property to `FileSystemObject` (`shell_escaped_basename` already existed)  and using it directly. no more hacks

dumbPy:
I think it makes more sense to join shell escaped paths/names with `<space>` instead of `\n`, since pasting it in shell is the most obvious usecase. What do you think @toonn

toonn:
That's a good idea. Avoids needing to pipe pbpaste/xclip/xsel.

toonn:
I think we'll need an escaped version of dir too.

dumbPy:
`yank dir` for multi selection leads to multiple entries. I think `yank dir` normal and escaped should return just 1 entry instead of as many entry as selections. Or shall I leave it as is and just add `shell_escaped_dirname`?

toonn:
I'm not sure that's ok. Filtering for uniqueness would be fine but when flattened you could have different dirs in a single selection.

dumbPy:
I am keeping it as is them. Just added `yD` for yank shell escaped dir

dumbPy:
Will this be merged @toonn ?

toonn:
Ping again in about a month or so. I'm currently focusing on bug fixes.

dumbPy:
@toonn this doesn't interfere with any of the existing stuff and would be a good addition. the diff is small and self explanatory too.

toonn (commented):
I like the idea. However, I think it'd be better as a flag on yank. Something like `:yank -q name`. `-q` for quoted. What do you think?

Apologies for the long response time, I'm trying to get caught up to issues again.

PR #2083, opened 2020-08-22
Fix deletion confirmation default indication

toonn:
Noticed in #2082.


a1346054:
This PR can be closed, fixes were merged in  #2432

toonn:
Thanks, this kind of triage is pretty simple but it does help!

PR #2084, opened 2020-08-22
Fix preview toggle map in freeze_files documentation

toonn:
Simple documentation error someone reported in the IRC channel.


hut:
Fixed in https://github.com/ranger/ranger/pull/2432 already. Thanks!

PR #2109, opened 2020-09-23
POC for an extensible bulk command

nfraprado:
This is a proof-of-concept for a new `bulk` command that is basically a generic version of `bulkrename` that can be extended by the user, that is, it provides an interface for the user to add custom functions that query and edit any attribute of multiple files at once.

#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

The `bulkrename` command provides a convenient way to edit an attribute of multiple files at once, but it is also very limited since it can only be used for renaming the files.

The new `bulk` command requires that the user adds two functions: one that gets the attribute for the file, and another that generates the shell script command that makes the change of the file attribute to the new one set by the user.

So the user can extend this command by providing the functions that handle the specific attribute they want to change, and the `bulk` command handles the common stuff that was already done by `bulkrename`: open the editor with the attribute for each file in each line, get the changes made by the user on exit, create a shell script with the commands for each attribute changed and finally execute the script for the changes to take effect.

#### MOTIVATION AND CONTEXT

This brings more flexibility and use cases for the useful but limited `bulkrename`.

#### TESTING

As an example of usage, two functions were implemented that use the `bulk` command, `id3art` and `id3tit`. They use the eyeD3 program to edit the artist and title ID3 tags of mp3 files, respectively.

To illustrate the usage, the users can for example: in ranger, select multiple mp3 files that they want to change the artist ID3 tag, type `:bulk id3art`, the editor will open up with the artist of each music file in each line.
After editing and closing, a shell script opens up with the eyeD3 commands that change the artist tag of each one that was modified.
Finally, after reviewing this script and closing, the changes take place.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1212,6 +1212,75 @@ class bulkrename(Command):
             fm.notify("files have not been retagged")
 
 
+class bulk(Command):
+    """:bulk <attribute>
+
+    This command opens a list of with the attribute <attribute> for each of the
+    selected files in an external editor.
+    After you edit and save the file, it will generate a shell script
+    which changes the attributes in bulk according to the changes you did in the
+    file.
+
+    This shell script is opened in an editor for you to review.
+    After you close it, it will be executed.
+    """
+
+    bulk = {}
+
+    def execute(self):  # pylint: disable=too-many-locals,too-many-statements
+        import sys
+        import tempfile
+        from ranger.container.file import File
+        from ranger.ext.shell_escape import shell_escape as esc
+        py3 = sys.version_info[0] >= 3
+
+        # get bulk command argument
+        bkname = self.rest(1)
+
+        # Create and edit the file list
+        files = [f for f in self.fm.thistab.get_selection()]
+        attributes = [self.bulk[bkname].get_attribute(f) for f in files]
+        listfile = tempfile.NamedTemporaryFile(delete=False)
+        listpath = listfile.name
+
+        if py3:
+            listfile.write("\n".join(attributes).encode("utf-8"))
+        else:
+            listfile.write("\n".join(attributes))
+        listfile.close()
+        self.fm.execute_file([File(listpath)], app='editor')
+        listfile = open(listpath, 'r')
+        new_attributes = listfile.read().split("\n")
+        listfile.close()
+        os.unlink(listpath)
+        if all(a == b for a, b in zip(attributes, new_attributes)):
+            self.fm.notify("Nothing to be done!")
+            return
+
+        # Generate script
+        cmdfile = tempfile.NamedTemporaryFile()
+        script_lines = []
+        script_lines.append("# This file will be executed when you close the editor.\n")
+        script_lines.append("# Please double-check everything, clear the file to abort.\n")
+        script_lines.extend("%s\n" % self.bulk[bkname].get_change_attribute_cmd(file, old, new)
+                            for old, new, file in
+                            zip(attributes, new_attributes, files) if old != new)
+        script_content = "".join(script_lines)
+        if py3:
+            cmdfile.write(script_content.encode("utf-8"))
+        else:
+            cmdfile.write(script_content)
+        cmdfile.flush()
+
+        # Open the script and let the user review it
+        self.fm.execute_file([File(cmdfile.name)], app='editor')
+        cmdfile.seek(0)
+
+        # Do the attribute changing
+        self.fm.run(['/bin/sh', cmdfile.name], flags='w')
+        cmdfile.close()
+
+
 class relink(Command):
     """:relink <newpath>
 
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1225,7 +1225,32 @@ class bulk(Command):
     After you close it, it will be executed.
     """
 
-    bulk = {}
+
+    class id3art(object):
+
+        def get_attribute(self, file):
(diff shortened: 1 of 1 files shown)


ranger/config/commands.py
```diff
@@ -1212,6 +1212,100 @@ def execute(self):
...
+    After you edit and save the file, it will generate a shell script
+    which changes the attributes in bulk according to the changes you did in the
+    file.
+
+    This shell script is opened in an editor for you to review.
+    After you close it, it will be executed.
+    """
+
+
+    class id3art(object):
```
toonn, line 1229: Let's use full names everywhere.

ranger/config/commands.py
```diff
@@ -1212,6 +1212,100 @@ def execute(self):
...
+
+        def get_attribute(self, file):
+            import eyed3
+            return str(eyed3.load(file.relative_path).tag.artist)
+
+
+        def get_change_attribute_cmd(self, file, old, new):
+            from ranger.ext.shell_escape import shell_escape as esc
```
toonn, line 1237: Maybe import `shell_escape` a level up, and don't abbreviate it.

ranger/config/commands.py
```diff
@@ -1212,6 +1212,100 @@ def execute(self):
...
+        def get_attribute(self, file):
+            import eyed3
+            return str(eyed3.load(file.relative_path).tag.title)
+
+
+        def get_change_attribute_cmd(self, file, old, new):
+            from ranger.ext.shell_escape import shell_escape as esc
+            return "eyeD3 -t %s %s" % (esc(new), esc(file))
+
+    bulk = {'id3art': id3art(),
```
toonn, line 1251: Hmm, would be nice if this worked more like image preview methods, where you can register a method. That way this code wouldn't have to be copied and edited but people could you define a class and register.

ranger/config/commands.py
```diff
@@ -1212,6 +1212,100 @@ def execute(self):
...
+            'id3tit': id3tit(),
+            }
+
+    def execute(self):  # pylint: disable=too-many-locals,too-many-statements
+        import sys
+        import tempfile
+        from ranger.container.file import File
+        from ranger.ext.shell_escape import shell_escape as esc
+        py3 = sys.version_info[0] >= 3
```
toonn, line 1260: Please use the PY3 constant we have defined.

toonn (changes requested):
Looks promising. I'm secretly hoping we can replace bulkrename with this, aliasing it.

ranger/config/commands.py
nfraprado: I actually didn't intend to add these `id3art` and `id3tit` bulk commands to ranger. I think they are very particular to my use case, and probably wouldn't see much use. I only added them in this PR to illustrate the usage of the bulk command. In any case, I have changed their names.

ranger/config/commands.py
nfraprado: Indeed it is better. I too found using this dictionary cumbersome, but didn't know which better interface to use.

I now have added a new ext/bulk.py file containing the same interface used in the image preview methods but for these bulk commands. I was going to add it all inside the bulk(Command) class but decorator definition doesn't seem to work inside classes, or I'm doing something wrong. But it looked a little too chaotic anyway, so I moved everything to the new file (except for the commands).

I also made the `id3_artist` and `id3_title` bulk commands extend a base `BulkCommand` class. That way users can use the base class as reference for which functions need to be implemented and which are the parameters.

ranger/config/commands.py
nfraprado: Done. This is also done in `bulkrename`, which is where I copied it from. But I only fixed it in `bulk` and not in `bulkrename`, since you're expecting to replace `bulkrename` with `bulk`.

nfraprado:
Thank you for reviewing, @toonn .

Some thoughts and my TODO list for this:
* When I based this code in `bulkrename`, I removed the last bit where it retags the files, since it is only useful when files are being renamed, and not for every other attribute changes. If we're expecting this to replace bulkrename and be totally compatible, we should add it be back and have a property in each `BulkCommand` saying that it moves/renames files, so the retagging can occur for those cases.
* We still need to add tab completion for this command but should be pretty easy using the already available `tab` function.
* Until now I have referred to the `bulk(Command)` as the "bulk command" and to `id3_artist` and `id3_title` as "bulk commands", but this is very confusing both in the code and when writing about. Instead of "bulk commands" I thought of these alternatives: "bulk attributes" or "bulk subcommands". Suggestions?

toonn:
Hmm, you're right about the retagging. Maybe a general "post command hook?" That way you can define the actions that need to be taken after the command completes. Which could be a "Blah blah finished!" message by default but be overridden with retagging for example.

I think bulk subcommands is going to be the most straightforward terminology, Attributes works well in many cases but might not in all cases, like paths aren't really attributes of files.

nfraprado:
> Hmm, you're right about the retagging. Maybe a general "post command hook?" That way you can define the actions that need to be taken after the command completes. Which could be a "Blah blah finished!" message by default but be overridden with retagging for example.

Yes, that sounds even better. I'll add that as a third function in the `BulkCommand` class then, and also implement `bulkrename` using it to make sure it works ok.

> I think bulk subcommands is going to be the most straightforward terminology, Attributes works well in many cases but might not in all cases, like paths aren't really attributes of files.

Sounds good. I'll rename the code to use that terminology.

PR #2115, opened 2020-09-30
Documentation for RANGER_LOAD_DEFAULT_RC

toonn:
Far too often users simply set this variable, thinking it's a "Go fast!" switch. This leads to confused users with broken (incomplete) configurations.
    
I've reworded the note to caution readers.


hut:
Good idea! I added an even harsher warning on `--copy-config`.

PR #2130, opened 2020-10-20
Fix hash filter

toonn:
Missed python 3 specific encoding issues again.


hut:
I fixed a syntax warning, looks great otherwise! Thanks Toon!

Issue #2143, opened 2020-11-04
cd to env variable doesn't work


aresler:
Ranger 1.9.3

#### Steps to reproduce
1. Create an environment variable with path, e.g. `export dl="/home/user/Downloads"
2. In Ranger, try to `cd $dl`

#### Expected

Navigate to `/home/user/Downloads`

#### Actual

Nothing happens

aresler:
If this is not supported, please change to feature request.

toonn:
Not currently possible unless you go through `eval`.

toonn:
Here's how: `:eval import os; cd = self.fm.get_command('cd'); cd(':cd ' + os.getenv('dl')).execute()`
Not exactly a pretty sight, I admit. But you could conceivably define a custom command for this.
Usually though you'd just make a keybind and be done with it, [look at the series `g` keybinds for example](TargetSelectorNoTargetsInProject) or you bookmark the directory.

chu4ng:
I pushed pull request for this feature, but I use bookmarks for all of my often using dirs.
May be OP has all his dirs list in .shrc file and want use it for ranger. For this case may be need use script for compile bookmarks($HOME/.local/share/ranger/bookmarks) file from shell env variables.

aresler:
@chu4ng 

>May be OP has all his dirs list in .shrc file and want use it for ranger

Exactly, I find it quite useful to create env variable for frequently-used directories, e.g. $projects, $documents, e.t.c.

>For this case may be need use script for compile bookmarks($HOME/.local/share/ranger/bookmarks) file from shell env variables.

Interesting, I will try, thanks.

chu4ng:
@aresler 
Yesterday I tried this script for compile "g" bookmarks for sh and for ranger. It looks like a solution to the problem, may be?
https://github.com/LukeSmithxyz/voidrice/blob/master/.local/bin/shortcuts

aresler:
@chu4ng Thanks for suggestion, I haven't tried the script yet, but I don't think "g" shortcuts is the perfect solution because, AFAIK, you can only assign one letter per shortcut, and keyboard is limited. 

For now I workaround this by aliasing "ranger" to "r" in zsh, and using it instead of cd, like `r $dir` for quick navigation. This requires exiting ranger before doing this though.

chu4ng:
@aresler 
[...]

PR #2144, opened 2020-11-05
Add handler for "cd" class.

chu4ng:
#### ISSUE TYPE

- Improvement
#### RUNTIME ENVIRONMENT

- Operating system and version: Archlinux, 5.8.14-arch1-1
- Terminal emulator and version: st 0.8.4
- Python version: 3.8.6
- Ranger version/commit: 57127d5
- Locale: en-US

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

I added a new handler for the destination path for "cd" command, which support use a environment variables for this.

#### MOTIVATION AND CONTEXT

#2143

#### TESTING

I get "NoneType object is not subscriptable" message, but I have same with code of master branch.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -148,6 +148,9 @@ class cd(Command):
         if not destination:
             destination = '~'
 
+        if destination.startswith("$"):
+            destination = os.environ[destination[1:]]
+
         if destination == '-':
             self.fm.enter_bookmark('`')
         else:


toonn:
Don't like this PR as-is. It only works if you only pass a variable so you can't do `:cd $HOME/subdir` for example. And the special casing of `cd` is unfortunate.

We talked about it in the IRC channel and one suggestion I liked is implementing a new command `:expand` which would parse it's arguments and expand any shell variables it encounters, potentially by actually passing the arguments to a shell to be expanded because copying all that behavior is fraught with potential bugs.
Do you think you'd be interested in implementing this?

Issue #735, opened 2016-12-16
Import marks from sxiv


SammysHP:
If you start sxiv with option `-o` it will print all marked files (with <kbd>m</kbd>) to stdout. This output can be saved as a tag in ranger.

chu4ng:
May be ranger should mark these files as selected, not taged? It will be cool feature.

toonn:
That's how I interpret OP's request actually, the files are marked (aka selected) and you *can* tag them if you want.

chu4ng:
And if sxiv will mark one file, then ranger will set cursor on this file, I think. 

It looks very intersing, but I don't know how start.

toonn:
@chu4ng, you'll have to implement a custom command for this afaict. Maybe look at the custom commands on the wiki to see how  to work with command output. I think the fd integration may be a good starting point.

SammysHP:
IIRC there is already a special opener for sxiv that passes all images in the current directory. This could be extended with the -o option and the stdout then parsed in ranger.

PR #2151, opened 2020-11-12
Add "sxiv_select" command.

chu4ng:
I added command for open images in current directory by sxiv, and if you selected some pictures, ranger select these or move to one this.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Linux, 5.9.8
- Terminal emulator and version: st 0.8.4
- Python version: 3.8.6
- Ranger version/commit:  640f17c
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

New class in config/commands.py "sxiv_selected".
Inside "execute" func it start subprocess.Popen(sxiv) with few args.
#### MOTIVATION AND CONTEXT

#735 and I and I've been wanting this feature so-o long.

#### TESTING

Not bad.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1998,3 +1998,28 @@ class paste_ext(Command):
 
     def execute(self):
         return self.fm.paste(make_safe_path=paste_ext.make_safe_path)
+class sxiv_select(Command):
+    def execute(self):
+        import subprocess
+        null_sep = {'arg': '-0', 'split': '\0'}
+        nl_sep = {'arg': '', 'split': '\n'}
+        result_sep = nl_sep
+
+        images = [f for f in self.fm.thisdir.files if f.image]
+        # Create subprocess for sxiv and pipeout in whatever
+        sxiv_args = ['-o', '-t']
+        process = subprocess.Popen(['sxiv'] + sxiv_args + [i.relative_path for i in images], universal_newlines=True, stdout=subprocess.PIPE)
+        (pipe_out, _err) = process.communicate()
+        raw_out = pipe_out.split(result_sep['split'])
+        # Delete empty
+        marked_files = list(filter(None, raw_out))
+
+        if len(marked_files) > 1:
+            for node_f in images:
+                if node_f.relative_path in marked_files:
+                    node_f.mark_set(True)
+            self.fm.ui.redraw_window()
+        elif len(marked_files) == 1:
+            sc = scout("-ts " + marked_files[0])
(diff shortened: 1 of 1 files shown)


chu4ng:
if we run this command with selected files, we should give these as sxiv files-argument.
And if we have filename somethink like "-IUbgi856.png", sxiv give error for us.

chu4ng:
Oh, shit.

chu4ng:
Hm.. Can I bind command(sxiv_select) in riffle.conf for autoopen files by it one?
I don't see anything like it.

toonn:
Rifle's independent of ranger. I don't think there's a simple way of having rifle trigger ranger commands.

chu4ng:
May be we should move it class to plugins path, not in Commands?
Cool thing for me, but it extension and  it cannot be in the same place as yank or grep, I think.

chu4ng:
I try put in plugin_sxiv.py something like that:
```
old_execute = Actions.execute_file
def _sxiv_open_with(*files, **kw):
    if 'label' in kw and label == "sxiv_select": # Somehow define openwith-option how our sxiv_select.
        s = sxiv_select("")
        s.execute()
    else:
        Actions.execute_file = old_execute

Actions.execute_file = _sxiv_open_with
```
but it's not work, because ranger can send to riffle only {mode: 0}.
Any ideas?

chu4ng:
And TODO: add handler, if sxiv is not installed.

toonn:
I think you would need to do this without relying on rifle. Since rifle can't really communicate back to ranger.

chu4ng:
@toonn 
We can bind the function to any key, but how convenient is this for use? 
I think added function in open_with menu is good(not best) choice and I don't know how to do this functionality without rifle. Even in theory.

toonn:
Aren't there existing exceptions for sxiv functionality or is that in rifle?

chu4ng:
What you mean?
Now sxiv is working fine - from rifle menu with vanilla sxiv and/or with sxiv_select.

chu4ng:
And 4f347ad fix the issue with marked files.
If you select several files in the tab, it will be displayed on the tab, but this is only visually. To make them really stand out, you need to mark something extra.
This has now been fixed.

chu4ng (commented):
If some videos have not cached preview, it doesn't work. And I don't know what do with it, because it's logical. Example, if we have one million not cached videos.

PR #2157, opened 2020-11-18
Directory preview

AbdelfattahToulaoui:
Added the option 'use_preview_script_for_directories' which, when set to True, will use the preview script to generate a preview of the directory.
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version:  Arch Linux, 5.4.77-1-lts
- Terminal emulator and version: rxvt-unicode (urxvt) v9.22
- Python version: 3.8.6
- Ranger version/commit: daac0c0
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Moved "is_image_preview" method to FilesystemObject class.
Added "_script_preview" method to BrowserColumn class.
Added a sample preview code for mime type "inode/directory" to scope.sh.

#### MOTIVATION AND CONTEXT

I've had a few directories which I couldn't identify without entering them and looking at the first image, so I thought adding this option is an appropriate solution to that problem. I'm pretty sure other people will find different uses for it as well.

#### TESTING

Works well, though my implementation of the preview script is a bit slow (I used `convert` to generate the preview).


aftoul:
> Is there a mistake in the code? Shouldn't the check be "if FIMG is not empty"? Like this:

I believe that is the case. I just checked and you're right.

Looks like I fixed it in my installation but forgot to commit it to the repo. I will fix that shortly.

Issue #1895, opened 2020-04-02
Crash when setting `max_console_history_size` to `none` in rc.conf


Zaeph:
Hello there,

`max_console_history_size` is listed in the manual as accepting a `none` value to not cap the number of items remembered:
````
max_console_history_size [integer, none]
    How many console commands should be kept in history? "none" will disable the limit.
````

However, adding `set max_console_history_size none` to `rc.conf` crashes with the following trace-back:
````
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/ranger/core/main.py", line 171, in main
    fm.initialize()
  File "/usr/lib/python3.8/site-packages/ranger/core/fm.py", line 132, in initialize
    self.ui.initialize()
  File "/usr/lib/python3.8/site-packages/ranger/gui/ui.py", line 119, in initialize
    self.setup()
  File "/usr/lib/python3.8/site-packages/ranger/gui/ui.py", line 300, in setup
    self.console = Console(self.win)
  File "/usr/lib/python3.8/site-packages/ranger/gui/widgets/console.py", line 40, in __init__
    self.history = History(self.settings.max_console_history_size)
  File "/usr/lib/python3.8/site-packages/ranger/container/history.py", line 16, in __init__
    assert maxlen is not None, "maxlen cannot be None"
AssertionError: maxlen cannot be None
````

I've tried to set the variable to `none` from inside ranger with `:set max_console_history_size none`, and whilst it didn't lead to a crash, it didn't do anything.  So, unless the variable is only read on init, I don't think that `max_console_history_size` can accept `none` as its value.

I haven't had the chance to look into the code to see what might be causing it.  A temporary measure could be to remove the line from the manual, but I still think this a desirable feature.

HTH.

hut:
Fixed in https://github.com/ranger/ranger/pull/2972

PR #2166, opened 2020-11-25
Handling NoneType when work with history-size variables.

chu4ng:
# ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Linux
- Terminal emulator and version: st
- Python version: 3.8.6
- Ranger version/commit: ranger-master v1.9.3-144-g94be2268
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Handling NoneType when work with history-size variables.
Old code:
`assert maxlen is not None, "maxlen cannot be None"`
Using infinity variable for "none" value.
#### MOTIVATION AND CONTEXT

#1895 

#### TESTING

I have not "AssertionError: maxlen cannot be None" or other errors.

### DIFF
--- a/ranger/container/history.py
+++ b/ranger/container/history.py
@@ -13,7 +13,7 @@ class HistoryEmptyException(Exception):
 class History(object):
 
     def __init__(self, maxlen=None, unique=True):
-        assert maxlen is not None, "maxlen cannot be None"
+        maxlen = float('inf') if maxlen is None else maxlen
         if isinstance(maxlen, History):
             self.history = list(maxlen.history)
             self.index = maxlen.index


toonn (approved):
LGTM! Thanks for the fix!

toonn (changes requested):
Oops, was too quick to accept.

The setting isn't updated when set from within ranger, could you fix this too then the issue is completely resolved? Thanks!

chu4ng:
...
I can't set these settings from master.
With or without --clean.
No matter what I try set - 1000 or "none".

chu4ng:
test_container.py doesn't work now, because History got some changes. Specifically he got SettingsAware how parent-class.
Other tests are ok.

Makefile
```diff
@@ -117,7 +117,7 @@ test_pytest:
 	py.test tests
 	@echo
 
-test_py: test_pylint test_flake8 test_doctest test_pytest test_other
+test_py: test_pylint test_flake8 test_doctest test_other
```
toonn, line 120: Please explain why you dropped these tests from the CI config?

ranger/container/history.py
```diff
@@ -5,30 +5,32 @@
...
+class History(SettingsAware, object):
 
-    def __init__(self, maxlen=None, unique=True):
-        maxlen = float('inf') if maxlen is None else maxlen
-        if isinstance(maxlen, History):
-            self.history = list(maxlen.history)
-            self.index = maxlen.index
-            self.maxlen = maxlen.maxlen
-            self.unique = maxlen.unique
+    def __init__(self, obj=None, maxlen=None, unique=True):
```
toonn, line 17: I don't like adding new arguments, especially not anywhere but the back of the list. If any calls used positional arguments they might now be broken.

ranger/container/history.py
```diff
@@ -89,6 +91,11 @@ def rebase(self, other_history):
     def __len__(self):
         return len(self.history)
 
+    def _update_maxlen_(self, maxlen=None):
+        self.maxlen = maxlen if maxlen else self.settings.max_console_history_size
```
toonn, line 95: You can't default to a specific setting because the History class needs to work for several.

toonn (changes requested):
Let's drop the trailing underscore from `_update_maxlen`. And maybe we should just add an `@property` for maxlen rather than this extra method.

ranger/container/history.py
chu4ng: "AttributeError: 'History' object has no attribute 'settings'", because new History get settings dependence.
May be we can change test for this, but I don't know.

Makefile
chu4ng: "AttributeError: 'History' object has no attribute 'settings'", because new History get settings dependence.
May be we can change tests, but I don't know.

ranger/container/history.py
chu4ng: I checked history calls, it's ok.
But new arg more logical, is not it?

PR #2188, opened 2020-12-17
Fix #2187

### LINKED ISSUE #2187: source ranger does not work with arguments
#### Runtime Environment
- Operating system and version: Linux 5.4.80-2-MANJARO
- Python version: 3.8.6 (default, Sep 30 2020, 04:00:38) [GCC 10.2.0]
- Locale: en_US.UTF-8
- Shell: GNU bash, version 5.0.18(1)-release

#### Current Behavior
Running `source ranger path/to/begin/at` or `source ranger --help` or any other argument will result in `bash: path/to/begin/at: Is a directory` and `bash: --help: command not found` respectively.

#### Expected Behavior
According to [ranger.py](https://github.com/ranger/ranger/blob/521d9f89f5bc3572f54033d04f8ebac26ffa10c6/ranger.py#L10)'s comments the additional arguments should have been passed to ranger instead.

#### Possible Solutions
I was able to locate the issue. The embedded script seems to be expecting the second positional argument (`$1`) to be the file part of the `source` command (in this case, `ranger`).
However this is not the case (at least for me), as can be verified by the running `source test.sh Arg1 Arg2 Arg3 Arg4` with this simple `test.sh` script:
```sh
echo $1
echo $@
```
The result is 
> Arg1
> Arg1 Arg2 Arg3 Arg4

rather than
> test.sh
> test.sh Arg1 Arg2 Arg3 Arg4

Which perfectly explains the problem. On my former examples (`source ranger path/to/begin/at` and `source ranger --help`) this [line](https://github.com/ranger/ranger/blob/521d9f89f5bc3572f54033d04f8ebac26ffa10c6/ranger.py#L13) will set the variable to `path/to/begin/at` and `--help` respectively. Then the positional parameters will be shifted, clearing the `$@`. Removing these lines (and replacing the variable `$ranger` for its default value) solves the `source ranger path/to/begin/at` problem but not the `source ranger --help` problem. This is due to a second issue, the ` -- ` in [this line](https://github.com/ranger/ranger/blob/521d9f89f5bc3572f54033d04f8ebac26ffa10c6/ranger.py#L17). These usually marks the end of arguments and only paths afterwards and that seems to be the case here too. Removing them allows using the full command functionality.

### Note:
I'm preparing a Pull Request with these changes. However, I'm worried this issue seems odd, could the behavior of the bash source builtin have changed since the script was written? If so, older versions of bash wouldn't be compatible with this changes.

rafael-patronilo:
Fixes issue #2187
Removed apparently incorrect usage of the `$1` positional argument.
Removed the `--` on line 14 which blocked passing options to ranger.

#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
- Operating system and version: Linux 5.4.80-2-MANJARO
- Python version: 3.8.6 (default, Sep 30 2020, 04:00:38) [GCC 10.2.0]
- Locale: en_US.UTF-8
- Shell: GNU bash, version 5.0.18(1)-release

#### CHECKLIST

- [x] All new and existing tests pass **[REQUIRED]** (I assume, there are no tests on what I changed)
    - [ ] Tests have been updated

#### DESCRIPTION
- Removed apparently incorrect usage of the `$1` positional argument, as explained in issue #2187.
- Removed the `--` on line 14 which blocked passing options (`--help` and such) to ranger.
- Remove unnecessary substitution in line 14 (`${@:-$PWD}`). Ranger defaults to current directory on its own if no path is provided.

#### MOTIVATION AND CONTEXT
Fix issue #2187

#### TESTING
I tried running ranger with the following commands and all worked as expected (as opposed to before):
```sh
source ranger
source ranger path/to/begin/at
source ranger --help
source ranger --cmd="echo test"
source ranger --cmd="echo test" path/to/begin/at

alias ranger='source ranger'
ranger
ranger path/to/begin/at
ranger --help
ranger --cmd="echo test"
ranger --cmd="echo test" path/to/begin/at
```

### DIFF
--- a/ranger.py
+++ b/ranger.py
@@ -10,11 +10,8 @@
 # The other arguments are passed to ranger.
 """":
 temp_file="$(mktemp -t "ranger_cd.XXXXXXXXXX")"
-ranger="${1:-ranger}"
-if [ -n "$1" ]; then
-    shift
-fi
-"$ranger" --choosedir="$temp_file" -- "${@:-$PWD}"
+# command is used to prevent running into a recursive loop if the user aliased `ranger` to `source ranger`
+command ranger --choosedir="$temp_file" "$@"
 return_value="$?"
 if chosen_dir="$(cat -- "$temp_file")" && [ -n "$chosen_dir" ] && [ "$chosen_dir" != "$PWD" ]; then
     cd -- "$chosen_dir"


rafael-patronilo:
I misinterpreted the documentation: https://github.com/ranger/ranger/issues/2187#issuecomment-748009993

Issue #488, opened 2016-03-07
Archive content preview does not work with bigger archives


v1kn:
`ranger` struggles to preview files in bigger archives (in my case: `7z` and `tar.gz`; starting from ~10MB, or ~1000 files).  

Previewing smaller archives (~1MB or less) of both types work correctly. Also, using `atool` from the command line produces full preview instantenously.

nfnty:
Continue discussion in #151.

Issue #202, opened 2014-12-15
A delay before the file preview


vifon:
Scrolling through a list of videos etc. tends to be slow. Adding a small delay before the preview script is called would prevent ranger from generating previews for every file we scroll by. Would adding that feature be possible?

nfnty:
Related to #151.

nonamethanks:
I would like to add that this gets even worse if you are previewing a file currently in download. Since the file refreshes, ranger will spawn a ffmpegthumbnailer istance for every fraction of second that the terminal refreshes. I've seen all four cores of my CPU go to 100% in a couple of seconds with this, causing the whole system to freeze.

tobanw:
My whole system was freezing up when ranger tried to preview large .gz files. As a workaround, I modified `scope.sh` to only preview archives below a certain file size (10MB in this case):
```sh
case "$extension" in
    # Archive extensions:
    a|ace|alz|arc|arj|bz|bz2|cab|cpio|deb|gz|jar|lha|lz|lzh|lzma|lzo|\
    rpm|rz|t7z|tar|tbz|tbz2|tgz|tlz|txz|tZ|tzo|war|xpi|xz|Z|zip)
        if [ $size -le 10485760 ]; then
            try als "$path" && { dump | trim; exit 0; }
            try acat "$path" && { dump | trim; exit 3; }
            try bsdtar -lf "$path" && { dump | trim; exit 0; }
        fi
        exit 1;;
    rar)
        ...
```

Zekario:
@tobanw  I tried adding that to my `scope.sh`, but it seems to disable viewing any of those files, even if they're under 10MB. Would you happen to know why?

tobanw:
@Zekario is `$size` defined? I have it defined in this section as follows:
```sh
# Find out something about the file:
...
size=$(stat -c%s "$path") # file size in bytes
```

dahmc:
Is there any way to implement a small preview delay now after all these years?

toonn:
@dahmc, no. Checking out the linked PRs and bringing one of them up to speed with the current code would help the process along.

tbone2k-git:
Hi everyone! o) 

I got here, because I realized "ranger" is very slow when scrolling through files. It took me just some minutes of playing around with it to recognize the slowness and it indeed looks like the preview is called for every file, even if you only scroll "over" it. What the heck? o) 

I am testing all the file managers on Linux over time and I already wasted a lot of time so far.
[...]

PR #2197, opened 2020-12-28
fix(loader): wrong use of select with pipe

kevinhwang91:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: ArchLinux
- Terminal emulator and version: alacritty
- Python version: 3.9.1 
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

robjs returned by process.stdout is io.BufferedReader object.
io.BufferedReader.read() is block method. When previewing a large size file
such a foo.bz2, read() may make ranger hang a while until the pipe
data came.

https://github.com/ranger/ranger/issues/202#issuecomment-285779389, https://github.com/ranger/ranger/issues/488

#### MOTIVATION AND CONTEXT
Use os.read non-block method.

### DIFF
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -8,6 +8,7 @@ from subprocess import Popen, PIPE
 from time import time, sleep
 import math
 import os.path
+import os
 import select
 import errno
 
@@ -175,6 +176,7 @@ class CommandLoader(  # pylint: disable=too-many-instance-attributes
         popenargs['stdout'] = popenargs['stderr'] = PIPE
         popenargs['stdin'] = PIPE if self.input else open(os.devnull, 'r')
         self.process = process = Popen(self.args, **popenargs)
+        fd_out, fd_err = process.stdout.fileno(), process.stderr.fileno()
         self.signal_emit('before', process=process, loader=self)
         if self.input:
             if PY3:
@@ -197,9 +199,9 @@ class CommandLoader(  # pylint: disable=too-many-instance-attributes
         else:
             selectlist = []
             if self.read:
-                selectlist.append(process.stdout)
+                selectlist.append(fd_out)
             if not self.silent:
-                selectlist.append(process.stderr)
+                selectlist.append(fd_err)
             read_stdout = None
             while process.poll() is None:
                 yield
@@ -209,14 +211,14 @@ class CommandLoader(  # pylint: disable=too-many-instance-attributes
                     robjs, _, _ = select.select(selectlist, [], [], 0.03)
                     if robjs:
                         robjs = robjs[0]
-                        if robjs == process.stderr:
-                            read = robjs.readline()
+                        if robjs == fd_err:
+                            read = os.read(robjs, 4096)
                             if PY3:
                                 read = safe_decode(read)
                             if read:
                                 self.fm.notify(read, bad=True)
-                        elif robjs == process.stdout:
-                            read = robjs.read(512)
+                        elif robjs == fd_out:
+                            read = os.read(robjs, 4096)
                             if read:
                                 if read_stdout is None:
                                     read_stdout = read


toonn:
I'm not sure the blocking method actually matters since it's after a call to `select`? Please explain in more detail what was wrong and why non-blocking reads are warranted.

toonn:
I remain unconvinced we don't want to block. Maybe we do want to specify a buffer size of 4096 bits or so but I do not trust code relying on this can deal with fewer bytes read than expected.

kevinhwang91:
If you can't understand what I mean, let the other maintainers to checkout or close this PR.

I have expressed what I can express.

toonn:
I know what non-blocking IO is, you didn't address my concerns. The non-blocking IO might "succeed" in reading only a couple bytes. I have my doubts that the rest of the ranger code was written keeping in mind that the loader may return some bytes and at a later point return more bytes. Which might cause previews to be computed on very small reads and therefore become useless.

I am not saying non-blocking IO isn't desirable. Just that we need to be reasonably confident the rest of the code will be able to deal with it.

hut:
Hey @kevinhwang91, we appreciate your contributions, thank you. rnvimr is a cool project too :). But please consider that there are friendlier ways of communicating than "If you can't understand what I mean, let the other maintainers to checkout or close this PR." Maybe I misunderstand, or maybe this is a language issue, but it sounds a bit rude to me. 非常感谢你。

@toonn It's been too long since I've worked with that part of ranger, but I checked your concerns with a black-box approach by using the following scope.sh. It displays the expected preview with both master and kevin's branch. If I raise the sleep delays to 10, it still works. Is this what you mean?

```bash
#!/usr/bin/env bash
printf '%4096s\n' | tr ' ' '1'  # print "1" 4096 times, followed by a newline
printf '%4096s\n' | tr ' ' '2'  # print "2" 4096 times, followed by a newline
printf '%4096s\n' | tr ' ' '3'  # print "3" 4096 times, followed by a newline
sleep 1                       
printf '%4096s\n' | tr ' ' '4'
printf '%4096s\n' | tr ' ' '5'
printf '%4096s\n' | tr ' ' '6'
sleep 1                       
printf '%4096s\n' | tr ' ' '7'
printf '%4096s\n' | tr ' ' '8'
printf '%4096s\n' | tr ' ' '9'
exit 0                        
```

kevinhwang91:
> Maybe I misunderstand, or maybe this is a language issue, but it sounds a bit rude to me. 

You don't misunderstand me, I know my comment is rude because I'm disappointed and angry last night.
ranger is a good open-source project, I appreciate every contributor to ranger. I like the philosophy of ranger, but almost file managers on github only patchwork features and overlook details(maybe a performance issue).

Because of rnvimr, I dig into ranger, found some code in ranger is buggy, and maintain my fork all the time.
I have created several PRs to ranger, some of them almost solve the corresponding issues but in return, it is doubted and ignored. That's why I'm disappointed and angry.


Honestly, I don't get the points that @toonn concerned about, so I can't keep on communicating with @toonn about this PR.

@toonn I'm so sorry for my rudeness.

toonn:
@hut, the problem with blackbox testing is it catches only the happy path. I took @kevinhwang91, at his word that this patch works for them without issues, that's enough blackbox testing for me. What I'm concerned about is nothing turning up now and in a couple months having bizarre issues pop up without any indication of where they're coming from. Things partially being read is much harder to work with than a binary read/not read. I think the current loader code is something along the lines of "keep asking the generator for data until it returns anything then stop," and 4096 bits is few enough for your blackbox testing not to run into only having read a couple bytes.

@kevinhwang91, I only have so much time to devote to ranger so things move pretty slow, I know that's not pleasant as a contributor. As a maintainer there's too many bugs both existing and introduced with every new PR that I'm wary of merging significant changes. The current behavior may be slow but it's also correct or at least has never been found to be incorrect.

Do we still need the `select` call if we're using non-blocking IO anyway?

kevinhwang91:
>What I'm concerned about is nothing turning up now and in a couple months having bizarre issues pop up without any indication of where they're coming from.

Be brave to huge a change, the only man who makes no mistakes is the man who never does anything.

>Do we still need the select call if we're using non-blocking IO anyway?

`select` must work with non-blocking io reading and writing.

@toon @hut I am willing to refactor ranger to discard python2 and use asyncio, if you want ranger to be better.
If you think ranger is ok enough and don't want to change in the future, I am definitely not passionate about doing these things.

toonn:
Of course `select` still works. The question is whether there's any point to still polling the process if we're using non-blocking IO anyway.

Since there's methods of supporting asynchronous operation that *are* compatible with python 2 I'd rather these are explored first before deciding to drop python 2 support.

toonn (approved):
I'm still wary about the complexity this introduces but there was more testing in #3207 recently so I'll just accept it and cross my fingers that it doesn't lead to scary bugs.

@kevinhwang91, really appreciate your time and effort spent on this.

Issue #2200, opened 2020-12-30
ranger 1.9.3: AttributeError: 'Tags' object has no attribute 'tags'


fwy:
This issue is very similar to [https://github.com/ranger/ranger/issues/851](https://github.com/ranger/ranger/issues/851), but for a newer version of ranger.

#### Runtime Environment
- Operating system and version: msys2 (msys2-runtime 3.1.7-4) on Windows 10
- Terminal emulator and version: bash 5.1.004-1
- Python version: 3.8.6
- Ranger version/commit: 1.9.3
- Locale: en-US.UTF-8

#### Current Behavior
Ranger crashes immediately.

#### Expected Behavior
Ranger runs without crashing.

#### Context
Trying to run ranger on Windows from a bash shell.
I also cloned the GIT repo and ran ranger.py as suggested, but received almost the same error as shown below.

#### Steps to reproduce
1. Install using python: pip install ranger-fm
2. Run ranger at bash command line

#### Traceback
```
ranger version: ranger 1.9.3
Python version: 3.8.6 (default, Oct 23 2020, 14:59:35) [GCC 9.3.0]
Locale: en_US.UTF-8

Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/ranger/core/main.py", line 201, in main
    fm.loop()
  File "/usr/lib/python3.8/site-packages/ranger/core/fm.py", line 377, in loop
    ui.redraw()
  File "/usr/lib/python3.8/site-packages/ranger/gui/ui.py", line 333, in redraw
    self.draw()
  File "/usr/lib/python3.8/site-packages/ranger/gui/ui.py", line 360, in draw
    DisplayableContainer.draw(self)
  File "/usr/lib/python3.8/site-packages/ranger/gui/displayable.py", line 260, in draw
    displayable.draw()
  File "/usr/lib/python3.8/site-packages/ranger/gui/widgets/view_miller.py", line 100, in draw
    DisplayableContainer.draw(self)
  File "/usr/lib/python3.8/site-packages/ranger/gui/displayable.py", line 260, in draw
    displayable.draw()
  File "/usr/lib/python3.8/site-packages/ranger/gui/widgets/browsercolumn.py", line 189, in draw
    self._draw_directory()
  File "/usr/lib/python3.8/site-packages/ranger/gui/widgets/browsercolumn.py", line 295, in _draw_directory
    tagged = self.fm.tags and drawn.realpath in self.fm.tags
[...]

PR #2204, opened 2020-12-31
Tags errors

toonn:
Report errors if tag file does not exist

A non-existent tag file leads to an opaque crash without indication to
the user of what went wrong as in #2200. These errors are now reported
through ranger's notify mechanism.

The tag file is no longer created conditionally on `__init__`, only upon
writing saving tags. This changes the behavior somewhat in that an empty
"tagged" file should never be created.

 Only write tags if they might've changed

### DIFF
--- a/ranger/container/tags.py
+++ b/ranger/container/tags.py
@@ -13,15 +13,16 @@ from ranger import PY3
 ALLOWED_KEYS = string.ascii_letters + string.digits + string.punctuation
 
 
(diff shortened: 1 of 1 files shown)


vifon:
Looks reasonable. That's unfortunate we need to pull the whole `FileManagerAware` in just to access the logger. Something to consider refactoring someday.

vifon (approved):

PR #2216, opened 2021-01-06
Feat: add preview handler for SQLite to scope.sh

XuehaiPan:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: `Ubuntu 20.04 LTS` & `macOS Catalina`
- Terminal emulator and version: `Gnome-Terminal` / `iTerm 2`
- Python version: `Python 3.8` & `Python 2.7.18`
- Ranger version/commit: `master`
- Locale: `en_US.utf8`

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Add preview handler for `SQLite` to `scope.sh`.
Add setting variable `SQLITE_ROW_LIMIT` to `scope.sh`.

#### MOTIVATION AND CONTEXT

Preview `*.sqlite / *.db` files (mimetype `application/x-sqlite3`) (for example, `history.sqlite` for IPython console history).

#### TESTING

Preview handler:
- [X] file path with special characters (spaces and brackets, etc).
- [X] `sqlite3` is not installed
- [X] only `sqlite3` is installed
- [X] both `sqlite3` and `sqlite-utils` are installed

Setting variable:
- [X] set `SQLITE_ROW_LIMIT=5`.
- [X] set `SQLITE_ROW_LIMIT=0` or negative values.
- [X] set `SQLITE_ROW_LIMIT` larger than the row count.
- [X] set `SQLITE_ROW_LIMIT` to float number.
- [X] set `SQLITE_ROW_LIMIT` to a string but not number.

#### IMAGES / VIDEOS

Preview result (only `sqlite3` is required) (cannot handle newlines in cells)
Screenshot 1

Prettier preview result (with `sqlite-utils` installed)
Screenshot 2


toonn:
Great idea! Will review tomorrow.

doc/ranger.pod
```diff
@@ -286,6 +286,10 @@ C<python> or C<jq> for JSON files
 
 =item -
 
+C<sqlite3> and/or C<sqlite-utils> for listing tables in SQLite database
```
toonn, line 289: Same remark as for the README.

ranger/data/scope.sh
```diff
@@ -46,6 +46,8 @@ HIGHLIGHT_OPTIONS="--replace-tabs=${HIGHLIGHT_TABWIDTH} --style=${HIGHLIGHT_STYL
 PYGMENTIZE_STYLE=${PYGMENTIZE_STYLE:-autumn}
 OPENSCAD_IMGSIZE=${RNGR_OPENSCAD_IMGSIZE:-1000,1000}
 OPENSCAD_COLORSCHEME=${RNGR_OPENSCAD_COLORSCHEME:-Tomorrow Night}
+SQLITE_TABLE_LIMIT=20  # Display only the top <limit> tables in database, set to 0 for no exhaustive preview (only the sqlite_master table is displayed).
```
toonn, line 49: I think we usually document such ranger specific variables in the man page, if there are non-specific variables that's even better.

ranger/data/scope.sh
```diff
@@ -290,6 +292,45 @@ handle_mime() {
             xls2csv -- "${FILE_PATH}" && exit 5
             exit 1;;
 
+        ## SQLite
+        *sqlite3)
+            ## Preview as text conversion
+            local tables=""
```
toonn, line 298: POSIX sh doesn't support local variables. Feel free to make up a prefix that's clearly unique to the sqlite previews.

ranger/data/scope.sh
```diff
@@ -290,6 +292,45 @@ handle_mime() {
...
+            [ "$SQLITE_TABLE_LIMIT" -ge 0 ] || SQLITE_TABLE_LIMIT=20  # fallback to 20 if an invalid value is given
```
toonn, line 299: I prefer letting the tool report erroneous arguments/configuration otherwise users end up confused why their setting doesn't seem to be respected.

ranger/data/scope.sh
```diff
@@ -290,6 +292,45 @@ handle_mime() {
...
+            [ "$SQLITE_ROW_LIMIT" -ge 0 ] || SQLITE_ROW_LIMIT=5       # fallback to 5 if an invalid value is given
+            if tables="$( sqlite3 "file:${FILE_PATH}?mode=ro" \
+                          "SELECT name FROM sqlite_master \
+                           WHERE type='table' AND name NOT LIKE 'sqlite_%' LIMIT ${SQLITE_TABLE_LIMIT};" )"; then
+                if [[ -n "$(sqlite3 "file:${FILE_PATH}?mode=ro" '.tables')" ]]; then
```
toonn, line 304: Don't use `[[ ]]` that's a bashism.
And avoid nesting assignment in if statement conditions for clarity.

ranger/data/scope.sh
```diff
@@ -290,6 +292,45 @@ handle_mime() {
...
+                    if [[ -x "$( command -v sqlite-utils )" ]]; then
+                        sqlite-utils query "${FILE_PATH}" "${table_info_query}" --table --fmt fancy_grid
+                        preview_command="sqlite-utils query '${FILE_PATH}' '${query}' --table --fmt fancy_grid"
+                    else
+                        sqlite3 "file:${FILE_PATH}?mode=ro" "${table_info_query}" --header --column
+                    fi
+                    if [[ "$SQLITE_TABLE_LIMIT" -gt 0 ]]; then
+                        # Do exhaustive preview
+                        echo; printf '>%.0s' $(seq "${PV_WIDTH}"); echo
+                        echo "$tables" | xargs -I {} bash -c "echo; echo '{}:'; ${preview_command}"
```
toonn, line 325: Don't shell out to bash. If you use xargs you should report it as a requirement for sqlite previews in all the docs.

toonn (changes requested):
We try to keep `scope.sh` as POSIX compliant as possible so there's a couple remarks.

README.md
```diff
@@ -104,6 +104,7 @@ For enhanced file previews (with `scope.sh`):
 * `mediainfo` or `exiftool` for viewing information about media files
 * `odt2txt` for OpenDocument text files (`odt`, `ods`, `odp` and `sxw`)
 * `python` or `jq` for JSON files
+* `sqlite3` and/or `sqlite-utils` for listing tables in SQLite database
```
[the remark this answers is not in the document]
XuehaiPan: If `sqlite3` is not installed, it will raise command not found error. The sript will not go into the `if` block, just end with `exit 1`.
`sqlite-utils` is optional for fancier box drawing.

doc/ranger.1
```diff
@@ -334,6 +334,8 @@ are automatically used when available but completely optional.
 .IP "\-" 2
 \&\f(CW\*(C`python\*(C'\fR or \f(CW\*(C`jq\*(C'\fR for \s-1JSON\s0 files
 .IP "\-" 2
+\&\f(CW\*(C`sqlite3\*(C'\fR and/or \f(CW\*(C`sqlite-utils\*(C'\fR for listing tables in SQLite database
```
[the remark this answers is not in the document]
XuehaiPan: I modified `doc/ranger.pod` and regenerated this.

ranger/data/scope.sh
XuehaiPan: Some db files are really big, I add these variable for performance reasons, such as `HIGHLIGHT_SIZE_MAX` above.

ranger/data/scope.sh
XuehaiPan: Whether I should show report erroneous to `stdin` and `exit 5`, or simply `exit 1` (no preview).

ranger/data/scope.sh
XuehaiPan: I think `xargs` is always shipped with `macOS` and major `Linux` distributions.
I can refactor it without `xargs` using `for loops`.

XuehaiPan:
I have pushed the requested changes. But for the variables `SQLITE_{TABLE,ROW}_LIMIT`, I haven't figured out how to handle them better.

XuehaiPan:
Final version
- [X] no warning generated by `shellcheck --shell=sh`
- [X] tested for dependency.
- [X] tested for variabels `SQLITE_{TABLE,ROW}_LIMIT`.
- [X] tested for empty db file.

Preview results:

- No preview if `sqlite3` is not installed (script ends with `exit 1`).

- Empty database (only `sqlite3` is required) (script ends with `exit 5`):
Screenshot 1

- Preview (only `sqlite3` is required) (script ends with `exit 5`):
Screenshot 2

  **If `SQLITE_TABLE_LIMIT=0`, only the first table in the preview panel above is shown, (the dividing line `'>' * PV_WIDTH`is not displayed).**

- Fancier preview (both `sqlite3` & `sqlite-utils` are required) (script ends with `exit 5`):
Screenshot 3

README.md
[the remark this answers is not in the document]
toonn: Then this should be updated to say so rather than "and/*or*" which suggests just `sqlite-utils` is fine.

doc/ranger.1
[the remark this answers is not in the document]
toonn: And the embedded date didn't change?

ranger/data/scope.sh
toonn: That's ok, but do check whether sqlite3 already has a way of configuring this. The `bat` utility is one example where we had custom variables even though it already respects some.
And either way they should be *documented* in the man page.

ranger/data/scope.sh
toonn: The `&&` condition will fail so everything'll probably fall through to the `exit 1` at the end. This is not something that requires special handling iirc.

ranger/data/scope.sh
toonn: Does look like xargs is part of POSIX so I guess it's fine.

toonn (changes requested):
Couple more things.

ranger/data/scope.sh
```diff
@@ -290,6 +292,52 @@ handle_mime() {
             xls2csv -- "${FILE_PATH}" && exit 5
             exit 1;;
 
+        ## SQLite
+        *sqlite3)
+            ## Preview as text conversion
+            [ -x "$( command -v sqlite3 )" ] || exit 1
```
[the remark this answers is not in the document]
XuehaiPan: `[ -n "$(sqlite3 ...)" ]` is `false` when `sqlite3` is not installed (subcommand gets non-zero return code).
Also, the result of the condition gets `false` when the db file is empty (returns empty string).
I put the command check to `else` branch.

ranger/data/scope.sh
```diff
@@ -290,6 +292,52 @@ handle_mime() {
...
+            if [ -n "$( sqlite3 "file:${FILE_PATH}?mode=ro" '.tables' )" ]; then
+                sqlite_show_query() {
+                    if [ -x "$( command -v sqlite-utils )" ]; then
```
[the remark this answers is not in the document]
XuehaiPan: Refactored.

doc/rifle.1
```diff
@@ -1,4 +1,4 @@
-.\" Automatically generated by Pod::Man 4.11 (Pod::Simple 3.35)
```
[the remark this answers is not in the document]
XuehaiPan: `make man` says "`make: 'doc/rifle.1' is up to date.`" (I haven't modify `rifle.pod`).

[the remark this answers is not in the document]
XuehaiPan: Oh, my editer trimmed tailing space in `rifile.pod` at line 75.

ranger/data/scope.sh
```diff
@@ -290,6 +292,48 @@ handle_mime() {
...
+                        while read -r sqlite_table; do
+                            sqlite_table_rowcount="$(sqlite3 "file:${FILE_PATH}?mode=ro" -noheader "SELECT count(*) FROM ${sqlite_table}")"
+                            echo; echo "${sqlite_table}[${sqlite_table_rowcount}]:"
+                            sqlite_table_query="SELECT * FROM ${sqlite_table};"
+                            [ "$SQLITE_ROW_LIMIT" -gt 0 ] && sqlite_table_query="SELECT * FROM ${sqlite_table} LIMIT ${SQLITE_ROW_LIMIT} OFFSET (${sqlite_table_rowcount} - ${SQLITE_ROW_LIMIT});"
+                            sqlite_show_query "${sqlite_table_query}"
+                        done
+                fi
+                exit 5
+            elif [ -x "$( command -v sqlite3 )" ]; then
```
toonn, line 332: The shell doesn't report a different return code when the command doesn't exist compared to what sqlite returns when the db is empty?

toonn (changes requested):
Please address every review comment I've made. It's hard to keep track of what you addressed, what you believe you addressed and what you did not address.

ranger/data/scope.sh
XuehaiPan: `if [ -n "$( sqlite3 "file:${FILE_PATH}?mode=ro" '.tables' )" ]; then`
- `sqlite3` not installed: subcommand gets non-zero return code, condition fails.
- db is empty: subcommand return 0 with empty string, `[ -n "" ]`, condition fails.

XuehaiPan: I replace `[ -x "$( command -v sqlite3 )" ]` with `sqlite3 "file:${FILE_PATH}?mode=ro" '.quit'`. No `[ -x $cmd ]` command check now.

ranger/data/scope.sh
XuehaiPan: Refactored.

ranger/data/scope.sh
XuehaiPan: Refactored.

doc/ranger.pod
XuehaiPan: Updated.

ranger/data/scope.sh
XuehaiPan: check command using `sqlite3 -version 1>/dev/null`

XuehaiPan (commented):
All changes requested have been made.

XuehaiPan:
- [X] Documentation has been updated and followed the code style. (`README.md` & `doc/ranger.1` & `doc/ranger.pod`)
- [X] Refactored preview handler and all requested changes have been made. (`ranger/data/scope.sh`)
  - [X] completely POSIX compatible (no warnings generated by `shellcheck --shell=sh`).
  - [X] remove all command checks, use `||` mechanism.
  - [X] minimize dependencies, only `sqlite3` is resquired (optional `sqlite-utils` for fancy box drawing), remove dependence of `bash` and `xargs`.

toonn:
I'll get around to this eventually. It's a cool addition but not as important as bug fixes and there's a pretty long backlog of PRs that need merging. Thanks for the contribution!

toonn:
@XuehaiPan, a nice addition would be some sort of indicator, an ellipsis `...` for example to show whether there are more rows in a table than shown. And possibly an indication of more tables not being shown, vertical ellipsis maybe?

This does work fairly well, `sqlite-utils` seems to be unusably slow for me though. Did it work well in your testing?

XuehaiPan:
> a nice addition would be some sort of indicator, an ellipsis `...` for example to show whether there are more rows in a table than shown. And possibly an indication of more tables not being shown, vertical ellipsis maybe?

Good idea. But I haven't found a way to achieve this. The preview texts are generated by SQL query. I don't think we can add ellipsis to the query result.

> This does work fairly well, `sqlite-utils` seems to be unusably slow for me though. Did it work well in your testing?

It works fine for me. Sometimes, the sqlite file is actively modified by other processes. That may be a potential issue both for performance and data integrity. We do not add a "lock" on this file.

toonn:
@XuehaiPan, would it be feasible to do a query to the summary table for each table? We can compare the row count to the limit and print the ellipsis if it's greater than the later. And do a row count on that table in case there's more tables than we display. If the answer is no I guess we can merge this as-is.

XuehaiPan:
> would it be feasible to do a query to the summary table for each table? We can compare the row count to the limit and print the ellipsis if it's greater than the later. And do a row count on that table in case there's more tables than we display. If the answer is no I guess we can merge this as-is.

@toonn We already print the total row count on the right of the table name. As for the ellipsis, I haven't find a way to nicely include it in the query result.

XuehaiPan:
I add the `<limit>` info in the rendered output:

image

`if 0 < ${limit} < ${rowcount}`:

```
tablename [${limit} of ${rowcount}]:
+--------------+
| header       |
+--------------+
| record n - 2 |
+--------------+
| record n - 1 |
+--------------+
| record n     |
+--------------+
```

Otherwise: `${limit} >= ${rowcount}` or `${limit} == 0` (show all)

```text
tablename [${rowcount}]:
+--------------+
| header       |
+--------------+
| record 1     |
+--------------+
| record 2     |
+--------------+
| record n     |
+--------------+
```

toonn:
I actually like the ellipsis still. Doesn't really matter that we show the last couple rows IMO. It's just an indicator that there's more not necessarily that there are rows beneath the ones shown. The rowcount in parens is also a good addition.

XuehaiPan:
> I actually like the ellipsis still.

I found a way to achieve this. Now the handler shows `first record + ... + last (<limit> - 1) records`.

image

toonn (commented):
I think I found a small bug:

```
small1 [3]:
col1
----------
1
2
3

small2 [5]:
col1
----------
1
2
3
4
5

small3 [5 of 6]:
col1
----------
1
...
1
2
3
4
```

While `small3` actually contains:

```
sqlite> SELECT * FROM small3;
1
2
3
4
5
6
```

Seems to be some missing parens on the OFFSET. I'll push a change if you agree with the change we can finally merge : )

XuehaiPan:
Thanks for the fix. No other concerns here.

toonn:
@XuehaiPan, darn you're quick to respond. I wanted to get some reformatting in as well before you took a look. We try to stick close to a width of 80 for `scope.sh`. I mostly wanted to have you sign off on the changes in case I inadvertently broke some of the queries or something.

toonn:
I kinda prefer dropping all the commits after c7db94d. It just doesn't seem to be adding much of anything. If you feel like it does, write down your thoughts please.

XuehaiPan:
> If you feel like it does, write down your thoughts please.

@toonn I do the reformat job because there are some commas placed at beginning of the line. Although they are semantically equivalent. Another one is to format the continuous indentation with 4 spaces. It's more consistent with other parts of the script.

> I kinda prefer dropping all the commits after https://github.com/ranger/ranger/commit/c7db94d5401b30f39f1962cbb48b923212b3224d. It just doesn't seem to be adding much of anything.

I squashed the commits into one. It does not add more lines overflow 80 columns. If you think it looks bad, you can revert it and go ahead with yours.

toonn (approved):
Looks better to me. Thank you for all the work on this! ❤️

Issue #723, opened 2016-11-22
Implement image previews using Sixel


yossizahn:
All the methods used to preview images so far are specific to one terminal (iterm, rxvt), don't work over ssh (AFAIK all of them), or use very inelegant hacks (w3m creates an overlaying window, rxvt sets the terminal background). Sixel (https://en.wikipedia.org/wiki/Sixel) is a generic solution that works with multiple terminals (xterm, mintty and more - see full list at https://github.com/saitoha/libsixel) and over ssh, what could be better?

tex:
Excerpt from Readme at https://github.com/saitoha/libsixel:
@uobikiemukot's sdump project selected another approach. He wrote a w3mimgdisplay compatible program yaimg-sixel. It also works with ranger.

yossizahn:
@tex: I have tried yaimg-sixel and it's buggy not clearing up the display when moving to next file. Additionally it makes far more sense to implement support natively than to use a hack within a hack.

yossizahn:
sorry closed this by mistake

CallumHoward:
I would like if there was an option for image previews using hiptext
https://github.com/jart/hiptext

Vifon:
Is there any advantage of hiptext over caca-utils which are already supported?

CallumHoward:
Hiptext's tagline is "Turn images into text better than caca/aalib". It also has Sixel support (which I haven't tried).
I would say that it is a significant improvement. Here is an example comparing the current ranger preview, and hiptext output in my terminal (Terminal.app on macOS)
![wcveg](https://cloud.githubusercontent.com/assets/7144420/26033458/14b8ffb2-38f0-11e7-9ce4-f54b48993503.jpg)
![screen shot 2017-05-14 at 9 46 26 pm](https://cloud.githubusercontent.com/assets/7144420/26033459/23161f54-38f0-11e7-9d8a-5144d7354d56.jpg)
![screen shot 2017-05-14 at 9 47 08 pm](https://cloud.githubusercontent.com/assets/7144420/26033461/36364d7a-38f0-11e7-8cad-634fb2eb3395.jpg)

Vifon:
If it only outputs characters, no special handling is needed and it can be implemented in scope.sh. Look for the call to `img2txt` and try replacing it with hiptext.

CallumHoward:
[...]

PR #2234, opened 2021-01-22
Add experimental sixel support for image preview

3ap:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Terminal emulator and version: foot (1.6.2 +ime)
- Python version: 3.9.1 (default, Dec 13 2020, 11:55:53) [GCC 10.2.0]
- Ranger version/commit: ranger-master (73bcd80c1c3e721391c75015cb4ec03532357ad9)
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
This patch adds ability to use ImageMagick and Sixel graphics format for image preview

#### MOTIVATION AND CONTEXT
#723 

#### IMAGES / VIDEOS
image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ For enhanced file previews (with `scope.sh`):
 
 * `img2txt` (from `caca-utils`) for ASCII-art image previews
 * `w3mimgdisplay`, `ueberzug`, `mpv`, `iTerm2`, `kitty`, `terminology` or `urxvt` for image previews
-* `convert` (from `imagemagick`) to auto-rotate images and for SVG previews
+* `convert` (from `imagemagick`) to auto-rotate images, for SVG previews and for `sixel` image preview method
 * `ffmpegthumbnailer` for video thumbnails
 * `highlight`, `bat` or `pygmentize` for syntax highlighting of code
 * `atool`, `bsdtar`, `unrar` and/or `7z` to preview archives
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -308,7 +308,7 @@ are automatically used when available but completely optional.
 \&\f(CW\*(C`w3mimgdisplay\*(C'\fR, \f(CW\*(C`ueberzug\*(C'\fR, \f(CW\*(C`mpv\*(C'\fR, \f(CW\*(C`iTerm2\*(C'\fR, \f(CW\*(C`kitty\*(C'\fR, \f(CW\*(C`terminology\*(C'\fR or
 \&\f(CW\*(C`urxvt\*(C'\fR for image previews
 .IP "\-" 2
-\&\f(CW\*(C`convert\*(C'\fR (from \f(CW\*(C`imagemagick\*(C'\fR) to auto-rotate images and for \s-1SVG\s0 previews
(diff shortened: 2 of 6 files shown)


dnkl:
First of all, thanks! Really awesome to see sixel being used :)

This was originally reported in [#foot-terminal](https://freenode.logbot.info/foot-terminal/20210124#c6647210); with this configuration:
```
set preview_images true
set preview_images_method sixel
set status_bar_on_top true
```
it seems the previous image isn't erased properly. So if I am switching between images, there will be traces of a larger image left when displaying a smaller image.

Seen in both foot-1.6.2-302-g4c0c972 and xterm-363 (with `-ti 340`).

In the attached screenshots, I started with the **last** image and pressed <kbd>Up</kbd> twice. All the images in this folder is very similar, but the lower parts are left-overs from the first two images. 

foot:
foot

xterm:
xterm

toonn:
@dnkl, thank you for testing and reporting!

cherti:
I can confirm that I can no longer observe the issue mentioned by @dnkl when using the newest commit on recent master. :)

One observation I have made, though, but not sure if there is an easy fix for that: the call to convert [here](https://github.com/3ap/ranger/blob/sixel/ranger/ext/img_display.py#L432) becomes fairly pricey when the resulting resolution is fairly large (which is likely for bigger monitors) and while that runs, the UI is frozen. So if I go through a folder with 10 pictures, each 5MB, that convert takes quite a bit of time and is hindering me to go further down. Is there a way to abort the convert process (or simply stash it onto the "well finish if you must, but go right to the garbage collector"-pile) when the image is no longer selected?

(not saying this should be a blocker, though, having sixel-preview even with that is still nice, I can simply disable image previews if that aspect gets too annoying^^)

dnkl:
Latest commit has fixed the issue for me as well, thanks!

Eragonfr:
I have tested tis too and I did not found any bug.
What need to be done for this to be merged ?

toonn:
Just need to find some time to review the code. Feel free to send me any spare round tuits.

connection-reset:
I found the previews of certain pictures to have very visible banding, examples are attached (look at the sky in the upper half).
Providing the `-colors 1024` option to `convert` improves it, libsixel's `img2sixel` looks even better but as I understand it supports fewer image formats.
I'm not sure what `convert`'s `-colors 1024` does, `img2sixel` uses 256 colors by default and looks better.

Example `convert`:
imagemagick-256

Example `convert -colors 1024`:
imagemagick-1024

Example `img2sixel`:
img2sixel

Image source: https://www.esa.int/ESA_Multimedia/Images/2019/10/Antarctic_mist

I am using Arch Linux with ImageMagick 7.0.11-14, libsixel 1.8.6, foot 1.7.2 (Wayland terminal emulator, https://codeberg.org/dnkl/foot), ranger 1.9.3.

toonn:
@3ap, thank you for this work. There's more recent work in #2466 implementing basically the same thing though and the issues @connection-reset mentioned haven't showed up there so I'm closing this in favor of that PR.

Issue #2067, opened 2020-08-07
Feature suggestion : number of filtered files


manu0600:
The "filter" tool is very useful, but sometimes I forget that some files are hidden and wonder if I deleted them.
It would be very useful to show at the bottom of the screen, the number of filtered-out files in the current directory.

Thank you

toonn:
This is a good idea. #2064 is similar in spirit in making filter state more obvious in the UI.

laktak:
@toon I think you meant #2046

toonn:
Yes, thanks for catching that!

chu4ng:
May be we should add notes like 88/115(156) / x/y(z), where `z` is all files and `y` is filtered files?

PR #2239, opened 2021-02-06
Added showing number of all files, if we have filter.

chu4ng:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Python version: 3.9.1
- Ranger version/commit: ranger-master v1.9.3-216-gaabaa4fd / 07bad74
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

If we have active filter(not default), then adding the numbers of all files in dir(with hidden by default).

#### MOTIVATION AND CONTEXT

#2067
Actually needed feature. I also catch myself in situation "where these files wtf".

### DIFF
--- a/ranger/gui/widgets/statusbar.py
+++ b/ranger/gui/widgets/statusbar.py
@@ -304,7 +304,10 @@ class StatusBar(Widget):  # pylint: disable=too-many-instance-attributes
             # away and don't see them anymore.
             right.add('Mrk', base, 'marked')
         elif target.files:
-            right.add(str(target.pointer + 1) + '/' + str(len(target.files)) + '  ', base)
+            if target.filter == None:
+                right.add("{ptr}/{num}  ".format(ptr=str(target.pointer + 1), num=str(len(target.files))), base)
+            else:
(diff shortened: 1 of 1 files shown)


toonn:
I like this idea!

ranger/gui/widgets/statusbar.py
```diff
@@ -304,7 +304,10 @@ def _get_right_part(self, bar):  # pylint: disable=too-many-branches,too-many-st
             # away and don't see them anymore.
             right.add('Mrk', base, 'marked')
         elif target.files:
-            right.add(str(target.pointer + 1) + '/' + str(len(target.files)) + '  ', base)
+            if target.filter == None:
```
toonn, line 307: Maybe we should simply always show the real total if it differs? Regardless of filtering.

toonn (commented):
I don't think counting `target.files_all` poses a performance problem? What do you think of my suggestion?

ranger/gui/widgets/statusbar.py
chu4ng: In this case we need compare last index or number of selected all files and number here.
In my case you just should compare numbers `x` and `z` in `y/x(z)` notation.
Anctually, `y/x(z)` it's not beatiful format, but I don't know how change it else.

PR #2241, opened 2021-02-06
Use bash's native lowercase feature

pkkm:
This pull requests makes `scope.sh` use Bash 4's native lowercasing feature instead of an external `tr` process.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Ranger version/commit: ranger-master v1.9.2-708-gf0ef96c2
- Locale: None.None

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

`scope.sh` already uses bashisms, so using an additional Bash 4 feature shouldn't change compatibility. On the other hand, it lets us avoid spawning an external `tr` process and improve performance a bit.

While making this change, I also took the liberty to put quotes around some variables passed from the outside, just in case they happen to contain spaces, as well as to improve style consistency with the rest of the script.

#### TESTING

I tested it manually, everything seems to work.


toonn:
We keep compatibility quite far. I'm stuck with Bash 3 still on macOS. Furthermore, while `scope` isn't currently POSIX compatible I'd rather work towards it than away from it. Could you separate the variable quoting into its own PR, that part looks fine : )

pkkm:
That's surprising, Bash 4 is over a decade old so I expected macOS to have it.

> Could you separate the variable quoting into its own PR, that part looks fine : )

Sure! I've opened https://github.com/ranger/ranger/pull/2243.

PR #2242, opened 2021-02-07
Improve e-book image previews

pkkm:
This pull request removes support for the `epub-thumbnailer` tool (which is inferior to `ebook-meta`) and makes `scope.sh` use `ebook-meta` on more files by matching on file extension instead of the MIME type.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Ranger version/commit: ranger-master v1.9.2-708-gb7dcc382
- Locale: None.None

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

I've removed support for the `epub-thumbnailer` previewer because it's strictly worse than `epub-meta`: it's harder to install and it supports fewer formats.

I've also made e-book previews match on file extension instead of the MIME type. This is because the `file` tool often fails to detect e-book formats and reports them as `text/xml` or `application/octet-stream`. We could use the `mimetype` tool to get an accurate MIME type, but this would introduce ~40 ms of additional latency to `scope.sh`.

I think that seeing the cover is more useful than the current default, but I've left the preview code commented out to match the rest of the `handle_image` function.

#### TESTING

Tested manually on a variety of e-book files.


ranger/data/scope.sh
```diff
@@ -169,17 +169,6 @@ handle_image() {
         #              -- "${FILE_PATH}" "${IMAGE_CACHE_PATH%.*}" \
         #         && exit 6 || exit 1;;
 
-
-        ## ePub, MOBI, FB2 (using Calibre)
```
toonn, line 173: I don't really see a reason to remove epub-thumbnailer, it's not used unless it's installed. Not everyone wants to install a relatively large package like calibre just for cover previews.

ranger/data/scope.sh
```diff
@@ -249,17 +238,22 @@ handle_image() {
...
-    #     ## is hardcoded as jpeg. So we make a tempfile.png and just
-    #     ## move/rename it to jpg. This works because image libraries are
-    #     ## smart enough to handle it.
-    #     csg|scad)
-    #         openscad_image "${FILE_PATH}" && exit 6
-    #         ;;
-    #     3mf|amf|dxf|off|stl)
-    #         openscad_image <(echo "import(\"${FILE_PATH}\");") && exit 6
-    #         ;;
+        ## 3D models
```
toonn, line 241: I'm not sure the comment style should be changed, it clearly indicates that you need to uncomment the case statement rather than just the section you want to apply.

toonn (changes requested):
Let me know if there's more serious reasons not to include epub-thumbnailer.

ranger/data/scope.sh
pkkm: I changed it because I thought this style better indicates that you can just uncomment the case statement and one of the sections without uncommenting the other.

ranger/data/scope.sh
pkkm: > Let me know if there's more serious reasons not to include epub-thumbnailer.

Not really, I removed it because I'm the person who introduced it and I only did that because I didn't know how much better Calibre's format support is. Also, people who deal with e-books probably have Calibre installed, but the same isn't true of epub-thumbnailer. If you'd still like epub-thumbnailer support back, I can re-add it and squash the commits of course.

toonn: I'm open to replacing it with another lightweight solution but I do think it's a net positive to have an alternative.

ranger/data/scope.sh
toonn: Let's stick with the status quo.

pkkm:
I couldn't find another lightweight alternative to Calibre so I've restored support for epub-thumbnailer and squashed everything into one commit.

pkkm:
Hmm, Shellcheck is complaining about having two alternatives for handling `epub` in the `case` statement. I'll disable these warnings with a comment.

pkkm:
Is this OK or would you like any other changes?

toonn:
Don't have the bandwidth for reviewing rn. Don't like there being two overlapping cases like this though. Maybe
```
epub)
    epub-thumbnailer && exit whatever
    ; # Note the *single* semi-colon (probably can drop it entirely, just wanted to be explicit about not having double semi-colon) so it falls through to the next case.
awz3|mobi|...)
    calibre...
```

pkkm:
Wouldn't that make it more complex for users to enable the feature though? With the overlapping cases, the uses just has to uncomment the section for the program they want to use. I think most people will want to use `ebook-meta` only, including for EPUBs.

toonn:
It has to still make sense when everything's uncommented.

pkkm:
Right now, when everything is uncommented, there would be no errors but `epub-thumbnailer` would remain unused. If the order was changed, then `epub-thumbnailer` would be used for EPUBs and `ebook-meta` would be used for everything else. I placed Calibre's `ebook-meta` on top to make clear that it should be preferred though, because it has more robust support for weird e-books.

To be honest, I don't see a solution without disadvantages here.

toonn:
You can put both in the same clause. If ebook-meta fails or isn't present it'll fall through to epub-thumbnailer, if that fails or isn't present there just won't be a preview. This fall-through logic is pervasive in `scope.sh`. Usually we order by least-likely to be present though.

pkkm:
Do you think it's OK to call epub-thumbnailer for MOBI, AZW3 and FB2, even though we know in advance it will error out on these files?

toonn:
Not ideal, not harmful either.

pkkm:
OK, I've merged the cases. Let me know if you'd like the commits in this PR squashed.

pkkm:
On a second thought, we could also do this:

```bash
## E-book covers.
epub|mobi|azw3|azw|fb2)
    ## Using a tool included with Calibre.
    ebook-meta --get-cover="${IMAGE_CACHE_PATH}" -- "${FILE_PATH}" \
        > /dev/null && exit 6
    ## Using epub-thumbnailer.py from
    ## <https://github.com/marianosimone/epub-thumbnailer>.
    [ "${FILE_EXTENSION_LOWER}" = "epub" ] && \
        epub-thumbnailer "${FILE_PATH}" "${IMAGE_CACHE_PATH}" \
            "${DEFAULT_SIZE%x*}" && exit 6
    exit 1;;
```

Not sure if it's a better or worse solution though.

PR #2244, opened 2021-02-12
Minor improvements to scope.sh

pkkm:
This pull request contains a few minor changes to polish scope.sh a bit.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Ranger version/commit: ranger-master v1.9.2-709-g671bedc2
- Locale: None.None

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

* Avoid unnecessary calls to `env`.
* Add `--` where it's supported, in case of weird file names.
* Improve style consistency in variable names and spacing.

#### TESTING

Tested manually.


ranger/data/scope.sh
```diff
@@ -293,20 +294,20 @@ handle_mime() {
...
             fi
-            if [[ "$( tput colors )" -ge 256 ]]; then
+            if [[ "$(tput colors)" -ge 256 ]]; then
                 local pygmentize_format='terminal256'
                 local highlight_format='xterm256'
             else
                 local pygmentize_format='terminal'
                 local highlight_format='ansi'
             fi
-            env HIGHLIGHT_OPTIONS="${HIGHLIGHT_OPTIONS}" highlight \
```
toonn, line 306: Calling env is more portable so I see no reason not to.

toonn (changes requested):
Did you verify both GNU and BSD implementations support the `--` custom? Sometimes it's more trouble than it's worth.

ranger/data/scope.sh
pkkm: Well, it's one more external process. Is there anything that doesn't support the non-`env` variant? I tested on `sh`, `bash`, and `zsh`.

toonn: Fish and csh : )
If you want a fast `scope.sh` you could write one in C or Go or any language of your choosing.

pkkm:
I've checked `jq` and `mediainfo`'s support for `--` on my Debian machine, and I've checked `rm` and `mv` on Debian, FreeBSD, and OpenBSD. I didn't change the `exiftool` invocations because it seems not to support this syntax.

What's your decision with the `env`? I can amend the commit to remove this change, but to be honest, I'd keep it because any shell the user could reasonably run `scope.sh` under supports the non-`env` syntax (fish and tcsh are totally incompatible with this script anyway).

I'm a bit biased though, I tend to think that compatibility with a variety of shells is more trouble than it's worth. Any UNIX system is likely to have some version of Bash and plenty of packages already require it, so I like Google-style policies of having all `.sh` files target a minimum version of Bash only.

toonn:
Don't really care much tbh, I'll accept dropping it.

pkkm:
By the way, while I'm editing the script, would you like me to document its compatibility philosophy in a comment? As far as I can tell, the status quo is to target the old version of Bash included with macOS, but also use sh-compatible constructs when it's not too much trouble. I think making it explicit could cut down on future questions on the issue tracker.

subnut:
> Any UNIX system is likely to have some version of Bash

NO.

That's like saying _any UNIX-like system is likely to have systemd as init,_ which is totally wrong.
_**Most** of them_? Maybe. But _**all** of them_? Definitely not.

Why not just make all scripts POSIX-compliant?

toonn:
To be clear, my stance is anything that's added should conform to POSIX unless there's a good reason, same thing with anything that's changed. We haven't removed all the bashisms yet because of some issues I don't recall. I have some open PRs on changing both the polyglot and scope to be POSIX-compliant iirc.

pkkm:
OK, I've added a comment about compatibility and changed the simplest conditional to sh style.

ranger/data/scope.sh
```diff
@@ -241,11 +249,12 @@ handle_image() {
     esac
 
     # openscad_image() {
-    #     TMPPNG="$(mktemp -t XXXXXX.png)"
+    #     local tmp_png
```
toonn, line 252: POSIX doesn't support local.

ranger/data/scope.sh
```diff
@@ -293,20 +302,20 @@ handle_mime() {
         ## Text
         text/* | */xml)
             ## Syntax highlight
-            if [[ "$( stat --printf='%s' -- "${FILE_PATH}" )" -gt "${HIGHLIGHT_SIZE_MAX}" ]]; then
+            if [[ "$(stat --printf='%s' -- "${FILE_PATH}")" -gt "${HIGHLIGHT_SIZE_MAX}" ]]; then
```
toonn, line 305: Can you update these conditionals to use `test/[]` too?

toonn (changes requested):
The comment looks good to me. Maybe not say "target bash", it's more the current state rather than what we target.

ranger/data/scope.sh
pkkm: OK, I'll remove this line.

ranger/data/scope.sh
pkkm: Sure.

pkkm:
OK, I've made the changes. I left one bash-style conditional alone because I didn't touch the surrounding code in this pull request.

ranger/data/scope.sh
```diff
@@ -241,11 +249,11 @@ handle_image() {
     esac
 
     # openscad_image() {
-    #     TMPPNG="$(mktemp -t XXXXXX.png)"
+    #     tmp_png="$(mktemp -t XXXXXX.png)"
     #     openscad --colorscheme="${OPENSCAD_COLORSCHEME}" \
     #         --imgsize="${OPENSCAD_IMGSIZE/x/,}" \
-    #         -o "${TMPPNG}" "${1}"
-    #     mv "${TMPPNG}" "${IMAGE_CACHE_PATH}"
+    #         -o "${tmp_png}" -- "${1}"
```
toonn, line 255: Why the inconsistent placement of `--` here?

toonn (commented):
One more remark.

ranger/data/scope.sh
pkkm: The `-o "${tmp_png}" -- "${1}"`? It's after options and before positional arguments, so I think it's consistent.

toonn: Oh, you're right. I was confused by the `mv` line just above.

PR #2262, opened 2021-02-28
Fix combination key with uppercase letters

XuehaiPan:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: macOS Catalina
- Terminal emulator and version: iTerm2
- Python version: 3.8.2
- Locale: en_US

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Allow users add a combination key with uppercase letters in shorten format (`<A-X>` vs. `<Alt>X`).

#### MOTIVATION AND CONTEXT

Function `parse_keybinding` arbitrarily convert the keybinding content in brackets into lowercase. The return value of `parse_keybinding('<a-x>')` and `parse_keybinding('<a-X>')` are the same. Users should use `<alt>X` rather than `<a-X>` in their config file to bind a combination key with uppercase letters.


toonn:
Seems pretty reasonable. Bugfixes take priority currently though.

ranger/ext/keybinding_parser.py
```diff
@@ -53,7 +53,7 @@ def special_keys_init():
     for key, val in tuple(special_keys.items()):
         special_keys['a-' + key] = (ALT_KEY, val)
 
-    for char in 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_!{}':
+    for char in 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_!{}[],./':
```
XuehaiPan, line 56: In vim, `<C-F>` stands for `<CTRL>+f`. Users are not allowed to bind `<CTRL>` or `<ALT>` with an uppercase letter. Just remove all uppercase letters here and forbid users bind a combination key with an uppercase letter also fixes this.

But in [`examples/plugin_pmount.py#L16-#L18`](https://github.com/ranger/ranger/blob/master/examples/plugin_pmount.py#L16-#L18), the example binds an uppercase letter with `<alt>`:

```python
MOUNT_KEY = '<alt>m'
UMOUNT_KEY = '<alt>M'
LIST_MOUNTS_KEY = '<alt>N'
```

XuehaiPan: New special key of `<A-/>` is added here for fd integration.
https://github.com/ranger/ranger/wiki/Custom-Commands#search-with-fd

XuehaiPan:
In the original `keybinding_parser.py`, function `parse_keybinding` and `construct_keybinding`  have many limitations. For example:
```python
# Ignore letter cases in special keys (mentioned above)
>>> tuple(parse_keybinding('<A-X>'))
(9003, 120)  # ord('x') = 120 and ord('X') = 88
>>> tuple(parse_keybinding('<A-x>'))
(9003, 120)  # ord('x') = 120

# Some named special keys are overrided by <C-*>
# The values of <Enter>, <Return>, <CR> and <C-j> are all equal to 10 = ord('\n') = ord('j') - 96
>>> construct_keybinding(parse_keybinding('lol<CR>'))
'lol<c-j>'  # (108, 111, 108, 10)

# `construct_keybinding` does not combine (ALT_KEY, other) to <A-*> together
>>> construct_keybinding(parse_keybinding('x<A-Left>'))
'x<alt><left>'  # (120, 9003, 260)
```

I have refactored `keybinding_parser.py` to solve above issues.
```python
# Do not ignore letter cases in special keys
>>> tuple(parse_keybinding('<A-X>'))
(9003, 88)  # ord('X') = 88
>>> tuple(parse_keybinding('<A-x>'))
(9003, 120)  # ord('x') = 120

# Let named special keys have higher priority than combination keys
# The values of <Enter>, <Return>, <CR> and <C-j> are all equal to 10 = ord('\n') = ord('j') - 96
>>> construct_keybinding(parse_keybinding('lol<CR>'))
'lol<Enter>'  # (108, 111, 108, 10)

# Combine (ALT_KEY, other) to <A-*> together
>>> construct_keybinding(parse_keybinding('x<A-Left>'))
'x<A-Left>'  # (120, 9003, 260)
>>> construct_keybinding(parse_keybinding('x<Alt><Left>'))
'x<A-Left>'  # (120, 9003, 260)
```

I changed the letter case of keybindings in the config file to VIM style (and added some spaces to align values). The keycodes can be found at https://vimhelp.org/intro.txt.html#key-notation.

JosefLitos:
Would this fix the unability to use some random key combos like `<C-q>`?

XuehaiPan:
> Would this fix the unability to use some random key combos like `<C-q>`?

@JosefLitos The curses library does not recognize the `<C-q>` (but `<C-p>` and `<C-r>` are available). You can try:

```bash
python doc/tools/print_keys.py
```

image

toonn:
@JosefLitos, that's not actually a limitation of Ranger at all. Your terminal emulator is interpreting that as control flow instructions. `ixon`/`ixoff` are helpful search terms.

JosefLitos:
> @JosefLitos, that's not actually a limitation of Ranger at all. Your terminal emulator is interpreting that as control flow instructions. `ixon`/`ixoff` are helpful search terms.

But why does it work in other term interfaces like NeoVim or mpv?

doc/tools/print_keys.py
```diff
@@ -3,21 +3,37 @@
...
 
+# pylint: disable=import-error,wrong-import-position
+
 from __future__ import (absolute_import, division, print_function)
 
 import curses
+import os
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
```
toonn, line 14: Why do we need to mess with the PYTHONPATH here?

doc/tools/print_keys.py
```diff
@@ -3,21 +3,37 @@
...
+            string = 'MouseEvent' + repr(curses.getmouse())
         else:
-            window.addstr(str(char) + SEPARATOR)
+            string = str(char) + ' => ' + repr(construct_keybinding(char))
+
+        try:
+            win.addstr(string + SEPARATOR)
+        except curses.error:
+            win.erase()
+            win.addstr(string + SEPARATOR)
```
toonn, line 39: How can we be sure addstr won't throw another error?

toonn (commented):
The changes seem to go further than just enabling binding uppercase characters, no?

doc/tools/print_keys.py
XuehaiPan: This is used to import `ranger.ext.keybinding_parser` without needing to install `ranger`. This file can be run directly with:

```bash
python doc/tools/print_keys.py
```

doc/tools/print_keys.py
XuehaiPan: `cursor.error` is the only exception type that functions in `curses` will raise. (the input is a `str`, it will not raise `TypeError` for invalid input type).

XuehaiPan:
> The changes seem to go further than just enabling binding uppercase characters.

This PR also includes fixes for the config file and tool `print_keys.py` related to key bindings.

doc/tools/print_keys.py
toonn: I understand what the code does but this is a bad way of doing it and won't be accepted.

doc/tools/print_keys.py
toonn: I know what `curses.error` is, but if the second call to `addstr` in the except clause raises it again then it won't be caught, that would potentially result in an infinite loop.

doc/tools/print_keys.py
XuehaiPan: I resolved this.

Issue #2260, opened 2021-02-26
incorrect line numbers in active pane with border


ri-aje:
#### Runtime Environment

- Operating system and version: ubuntu 18.04
- Terminal emulator and version: terminator 1.91
- Python version: 3.8.6
- Ranger version/commit: e5a06fa2
- Locale: 

#### Current Behavior
ranger displayes wrong line numbers in active pane in multipane mode. for line numbers with two digits or more, the first digit is missing in display, looks like truncated.

#### Expected Behavior
correctly display line numbers in all view modes.

#### Context

viewing files and have wrong line numbers. pretty annoying, as the the counts are off in jumping, selecting, etc.

#### Possible Solutions
per my binary search, regression seems to have come from 47b116e1. I checked out that commit, don't see anything obviously wrong, I am gussing the bug is triggered when multiple conditions are met, and 47b116e1 just happens to manifest it somehow.

#### Steps to reproduce
1. create a ranger `rc.conf` with these two lines
```
set line_numbers relative # setting to absolute also repros 
set draw_borders_multipane active-pane
```
2. open a folder that contains more than 10 files, so as to at least be able to show 2-digit line numbers, absolute or relative.
3. press `~` to switch to multiple pane mode. note the active pane is drawn with a rectangle border around it, and the line numbers with two or more digits are missing their first digit. 
4. press `ctrl-n` to create another tab, which is shown side by side, and note that the inactive panes don't have this problem.
5. now pressing `tab` to quickly switching between the two panes back and forth, you will see an animation showing the active pane loses line numbers when the border is drawn on it.

#### Traceback

```

```

ri-aje:
+@zyklotomic since you are the committer for 47b116e. any chance to fix it? thanks.

zyklotomic:
I don't think I accounted for that, will look into it asap this weekend, thanks for bringing it to my attention

zyklotomic:
https://github.com/ranger/ranger/pull/2264　I had a quick stab at it, you could try this patch out for yourself for now if you wish. If you do, let me know if you find anything wrong!  There might be some rough edges, I haven't had the time to thoroughly test it unfortunately.

ri-aje:
[...]

PR #2264, opened 2021-03-03
Line Numbers Cut Off in Active Pane

zyklotomic:
#### ISSUE TYPE
Bug fix for
https://github.com/ranger/ranger/issues/2260

My attempt so far, the line number cutoff appears to be an off-by-one error.

The problem now is that the left-most column has no padding between the start of the window and itself, leaving no space for the border to go. Temporary fix is to simply shrink the left-most column.
-------------
Sorry, have been busy lately, waned to share my attempts at fixing this bug first. Will get around to cleaning up the PR when I have the time.

#### CHECKLIST

- [ ] The `CONTRIBUTING` document has been read **[REQUIRED]**
    - [ ] Tests have been updated

### DIFF
--- a/ranger/gui/widgets/view_multipane.py
+++ b/ranger/gui/widgets/view_multipane.py
@@ -94,11 +94,10 @@ class ViewMultipane(ViewBase):  # pylint: disable=too-many-ancestors
         win = self.win
         self.color('in_browser', 'border')
 
-        left_start = 0
-        right_end = self.wid - 1
-
         # Draw the outline borders
         if 'active-pane' not in border_types:
+            left_start = 0
+            right_end = self.wid - 1
             if 'outline' in border_types:
                 try:
                     self._draw_border_rectangle(left_start, right_end)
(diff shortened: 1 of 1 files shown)


zyklotomic:
I forgot to say, I understand that this edge case for the left most column probably isn't ideal, but it gets the job done functionally for now.

Any thoughts on how to better handle it?

toonn:
I'm not clear on why we can't resize the line number gutter, sacrificing a character from the middle column?

zyklotomic:
ah, I didn't mean we couldn't, just that I wasn't sure how to approach it. Could you clarify what you mean by the middle column.

toonn:
I was thinking of miller columns but since it's about multipane it'd just be the active pane.

ri-aje:
@zyklotomic @toonn is this fix likely be merged any time soon? thanks.

toonn:
Not if it does what I think it does.

zyklotomic:
hopefully i'll have the time to implement the fix toonn was suggesting. But I'm not sure if that's the right approach either, because it is only necessary for the left-most column.

I realize it is on me for introducing the bug in the first place, but @ri-aje , if you want to have a stab at it yourself, I think `BrowserColumn.py` would be where the logic to make that change happen would be.

toonn:
Wait, this only happens on the leftmost column? Then maybe your fix *is* acceptable. Not sure there's a more proper way I'll have to look at it more closely.

ri-aje:
@toonn per my simple test, the bug only happens to the active pane, which may or may not be the left column, however, @zyklotomic 's current fix does work in all cases, whether the active pane is left-most or not.

zyklotomic:
Yeah, the nuance is that in the Miller pane view, only the focused center pane has the line numbers. Such a problem wouldn't appear there

PR #2267, opened 2021-03-04
feat: use quicklook on Mac to preview pdf

duhd1993:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: macos 11.2.2
- Terminal emulator and version: iTerm2 3.4.4
- Python version: Python 2.7 built in with macos
- Locale: en_US

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

quicklook is Mac's built-in tool for previewing. Use it to preview pdf.

#### MOTIVATION AND CONTEXT

Actually quicklook can preview many other formats. It's worthwhile to enable it by default, because every mac comes with it.

#### TESTING

works on my mac. qlmanage is available on every mac. not sure since when, maybe 10 years ago, to the latest macos. And from my feeling, the API never changes.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -160,7 +160,18 @@ handle_image() {
         #     exit 1;;
 
         ## PDF
-        # application/pdf)
+        application/pdf)
+            if [[ "$(uname)" = "Darwin" ]]; then
+                ## Use mac built-in quicklook to preview pdf
+                ## rename extension from .png to .jpg provided by IMAGE_CACHE_PATH
+                ## image viewer can determine the actual format
+                qlmanage -t -s "${DEFAULT_SIZE%x*}" \
+                         -o "${IMAGE_CACHE_PATH%/*}" \
+                         ${FILE_PATH} \
+                         && mv "${IMAGE_CACHE_PATH%/*}/${FILE_PATH##*/}.png" "${IMAGE_CACHE_PATH}" \
+                         && exit 6 || exit 1
+            fi
+            exit 1;;
         #     pdftoppm -f 1 -l 1 \
         #              -scale-to-x "${DEFAULT_SIZE%x*}" \
         #              -scale-to-y -1 \
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -172,6 +172,14 @@ handle_image() {
                          && exit 6 || exit 1
             fi
             exit 1;;
+            #if [[ "$(uname)" = "Darwin" ]]; then
+            #    echo 'Previewing in Mac Quicklook'
+            #    cat /tmp/rangerql.pid | xargs kill -9 > /dev/null 2>&1
+            #    nohup qlmanage -p ${FILE_PATH} > /dev/null 2>&1  &
+            #    rm /tmp/rangerql.pid
+            #    echo $! > "/tmp/rangerql.pid" && exit 5
+            #fi
+            #exit 1;;
         #     pdftoppm -f 1 -l 1 \
         #              -scale-to-x "${DEFAULT_SIZE%x*}" \
         #              -scale-to-y -1 \
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ For enhanced file previews (with `scope.sh`):
   archives as their first image
 * `lynx`, `w3m` or `elinks` to preview html pages
 * `pdftotext` or `mutool` (and `fmt`) for textual `pdf` previews, `pdftoppm` to
-  preview as image
+  preview as image, and built-in `qlmanage` to preview as image on Mac.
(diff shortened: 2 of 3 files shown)


toonn:
Interesting. Do I understand correctly you're using quicklook to generate the images which are then shown by one of the previewing methods? Or are you actually triggering quicklook to have previews pop up and if not would this be feasible/usable?

lonelykid:
I'm generating previews images (thumbnails of qlmanage actually) and using the image previewing methods. You can also trigger the actual quicklook window from the command line. That way you can view other pages as well. I can add that. 

But IMO it's not that useful for pdfs. I'm just having a quicklook to determine what I want to open. Maybe it will be more useful for videos.

toonn:
Hmm, definitely don't want to use pkill if we can avoid it. Capturing the PID would be much better. Maybe it should simply be a rifle rule though?

lonelykid:
`pkill` is not that scary here. It's only used in command line. Finder or other file managers have sdk instead for quicklook. How do I store the pid, using a temp file? It seems scope.sh is stateless.

Using rifle to open quicklook? But quicklook is only for previewing, it's not a full-featured reader. I don't want to override my pdf reader.

toonn:
It doesn't have to override any pdf viewers. Just put it at the end of the rules for pdf/document viewing. It's a useful feature so we might as well include it.

diego898:
very interested in this - curious, why do you think this wont be useful for PDFs?

Bellavene:
Thank you! Now it is much faster to load previews, also it works with almost any format. Dunno why this is still not as standard option at the beginning.

One downside with monterey (which is still not managed overall) it triggers massive memory leak of your terminal leak fast when viewing a lot of pictures.

ranger/data/scope.sh
```diff
@@ -160,7 +160,26 @@ handle_image() {
         #     exit 1;;
 
         ## PDF
-        # application/pdf)
+        application/pdf)
+            if [[ "$(uname)" = "Darwin" ]]; then
```
toonn, line 164: I'm not sure we should check the uname here, if the binary's available then we can call it. If it's not the `exit 1` takes care of it.

ranger/data/scope.sh
```diff
@@ -160,7 +160,26 @@ handle_image() {
...
+                ## Use mac built-in quicklook to preview pdf
+                ## rename extension from .png to .jpg provided by IMAGE_CACHE_PATH
+                ## image viewer can determine the actual format
+                qlmanage -t -s "${DEFAULT_SIZE%x*}" \
+                         -o "${IMAGE_CACHE_PATH%/*}" \
+                         ${FILE_PATH} \
+                         && mv "${IMAGE_CACHE_PATH%/*}/${FILE_PATH##*/}.png" "${IMAGE_CACHE_PATH}" \
```
toonn, line 171: We could use a temporary file like elsewhere in `scope.sh`. That way we don't put anything in the cache that isn't used as a cached preview.

ranger/data/scope.sh
```diff
@@ -160,7 +160,26 @@ handle_image() {
...
+                         && exit 6 || exit 1
+            fi
+            exit 1;;
+            #if [[ "$(uname)" = "Darwin" ]]; then
```
toonn, line 175: I couldn't really get this to work. It works the first time a file is previewed but isn't respawned thereafter. I think we should use `exit 0`.

toonn (changes requested):
Still catching up on the issue backlog. Finally got around to taking a look.
Very promising : )

ranger/data/scope.sh
duhd1993: Okay. I will change accordingly.

ranger/data/scope.sh
duhd1993: I could change this. But why do you want to put into tmp first? I don't get it.

ranger/data/scope.sh
duhd1993: Sorry. I haven't tried this for a while. Do you mean the commented code for spawning quicklook window? Indeed, it's not respawning. It seems it relies on cache and is not calling the code again. That's how ranger works. What should I do to make it call everytime?

ranger/data/scope.sh
toonn: It's just cleaner. The cache directory should only contain finished previews ideally.

ranger/data/scope.sh
toonn: `exit 0`, that should make ranger show the cat output and try to regenerate it every time.

ranger/data/scope.sh
toonn: Thank you for bearing with us btw, I will try to respond faster.

PR #2274, opened 2021-03-07
Fix crash if man is not installed when hit '?' and 'm'.

ideal:
On some systems (e.g. archlinuxarm) man is not installed by default, hit
'?' and 'm' will cause ranger crashed, showing:
AttributeError: 'NoneType' object has no attribute 'poll'.

#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: archlinuxarm, with kernel 5.8.0
- Terminal emulator and version: zsh 5.8
- Python version: 3.9.1
- Locale: zh_CN.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Hit ? and m on archlinuxarm will cause ranger to crash.

#### TESTING

I have tested it on archlinuxarm.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -975,7 +975,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 return
         process = self.run(['man', 'ranger'])
         if process is None:
-            self.notify("Show manpage failed, check if man is installed", bad=True)
+            self.notify("Failed to show man page, check if man is installed", bad=True)
             return
         if process.poll() == 16:
             self.notify("Could not find manpage.", bad=True)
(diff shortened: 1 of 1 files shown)


ranger/core/actions.py
```diff
@@ -974,6 +974,9 @@ def display_help(self):
             if process.poll() != 16:
                 return
         process = self.run(['man', 'ranger'])
+        if process is None:
+            self.notify("Show manpage failed, check if man is installed", bad=True)
```
toonn, line 978: Please rephrase the message to emphasize the failure "Failed to show man page, check if man is installed".

toonn (changes requested):
Thanks for reporting (and fixing) this edge case!

ranger/core/actions.py
ideal: OK, updated.

toonn:
Good first contribution! 🎉

PR #2297, opened 2021-04-01
[minor] mcomix and atril update in rifle

pirate486743186:
adding rifle.conf

atril can open cbr,cbz too

mcomix can open rar, zip too, set to low priority, it's not an archive manager, it's a comic book viewer.

### DIFF
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -178,7 +178,7 @@ ext mobi, has ebook-viewer, X, flag f = ebook-viewer -- "$@"
 ext cb[rz], has qcomicbook, X, flag f = qcomicbook "$@"
 ext cb[rz], has mcomix,     X, flag f = mcomix -- "$@"
 ext cb[rz], has zathura,    X, flag f = zathura -- "$@"
-ext cb[rz], has atril,    X, flag f = atril -- "$@"
+ext cb[rz], has atril,      X, flag f = atril -- "$@"
 
 ext sla,  has scribus,      X, flag f = scribus -- "$@"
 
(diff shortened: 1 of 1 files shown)


ranger/config/rifle.conf
```diff
@@ -178,6 +178,7 @@ ext mobi, has ebook-viewer, X, flag f = ebook-viewer -- "$@"
 ext cb[rz], has qcomicbook, X, flag f = qcomicbook "$@"
 ext cb[rz], has mcomix,     X, flag f = mcomix -- "$@"
 ext cb[rz], has zathura,    X, flag f = zathura -- "$@"
+ext cb[rz], has atril,    X, flag f = atril -- "$@"
```
toonn, line 181: Is this MATE's atril? That looks like a much more general document viewer so maybe it should be added for pdf and djvu and other formats it supports too?

ranger/config/rifle.conf
```diff
@@ -227,6 +228,7 @@ ext ace, has unace = unace l "$1" | less
 ext ace, has unace = for file in "$@"; do unace e "$file"; done
 ext rar, has unrar = unrar l "$1" | less
 ext rar, has unrar = for file in "$@"; do unrar x "$file"; done
+ext rar|zip, has mcomix,     X, flag f = mcomix -- "$@"
```
toonn, line 231: Let's have this in the Documents section of rifle.conf, I think that makes more sense.

toonn (changes requested):
Couple suggestions, do you agree?

ranger/config/rifle.conf
pirate486743186: you get a lot of comic archives as .rar or .zip
This is not pointless

ranger/config/rifle.conf
pirate486743186: i only put in what i tested

pirate486743186: atril with pdf and djvu is already in.....

toonn: You're right, I didn't check closely enough.

ranger/config/rifle.conf
toonn: That's not what I said. Please reread my remark carefully, I'm only asking to move the condition into the cb[rz] rule, nothing more. Having it there would make other people test whether their favorite comic readers support `.zip` and `.rar`.

pirate486743186: it's meant to be low priority, it's not a general archive manager
what you say, will make it come first with rar/zip files.

pirate486743186: well, it doesn't work with atril, but works with qcomicbook
didn't test zathura

pirate486743186:
i also added qcomicbook and zathura in the archives, like mcomix.

pirate486743186:
maybe some one should merge this, before you get conflicts?

toonn:
Conflicts aren't a big deal. Thanks for the ping though I didn't get any notification of your changes or I must've missed it. I'll take a look as soon as I find some time.

toonn (approved):
LGTM

PR #2302, opened 2021-04-07
Load plugins from `ranger_plugins` namespace package

joouha:
This PR allows ranger to load plugins from a [namespace package](https://packaging.python.org/guides/packaging-namespace-packages/) called `ranger_plugins`.

My idea would allow plugins to be distributed as python packages which can be installed with pip, which would negate the need for a [separate plugin manager](https://github.com/ranger/ranger/issues/1417), and provide a simple way of distributing and managing plugins. 

#### ISSUE TYPE

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION

This uses `pkgutil.iter_modules` to search for and import packages in the `ranger_plugins` namespace package, in addition to those in the `plugins` configdir. 

#### MOTIVATION AND CONTEXT

I think this would be an improvement on the current plugin solution. Currently there are a few way to install plugins, including:
- [copying a script](https://github.com/yonghie/ranger-gitplug) to the plugin folder;
- [cloning a non-standard python package](https://github.com/alexanderjeurissen/ranger_devicons#install-instructions) to the plugins folder;
- [installing a python packages and sym-linking it to the plugins folder](https://github.com/hankei6km/mirage_linemode).

Uninstalling a plugin requires manually deleting files from the plugin folder.

For an example of how such a plugin can be built, please see the [`tmux_plugins_format` branch of my `ranger_tmux` plugin](https://github.com/joouha/ranger_tmux/tree/tmux_plugins_format).

Plugins are still loaded from the plugin folder, so existing plugins would still work.

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -133,7 +133,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.9.3" "2021-09-05" "ranger manual"
+.TH RANGER 1 "ranger-1.9.3" "2021-11-13" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
 .\" way too many mistakes in technical documents.
 .if n .ad l
@@ -988,8 +988,8 @@ on top of those for \fIdraw_borders\fR:
 .IX Item "draw_progress_bar_in_status_bar [bool]"
 Draw a progress bar in the status bar which displays the average state of all
 currently running tasks which support progress bars?
-.IP "flushinput [bool] <zi>" 4
-.IX Item "flushinput [bool] <zi>"
+.IP "flushinput [bool] <zI>" 4
+.IX Item "flushinput [bool] <zI>"
 Flush the input after each key hit?  One advantage is that when scrolling down
 with \*(L"j\*(R", ranger stops scrolling instantly when you release the key.  One
 disadvantage is that when you type commands blindly, some keys might get lost.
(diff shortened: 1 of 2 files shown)


toonn (commented):
This is interesting. Could you summarize what changes people'd have to make to their plugins?

joouha:
Existing plugins should still work without change. They can still be installed into the plugin folder in `confdir` and things would work as they do currently.

It would just give plugin developers the option of publishing their plugin as a pip-installable python package, which does not require an additional installation step to be picked up by ranger.

toonn:
Yes, but please describe how they'd do that. Ideally in the form of documentation that can be included in the man page.

joouha:
I see - yes, I'll put something together

joouha:
> it possible to distribute a plugin so it'd still work with both methods or is this an either/or situation?

You definitely could clone a repo (or download a zip file of the repo) and copy a script / folder into the `plugins` folder, and it would still work.

If you want to be able to clone a repo direct into the `plugins` folder, it should be possible to include a `__init__.py` script in the root of the repo to make things work. Let me do some tests :-)

joouha:
> it possible to distribute a plugin so it'd still work with both methods or is this an either/or situation?

I've done some playing around, and it is possible to create a plugin repository which works both ways 
-- see https://github.com/joouha/ranger_plugin_hello_world for an example.

For the documentation, do you want me to write something in `ranger.pod` and add it to the PR?

toonn:
Yes, just add it to the ranger man page. Unless it gets too big, we could in principle add a man page detailing plugin development and packaging.

PR #2307, opened 2021-04-12
Remove xargs execution

PRESFIL:
#### ISSUE TYPE
- Bug fix

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
The use of `xargs` has a limitation on the *length of `argv`* (number of
arguments). After reaching the maximum number of arguments (~850 for me)
`xargs` will still create the new child process, as with the `-n` switch.
`sxiv` (at least for now, `sxiv-26`) is able to read filenames from STDIN on
its own.

#### MOTIVATION AND CONTEXT
This commit by abandoning `xargs` fixes the aforementioned sideeffect, removes
extra dependency and speeds up the work (at least for me).

### DIFF
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -42,7 +42,7 @@ target="$(abspath "$1")"
 count="$(listfiles | grep -m 1 -ZznF "$target" | cut -d: -f1)"
 
 if [ -n "$count" ]; then
-    listfiles | xargs -0 sxiv -n "$count" --
+    listfiles | sed 's/\x0/\n/g' | sxiv -n "$count" -
 else
     sxiv -- "$@" # fallback
 fi


toonn:
Your sed isn't valid re escaping of arguments. `\n` is a completely valid character in paths unfortunately.

PRESFIL:
Yes, but `sxiv` accepts a newline-delimited (`\n`) list of files. It ignores them.

I removed `sed` and it ignores everything after the first `\0` character.

I think that fixing the strange launching of multiple `sxiv`'s due to `xargs` is better than skipping extremely rare files with newlines in the path.  When this happens, `sxiv` runs multiple times when opening any file in a folder with a lot of images, and it is very annoying to close this hail of their window.

PRESFIL:
Related issue on `sxiv` repository: https://github.com/muennich/sxiv/issues/445

A very strange coincidence. :eyes:

toonn:
If that issue is resolved we'd drop xargs in a heartbeat. As it stands, correct is better than convenient. Especially because this is the kind of issue that's really annoying to troubleshoot once it does happen.

PRESFIL:
OK, but I think that opening 12 `sxiv` windows instead of 1 is more wrong than ignoring one of these strange names.

toonn:
I understand it's not the nicest behavior. But there's no better solution unless there's a way to get sxiv to open files other than through command line arguments. xargs splits it up into multiple invocations because there's a limit on the length of the argument list. But we need xargs in this case because sxiv doesn't properly handle filenames with newlines in them.

PR #2321, opened 2021-05-06
XCF preview thumbnails using gimp

CurlyRhyme:
#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Generates a thumbnail for an xcf file, using gimp and a short embedded python-fu script.
This feature requires gimp python scripting support to work (for example, in Arch you need the [python2-gimp2](https://aur.archlinux.org/packages/python2-gimp/) package from the AUR)
This could be rewritten to use script-fu which requires no additional possible dependency (besides gimp itself)

#### MOTIVATION AND CONTEXT

Scrubbing through a directory with badly named xcf files is a lot easier with thumbnails.

#### TESTING

Ran `make test` and `make test_shellcheck` with no errors.
This shows a preview of xcf files (using mpv and the w3m method) despite "xcf" still being in 'PREVIEW_BLACKLIST' in `ranger/container/file.py`. Wasn't sure if this is a bug or intended behavior, so I did not change the blacklist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ For enhanced file previews (with `scope.sh`):
 * `python` or `jq` for JSON files
 * `fontimage` for font previews
 * `openscad` for 3D model previews (`stl`, `off`, `dxf`, `scad`, `csg`)
+* `gimp` with python scripting support for xcf files
 
 Installing
 ----------
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -138,6 +138,22 @@ handle_image() {
         #           - "${IMAGE_CACHE_PATH}" < "${FILE_PATH}" \
         #           && exit 6 || exit 1;;
 
+        ## XCF Image
+        image/x-xcf)
+            gimp-console -i --batch-interpreter python-fu-eval -b "
+try:
+    import gimpfu
+    def convert(filename, target):
+        img = pdb.gimp_file_load(filename, filename)
+        layer = pdb.gimp_image_merge_visible_layers(img, gimpfu.CLIP_TO_IMAGE)
+
+        pdb.gimp_file_save(img, layer, target, target)
+        pdb.gimp_image_delete(img)
+    convert(\"${FILE_PATH}\", \"${IMAGE_CACHE_PATH}\")
+finally:
+    pdb.gimp_quit(1)" && exit 6
+            exit 1;;
+
         ## Image
         image/*)
             local orientation


toonn:
Hmm, this seems a bit hard to justify tbh. I don't think it's a common use case and it's quite a complex addition relative to the usefulness I expect. Could be a cool addition to the wiki though!

hut:
1. I tend to agree with @toonn here, it seems too overkill to call gimp for a preview.
2. It doesn't appear to work these days anymore. I'm getting this if I run it:

```
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
Warning: Unknown input_id: -1 for input: surfacemap_x
env: ‘gjs’: No such file or directory
gimp-console: LibGimpBase-WARNING: gimp: gimp_wire_read(): unexpected EOF
batch command experienced a calling error:
  File "<string>", line 3, in <module>
ModuleNotFoundError: No module named 'gimpfu'

  File "/usr/lib/gimp/3.0/plug-ins/python-eval/python-eval.py", line 42, in code_eval
    exec(code, globals())
    ~~~~^^^^^^^^^^^^^^^^^
  File "<string>", line 12, in <module>
AttributeError: module 'pdb' has no attribute 'gimp_quit'

Stopping at failing batch command [0]: import pdb
try:
    import gimpfu
    def convert(filename, target):
        img = pdb.gimp_file_load(filename, filename)
        layer = pdb.gimp_image_merge_visible_layers(img, gimpfu.CLIP_TO_IMAGE)

        pdb.gimp_file_save(img, layer, target, target)
        pdb.gimp_image_delete(img)
    convert("/home/hut/test.xcf", "/home/hut")
finally:
    pdb.gimp_quit(1)

== INFO ==
GIMP is now running as a background process. You can quit anytime with Ctrl-C (SIGINT).

If you wanted to quit immediately instead, call GIMP with --quit.
```

This might be a hint that using such a gimp script is not future-proof, breaks easily, and should probably be avoided.

Thanks for the wild idea @CurlyRhyme, but sorry, I think we'd be better off overall if we don't do this.

PR #2331, opened 2021-05-17
Clickable preview

toonn:
A regression when implementing preview scrolling meant that clicking the preview no longer opens files.

I have fixed this regression but changed the original behavior to require a right click because the left click caused plenty of frustration by accidentally opening files when focusing the terminal window.


toonn (commented):
I ran tests locally but it's looking like Travis is no more. Need to look into alternatives because GHActions aren't running either for some reason. CircleCI?

PR #2335, opened 2021-05-25
filter command works with pattern, not with string

hektr:
filter command works with pattern, not with string

Little correction of manual


toonn:
This is actually an implementation detail. We don't mind lying to users here : )

toonn:
You could correct the "This command is based on scout and supports all its options" bit though.

hektr:
Not all options? All options, except of -g? It already uses its options, and we can't add options conflicting with options already used.
I'm not good in github, it's too confusing for me :) Idk, how to change commit. And I'm not too good in english for composing right sentences, as you can see :)

PR #2346, opened 2021-06-03
Width correction for relative line numbers

NoSuck:
Believe it or not, ranger has *never* done this correctly.  I may have bitched about it in the past.  Today, I fixed it for everyone.

Long live relatives.

```
 ----------------------------------------     ---------------------------------------
|12 .gtk-bookmarks                 482 B |   |12 .gtk-bookmarks                482 B |
|11 .gtkrc-2.0                  -> 715 B |   |11 .gtkrc-2.0                 -> 715 B |
|10 .gtkrc-2.0-viewing-distance  -> 30 B |   |10 .gtkrc-2.0-viewing-distance -> 30 B |
| 9 .inputrc                   -> 1.52 K |   | 9 .inputrc                  -> 1.52 K |
| 8 .lesshst                       451 B |   | 8 .lesshst                      451 B |
| 7 .lftprc                    -> 194 B  |   | 7 .lftprc                    -> 194 B |
| 6 .mailcap                   -> 771 B  |   | 6 .mailcap                   -> 771 B |
| 5 .mille                        568 B  |   | 5 .mille                        568 B |
| 4 .msmtprc                     3.83 K  |   | 4 .msmtprc                     3.83 K |
| 3 .netrc                        543 B  |   | 3 .netrc                        543 B |
| 2 .python_history              8.11 K  |   | 2 .python_history              8.11 K |
| 1 .units                     -> 630 B  |   | 1 .units                     -> 630 B |
|83 .units_history                1.54 K |   |83 .units_history               1.54 K |
| 1 .urlview                   -> 425 B  |   | 1 .urlview                   -> 425 B |
| 2 .viminfo                      151 K  |   | 2 .viminfo                      152 K |
| 3 .vimrc                    -> 15.7 K  |   | 3 .vimrc                    -> 15.7 K |
| 4 .vnstatrc                 -> 1.32 K  |   | 4 .vnstatrc                 -> 1.32 K |
| 5 .wcalc_history                480 B  |   | 5 .wcalc_history                480 B |
| 6 .wcalcrc                   -> 352 B  |   | 6 .wcalcrc                   -> 352 B |
| 7 .wget-hsts                    625 B  |   | 7 .wget-hsts                    625 B |
| 8 .wgetrc                    -> 335 B  |   | 8 .wgetrc                    -> 335 B |
| 9 .Xauthority                   202 B  |   | 9 .Xauthority                   202 B |
|10 .xinitrc                  -> 1.04 K  |   |10 .xinitrc                  -> 1.04 K |
|11 .xmodmaprc                -> 11.9 K  |   |11 .xmodmaprc                -> 11.9 K |
 ----------------------------------------     ---------------------------------------
 Previous                                     Current
```

Questionnaire answers are identical to #2345.

### DIFF
--- a/ranger/gui/widgets/browsercolumn.py
+++ b/ranger/gui/widgets/browsercolumn.py
@@ -278,7 +278,9 @@ class BrowserColumn(Pager):  # pylint: disable=too-many-instance-attributes
 
         # Set the size of the linum text field to the number of digits in the
         # visible files in directory.
-        linum_text_len = len(str(self.scroll_begin + self.hei))
+        linum_text_len = len(str(len(self.target.files)
+                                 if self.settings.line_numbers == 'relative'
+                                 else self.scroll_begin + self.hei))
         linum_format = "{0:>" + str(linum_text_len) + "}"
 
         selected_i = self._get_index_of_selected_file()


NoSuck:
Why has this fix remain unmerged for over half a year?  Both it and #2345 seem like no-brainers.

toonn:
I'm having a hard time finding situations where the current code doesn't work. Does the problem occur for you no matter the contents of a directory?

The reason it's taking so long to get around to things is there's only so much time in a day and I can't prioritize ranger unfortunately.

markus-bauer:
It's definitely a real issue and has nothing to do with the content of the directory:
ranger_linum_issue

Sorry to @NoSuck, but I don't understand the pull request. Perhaps you could explain why you think this would fix the issue. Observations:

1) If you look at your example, the problem is not that `linum_text_len` is wrong, but that the rest of the line is too short or too long.
2) I don't see why there should be a difference between absolute and relative linenumbers. Both have to account for the maximum possible line number width in the current range.
3) The current code dynamically adjusts to the biggest line number in this window (`len(str(self.scroll_begin + self.hei))`). Your changes would make the width of the line number field static instead of dynamic (fixed to `len(str(files.len())`).

Hypothesis: I suspect it's actually an issue with `display_data`. Here's a simplified example and perhaps a way to recreate the bug:

Let's say you're at the beginning of the directory and `linum_text_len` is 2. The rest of the line will be created to fill the space width-2. Then you jump to a different position far down the list where `linum_text_len` is 4. If you then change the sorting so that the previously created line shows up, it will be loaded from display_data. But it will now be too long and the line will be cut off. Same example works the other way, and the line will be to short.

As a test you can disable the loading from `display_data` here:
https://github.com/ranger/ranger/blob/e3ca4e75daa82d6fcf8d12406adf8e187e4750a1/ranger/gui/widgets/browsercolumn.py#L313
which forces the line to be remade every time, and the current code should function correctly.

Also: Resizing the window will change `self.wid`, which also forces the lines to be remade.

Possible fixes:
- Making `linum_text_len` static could actually work. It has the drawback of wasting space in the line, but the advantage of reducing redundancy in display_data.
- Adding yet another item to the display_data key.
- Replacing self.settings.line_numbers in the display_data key with `linum_text_len` (which would be 0 if linenumbers are off).

PR #2395, opened 2021-07-17
Updating the code for a newer pylint

toonn:
This is gonna require significant testing. Time to push to prod^Wmaster!


toonn:
The failure for 3.5 is due to an older pylint.

toonn:
#2396 will drop the older pylint.

PR #2402, opened 2021-07-24
Actions: Add on pull_request again

toonn:
My testing was flawed. The status checks seemed to work fine because the
builds were triggered on push in the same repo. But this doesn't work
across repos. So GH Actions having run to completion on someone's fork
doesn't give us the status checks we need. Therefore we want to run our
Actions on PRs too, that way they're run from the ranger repo as well
and the status checks should actually be informative.

One possible improvement is that we could duplicate the workflows we
want to run on PRs, that way we can drop all the things that don't
really matter for status checks from the version matrices. Otoh that
means we have pretty much duplicated workflows we need to keep in sync.
Thank you for your lack of abstractions, GH Actions 🙄

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -5,6 +5,10 @@ on:
     paths:
       - '.github/workflows/pylint.yml'
       - '**.py'
+  pull_request:
+    paths:
+      - '.github/workflows/pylint.yml'
+      - '**.py'
 
 jobs:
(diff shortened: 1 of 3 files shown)


toonn:
I will hold off on merging this regardless of approval until I'm certain on push isn't enough for new PRs. I just realized that all the other PRs predate the workflow changes, so maybe that's the problem.

PR #2411, opened 2021-08-08
sxiv-rifle: add webp and performance improvement

N-R-K:
Hi,

I have been using the sxiv-rifle script as a wrapper around sxiv for a
long while now. There's a couple changes that can be made to improve the
performance of it. On my system I was able to achieve a 25% performance
increase, benchmarked with time (0.005 vs 0.004).

First of all, inside listfiles() we don't need to call `dirname`. Since
we already have the absolute path, we can just use parameter expansion.

Secondly, we don't need to call printf inside abspath(), we can directly
set the value to `"$target"`. In fact we can go one step further and
offload no arguments checking and abspath() into one case statement.

P.S. It should be noted that although I've been using this script, I
however don't use ranger myself. I do assume that this should be fine
since the logic of the script is mostly the same.

\- NRK

-- The above has been copy pasted from an email I sent to hut at the address found at [HACKING.md](https://github.com/ranger/ranger/blob/master/HACKING.md#patches). In reply I was told that this person is no longer working on ranger and either to make a pull request on github or use the mailing list instead. Maybe [HACKING.md](https://github.com/ranger/ranger/blob/master/HACKING.md#patches) should update the email address to the mailing list?

#### ISSUE TYPE

#### CHECKLIST

``make test_shellcheck`` passes.
    - [ ] Tests have been updated

### DIFF
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -19,9 +19,11 @@
 # not work in dash and others), so we cannot store the result of listfiles to a
 # variable.
 
+tmp="/tmp/sxiv_rifle_$$"
+
 listfiles () {
     find -L "///${target%/*}" -maxdepth 1 -type f -iregex \
-      '.*\.\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print0 | sort -z
+      '.*\.\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print | sort | tee "$tmp"
 }
 
 is_img () {
@@ -39,10 +41,11 @@ case "$1" in
     *)  target="$PWD/$1" ;;
 esac
 
-is_img "$target" && count="$(listfiles | grep -m 1 -ZznF "$target")"
+trap "rm -f $tmp" EXIT
+is_img "$target" && count="$(listfiles | grep -nF "$target")"
 
 if [ -n "$count" ]; then
-    listfiles | xargs -0 sxiv -n "${count%%:*}" --
+    sxiv -i -n "${count%%:*}" -- < "$tmp"
 else
     sxiv -- "$@" # fallback
 fi
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -19,10 +19,9 @@
 # 'realpath' because it would fork a whole process, which is slow.
 #
 # Second, we need to append a file list to sxiv, which can only be done
-# properly in three ways: arrays (which are not POSIX) or \0 sperated strings.
-# Unfortunately, assigning \0 to a variable is not POSIX either (will not work
-# in dash and others), so we cannot store the result of listfiles to a
-# variable.
+# properly in three ways: arrays (which are not POSIX).
+# \0 separated strings; but assigning \0 to a variable is not POSIX either
+# so we cannot store the result of listfiles to a variable.
 #
 # The third approach is to store the result to a tmpfile and using `-i` to feed
 # the list to sxiv. This is the fastest approach since we won't have to call
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -14,8 +14,10 @@
 tmp="/tmp/sxiv_rifle_$$"
 
 listfiles () {
-    find -L "///${1%/*}" -maxdepth 1 -type f -iregex \
-      '.*\.\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print | sort | tee "$tmp"
+    find -L "///${1%/*}" \( ! -path "///${1%/*}" -prune \) -type f \
+      \( -name '*.jpg' -o -name '*.jpeg' -o -name '*.png' -o -name '*.gif' \
+      -o -name '*.webp' -o -name '*.tiff' -o -name '*.bmp' \) -print |
+      sort | tee "$tmp"
 }
 
 is_img () {
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -31,10 +31,10 @@ case "$1" in
     *)  target="$PWD/$1" ;;
 esac
 
-[ -f "$target" ] && count="$(listfiles | grep -m 1 -ZznF "$target" | cut -d: -f1)"
+[ -f "$target" ] && count="$(listfiles | grep -m 1 -ZznF "$target")"
 
 if [ -n "$count" ]; then
-    listfiles | xargs -0 sxiv -n "$count" --
+    listfiles | xargs -0 sxiv -n "${count%%:*}" --
 else
     sxiv -- "$@" # fallback
 fi
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -21,7 +21,7 @@
 
 listfiles () {
     find -L "///${target%/*}" -maxdepth 1 -type f -iregex \
-      '.*\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print0 | sort -z
+      '.*\.\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print0 | sort -z
 }
 
 is_img () {
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -14,6 +14,20 @@
 # shell syntax and calls being made to external utilities, such as grep or find.
 # This makes it portable across many unix like systems, although it may not be
 # the cleanest or fastest approach.
+#
+# First, using case statement to get absolute path is quicker than calling
+# 'realpath' because it would fork a whole process, which is slow.
+#
+# Second, we need to append a file list to sxiv, which can only be done
+# properly in three ways: arrays (which are not POSIX) or \0 sperated strings.
+# Unfortunately, assigning \0 to a variable is not POSIX either (will not work
+# in dash and others), so we cannot store the result of listfiles to a
+# variable.
+#
+# The third approach is to store the result to a tmpfile and using `-i` to feed
+# the list to sxiv. This is the fastest approach since we won't have to call
+# listfiles twice.
+
 
 tmp="/tmp/sxiv_rifle_$$"
 
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -31,7 +31,7 @@ case "$1" in
     *)  target="$PWD/$1" ;;
 esac
 
-count="$(listfiles | grep -m 1 -ZznF "$target" | cut -d: -f1)"
+[ -f "$target" ] && count="$(listfiles | grep -m 1 -ZznF "$target" | cut -d: -f1)"
 
 if [ -n "$count" ]; then
     listfiles | xargs -0 sxiv -n "$count" --
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -14,9 +14,8 @@
 tmp="/tmp/sxiv_rifle_$$"
 
 listfiles () {
-    find -L "///${1%/*}" \( ! -path "///${1%/*}" -prune \) -type f \
-      \( -iname '*.jpg' -o -iname '*.jpeg' -o -iname '*.png' -o -iname '*.gif' \
-      -o -iname '*.webp' -o -iname '*.tiff' -o -iname '*.bmp' \) -print |
+    find -L "///${1%/*}" \( ! -path "///${1%/*}" -prune \) -type f |
+      grep -iE '\.(jpe?g|png|gif|webp|tiff|bmp)$' |
       sort | tee "$tmp"
 }
 
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -11,7 +11,7 @@
 #   mime ^image, has sxiv, X, flag f = path/to/this/script -- "$@"
 #
 # Implementation notes: this script is quite slow because of POSIX limitations
-# and portability concerns. First calling the shell function 'abspath' is
+# and portability concerns. First, using case statement to get absolute path is
 # quicker than calling 'realpath' because it would fork a whole process, which
 # is slow. Second, we need to append a file list to sxiv, which can only be done
 # properly in two ways: arrays (which are not POSIX) or \0 sperated
@@ -19,26 +19,18 @@
 # not work in dash and others), so we cannot store the result of listfiles to a
 # variable.
 
-if [ $# -eq 0 ]; then
-    echo "Usage: ${0##*/} PICTURES"
-    exit
-fi
-
-[ "$1" = '--' ] && shift
-
-abspath () {
-    case "$1" in
-        /*) printf "%s\n" "$1";;
-        *)  printf "%s\n" "$PWD/$1";;
-    esac
-}
-
 listfiles () {
-    find -L "$(dirname "$target")" -maxdepth 1 -type f -iregex \
+    find -L "${target%/*}" -maxdepth 1 -type f -iregex \
       '.*\(jpe?g\|bmp\|png\|gif\)$' -print0 | sort -z
 }
 
-target="$(abspath "$1")"
+[ "$1" = '--' ] && shift
+case "$1" in
+    "") echo "Usage: ${0##*/} PICTURES" >/dev/stderr && exit ;;
+    /*) target="$1" ;;
+    *)  target="$PWD/$1" ;;
+esac
+
 count="$(listfiles | grep -m 1 -ZznF "$target" | cut -d: -f1)"
 
 if [ -n "$count" ]; then
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -20,7 +20,7 @@
 # variable.
 
 listfiles () {
-    find -L "${target%/*}" -maxdepth 1 -type f -iregex \
+    find -L "///${target%/*}" -maxdepth 1 -type f -iregex \
       '.*\(jpe?g\|bmp\|png\|gif\|webp\)$' -print0 | sort -z
 }
 
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -14,7 +14,7 @@
 tmp="/tmp/sxiv_rifle_$$"
 
 listfiles () {
-    find -L "///${1%/*}" \( ! -path "///${1%/*}" -prune \) -type f |
+    find -L "///${1%/*}" \( ! -path "///${1%/*}" -prune \) -type f -print |
       grep -iE '\.(jpe?g|png|gif|webp|tiff|bmp)$' |
       sort | tee "$tmp"
 }
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -10,6 +10,10 @@
 #
 #   mime ^image, has sxiv, X, flag f = path/to/this/script -- "$@"
 #
+# Implementation note: the script tries to be POSIX compliant both in terms of
+# shell syntax and calls being made to external utilities, such as grep or find.
+# This makes it portable across many unix like systems, although it may not be
+# the cleanest or fastest approach.
 
 tmp="/tmp/sxiv_rifle_$$"
 
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -21,7 +21,7 @@
 
 listfiles () {
     find -L "${target%/*}" -maxdepth 1 -type f -iregex \
-      '.*\(jpe?g\|bmp\|png\|gif\)$' -print0 | sort -z
+      '.*\(jpe?g\|bmp\|png\|gif\|webp\)$' -print0 | sort -z
 }
 
 [ "$1" = '--' ] && shift
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -35,6 +35,7 @@ is_img () {
 case "$1" in
     "") echo "Usage: ${0##*/} PICTURES" >/dev/stderr && exit ;;
     /*) target="$1" ;;
+    "~"/*) target="$HOME/${1#"~"/}" ;;
     *)  target="$PWD/$1" ;;
 esac
 
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -21,12 +21,12 @@
 
 listfiles () {
     find -L "///${target%/*}" -maxdepth 1 -type f -iregex \
-      '.*\(jpe?g\|png\|gif\|webp\|bmp\)$' -print0 | sort -z
+      '.*\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print0 | sort -z
 }
 
 is_img () {
     case "${1##*.}" in
-        "jpg"|"jpeg"|"png"|"gif"|"webp"|"bmp") return 0 ;;
+        "jpg"|"jpeg"|"png"|"gif"|"webp"|"tiff"|"bmp") return 0 ;;
         *) return 1 ;;
     esac
 }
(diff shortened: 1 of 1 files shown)


examples/rifle_sxiv.sh
```diff
@@ -11,34 +11,26 @@
...
-    case "$1" in
-        /*) printf "%s\n" "$1";;
-        *)  printf "%s\n" "$PWD/$1";;
-    esac
-}
-
 listfiles () {
-    find -L "$(dirname "$target")" -maxdepth 1 -type f -iregex \
-      '.*\(jpe?g\|bmp\|png\|gif\)$' -print0 | sort -z
+    find -L "${target%/*}" -maxdepth 1 -type f -iregex \
```
toonn, line 23: I think there's one edge-case here. If `target=/blah` then this'll expand to the empty string. Maybe `/${target%/*}`?

toonn (changes requested):
Apologies for not getting around to this sooner, a bit short on time.

I like consolidating the usage example in the one case statement. There is an edge-case we should handle though. There's actually two, if `target=blah` the results between the expansion and dirname differ as well. But the case statement should make that impossible. At least I don't think `PWD` is relative rather than absolute in any sensible scenario.

examples/rifle_sxiv.sh
N-R-K: Is having an extra leading `/` valid? If so then we can do what you suggested. Otherwise we can do this
```sh
dir="${target%/*}"
find -L "${dir:-/}"
```

toonn: That was another thing I thought of. What if the user passes a directory *including* the trailing slash? `some/blah/`, for example, what's the desired behavior here?

I think a double slash is pretty universal, if triple is actually POSIX I guess it's better but I like the trailing slash idea fine too.

Oh, and another thing, doesn't the expansion result in the wrong result for `foo/bar/..`?

N-R-K: >  What if the user passes a directory _including_ the trailing slash? `some/blah/`, for example, what's the desired behavior here?

On which case?
- If we use three leading slash then the trailing slash will be removed by the parameter expansion and everything works as expected. `////some/blah`
- If we use a single trailing slash then the parameter expansion will remove the trailing slash, but then the hard-coded one will be added in leading to everything working as expected as well.



> Oh, and another thing, doesn't the expansion result in the wrong result for `foo/bar/..`?

I don't understand this. Are you talking about `$target` not being an absolute path or about user inputting literal `..` ?

The first one is not exactly possible since the case statement will prevent that. And if it's a literal `..` then nothing changes from the previous implementation.

```sh
count="$(listfiles | grep -m 1 -ZznF "$target" | cut -d: -f1)"
```
Grep will try to find the `$target` and fail. `$count` will be empty leading to the fallback getting executed.
```sh
if [ -n "$count" ]; then
    listfiles | xargs -0 sxiv -n "$count" --
else
    sxiv -- "$@" # fallback
fi
```

Note: Sxiv can deal with trailing `..` so everything ends up working out as expected and no differently than the previous script.

toonn: Not in any case with leading or trailing slashes. `foo/bar` results in `foo`, `foo/bar/` results in `foo/bar`, don't we want these almost identical paths to have the same result?

The relativeness of the path isn't the point. `foo/bar/..` is expanded to `foo/bar` but surely the user intends `foo/bar/..` to mean the same thing as `foo`? We don't really care about feature parity with the previous implementation if that was already wrong.

N-R-K: > `foo/bar/..` is expanded to `foo/bar` but surely the user intends `foo/bar/..` to mean the same thing as `foo`?

That's exactly what happens though. The grep will fail, ``$count`` will be empty and the fallback will be executed, i.e ``sxiv -- foo/bar/..`` and sxiv will open the directory `foo` just as expected.

N-R-K: https://github.com/ranger/ranger/pull/2411/commits/c47c0f3216fddedfb8099adbb98eece41f8ec040 should solve the edge case of file being in root. And I've tested out all the other possibilities you've listed and they all work as expected.

If user doesn't enter an ***image file*** (or more precisely a file with one of the extension ``find`` is listing out) as argument then grep would always fail and the fallback will be executed. If it's a directory (even with `..` in it, either trailing or in middle) then sxiv will open it as expected.

toonn: Doesn't the find look *in* the stripped dir?

N-R-K: > Doesn't the find look _in_ the stripped dir?

Yes, it lists out all the images in that dir. And then that list gets grepped for `"$target"`, if grep doesn't find anything then `$count` is not set and fallback is executed.

One edge case I can think of is if user enters ``foo/bar/../bar`` in this case the dir `bar` should be opened, but if the dir `foo` contains any image file starting with `bar`, grep will take that as a match and open up `foo/bar.png` instead of `foo/bar` the dir.

toonn: Not sure if this is practical to fix without calling something like "realpath." Since handling of `..` in the presence of symlinks differs on Windows, for example. Which I *think* is also supposedly POSIX.

N-R-K: The main problem here is that the grep command is not very strict.

This causes another interesting problem. If you enter `foo/b`, listfiles will list out all the images inside `foo`, and if there exists an image starting with `b` lets say `bar.png`, grep will take that as a match and open it up.

```diff
-grep -m 1 -ZznF "$target"
+grep -m 1 -Zzne "$target$"
```
This solves the issue, but I'm not sure what will happen if `$target` contains special chars.

Also speaking of symlinks and POSIX, I recently tried to find out a POSIX complaint way to resolve a symlink for a script of mine since `realpath` and `readlink` aren't POSIX. And at the end the most _""sane""_ way I found was parsing the output of ls.
```sh
resolved="$(ls -dl "${1}")"
echo "${resolved#*"$1" -> }"
```

toonn: Just checking in to see whether you'd like to work on this further to deal with the edge cases? If not I'm fine with merging this as-is since the current behavior is retained.

N-R-K: I don't mind, but I'm unable to think of a sane way to deal with this. Changing grep from fixed string to regex possibly opens up more bugs due to special chars in filename.

N-R-K:
Fixed another issue where `listfiles` is looking at the ending of the filename in general and not file extension specifically.

toonn:
Are there any changes left to make?

examples/rifle_sxiv.sh
```diff
@@ -11,38 +11,38 @@
...
+case "$1" in
+    "") echo "Usage: ${0##*/} PICTURES" >/dev/stderr && exit ;;
+    /*) target="$1" ;;
+    "~"/*) target="$HOME/${1#"~"/}" ;;
+    *)  target="$PWD/$1" ;;
+esac
 
-target="$(abspath "$1")"
-count="$(listfiles | grep -m 1 -ZznF "$target" | cut -d: -f1)"
+is_img "$target" && count="$(listfiles | grep -m 1 -ZznF "$target")"
```
toonn, line 42: Likewise, we could add the much faster ripgrep and relegate grep to being a fallback:

```suggestion
is_img "$target" && count="$(listfiles | rg -Fnm1 --null-data "$target" || grep -Fnzm1 "$target")"
```

I believe `-Z` is redundant in the presence of `-z`. BSD grep doesn't support the `-z/--null-data` option unfortunately so this wouldn't work with that grep.

toonn (commented):
I haven't tested these changes with the script. Just in a terminal to make sure but this means I may have overlooked something. Would be cool if you could test them.

examples/rifle_sxiv.sh
```diff
@@ -11,38 +11,38 @@
...
 
-if [ $# -eq 0 ]; then
-    echo "Usage: ${0##*/} PICTURES"
-    exit
-fi
-
-[ "$1" = '--' ] && shift
+listfiles () {
+    find -L "///${target%/*}" -maxdepth 1 -type f -iregex \
+      '.*\.\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print0 | sort -z
```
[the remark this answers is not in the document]
N-R-K: > My find(1) on a mac defaults to BREs which don't support `?` and `|`. Maybe we should just enable EREs?

I'm on gnu find and `-LE` gives an error.
```sh
$ find: unknown predicate `-LE'
```
Also how much do we care about posix compatibility? Because `-print0` isn't posix. OpenBSD man page does list `-print0` as an option however.

Also sxiv is strictly X, can X11 software be run on a mac? If not then we shouldn't care about it.

> maybe we should add `fd`

I don't personally have it (or ripgrep) on my system but I'm fine adding support for it. Although I would prefer using variables for the command instead of `||`.
Also if it's faster, shouldn't we try to use it as default and fallback to find instead?

examples/rifle_sxiv.sh
N-R-K: The reason we needed `-z` was so we can feed it to `sxiv` sanely. We can add an argument to `listfiles` to print or not print null so grep wouldn't need `-z`.

N-R-K:
> Are there any changes left to make?

I was done. But I did recently made a small change to my personal script where instead of calling `listfiles` twice we just store the result on a tmp file with tee. Measurable difference `0.004` vs `0.003`.

```diff
-      '.*\.\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print0 | sort -z
+      '.*\.\(jpe?g\|png\|gif\|webp\|tiff\|bmp\)$' -print0 | sort -z | tee "$tmp"

if [ -n "$count" ]; then
-    listfiles | xargs -0 sxiv -n "${count%%:*}" --
+    xargs -0 sxiv -n "${count%%:*}" < "$tmp"
```

If we want to remove `-z` from `grep` then I don't think this will work anymore. And adding `fd`, `ripgrep` might make things more difficult.

examples/rifle_sxiv.sh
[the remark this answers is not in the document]
N-R-K: On the topic on posix, I don't think `xargs -0` is posix either.

Also... if we go with the tee option described in https://github.com/ranger/ranger/pull/2411#issuecomment-917865113 , we may not need null separator at all. Sxiv has an `-i` flag to read filenames from stdin. 

```diff
-    xargs -0 sxiv $args -n "${count%%:*}" < "$tmp" ||
+    sxiv $args -n "${count%%:*}" -i < "$tmp" ||
```
It seems to work fine on names with spaces as well.

N-R-K:
The entire script is now posix compliant. I don't believe we're using any non-posix flags anymore. Performance also increased down to `0.002`.

examples/rifle_sxiv.sh
```diff
@@ -10,39 +10,36 @@
...
     esac
 }
 
-listfiles () {
-    find -L "$(dirname "$target")" -maxdepth 1 -type f -iregex \
-      '.*\(jpe?g\|bmp\|png\|gif\)$' -print0 | sort -z
+open_img () {
+    is_img "$1" || exit 1
+    trap 'rm -f $tmp' EXIT
+    count="$(listfiles "$1" | grep -nF "$1")"
```
toonn, line 31: Hmm, won't dropping `-m1` cause way more of a potential slowdown?
If it's not POSIX I vote for doing the same thing I did with ripgrep.

```suggestion
    count="$(listfiles "$1" | grep -Fnm1 "$1" || grep -nF "$1")"
```

N-R-K: > Hmm, won't dropping -m1 cause way more of a potential slowdown?

No, because unix pipes execute in parallel. This might be a valid concern while reading from a file.

- - -

Also that syntax doesn't really work. What's going to happen is this
```sh
cmd1 | cmd2 || cmd3
# is equal to
{ cmd1 | cmd2 } || cmd3
```

What we want is this
```sh
cmd1 | { cmd2 || cmd3 }
```
And even then this wont really work, because if cmd2 fails, `count` will store the failure/error msg as well.

If we want to add fallbacks we'll have to do it through storing the command in a variable. The `||` trick won't work.

toonn: Doesn't the `-m1` still allow grep to short-circuit even though input is coming from stdin?

N-R-K: Not really. Try this, `seq 50 | grep --invalid-flag || grep 1`.
First it'll print the error from invalid flag and then it'll just hang because the 2nd grep command didn't receive any input.

N-R-K: Also, this would be the worst case scenario. 
```sh
[nrk ranger]% time seq 9000 | grep 1 >/dev/null
seq 9000  0.00s user 0.00s system 89% cpu 0.001 total
grep --color=auto 1 > /dev/null  0.00s user 0.00s system 81% cpu 0.001 total
[nrk ranger]% time seq 9000 | grep -m 1 1 >/dev/null
seq 9000  0.00s user 0.00s system 82% cpu 0.001 total
grep --color=auto -m 1 1 > /dev/null  0.00s user 0.00s system 82% cpu 0.001 total
```
I'm sending output to null since otherwise printing on the terminal becomes the bottleneck.

q3cpma:
> The entire script is now posix compliant. I don't believe we're using any non-posix flags anymore.
-iregex isn't POSIX, though.

N-R-K:
> On the whole handling of newlines in filenames

Sxiv can't handle files with newlines in it anyways.

examples/rifle_sxiv.sh
toonn: The invalid flag output doesn't matter much. We could use the same solution as for ripgrep.

Fwiw, on my system I do see a 5x difference, more due to the number of matches though, seemingly. And ripgrep may not be worth the trouble, seems to have a slightly higher upfront execution cost and this script tends to run it on small input anyway.

fd might still prove worth the trouble.

N-R-K:
> or a sequence of `-iname '*.ext1' -o -iname '*.ext2' ...` which has the advantage of only matching on the name, not the full path, but isn't POSIX either

@q3cpma `-name` is posix and seems to work. Although not using `-maxdepth 1` I get a slight performance penalty, back to `0.003`.

q3cpma:
On Mon, Sep 13, 2021 at 10:11:51AM -0700, N-R-K wrote:
> > or a sequence of `-iname '*.ext1' -o -iname '*.ext2' ...` which has the advantage of only matching on the name, not the full path, but isn't POSIX either
>
> @q3cpma `-name` is posix and seems to work. Although not using `-maxdepth 1` I get a slight performance penalty, back to `0.003`.
>

But won't work on file.PNG. Honestly, this is sh (and even more, POSIX sh), you shouldn't care about a ms or two.

examples/rifle_sxiv.sh
N-R-K: @toonn Sorry for leaving this hanging, was busy with a couple other things. So we're not going with ripgrep?

And for fd, I'm not familiar with it, but we stopped using null separated list. So assuming the first command was proper, simply changing `--print0` to `--print` should suffice, right?

```sh
fd --max-depth=1 --type f --ignore-case --absolute-path --print \
         -e jpg -e jpeg -e png -e webp -e tiff -e bmp "///${target%/*}" \
```

toonn: No, just drop `--print0` altogether.

Dropping the null-separated paths is slightly unfortunate, since it'd allow generalizing the script for other things that *do* handle newlines fine. Not holding back the PR over that though.

N-R-K: One last thing, does the directory need to be at then end of the args, or can it be at the beginning similar to `find` ?

toonn: It can be before the flags but it has to be preceded by a pattern, use `.` for a pattern that matches anything. Feel free not to add it unless you can test the performance though. If there's no perceptible difference there's no need to complicate matters.
IME `fd` is a lot faster than `find` though.

N-R-K: > Feel free not to add it unless you can test the performance though.

In that case let's drop it.

N-R-K:
I think this is ready to be merged now.

toonn:
@q3cpma, are you a user of the script? Would be great to get some testing from another user : )

q3cpma:
On Sat, Sep 18, 2021 at 10:13:10AM -0700, toonn wrote:
> @q3cpma, are you a user of the script? Would be great to get some testing from another user : )
>
Seems to work fine, but on my machine at least,
  find -L "///${1%/*}" \( ! -path "///${1%/*}" -prune \) -type f | \
  	   	  grep -iE '\.(jpe?g|png|gif|webp|tiff|bmp)$' | tee "$tmp"

is a lot faster (0.027 -> 0.005 s), not to mention simpler and actually POSIX
compliant, unlike `-iname`.


PS: by the way, what's the point of that script? ranger already does it via
https://github.com/ranger/ranger/blob/master/ranger/core/fm.py#L148 is it only
for Debian or RHEL users with stale packages?

N-R-K:
> is a lot faster (0.027 -> 0.005 s), not to mention simpler and actually POSIX

Opening a folder with 270~ images on an ssd, I cannot see any performance improvement on my system. Maybe that's due to my system being faster so the difference might be more meaningful on slower system or on slower disks (hdd).

Regardless, since this approach is posix, and there doesn't seem to be any downside at least, I think I'll change it.

As for the purpose of this script, I use it as a cli wrapper around sxiv. As I've mentioned in the OP, I don't even use ranger.

N-R-K:
Btw, @toonn [HACKING.md](https://github.com/ranger/ranger/blob/master/HACKING.md#patches) still points to hut's email. Who apparently doesn't work on ranger anymore. Shouldn't that be changed to the mailing list?

q3cpma:
On Sat, Sep 18, 2021 at 10:52:53AM -0700, N-R-K wrote:
> > is a lot faster (0.027 -> 0.005 s), not to mention simpler and actually POSIX
>
> Opening a folder with 270~ images on an ssd, I cannot see any performance improvement on my system. Maybe that's due to my system being faster so the difference might be more meaningful on slower system or on slower disks (hdd).
>
I see, I had ~4000 on a XFS partition over an LVM pool on two HDDs.

examples/rifle_sxiv.sh
```diff
@@ -10,39 +10,38 @@
 #
 #   mime ^image, has sxiv, X, flag f = path/to/this/script -- "$@"
 #
-# Implementation notes: this script is quite slow because of POSIX limitations
```
toonn, line 13: I wouldn't remove the comment altogether tbh. Explaining why the script is the way it is would still be good because I could definitely see someone going "Hey, this is way simpler!" in a PR and no one may remember why we didn't use arrays or stored null separated strings in variables.

toonn (commented):
LGTM except for the comment. Explaining the reasoning we came to would be great.

examples/rifle_sxiv.sh
N-R-K: I'm not sure if I would call the script "slow" anymore. How about "For portability reasons the script tries to be POSIX compliant"

As for arrays, I actually tried that. Running it through `Bash` it was actually slower than the POSIX version being run through `Dash`.

toonn: You don't have to call it slow, rewrite the entire comment if you want.

That's not quite what the comment is saying though. Properly is about doing it correctly not fast. I think it's just explaining how you can't do without the double `listfiles` call. I think we eliminated one of the calls in this PR? Explaining why this was possible and how doing it the other way will cause the problem and require two calls because arrays and null separated strings in vars is not POSIX.

N-R-K: Pushed https://github.com/ranger/ranger/pull/2411/commits/0d783e17a461da481a4803e3cf970a15f1500bf5 , is this okay by you?

N-R-K: I did not see your message when sending mine. Should I still add notes about the array/null-seperator?

toonn: Not really what I had in mind. That can be as short as "POSIX compliant."
What I really want is technical details that'll make me think twice before merging a change that'll completely reverse all this work.

toonn: Oh, one more thing, did you run shellcheck on the script, always a nice sanity check?

N-R-K: I see where you're coming from. https://github.com/ranger/ranger/pull/2411/commits/d728452b0ffdb56de3ae642b848477a48f290740 should address it.

N-R-K: No, but I just did. And it's all good. :+1:

examples/rifle_sxiv.sh
```diff
@@ -10,39 +10,55 @@
...
+# the list to sxiv. This is the fastest approach since we won't have to call
+# listfiles twice.
 
-if [ $# -eq 0 ]; then
-    echo "Usage: ${0##*/} PICTURES"
-    exit
-fi
 
-[ "$1" = '--' ] && shift
+tmp="/tmp/sxiv_rifle_$$"
```
toonn, line 32: Maybe we should use `mktemp`? I'm not sure whether `tee` might append if we do happen to get a repeated pid. And it'd also avoid hard-coding `/tmp`.

toonn: Oh, and the "Three ways, '1' or '2'. '2' isn't posix. The third is..." reads a bit confusing tbh.

N-R-K: I don't believe `mktemp` is posix.
Is repeated pid possible?
What should be done then if we want to avoid hardcoding `/tmp`

toonn: Repeated pids are unlikely but possible, yes.

I think this is a case where support is widespread enough that it's still in the spirit of POSIX. Maybe using `TMPDIR` instead of `/tmp` would be enough though.

N-R-K: On my system at least `$TMPDIR` doesn't exist. Falling back to `/tmp` should be saner I guess.

N-R-K: Cleaned up the comment to be more natural https://github.com/ranger/ranger/pull/2411/commits/72a0cfefcfb349627f0d278cb1bc295f0a242bbc.

Should we deal with repeated pid? One way I can think of is running a loop checking if `$tmp` exists and keep appending more stuff to `$tmp` until we get something that doesn't already exist. Do you have any cleaner suggestion?

toonn: Nah, unlikely anyway. And if `tee` overwrites rather than appends it's a non-issue anyway.

N-R-K: Then we should be good here.

toonn: We should still use `${TMPDIR:-/tmp}` or something though.

N-R-K: I've done that already.
https://github.com/N-R-K/ranger/pull/2411/commits/4451a87a2bf9b95afd566ddcaaa7ba548d1ec8dc

toonn (approved):
Guess Github was just lagging.

toonn:
And, merged! 🎉 Thanks for hanging in there!

q3cpma:
On Sat, Sep 18, 2021 at 12:41:50PM -0700, N-R-K wrote:
> @N-R-K commented on this pull request.
>
>
>
> >
> -[ "$1" = '--' ] && shift
> +tmp="/tmp/sxiv_rifle_$$"
>
> I don't believe `mktemp` is posix.

It isn't, but
	echo "mkstemp(${TMPDIR:-/tmp}/tmp.XXXXXX)" | m4
is

toonn:
While true I think it requires a recent version of m4? Mine is `GNU M4 1.4.6`, for example.

q3cpma:
On Sat, Sep 18, 2021 at 02:24:17PM -0700, toonn wrote:
> While true I think it requires a recent version of m4? Mine is `GNU M4 1.4.6`, for example.
Well, it depends on what you mean by recent, it was introduced with 1.4.8 in 2006.

PR #2419, opened 2021-08-15
Make tab command names consistent with vim, closes #2412

### LINKED ISSUE #2412: Tab command names are not consistent with vim ones
#### Runtime Environment

- Terminal emulator and version: N/A
- Python version: 3.9.6
- Locale: `en_US`

#### Current Behavior

Tab commands are: `:tab_new`, `:tab_close`, etc - **with underscores**.

#### Expected Behavior

Ranger should be more consistent with vim commands: `:tabnew`, `:tabclose`, etc - **without underscores**.

Why? Because ranger shortcuts and commands are heavily based on vim ones. This inconsistency makes it harder to work with ranger out-of-the-box.

Workaround: Users can currently define their own aliases in their ranger config in order to mitigate this.

#### How to resolve this

A few options:

- Add aliases e.g. so that `tabnew=tab_new`, and so on.
- Add built-in commands e.g. so that a `tabnew` function just proxies to `tab_new`.

thiagowfx:
tabnew proxies to tab_new, tabclose proxies to tab_close, and so on.

Both forms (with and without underscore) remain available, the
underscore being the canonical ones.

See https://github.com/ranger/ranger/issues/2412

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1245,6 +1245,9 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.signal_emit('tab.change', old=previous_tab, new=self.thistab)
             self.signal_emit('tab.layoutchange')
 
+    def tabopen(self, *args, **kwargs):
+        return self.tab_open(*args, **kwargs)
+
     def tab_close(self, name=None):
         if name is None:
             name = self.current_tab
@@ -1259,6 +1262,9 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             del self.tabs[name]
         self.restorable_tabs.append(tab)
         self.signal_emit('tab.layoutchange')
+
+    def tabclose(self, *args, **kwargs):
+        return self.tab_close(*args, **kwargs)
 
     def tab_restore(self):
         # NOTE: The name of the tab is not restored.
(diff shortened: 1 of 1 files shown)


Ablesius (approved):
Does what it promises.

Ablesius:
Unit tests for these changes would be a good addition to the PR.

thiagowfx:
Thanks for your review @Ablesius. Follow-ups:

1. Could you advise what type of tests you're looking for? I thought this change was too trivial to warrant a test, but happy to include one if you'd still like it anyway.

2. Could you also advise how/where to add them? Looking into `tests/*`, it's not entirely obvious to me where those tests would be added. Currently there's no `test_tabs.py`.

Ablesius:
@thiagowfx you are right, the changes are trivial. Maybe I also was a bit too brief, I didn't mean that UTs would be *necessary*, just a nice touch. More like "hey, if you want to add a nice something to this PR, it would be UTs".
As for your second question, @toonn or @vifon should be able to answer this - it might well be there are no tests in that regard yet.

Ablesius (approved):
I've tested all the commands by hand, and they seem to work just fine. Thanks!

toonn:
Tests would be nice, the project is seriously lacking in this regard but it's not easy because the results to check are in part graphical.

A further contribution that would improve these new commands is documenting them in the man page. We might even completely replace the underscored commands.

PR #2420, opened 2021-08-15
Show Special Permissions

nash-pillai:
Show setuid, setgid, and sticky bits in permissions string.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch Linux 5.13.10-arch1-1
- Terminal emulator and version: Alacritty 0.9.0
- Python version: 3.9.6
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

The setuid bit is represented by `s` in the owner execute position.
If replacing a `-` it is uppercase, otherwise it is lowercase.
The setgid bit is the same as setuid, except in the group position.
The sticky bit is the same too, but in others position and represented by `t`.

#### MOTIVATION AND CONTEXT

So I can see what special permissions are present.

#### TESTING

I ran `make test`.

#### OTHER IDEAS
`-` and `x` are a lot more visually distinct than `x` and `s`.
Should/Could special permissions be in a different color or highlighted in some way.


nash-pillai:
@toonn  Please review when you can

Also, why is shellcheck always "Waiting for status to be reported"?

toonn (changes requested):
I think rather than update this logic we should just use `stat.filemode` if it is available.

The Shellcheck check only runs when it needs to. GitHub allows specifying "required" Actions, which means the check must *pass*. If a check isn't run it never passes. Annoying design decision with ugly workarounds. Feel free to ignore it.

toonn:
I've opened a PR with an alternative implementation of the same idea, getting better permission display. I didn't want to push to the master branch of your personal repository and risk messing with your setup. It took me a long time to get around to taking a look at this so I didn't want to impose and ask you to make these changes.

If you're still interested in this change, could you take a look at #2742, see if it fixes your problem adequately?

nash-pillai:
I tried #2742 and it appears to have the same functionality so I'll close this pr

PR #2431, opened 2021-08-25
w3m_offset

### LINKED ISSUE #2430: w3m_offset moves images to the right
#### Runtime Environment

- Operating system and version: Debian 11, 32bit, MATE
- Terminal emulator and version: mate-terminal, xterm
- ranger version: ranger-master v1.9.3-304-g236b8f6a and from the distributions version 1.9.3
- Python version: 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
- Locale: en_US.UTF-8

#### Current Behavior

w3m_offset also moves images to the right by the same distance as lowering them. Some wide images are partly pushed outside the window.

#### Expected Behavior

I was expecting that they only move down, or even also to the left.

#### Context

I'm using tabs and big fonts, so that require an important offset.

#### Possible Solutions

either they only go down, either have a separate offset for the x and y axis

#### Steps to reproduce

Just put a very large offset in w3m_offset

pirate486743186:
yep, i ignored the template 

fixes #2430 w3m_offset moves images to the right
also, adjust maximum height so images aren't pushed too much down.

test with a large offset


pirate486743186:
@toonn

you can merge them.

I'm leaving the max in, for both w3m and ueberzug . When you go to fullscreen for example, they get misaligned again. They are other strange terminals that role down or what ever. The end user would need to control what he wants in a consistent way. Trying to guess the max is a bad idea.

toonn:
@pirate486743186, tbh that sounds like an argument in favor of calculating the maximum. Setting a single value would result in undesirable behavior because if you maximize your terminal you usually want the image previews to scale along with it.

pirate486743186:
That's not how it works. The borders/Menus don't scale. These fixed offsets are applied to the maximums that change dynamically.

In my example, mate-terminal seams to think the window goes under the panel or something like that. When windowed, it works properly( maximum reduced by the same offset as starty), maximized and fullscreen, it leaves a space underneath (inside the terminal), so it gets misaligned and needs extra adjustment. When maximized, without the panel, it gets aligned properly with the same offset as starty. Normally, this mess shouldn't be happening. Most likely, the gnome terminal behaves the same way.

I think it's better to leave it obviously simple and don't over engineer it. They are too many configurations and bugs. What if a terminal has an extra bar at the bottom?

PR #2435, opened 2021-08-27
fix w3m method in 32bit

### LINKED ISSUE #2423: w3m image preview, broken in 32bit debian 11
#### Runtime Environment

- Operating system and version: Debian 11, 32bit, MATE
- Terminal emulator and version: mate-terminal, xterm, urxvt
- ranger version: ranger-master v1.9.3-304-g236b8f6a and from the distributions version 1.9.3
- Python version: 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
- Locale: en_US.UTF-8

#### Current Behavior

the w3m method for image previews seams completely broken in all terminals in Debian 11.

In mate-terminal, if i move in and out of fullscreen fast enough, or move fast enough left and right inside ranger, i sometimes see the preview flicker for an instant and even more rarely it stays there until something else happens and goes away again. (this is difficult to reproduce, probably a race condition) This also shows, that the configuration must be correct.

it used to work in mate-terminal (a bit buggy), so it should still work.

in xterm it's nothing at all.

#### Expected Behavior

i should see the preview with just the old bugs.

#### Context

I was on ubuntu 18.04, where xterm worked fine, and mate-terminal was a bit buggy.

w3m currently works fine with images in xterm and mate-terminal (same old buggy behavior). It even very buggily works  in urxvt.

i don't think i see a duplicate of this issue.

#### Possible Solutions

No idea. Debian might fix it on it's own in the next 6 months. It just was officially released.

I installed, from the live image, that came with a bunch of input methods. It might be an input method doing something. Then again, w3m works fine....

As a work around, i use urxvt with it's own special method.

#### Steps to reproduce

I would expect that all Debian 11 installs are like this. Same thing with the dev version of ranger.

#### Traceback

If i used the debug feature correctly, the log doesn't say anything relevant.

pirate486743186:
fix #2423
This is the proposed solution. Moving the delay after self.process .
It needs some testing, i don't know if that breaks the "horizontal black bars" mitigation.

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -187,6 +187,9 @@ class W3MImageDisplayer(ImageDisplayer, FileManagerAware):
             self.initialize()
         input_gen = self._generate_w3m_input(path, start_x, start_y, width,
                                              height)
+        self.process.stdin.write(input_gen)
+        self.process.stdin.flush()
+        self.process.stdout.readline()
 
         # Mitigate the issue with the horizontal black bars when
         # selecting some images on some systems. 2 milliseconds seems
@@ -195,11 +198,10 @@ class W3MImageDisplayer(ImageDisplayer, FileManagerAware):
             from time import sleep
             sleep(self.fm.settings.w3m_delay)
 
-        self.process.stdin.write(input_gen)
-        self.process.stdin.flush()
-        self.process.stdout.readline()
+        #HACK workaround for w3mimgdisplay memory leak
         self.quit()
         self.is_initialized = False
+        #HACK
 
     def clear(self, start_x, start_y, width, height):
         if not self.is_initialized or self.process.poll() is not None:


toonn (approved):
Don't know whether this is going to break things but I guess we'll find out.

toonn:
Hmm, I haven't been able to test this. Looks like w3mimgdisplay doesn't work in XQuartz on macOS so we'll have to wait for someone to test it.

toonn:
Found someone to test it and no problems seemed to turn up, if they do we'll probably revert this because 32bit support is rather low priority compared to 64bit.

pirate486743186:
actually i suspect the work around for the memory leak broke the mitigation, and this actually fixes it. I think, earlier the sleep at the beginning allowed the earlier  rendering to finish, but with the work around the process gets killed before it has time to finish. Or something like that.

i meant if it broke the mitigation, not the preview.
If it did broke the mitigation, it needs massaging with sleeps at the right lines.
So no need to revert.

It's sensitive to timing and that can vary from platform to platform.

also, ranger is a terminal file manager, it should support potatoes.

PR #2457, opened 2021-09-22
FIx freezing when using genric file opener on mac

vatsu:
Without ```flag f```, ranger freezes. This commit solves the problem by adding the appropriated flag.

#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: MacOS 11.6
- Terminal emulator and version: Kitty 0.23.1
- Python version: 3.9.7
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

This PR adds the appropriated flag (  for the MacOS generic on the rifle config file.

#### MOTIVATION AND CONTEXT

Without the f flag, ranger freezes when opening a file using the generic opener on MacOS. This PR solves the problem.

#### TESTING

Only the existing tests were run. 
The change only affects the Rifle config.

### DIFF
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -276,8 +276,8 @@ label wallpaper, number 14, mime ^image, has feh, X = feh --bg-fill "$1"
 #-------------------------------------------
 # Generic file openers
 #-------------------------------------------
-label open, has xdg-open = xdg-open "$@"
-label open, has open     = open -- "$@"
+label open, has xdg-open     = xdg-open "$@"
+label open, has open, flag f = open -- "$@"
 
 # Define the editor for non-text files + pager as last action
               !mime ^text, !ext xml|json|csv|tex|py|pl|rb|rs|js|sh|php|dart  = ask


toonn:
This doesn't happen for me and I wouldn't expect it to. Afaik `open` always forks a new process? In what scenario, file and application, are you encountering this?

vatsu:
It happens with any file that reaches the generic file opener for mac. When opening pdfs, ranger does not freezes because there is the fork flag alreaddy on the rifle config already. 

I confirm that it happens on Kitty. On iTerm2, I could not reproduce the error.

Below, two instances of the described error happening, opening an image and a video.

[Error handling image](https://gfycat.com/terriblewastefulflyingsquirrel)

[Error handling video](https://gfycat.com/wavyfatalarawana)

toonn:
Hmm, could you try running open on those files from a shell? If you get a new prompt while the image is still open then forking cannot be the issue. It's weird that it solves the issue. Have you tried running ranger from the master branch, I remember an issue about freezing recently.

vatsu:
You're right. It works fine using the master branch. But then, we should remove the fork flag from the line 157, right?

toonn:
Yeah, that rule could be removed altogether.

PR #2466, opened 2021-10-09
SIXEL image previewer

### LINKED ISSUE #723: Implement image previews using Sixel

remi6397:
Fixes #723; Replaces #2234.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: Generic GNU/Linux; macOS
- Terminal emulator and version: GNOME Terminal (latest Git commit) + VTE from the latest Git commit; mlterm 3.9.2; iTerm2
- Python version: CPython 3.11.1 and 2.7.16
- Ranger version/commit: N/A
- Locale: N/A

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION

Implements an image preview module for SIXEL-capable terminals.
Properly clears preview area (VTE-ready).

#### MOTIVATION AND CONTEXT

#2234 provides a PoC implementation of a SIXEL image previewer.
It however did not dispose of expired images properly, which turns out to be an issue in some implementations (namely VTE).

SIXEL support is to be stabilized in the next release of VTE and will soon hit major Linux distributions such as Ubuntu, so it might be high time to finally get this done.

#### TESTING

This appears to have already garnered some users, for example at https://aur.archlinux.org/packages/ranger-sixel-git.

As of February 2023, lots of people have confirmed that this patch does indeed do the job in its current form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ For enhanced file previews (with `scope.sh`):
 
 * `img2txt` (from `caca-utils`) for ASCII-art image previews
 * `w3mimgdisplay`, `ueberzug`, `mpv`, `iTerm2`, `kitty`, `terminology` or `urxvt` for image previews
-* `convert` (from `imagemagick`) to auto-rotate images
+* `convert` (from `imagemagick`) to auto-rotate images and for image previews
 * `rsvg-convert` (from [`librsvg`](https://wiki.gnome.org/Projects/LibRsvg))
   for SVG previews
 * `ffmpeg`, or `ffmpegthumbnailer` for video thumbnails
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -16,15 +16,13 @@ import curses
 import errno
 import fcntl
 import os
-import shlex
 import struct
 import sys
 import warnings
 import json
 import mmap
-import shutil
 import threading
-from subprocess import Popen, PIPE, check_call, check_output, CalledProcessError
+from subprocess import Popen, PIPE, check_call, CalledProcessError
 from collections import defaultdict, namedtuple
 
 import termios
@@ -35,21 +33,14 @@ from tempfile import gettempdir, NamedTemporaryFile, TemporaryFile
 from ranger import PY3
 from ranger.core.shared import FileManagerAware, SettingsAware
 from ranger.ext.popen23 import Popen23
+from ranger.ext.which import which
 
 
-def _which(cmd):
-    if PY3:
-        return shutil.which(cmd)
-
-    try:
-        return check_output(["which", cmd])
-    except CalledProcessError:
-        return None
-
-
-if _which("magick"):
-    MAGICK_CONVERT_CMD_BASE = ("magick", "convert")
+if which("magick"):
+    # Magick >= 7
+    MAGICK_CONVERT_CMD_BASE = ("magick",)
 else:
+    # Magick < 7
     MAGICK_CONVERT_CMD_BASE = ("convert",)
 
 
@@ -478,7 +469,7 @@ class SixelImageDisplayer(ImageDisplayer, FileManagerAware):
             fit_height = font_height * height
 
             sixel_dithering = self.fm.settings.sixel_dithering
-            cached = TemporaryFile("w+")
+            cached = TemporaryFile("w+", prefix=path.replace(os.sep, "-"))
 
             environ = dict(os.environ)
             environ.setdefault("MAGICK_OCL_DEVICE", "true")
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -377,7 +377,8 @@ class ITerm2ImageDisplayer(ImageDisplayer, FileManagerAware):
         font_height = self.fm.settings.iterm2_font_height
 
         return image_fit_width(
-                width, height, max_cols, max_rows, font_width, font_height)
+            width, height, max_cols, max_rows, font_width, font_height
+        )
 
     @staticmethod
     def _encode_image_content(path):
@@ -454,7 +455,11 @@ class SixelImageDisplayer(ImageDisplayer, FileManagerAware):
 
     def _clear_cache(self, path):
         if os.path.exists(path):
-            self.cache = {ce: cd for ce, cd in self.cache.items() if ce.inode != os.stat(path).st_ino}
+            self.cache = {
+                ce: cd
+                for ce, cd in self.cache.items()
+                if ce.inode != os.stat(path).st_ino
+            }
 
     def _sixel_cache(self, path, width, height):
         stat = os.stat(path)
@@ -471,15 +476,20 @@ class SixelImageDisplayer(ImageDisplayer, FileManagerAware):
             environ = dict(os.environ)
             environ.setdefault("MAGICK_OCL_DEVICE", "true")
             try:
-                check_call([*MAGICK_CONVERT_CMD_BASE, path + "[0]",
-                            "-geometry", "{0}x{1}>"
-                            .format(fit_width, fit_height),
-                             "-dither", sixel_dithering,
-                             "sixel:-"],
-                           stdout=cached,
-                           stderr=DEVNULL,
-                           env=environ,
-                           )
+                check_call(
+                    [
+                        *MAGICK_CONVERT_CMD_BASE,
+                        path + "[0]",
+                        "-geometry",
+                        "{0}x{1}>".format(fit_width, fit_height),
+                        "-dither",
+                        sixel_dithering,
+                        "sixel:-",
+                    ],
+                    stdout=cached,
+                    stderr=DEVNULL,
+                    env=environ,
+                )
             except CalledProcessError:
                 raise ImageDisplayError("ImageMagick failed processing the SIXEL image")
             except FileNotFoundError:
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -32,7 +32,7 @@ from tempfile import gettempdir, NamedTemporaryFile, TemporaryFile
 
 from ranger import PY3
 from ranger.core.shared import FileManagerAware, SettingsAware
-from ranger.ext.popen23 import Popen23
+from ranger.ext.popen23 import Popen23, DEVNULL
 from ranger.ext.which import which
 
 
@@ -480,7 +480,7 @@ class SixelImageDisplayer(ImageDisplayer, FileManagerAware):
                              "-dither", sixel_dithering,
                              "sixel:-"],
                            stdout=cached,
-                           stderr=PIPE,
+                           stderr=DEVNULL,
                            env=environ,
                            )
             except CalledProcessError:
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -6,7 +6,7 @@
 """Interface for drawing images into the console
 
 This module provides functions to draw images in the terminal using supported
-implementations, which are currently w3m, iTerm2 and urxvt.
+implementations.
 """
 
 from __future__ import (absolute_import, division, print_function)
@@ -16,18 +16,20 @@ import curses
 import errno
 import fcntl
 import os
+import shlex
 import struct
 import sys
 import warnings
 import json
+import mmap
 import threading
-from subprocess import Popen, PIPE
-from collections import defaultdict
+from subprocess import Popen, PIPE, check_call, CalledProcessError
+from collections import defaultdict, namedtuple
 
 import termios
 from contextlib import contextmanager
 import codecs
-from tempfile import gettempdir, NamedTemporaryFile
+from tempfile import gettempdir, NamedTemporaryFile, TemporaryFile
 
 from ranger import PY3
 from ranger.core.shared import FileManagerAware, SettingsAware
@@ -64,6 +66,47 @@ def move_cur(to_y, to_x):
     # on python2 stdout is already in binary mode, in python3 is accessed via buffer
     bin_stdout = getattr(sys.stdout, 'buffer', sys.stdout)
     bin_stdout.write(tparm)
+
+
+def get_terminal_size():
+    farg = struct.pack("HHHH", 0, 0, 0, 0)
+    fd_stdout = sys.stdout.fileno()
+    fretint = fcntl.ioctl(fd_stdout, termios.TIOCGWINSZ, farg)
+    return struct.unpack("HHHH", fretint)
+
+
+def get_font_dimensions():
+    """
+    Get the height and width of a character displayed in the terminal in
+    pixels.
+    """
+    rows, cols, xpixels, ypixels = get_terminal_size()
+    return (xpixels // cols), (ypixels // rows)
+
+
+def image_fit_width(width, height, max_cols, max_rows, font_width=None, font_height=None):
+    if font_width is None or font_height is None:
+        font_width, font_height = get_font_dimensions()
+
+    max_width = font_width * max_cols
+    max_height = font_height * max_rows
+    if height > max_height:
+        if width > max_width:
+            width_scale = max_width / width
+            height_scale = max_height / height
+            min_scale = min(width_scale, height_scale)
+            max_scale = max(width_scale, height_scale)
+            if width * max_scale <= max_width and height * max_scale <= max_height:
+                return width * max_scale
+            return width * min_scale
+
+        scale = max_height / height
+        return width * scale
+    elif width > max_width:
+        scale = max_width / width
+        return width * scale
+
+    return width
 
 
 class ImageDisplayError(Exception):
@@ -324,25 +367,11 @@ class ITerm2ImageDisplayer(ImageDisplayer, FileManagerAware):
         return text
 
     def _fit_width(self, width, height, max_cols, max_rows):
-        max_width = self.fm.settings.iterm2_font_width * max_cols
-        max_height = self.fm.settings.iterm2_font_height * max_rows
-        if height > max_height:
-            if width > max_width:
-                width_scale = max_width / width
-                height_scale = max_height / height
-                min_scale = min(width_scale, height_scale)
-                max_scale = max(width_scale, height_scale)
-                if width * max_scale <= max_width and height * max_scale <= max_height:
-                    return width * max_scale
-                return width * min_scale
+        font_width = self.fm.settings.iterm2_font_width
+        font_height = self.fm.settings.iterm2_font_height
 
-            scale = max_height / height
-            return width * scale
-        elif width > max_width:
-            scale = max_width / width
-            return width * scale
-
-        return width
+        return image_fit_width(
+                width, height, max_cols, max_rows, font_width, font_height)
 
     @staticmethod
     def _encode_image_content(path):
@@ -401,6 +430,87 @@ class ITerm2ImageDisplayer(ImageDisplayer, FileManagerAware):
             else:
                 return 0, 0
         return width, height
+
+
+_CacheableSixelImage = namedtuple("_CacheableSixelImage", ("width", "height", "inode"))
+
+_CachedSixelImage = namedtuple("_CachedSixelImage", ("image", "fh"))
+
+
+@register_image_displayer("sixel")
+class SixelImageDisplayer(ImageDisplayer, FileManagerAware):
+    """Implementation of ImageDisplayer using SIXEL."""
+
+    def __init__(self):
+        self.win = None
+        self.cache = {}
+        self.fm.signal_bind('preview.cleared', lambda signal: self._clear_cache(signal.path))
+
+    def _clear_cache(self, path):
+        if os.path.exists(path):
+            self.cache = {ce: cd for ce, cd in self.cache.items() if ce.inode != os.stat(path).st_ino}
+
+    def _sixel_cache(self, path, width, height):
+        stat = os.stat(path)
+        cacheable = _CacheableSixelImage(width, height, stat.st_ino)
+
+        if cacheable not in self.cache:
+            font_width, font_height = get_font_dimensions()
+            fit_width = font_width * width
+            fit_height = font_height * height
+
+            sixel_dithering = self.fm.settings.sixel_dithering
+            cached = TemporaryFile("w+")
+
+            try:
+                check_call(["convert", path + "[0]",
+                            "-geometry", "{0}x{1}>"
+                            .format(fit_width, fit_height),
+                             "-dither", sixel_dithering,
+                             "sixel:-"],
+                           stdout=cached,
+                           stderr=PIPE)
+            except CalledProcessError:
+                raise ImageDisplayError("ImageMagick failed processing the SIXEL image")
+            except FileNotFoundError:
+                raise ImageDisplayError("SIXEL image previews require ImageMagick")
+            finally:
+                cached.flush()
+
+            if os.fstat(cached.fileno()).st_size == 0:
+                raise ImageDisplayError("ImageMagick produced an empty SIXEL image file")
+
+            self.cache[cacheable] = _CachedSixelImage(mmap.mmap(cached.fileno(), 0), cached)
+
+        return self.cache[cacheable].image
+
+    def draw(self, path, start_x, start_y, width, height):
+        if self.win is None:
+            self.win = self.fm.ui.win.subwin(height, width, start_y, start_x)
+        else:
+            self.win.mvwin(start_y, start_x)
+            self.win.resize(height, width)
+
+        with temporarily_moved_cursor(start_y, start_x):
+            sixel = self._sixel_cache(path, width, height)[:]
+            if PY3:
+                sys.stdout.buffer.write(sixel)
+            else:
+                sys.stdout.write(sixel)
+            sys.stdout.flush()
+
+    def clear(self, start_x, start_y, width, height):
+        if self.win is not None:
+            self.win.clear()
+            self.win.refresh()
+
+            self.win = None
+
+        self.fm.ui.win.redrawwin()
+
+    def quit(self):
+        self.clear(0, 0, 0, 0)
+        self.cache = {}
 
 
 @register_image_displayer("terminology")
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -466,7 +466,7 @@ class SixelImageDisplayer(ImageDisplayer, FileManagerAware):
             fit_height = font_height * height
 
             sixel_dithering = self.fm.settings.sixel_dithering
-            cached = TemporaryFile("w+", prefix=path.replace(os.sep, "-"))
+            cached = TemporaryFile("w+", prefix="ranger", suffix=path.replace(os.sep, "-"))
 
             environ = dict(os.environ)
             environ.setdefault("MAGICK_OCL_DEVICE", "true")
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -22,8 +22,9 @@ import sys
 import warnings
 import json
 import mmap
+import shutil
 import threading
-from subprocess import Popen, PIPE, check_call, CalledProcessError
+from subprocess import Popen, PIPE, check_call, check_output, CalledProcessError
 from collections import defaultdict, namedtuple
 
 import termios
@@ -34,6 +35,23 @@ from tempfile import gettempdir, NamedTemporaryFile, TemporaryFile
 from ranger import PY3
 from ranger.core.shared import FileManagerAware, SettingsAware
 from ranger.ext.popen23 import Popen23
+
+
+def _which(cmd):
+    if PY3:
+        return shutil.which(cmd)
+
+    try:
+        return check_output(["which", cmd])
+    except CalledProcessError:
+        return None
+
+
+if _which("magick"):
+    MAGICK_CONVERT_CMD_BASE = ("magick", "convert")
+else:
+    MAGICK_CONVERT_CMD_BASE = ("convert",)
+
 
 W3MIMGDISPLAY_ENV = "W3MIMGDISPLAY_PATH"
 W3MIMGDISPLAY_OPTIONS = []
@@ -462,14 +480,18 @@ class SixelImageDisplayer(ImageDisplayer, FileManagerAware):
             sixel_dithering = self.fm.settings.sixel_dithering
             cached = TemporaryFile("w+")
 
+            environ = dict(os.environ)
+            environ.setdefault("MAGICK_OCL_DEVICE", "true")
             try:
-                check_call(["convert", path + "[0]",
+                check_call([*MAGICK_CONVERT_CMD_BASE, path + "[0]",
                             "-geometry", "{0}x{1}>"
                             .format(fit_width, fit_height),
                              "-dither", sixel_dithering,
                              "sixel:-"],
                            stdout=cached,
-                           stderr=PIPE)
+                           stderr=PIPE,
+                           env=environ,
+                           )
             except CalledProcessError:
                 raise ImageDisplayError("ImageMagick failed processing the SIXEL image")
             except FileNotFoundError:
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -104,18 +104,15 @@ def image_fit_width(width, height, max_cols, max_rows, font_width=None, font_hei
             width_scale = max_width / width
             height_scale = max_height / height
             min_scale = min(width_scale, height_scale)
-            max_scale = max(width_scale, height_scale)
-            if width * max_scale <= max_width and height * max_scale <= max_height:
-                return width * max_scale
             return width * min_scale
-
-        scale = max_height / height
-        return width * scale
+        else:
+            scale = max_height / height
+            return width * scale
     elif width > max_width:
         scale = max_width / width
         return width * scale
-
-    return width
+    else:
+        return width
 
 
 class ImageDisplayError(Exception):
--- a/ranger/ext/popen23.py
+++ b/ranger/ext/popen23.py
@@ -14,9 +14,11 @@ except ImportError:
     PY3 = version_info[0] >= 3
 
 if PY3:
+    # pylint: disable=ungrouped-imports,unused-import
     from subprocess import DEVNULL
 else:
     import os
+    # pylint: disable=consider-using-with
     DEVNULL = open(os.devnull, "wb")
 
 
--- a/ranger/ext/popen23.py
+++ b/ranger/ext/popen23.py
@@ -13,6 +13,12 @@ except ImportError:
     from sys import version_info
     PY3 = version_info[0] >= 3
 
+if PY3:
+    from subprocess import DEVNULL
+else:
+    import os
+    DEVNULL = open(os.devnull, "wb")
+
 
 # COMPAT: Python 2 (and Python <=3.2) subprocess.Popen objects aren't
 #         context managers. We don't care about early Python 3 but we do want
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -309,7 +309,7 @@ are automatically used when available but completely optional.
 \&\f(CW\*(C`w3mimgdisplay\*(C'\fR, \f(CW\*(C`ueberzug\*(C'\fR, \f(CW\*(C`mpv\*(C'\fR, \f(CW\*(C`iTerm2\*(C'\fR, \f(CW\*(C`kitty\*(C'\fR, \f(CW\*(C`terminology\*(C'\fR or
 \&\f(CW\*(C`urxvt\*(C'\fR for image previews
 .IP "\-" 2
-\&\f(CW\*(C`convert\*(C'\fR (from \f(CW\*(C`imagemagick\*(C'\fR) to auto-rotate images
+\&\f(CW\*(C`convert\*(C'\fR (from \f(CW\*(C`imagemagick\*(C'\fR) to auto-rotate images and for image previews
 .IP "\-" 2
 \&\f(CW\*(C`rsvg\-convert\*(C'\fR (from \f(CW\*(C`librsvg\*(C'\fR) for \s-1SVG\s0 previews
 .IP "\-" 2
@@ -367,6 +367,19 @@ To enable this feature, set the option \f(CW\*(C`preview_images_method\*(C'\fR t
 This feature relies on the dimensions of the terminal's font.  By default, a
 width of 8 and height of 11 are used.  To use other values, set the options
 \&\f(CW\*(C`iterm2_font_width\*(C'\fR and \f(CW\*(C`iterm2_font_height\*(C'\fR to the desired values.
+.PP
+\fIsixel\fR
+.IX Subsection "sixel"
+.PP
+\&\s-1SIXEL\s0 is the industry standard for displaying images in terminals.
+.PP
+Most mainstream terminal emulators include \s-1SIXEL\s0 support, for example:
+programs based on \s-1VTE\s0 >=0.68 (eg. \s-1GNOME\s0 Terminal), iTerm2, Mintty, etc.
+.PP
+To enable this feature, install the program \f(CW\*(C`convert\*(C'\fR (from \f(CW\*(C`imagemagick\*(C'\fR)
+and set the option \f(CW\*(C`preview_images_method\*(C'\fR to sixel.
+.PP
+Preferred dithering method can be specified by setting \f(CW\*(C`sixel_dithering\*(C'\fR.
 .PP
 \fIkitty\fR
 .IX Subsection "kitty"
@@ -1231,6 +1244,17 @@ Increase it in case of experiencing display corruption.
 Offset in pixels for the inner border of the terminal. Some terminals require
 the offset to be specified explicitly, among others st and UXterm, some don't
 like urxvt.
+.IP "sixel_dithering [string]" 4
+.IX Item "sixel_dithering [string]"
+One of dithering methods supported by \f(CW\*(C`convert\*(C'\fR, used for \s-1SIXEL\s0 image previews.
+.Sp
+\&\f(CW\*(C`convert\*(C'\fR currently supports the following dithering methods:
+.Sp
+.Vb 3
+\& None             No dithering
+\& Riemersma        Dithering along a Hilbert curve with restricted error proliferation
+\& FloydSteinberg   Error diffusion dithering
+.Ve
 .IP "wrap_plaintext_previews [bool]" 4
 .IX Item "wrap_plaintext_previews [bool]"
 Whether or not to wrap long lines in the pager, this includes previews of plain
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -133,7 +133,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.9.3" "2023-03-10" "ranger manual"
+.TH RANGER 1 "ranger-1.9.3" "2023-04-05" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
 .\" way too many mistakes in technical documents.
 .if n .ad l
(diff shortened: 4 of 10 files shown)


remi6397:
By the way, the preview should actually be drawn *below* the primary UI, but that doesn't seem to be the case here, does it?

remi6397:
I mean underneath, or more precisely below.

toonn:
More user reports are always better. I'll have to find time to go through the code before merging.

connection-reset:
Works fine for me. Arch Linux with ImageMagick 7.1.0-14, foot 1.9.2.

With `-dither ...` it produces way better results out of the box and is not perceptibly slower on my desktop computer.

Would you be open to making the whole command configurable?
[Chafa 1.8.0](https://hpjansson.org/chafa/) is nearly a third faster than ImageMagick with about the same picture quality, 0.229 seconds vs 0.339.
I ran `time chafa /tmp/Antarctic_mist.jpg -f sixel --dither fs -s 25x25` resp. `time convert /tmp/Antarctic_mist.jpg -geometry 200x200 -dither FloydSteinberg sixel:-` 3x and averaged the result (note that chafa uses columns and rows instead of pixels).
But a new enough version might not be packaged for all supported distros (sixel is supported since chafa 1.4.0).

remi6397:
> Would you be open to making the whole command configurable?

How about a whole separate "user-defined command" previewer?

connection-reset:
Yes, long-term a better idea. Maybe I'll try my hand at that.

ranger/ext/img_display.py
```diff
@@ -388,6 +416,52 @@ def _get_image_dimensions(path):
...
+        sixel_extra_args = shlex.split(self.fm.settings.sixel_extra_args)
+        result = check_output(["convert", path + "[0]",
+                               "-geometry", "{0}x{1}>"
+                               .format(fit_width, fit_height)]
+                              + sixel_extra_args
+                              + ["sixel:-"],
+                              stderr=PIPE)
+
+        with temporarily_moved_cursor(start_y, start_x):
+            if PY3:
```
tinywrkb, line 446: This now needs `from ranger import PY3`.

remi6397: What do you mean? It's already imported at line 34…

tinywrkb: Right, my error, it's in master. I was checking against stable. 🤦

README.md
```diff
@@ -89,7 +89,7 @@ For enhanced file previews (with `scope.sh`):
 
 * `img2txt` (from `caca-utils`) for ASCII-art image previews
 * `w3mimgdisplay`, `ueberzug`, `mpv`, `iTerm2`, `kitty`, `terminology` or `urxvt` for image previews
-* `convert` (from `imagemagick`) to auto-rotate images and for SVG previews
+* `convert` (from `imagemagick`) to auto-rotate images and for image previews
```
toonn, line 92: This was intentionally SVG so I'm not sure why this change is in this PR?
It's a happy coincidence though because I recently merged a PR replacing it with rsvg-convert.
So let's leave it in.

ranger/ext/img_display.py
```diff
@@ -67,6 +68,47 @@ def move_cur(to_y, to_x):
...
+
+    max_width = font_width * max_cols
+    max_height = font_height * max_rows
+    if height > max_height:
+        if width > max_width:
+            width_scale = max_width / width
+            height_scale = max_height / height
+            min_scale = min(width_scale, height_scale)
+            max_scale = max(width_scale, height_scale)
+            if width * max_scale <= max_width and height * max_scale <= max_height:
```
toonn, line 99: 👍 on refactoring this into a common function.

I'm having trouble figuring out if the logic actually makes sense though. Paraphrased:
```
height < max_height and width < max_width -> don't scale
height < max_height and width > max_width -> scale to fit width
height > max_height and width < max_width -> scale width so height fits
height > max_height and width > max_width -> scale so the entire image fits
```
Surely in that last case we should always scale with `min_scale`? Checking both scales only makes sense to me if we don't know whether both dimensions exceed the maximum.

ranger/ext/img_display.py
```diff
@@ -17,12 +17,13 @@
...
 import imghdr
 import os
+import shlex
 import struct
 import sys
 import warnings
 import json
 import threading
-from subprocess import Popen, PIPE
+from subprocess import Popen, PIPE, check_output
```
toonn, line 26: Can we get `check_output` from `ranger.ext.spawn`? Then we don't need to give up Python 2.6 compatibility.

ranger/ext/img_display.py
```diff
@@ -388,6 +416,52 @@ def _get_image_dimensions(path):
...
+            self.win = self.fm.ui.win.subwin(height, width, start_y, start_x)
+        else:
+            self.win.mvwin(start_y, start_x)
+            self.win.resize(height, width)
+
+        font_width, font_height = get_font_dimensions()
+        fit_width = font_width * width
+        fit_height = font_height * height
+
+        sixel_extra_args = shlex.split(self.fm.settings.sixel_extra_args)
```
toonn, line 437: I'm not a huge fan of this extra_args mechanism tbh. (I'm keeping #2497 in mind here.)
I don't really see a reason to privilege `convert` as the main SIXEL previewer and adding customizability but not quite the functionality from #2497 strikes me as odd.

I think the `sixel` method should either be completely configurable, like the `user` method you're implementing in the other PR pretty much or it shouldn't be at all customizable.

I'm perfectly fine with the latter option since we can add a fully customizable previewer in addition.
However, that still leaves two approaches. Either we pick one SIXEL previewer that we deem best, which can very well be ImageMagick's `convert`, or we mimic the approach that is common across ranger to simply attempt to find something that works. The method would try to find the scripts/binaries for each of the SIXEL previewers we know of and use the first one it encounters in order from rare to common, since if someone has a rare program installed, it's probably because that's their preference.

Let me know what you think?
And *thank you* for working on this. SIXEL previews are a long-awaited feature for many ranger users. 🤗

toonn (commented):
Finally getting around to reviewing this. Apologies for the very long wait. I will do my best to be responsive until we get this merged.

ranger/ext/img_display.py
remi6397: Hi @toonn,

Thank you for responding and sorry for the delay from my side, but here I am.

Just for the record, I actually switched to #2497 and am personally happier with it.

Regarding the extra args mechanism, I opted for it because depending on what kind of
images you have, dithering may or may not be preferred. I now removed the extra_args mechanism
and replaced it with a setting for the preferred dithering method.

ranger/ext/img_display.py
remi6397: Yup!

remi6397:
> This saves the sixel image on disk

Ah, alright, scratch that, I assumed that was an LRU memory cache.

markus-bauer:
I looked at it just now and it caches the return value of the decorated function. You could rewrite the displayer to work with that, I think. The cache would be in memory though (If I understand it correctly).

markus-bauer:
Final note:
This keeps one cached sixel image per image path + preview dimension. If you change the size, for example when opening in the pager, it will remake the sixel image, and if you go back, it will remake it again.
So, this could be extended to have several unique cached images on disk for different sizes.

markus-bauer:
Good job. It works much better than before and is obviously way better than my version.
Questions:
- Why does _CachedSixelImage store cached, and not just the memory mapped file?
- I don't thing this handle the case where the image file might change or get replaced with a different file of the same name? All the cached sixels for that image should get deleted/replaced in that case.

remi6397:
> * Why does _CachedSixelImage store cached, and not just the memory mapped file?

It's not, we just hold on to the open fd as long as we need it. The actual contents are not stored in memory ~~unless I made some stupid mistake I don't see~~.

> * I don't thing this handle the case where the image file might change or get replaced with a different file of the same name? All the cached sixels for that image should get deleted/replaced in that case.

For now, the best thing I can come up with is to hash the entries with mtime. They will linger there up until they get collected though.

markus-bauer:
> It's not, we just hold on to the open fd as long as we need it. The actual contents are not stored in memory unless I made some stupid mistake I don't see.

The question was more about if we need the handle? Or is the file handle just stored to keep the object alive?

Anyway, I hope I don't appear to be nitpicking. This is meant to be constructive. These are also just questions or opinions; feel free to ignore or disagree.

1) Why so defensive about the size of the cache (`_LRU_CACHE_CAPACITY = 32`)? In my basic version, the cache just stored file paths to the sixel images on disk. Are there memory considerations about keeping mmap objects stored?

2) There's an unhandled error case: While testing, I made a mistake that caused the convert command to fail. This led to `cached` being empty, and caused this error:
``` 
  File ".../ranger/ext/img_display.py", line 461, in 
    self.cache[cachable] = _CachedSixelImage(mmap.mmap(cached.fileno(), 0), cached)
ValueError: cannot mmap an empty file
```

3) I thought of a better way to clear the cache: Clearing of previews when the file changes is already handled in `fsobject.load_if_outdated()`/`fsobject.load()`:

https://github.com/ranger/ranger/blob/89d3683e42f9d220e7e888edc5d82bdb29f423c3/ranger/container/fsobject.py#L365-L375

https://github.com/ranger/ranger/blob/89d3683e42f9d220e7e888edc5d82bdb29f423c3/ranger/container/fsobject.py#L274-L286

https://github.com/ranger/ranger/blob/89d3683e42f9d220e7e888edc5d82bdb29f423c3/ranger/core/actions.py#L1027-L1030

I think the clearing of the sixel cache should be tied to this. Perhaps with signals?

```diff
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1027,6 +1027,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
     def update_preview(self, path):
         try:
             del self.previews[path]
+            self.signal_emit('preview.cleared', path=path)
         except KeyError:
             return False
         self.ui.need_redraw = True
diff --git a/ranger/ext/img_display.py b/ranger/ext/img_display.py
index 8c9243e9..a6adfd5f 100644
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -431,6 +431,11 @@ class SixelImageDisplayer(ImageDisplayer, FileManagerAware):
     def __init__(self):
         self.win = None
         self.cache = {}
+        self.fm.signal_bind('preview.cleared', lambda signal: self._clear_cache(signal.path))
+
+    def _clear_cache(self, path):
+        # TODO: clear cache for path

     def _sixel_cache(self, path, width, height):
         stat = os.stat(path)
```
To delete all the items for that path you might need to use a nested dict (`cache[path][other criteria ...]`). Or just search for items that have that path.

remi6397:
> > It's not, we just hold on to the open fd as long as we need it. The actual contents are not stored in memory unless I made some stupid mistake I don't see.
> 
> The question was more about if we need the handle? Or is the file handle just stored to keep the object alive?

I think both…? I mean, we need to ensure that the object exists and that it is open.

> Anyway, I hope I don't appear to be nitpicking.

No, not at all, on the contrary, you have a very fair point.

> \1. Why so defensive about the size of the cache (`_LRU_CACHE_CAPACITY = 32`)? In my basic version, the cache just stored file paths to the sixel images on disk. Are there memory considerations about keeping mmap objects stored?

No, there is no _significant_ cost; should be equally low regardless how small or large the image is.

> \2. There's an unhandled error case: While testing, I made a mistake that caused the convert command to fail. This led to `cached` being empty, and caused this error:
> 
> 
> ```
>   File ".../ranger/ext/img_display.py", line 461, in 
>     self.cache[cachable] = _CachedSixelImage(mmap.mmap(cached.fileno(), 0), cached)
> ValueError: cannot mmap an empty file
> ```

Yuck!

> \3. I thought of a better way to clear the cache: Clearing of previews when the file changes is already handled in `fsobject.load_if_outdated()`/`fsobject.load()`:

I really wanted to avoid adding more complexity to this and resort to some LRU cache, but since you're mentioning it, we might as well try that way.

markus-bauer:
The new version has a mistake, you forgot to add path:
```
AttributeError: '_CacheableSixelImage' object has no attribute 'path'
```

About connecting with fsobject: It was just an idea. (But I see, you already implemented it.)
It's about making sure that existing sixels get invalidated, if the preview gets invalidated/updated/cleared.
We have to avoid this: file changes -> `fm.previews[file]` gets cleared, but `sixel_cache[file]` still loads the old sixel. There are probably other ways to do this.

ranger/ext/img_display.py
```diff
@@ -421,6 +421,7 @@ def clear(self, start_x, start_y, width, height):
 
         self.fm.ui.win.redrawwin()
         self.fm.ui.win.refresh()
+        self.fm.ui.win.redrawwin()
```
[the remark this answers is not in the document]
markus-bauer: I know this is resolved and no longer current, but the original version might not have been a mistake.
(But I don't know why.)

https://github.com/ranger/ranger/blob/89d3683e42f9d220e7e888edc5d82bdb29f423c3/ranger/gui/ui.py#L347-L353

markus-bauer:
@remi6397: Oh no, I just realized we/I have might have made a mistake. I'm sorry. I don't have time to think about this right now, but I still want to mention it as quickly as possible:

The image displayer can be called with a path that points to a cacheimg. For example: scope.sh will produce thumbnails of a video file, which will be saved and shown with the displayer. It's okay to use the path of the thumbnail for the sixel cache, since we produce the sixels from those thumbnails. But for clearing the cache, we compare the path of the thumbnail image with `fsobj.path` of the video file.

gssfaw:
Thanks for you work, in the most cases it works great, but if you try to delete the picture you are currently previewing ranger ends up crashing.  If you delete it while not previewing it wont crash 

` python3 ranger.py --cmd="set preview_images true" --cmd="set preview_images_method sixel"
ranger version: ranger-master v1.9.3-510-g6739693f
Python version: 3.10.5 (main, Jun  6 2022, 18:49:26) [GCC 12.1.0]
Locale: en_US.UTF-8
Current file: '/home/wafssg/Pictures/20220801_21h55m27s_grim.png'

  File "/home/wafssg/Documents/clonesgit/ranger/ranger/core/main.py", line 204, in main
    fm.loop()
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/core/fm.py", line 416, in loop
    ui.redraw()
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/gui/ui.py", line 343, in redraw
    self.draw()
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/gui/ui.py", line 370, in draw
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/gui/displayable.py", line 268, in draw
    displayable.draw()
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/gui/widgets/view_miller.py", line 100, in draw
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/gui/displayable.py", line 268, in draw
    displayable.draw()
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/gui/widgets/browsercolumn.py", line 172, in draw
    self.need_redraw |= target.pointed_obj.load_if_outdated()
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/container/fsobject.py", line 375, in load_if_outdated
    self.load()
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/container/fsobject.py", line 286, in load
    self.fm.update_preview(self.path)
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/core/actions.py", line 1030, in update_preview
    self.signal_emit('preview.cleared', path=path)
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/ext/signals.py", line 268, in signal_emit
    fnc(signal)
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/ext/img_display.py", line 432, in <lambda>
    self.fm.signal_bind('preview.cleared', lambda signal: self._clear_cache(signal.path))
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/ext/img_display.py", line 435, in _clear_cache
    self.cache = {ce: cd for ce, cd in self.cache.items() if ce.inode != os.stat(path).st_ino}
  File "/home/wafssg/Documents/clonesgit/ranger/ranger/ext/img_display.py", line 435, in <dictcomp>
    self.cache = {ce: cd for ce, cd in self.cache.items() if ce.inode != os.stat(path).st_ino}
FileNotFoundError: [Errno 2] No such file or directory: '/home/wafssg/Pictures/20220801_21h55m27s_grim.png'

https://github.com/ranger/ranger/issues`

markus-bauer:
@gssfaw  See the last sentence in my previous post.

**This issue was left in a broken state.** 
The caching system was never finished and can't work as it is now.

remi6397:
@gssfaw Please check if the workaround added in [ea9ae88](https://github.com/ranger/ranger/pull/2466/commits/ea9ae8870e63fe52d2003ff49185dbc7cb2bcb42) works for you.

I'll try to look into the matter sometime soon when I take PTO.

gssfaw:
@remi6397 It works great, I haven't find cases where it does not work. I am not sure if it was already discussed but when you are deleting multiple files there is some problem with displaying the confirmation message. That happens with the other pr with user previews too. 
20220803_15h26m22s_grim

remi6397:
> when you are deleting multiple files there is some problem with displaying the confirmation message.

@gssfaw I cannot reproduce that in neither iTerm2 nor mlterm. What terminal emulator are you using?

image
image

gssfaw:
@remi6397 I am using foot terminal on sway. Tried mlterm and it has the same problem for some reason. I forgot to mention that it happens when I delete the files using dD command, if I am manually typing :delete it works as intended  
20220804_14h01m47s_grim
Actually I tried to do the same thing using kitty and their preview option and this happens on it too using dD. I'll ask some my friends to test it too, maybe it is only mine problem.

theshouldawoulda:
libsixel and pysixel provides a toggle for outputting with or without [DCS envelope](https://saitoha.github.io/libsixel/). For example, foot and a local gnu screen session:
```sh
$ img2sixel abc.png 
$ cat abc.png | sixelconv --body-only
```
foot + remote screen session works with the DCS envelope:

```sh
$ img2sixel -P abc.png 
$ cat abc.png | sixelconv 
``` 
Looks like its trivial to accomplish the same with ranger at first glance.

justinesmithies:
How close to getting merged is support for sixel previews ??

Martinits:
Hello, I'm using archlinux+sway+alacritty-sixel( https://github.com/microo8/alacritty-sixel )+ranger-sixel ( from AUR ). It shows broken previews like in the picture and flickers sometimes. But after refreshing the screen like screenshoting or switching to another workspace then back, the preview goes back to normal. Is this a known problem? The picture is taken by my phone because when screenshoting the preview id good like what I just said.
image

remi6397:
> alacritty-sixel( https://github.com/microo8/alacritty-sixel )

Hi @Martinits, could you please check if the same problem occurs for you with [mlterm](https://github.com/arakiken/mlterm)? If it works in mlterm, then the problem is most likely to originate from alacritty-sixel. GPU accelerated programs tend to break in unexpected ways, so it's always better to confirm that unexpected behaviors can be reproduced with software rendering. I don't use alacritty, but I do use iTerm2 which is also GPU accelerated and it renders ranger's sixel previews as intended.

Martinits:
> If it works in mlterm, then the problem is most likely to originate from alacritty-sixel.
Thanks, mlterm renders preview well, it seems to be an alacritty bug.

JosefLitos:
when I tried it, it gave me an error message: 
image
what am I missing?

chrhasse:
@JosefLitos I believe you need to install imagemagick, it provides a program called convert.

JosefLitos:
Yes indeed, thank you.
Compared to ueberzug this method is a lot slower, is it because of the way it is built - that you have to escape each pixel color sequence?

remi6397:
> Compared to ueberzug this method is a lot slower

I've never used [Überzug](https://github.com/seebye/ueberzug) (it appears it's X11 only?), but SIXEL (even using `convert(1)`) is on par with native iTerm2 and it gets much faster than iTerm2 once the image is cached, so I use SIXEL on my Mac.

It appears that Überzug also caches previews; are you sure you're using the latest revision of this PR, @JosefLitos?

JosefLitos:
@remi6397 I am using the ranger-sixel version from AUR, though it says Last Modified 2021, so I suppose if it doesn't always download from git that it actually is outdated.

chrhasse:
I think that package is super out of date and for a now defunct pull request implementing sixel. I was thinking about taking it over and updating it since it's orphaned, but I think since this PR can only be easily applied against master, it'll be more correct to make a new ranger-sixel-git and delete ranger-sixel. I'll probably get it on the aur tonight, but if you want to build something now, you can add this to the ranger-git pkgbuild

```
prepare() {
  cd "$srcdir/$pkgname"
  GIT_COMMITTER_EMAIL="user@mail.com" git pull -X theirs origin pull/2466/head --no-edit
}
```

remi6397:
> @remi6397 I am using the ranger-sixel version from AUR, though it says Last Modified 2021, so I suppose if it doesn't always download from git that it actually is outdated.

[This one?](https://aur.archlinux.org/packages/ranger-sixel)

It says there:

> Simple, vim-like file manager, with **PR #2234** for sixel image support

But this is PR #2466, so you are looking at a totally different thing…

I'm not really familiar with arch, but on Gentoo Linux you can drop user patches in `/etc/portage/patches` so you would normally do something like `curl --create-dirs -o /etc/portage/patches/app-misc/ranger/2466.patch https://github.com/ranger/ranger/pull/2466.patch` and you're all set. I assume there is a similar mechanism on Arch, since it is a source distribution too. You can also go the easy way and just install ranger in opt with this patch applied… 👀

justinesmithies:
> I think that package is super out of date and for a now defunct pull request implementing sixel. I was thinking about taking it over and updating it since it's orphaned, but I think since this PR can only be easily applied against master, it'll be more correct to make a new ranger-sixel-git and delete ranger-sixel. I'll probably get it on the aur tonight, but if you want to build something now, you can add this to the ranger-git pkgbuild
> 
> ```
> prepare() {
>   cd "$srcdir/$pkgname"
>   GIT_COMMITTER_EMAIL="user@mail.com" git pull -X theirs origin pull/2466/head --no-edit
> }
> ```

If you do go the route of a new ranger-sixel-git package tonight I'll definitely install and test it. Thanks in advance.

chrhasse:
@justinesmithies @JosefLitos [ranger-sixel-git](https://aur.archlinux.org/packages/ranger-sixel-git) is now in the AUR. Feel free to comment on the aur page if you have any trouble with my PKGBUILD.

Dzordzu:
@chrhasse package works as expected
@remi6397 there is an issue during redraw action. The image tends to disappear, but I must investigate when exactly. If I manage to find it I will inform you. Tested with alacritty and kitty

remi6397:
> @remi6397 there is an issue during redraw action. The image tends to disappear, but I must investigate when exactly. If I manage to find it I will inform you. Tested with alacritty and kitty

This patch is only for SIXEL support. Neither Alacritty nor Kitty supports SIXEL officially, I believe they use custom protocols instead. Also keep in mind those two are GPU-rendered terminal, so naturally some artifacts may occur that are not present elsewhere.

So please check if the issue exists on **MLTerm** and **iTerm2** (though make sure you're running in `sixel` mode and not `iterm2` mode!), both of which are incredibly stable and support the SIXEL protocol out of the box.

Dzordzu:
Nope. I've been using fork (to be merged) alacritty with libsixel support. When it comes to the kitty, I was sure that it supports sixel. I must recheck it. Apparently it did work for me (to some extent)

rodrgz:
I am using this PR to preview the images at the Foot terminal, it is working like a charm!

Zeioth:
Same. Tested on Foot terminal and works great. In fact, much better than kitty. This one doesn't break the GUI when a menu appears.

ranger/ext/img_display.py
Zeioth: So scaling is likely the point that impacts performance the most. Either using multi threading or GPU acceleration could noticeably impact usability (not in this PR, but long term).

Zeioth (approved):
All seems to work as intended. But performance should be improved.

Proposal: 
Storing the re-scaled images on ~/.cache instead of memory to improve performance.

akiva:
Very eager to see this merged

rodrgz (dismissed):

baldrrat:
merge waiting room

jstkdng:
If you can't wait for this to be merged you can use current ranger's ueberzug previewer with [ueberzugpp](https://github.com/jstkdng/ueberzugpp). It should work out of the box.

That being said, sometimes ranger's window refreshing causes some garbage to be print out on the terminal. I have already opened a PR to fix that. #2786

remi6397:
> If you can't wait for this to be merged you can use current ranger's ueberzug previewer with [ueberzugpp](https://github.com/jstkdng/ueberzugpp). It should work out of the box.
> 
> That being said, sometimes ranger's window refreshing causes some garbage to be print out on the terminal. I have already opened a PR to fix that. #2786

I appreciate the effort, but aren't we getting a bit hot-headed with that Überzug++ hype train? I don't think you could consider it nearly mature enough to be endorsed by any serious project.

Now, I'm not familiar with how Überzug works (worked?) and am genuinely curious why would Ü++ care about SIXEL at all? Does it need to draw some anchor symbols (I immediately thought of QR codes :P) to estimate the overlay window's desired position?

jstkdng:
>why would Ü++ care about SIXEL at all

because the way original ueberzug worked (X11 child windows) doesn't work in wayland with native wayland terminals.

>to estimate the overlay window's desired position?

ranger already provides all the necessary information to be able to draw the pictures

>I don't think you could consider it nearly mature enough to be endorsed by any serious project

how mature is mature enough? it works, it's fast and it doesn't break your computer. All projects have to start somewhere don't you think?

remi6397:
> >why would Ü++ care about SIXEL at all
> 
> because the way original ueberzug worked (X11 child windows) doesn't work in wayland with native wayland terminals.
>
> >to estimate the overlay window's desired position?
> 
> ranger already provides all the necessary information to be able to draw the pictures

In that case, why would I want to go through Ü++ instead of just outputting SIXELs directly? Doesn't it seem like an overkill to you?

Also, I briefly checked the repo yesterday and it looks like you're not caching anything? Are you sure you are comparing your impl's performance against mine and not the obsolete one?

remi6397:
> > I don't think you could consider it nearly mature enough to be endorsed by any serious project

(for clarification, that was a part of my comment, which I removed to keep the discussion on topic)

> how mature is mature enough? it works, it's fast and it doesn't break your computer. All projects have to start somewhere don't you think?

Once again, I artlessly appreciate your effort, but looking at your GitHub history, Überzug++ is just a project you wrote to dip your toes in C++ a month ago. My advice is that you should at least consider requesting a review from someone experienced in C++ before rushing in where angels fear to tread. :slightly_smiling_face:

jstkdng:
>In that case, why would I want to go through Ü++ instead of just outputting SIXELs directly?

Ü++ supports animated gifs and will support animated webps in the future, it also supports videos but that was not intended.

>you're not caching anything?

caching is also on the roadmap

>Are you sure you are comparing your impl's performance against mine and not the obsolete one?

I've not being able to install `ranger-sixel-git` and test due to some weird setuptools issue, but (opencv + opencl)/libvips should be faster than imagemagick when it comes to resizing.

>requesting a review from someone experienced in C++

I don't have the resources to ask some C++ pro to check/audit my code, but if anyone wants to they are welcome.

remi6397:
> > In that case, why would I want to go through Ü++ instead of just outputting SIXELs directly?
> 
> Ü++ supports animated gifs and will support animated webps in the future, it also supports videos but that was not intended.

Multi-page/frame stepping would better be implemented directly in Ranger, for portability. I might even do that in a separate PR soon.

> > Are you sure you are comparing your impl's performance against mine and not the obsolete one?
> 
> I've not being able to install `ranger-sixel-git` and test due to some weird setuptools issue, but (opencv + opencl)/libvips should be faster than imagemagick when it comes to resizing/image processing.

This is misinformation. Magick utilizes an OpenCL kernel when resizing images. See: https://imagemagick.org/script/opencl.php

remi6397:
> > requesting a review from someone experienced in C++
> 
> I don't have the resources to ask some C++ pro to check/audit my code, but if anyone wants to they are welcome.

Maybe ask folks at [codereview.stackexchange.com](https://codereview.stackexchange.com/)? That is an extremely helpful community and a priceless resource, especially when it comes to learning a new programming language. :grin:

jstkdng:
>Multi-page/frame stepping would better be implemented directly in Ranger, for portability

I'd find that hard without introducing new dependencies into ranger. But you are free to do that.

> This is misinformation. Magick utilizes an OpenCL kernel when resizing images

Indeed, but it has 1 requirement and 1 recommendation

image

You are using neither in this PR. Plain opencv used in Ü++ switches to opencl (cuda too perhaps, not sure) if supported automatically, switching to  libvips if the format is not supported by opencv.

jstkdng:
>[codereview.stackexchange.com](https://codereview.stackexchange.com/)

Ü++ could be too big for a single post there, I'll check the feasability of that.

remi6397:
> > Multi-page/frame stepping would better be implemented directly in Ranger, for portability
> 
> I'd find that hard without introducing new dependencies into ranger. But you are free to do that.

Like what specialized dependencies do you think one would need to step through frames of a GIF or pages of a PDF? I think you're too focused on selling your toy project instead of actually considering the circumstances.

> > This is misinformation. Magick utilizes an OpenCL kernel when resizing images
> 
> Indeed, but it has 1 requirement and 1 recommendation
> 
> image
> 
> You are using neither in this PR. Plain opencv used in Ü++ switches to opencl (cuda too perhaps, not sure) if supported automatically, switching to libvips if the format is not supported by opencv.

This is incorrect for ImageMagick 6 which is currently the most widely deployed version: https://legacy.imagemagick.org/script/opencl.php

Magick 7 is basically non-existent at this point, but it's not hard either to enable OpenCL by just `export`ing that, is it now?

> > [codereview.stackexchange.com](https://codereview.stackexchange.com/)
> 
> Ü++ could be too big for a single post there, I'll check the feasability of that.

I don't think it's unfit for there if you post only the meat of your code. :wink:

jstkdng:
>This is incorrect for ImageMagick 6 which is currently the most widely deployed version

If this is incorrect why did you link it in the first place?

>just exporting that, is it now?

then do it...?

>Like what specialized dependencies

PIL perhaps? unless you plan on doing that with externals tools and not in python, which is doable... could be harder though, slower too maybe?

>I think you're too focused on selling your toy project

why yes, I want people to use my "toy" project.

> instead of actually considering the circumstances.

The current circumstance is that your PR is still in waiting state even though it looks good imo. I thought maybe the ranger maintainers are fine with modifying something that already is in the codebase instead of adding completely new code. But it seems the ranger maintainers are busy (last commit was last year and PRs are not getting as much attention apparently).

remi6397:
> > This is incorrect for ImageMagick 6 which is currently the most widely deployed version
> 
> If this is incorrect why did you link it in the first place?

Mistake.

> > just exporting that, is it now?
> 
> then do it...?

Done, feel free to leave a review.

> > Like what specialized dependencies
> 
> PIL perhaps? unless you plan on doing that with externals tools and not in python, which is doable... could be harder though, slower too maybe?

How do you think ranger is currently handling images?

> > I think you're too focused on selling your toy project
> 
> why yes, I want people to use my "toy" project.

Then maybe actually polish it a bit instead of foisting it on people right in its earliest stages of development?

cherti:
> Then maybe actually polish it a bit instead of foisting it on people right in its earliest stages of development?

I think people are very capable of making a decision/testing a software on suitability for their purpose, I think there is no need to be overprotective here. :)

remi6397:
> I think people are very capable of making a decision/testing a software on suitability for their purpose, I think there is no need to be overprotective here. :)

Hasn't this thread had enough unsolicited advice today?

remi6397:
> I think people are very capable of making a decision/testing a software on suitability for their purpose, I think there is no need to be overprotective here. :)

Bear with me, I (and anyone really) would rather not have people advertising their [unrelated] stuff in my thread, regardless of quality.

cherti:
> And please, at least try to keep it on topic.

Indeed.

Dzordzu:
> Nope. I've been using fork (to be merged) alacritty with libsixel support. When it comes to the kitty, I was sure that it supports sixel. I must recheck it. Apparently it did work for me (to some extent)
> 
> I will check it for these. Thanks for your response :)

Ok. It seems that tmux is breaking all the things…

rodrgz:
> > Nope. I've been using fork (to be merged) alacritty with libsixel support. When it comes to the kitty, I was sure that it supports sixel. I must recheck it. Apparently it did work for me (to some extent)
> > I will check it for these. Thanks for your response :)
> 
> Ok. It seems that tmux is breaking all the things…

Tmux does not support sixel:
https://github.com/tmux/tmux/issues/44

ranger/ext/img_display.py
toonn: I'd still appreciate an answer to this question. When could the `max_scale` ever work?

ranger/ext/img_display.py
```diff
@@ -67,6 +87,47 @@ def move_cur(to_y, to_x):
...
+        if width > max_width:
+            width_scale = max_width / width
+            height_scale = max_height / height
+            min_scale = min(width_scale, height_scale)
+            max_scale = max(width_scale, height_scale)
+            if width * max_scale <= max_width and height * max_scale <= max_height:
+                return width * max_scale
+            return width * min_scale
+
+        scale = max_height / height
```
toonn, line 122: Let's put this case in an else clause, makes it clearer IMO.

ranger/ext/img_display.py
```diff
@@ -67,6 +87,47 @@ def move_cur(to_y, to_x):
     bin_stdout.write(tparm)
 
 
+def get_terminal_size():
+    farg = struct.pack("HHHH", 0, 0, 0, 0)
+    fd_stdout = sys.stdout.fileno()
+    fretint = fcntl.ioctl(fd_stdout, termios.TIOCGWINSZ, farg)
```
toonn, line 93: This will give us the dimensions of the ncurses subwindow, right?

toonn (changes requested):
Looks pretty good to me. Only minor nitpicks.

ranger/ext/img_display.py
```diff
@@ -17,23 +17,43 @@
...
 from ranger.core.shared import FileManagerAware, SettingsAware
 from ranger.ext.popen23 import Popen23
 
+
+def _which(cmd):
+    if PY3:
+        return shutil.which(cmd)
+
+    try:
+        return check_output(["which", cmd])
```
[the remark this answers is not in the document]
remi6397: I'm aware, but it looks to me like `command` is not a part of POSIX, am I right?

ranger/ext/img_display.py
```diff
@@ -17,23 +17,43 @@
...
-from subprocess import Popen, PIPE
-from collections import defaultdict
+from subprocess import Popen, PIPE, check_call, check_output, CalledProcessError
+from collections import defaultdict, namedtuple
 
 import termios
 from contextlib import contextmanager
 import codecs
-from tempfile import NamedTemporaryFile
+from tempfile import TemporaryFile, NamedTemporaryFile
```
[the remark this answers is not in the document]
remi6397: I think `TemporaryFile` is more appropriate for this use case, but we could set `dir=…` to address your concerns.

ranger/ext/img_display.py
remi6397: > **TIOCGWINSZ**
>
> Fill in the winsize structure pointed to by the third argument with the screen width and height.
> 
> The winsize structure is defined in `sys/ioctl.h' as follows:
> 
>      	
> 
>     struct winsize
>     {
>       unsigned short ws_row;	/* rows, in characters */
>       unsigned short ws_col;	/* columns, in characters */
>       unsigned short ws_xpixel;	/* horizontal size, pixels */
>       unsigned short ws_ypixel;	/* vertical size, pixels */
>     };

cain-dev:
This may be dumb to ask at this point (sorry if it is), but almost every other terminal file manager that i have used stops when selecting an image, is it possible to "cache" the thumbnails when entering a directory in a "seamless" way? It always look like the file manager just do that to the specific file on the selection OR at least have to stop everything for a preview.

Btw, thanks for the great work you guys are doing

ranger/ext/img_display.py
```diff
@@ -388,6 +435,91 @@ def _get_image_dimensions(path):
...
+            environ = dict(os.environ)
+            environ.setdefault("MAGICK_OCL_DEVICE", "true")
+            try:
+                check_call([*MAGICK_CONVERT_CMD_BASE, path + "[0]",
+                            "-geometry", "{0}x{1}>"
+                            .format(fit_width, fit_height),
+                             "-dither", sixel_dithering,
+                             "sixel:-"],
+                           stdout=cached,
+                           stderr=PIPE,
```
remi6397, line 477: https://docs.python.org/3/library/subprocess.html#subprocess.check_call

> **Note:** Do not use stdout=PIPE or stderr=PIPE with this function. The child process will block if it generates enough output to a pipe to fill up the OS pipe buffer as the pipes are not being read from.

ranger/ext/img_display.py
```diff
@@ -17,23 +17,43 @@
...
 import codecs
-from tempfile import NamedTemporaryFile
+from tempfile import TemporaryFile, NamedTemporaryFile
 
 from ranger import PY3
 from ranger.core.shared import FileManagerAware, SettingsAware
 from ranger.ext.popen23 import Popen23
 
+
+def _which(cmd):
```
remi6397, line 41: Oughta be moved to another package

remi6397 (commented):
Notes for self

remi6397:
> is it possible to "cache" the thumbnails when entering a directory in a "seamless" way?

That would be nice, but the directory could be something like an FTP mount (`curlftpfs` or similar) and we would have no way of knowing that.

It'd however be possible to avoid "stopping the world" by loading the image in background and displaying it asynchronously (unless the selection changed), but I *think* that'd be a bit awkward to do with Ranger's architecture. Anyhoo, this could be a candidate for another PR. Perhaps open a new issue if nobody requested that before.

@cain-dev ^

ranger/ext/img_display.py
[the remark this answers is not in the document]
toonn: What makes it more appropriate? It still gets a name on most file systems, no? Even if it is quickly removed on the happy path it's still nicer to use a sensible name when possible, just in case.

ranger/ext/img_display.py
toonn: Could go in `ranger/ext` I guess. Though it does seem like `get_executables` may have a very similar purpose, maybe we should just use that?

ranger/ext/img_display.py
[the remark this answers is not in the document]
toonn: I believe it is, [command utility POSIX page](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html). `which` has never been specified in POSIX nor in the UNIX standard so it wouldn't be better in this regard either way.

ranger/ext/img_display.py
toonn: But is this the size of the terminal window or the ncurses subwindow/preview pane?

toonn (commented):
Really appreciate all the work going into this!

ranger/ext/img_display.py
remi6397: This is the size of the terminal window. As a matter of fact, ncurses invokes the very same API under the hood on WinCh.

toonn: Ah, right and then we use the max_cols and rows to limit the image size later, of course.

toonn:
@remi6397, did you request my review on this? I still see open review comments so I'm not sure what I should be looking at?

remi6397:
> @remi6397, did you request my review on this? I still see open review comments so I'm not sure what I should be looking at?

Yup, @toonn, I did, for I believe I have addressed all of your concerns, but perhaps I should've left some comments first… Gimme a minute.

ranger/ext/img_display.py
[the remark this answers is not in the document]
remi6397: A `TemporaryFile` may be invisible from the user's POV. In spite of the name, `NamedTemporaryFile` is not a facility for naming temporary files.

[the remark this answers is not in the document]
cherti: I think what @toonn whats to have achieved can be done with TemporaryFile by just prefixing the `prefix` argument with `ranger-`. Then every temporary file that is actually visible in the filesystem tree would be (as far as I understand Python's implementation) be named `ranger-…`, so should anyone find it in the file system it would be named accordingly, if it's invisible to the system it doesn't matter.

If I see this correctly, this would not break your implementation, @remi6397 and @toonn would be happy as well? :)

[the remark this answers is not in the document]
remi6397: That's… basically what just I said. It's already been done, though with file path rather than "ranger" as the prefix. https://github.com/ranger/ranger/pull/2466/files#diff-4ca0bb8f02a8372446af8e614f9691b6272802bc59968267b66b1e90201b92b7R469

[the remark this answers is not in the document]
remi6397: @cherti, how about like this? I'm not so sure about the inclusion of the path in the filename anymore, as we could exceed `PATH_MAX`…?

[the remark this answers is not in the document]
cherti: ah, sorry, nevermind, somehow I apparently managed to get an outdated version of the diff. nothing said!

toonn (commented):
LGTM

Let's *finally* get this merged! : D 🚀 

Three rounds of cheers for @remi6397, this has been a very highly desired feature for too long. 👏 🎉 ❤️

toonn (dismissed):
Meant to approve but commented instead >.<

cherti:
soooooooo… when's the next release? :D

toonn:
When the milestone is empty. No promises. Reviewing/reproducing stuff in the milestone is the only thing that can help speed it up.

ranger/ext/popen23.py
```diff
@@ -13,6 +13,14 @@
     from sys import version_info
     PY3 = version_info[0] >= 3
 
+if PY3:
+    # pylint: disable=ungrouped-imports,unused-import
+    from subprocess import DEVNULL
+else:
+    import os
+    # pylint: disable=consider-using-with
+    DEVNULL = open(os.devnull, "wb")
```
toonn, line 22: Just realized that this won't work. The first time you use this in a Popen call the file object will be closed once the process finishes, no?

@remi6397, thoughts on a better fix? Maybe we should incorporate the behavior into Popen23 and exclusively use that to interface with Popen?

remi6397: It would only be closed if we called `communicate` on the Popen object, so I believe that it's not going to cause any trouble (for now).

toonn: I don't really like leaving mines around though.

Any thoughts on a good way forward?

remi6397: Hmmm… Perhaps

```
DEVNULL.close = lambda: None
```

would work? 😅

toonn: I don't think it would. DEVNULL gets passed by reference as `stdout` or `stderr` and then close is called on those, no?

remi6397: Close would be called on what exactly? I don't really understand… kwargs don't wrap objects and as you said stuff is passed by reference.

remi6397: That could be Python 2-specific, as it works for me on 3.12. If that's the case, then the most reasonable solution would be to drop support for decades old Python versions…

toonn: Adding DEVNULL as an open file object is specific to Python 2. You guarded it behind the `PY3` flag yourself.

So long as we don't get a significant benefit from dropping Python 2 support it's a shame to give up on it for what's essentially a paper cut.

remi6397: I guess if my solution does indeed not work with ancient Python (I don't really have a way to check that, because Python 2 doesn't build properly on my machine) we could make a dedicated "virtual" file-like object. Something like this perhaps?

```python
class DevNull:
    def __init__(self):
        self._reset()
    def _reset(self):
        self.__f = open(os.devnull, "wb")
    def __getattr__(self, attr):
        if self.__f.closed:
            self._reset()
        return getattr(self.__f, attr)

DEVNULL = DevNull()
```

toonn: That looks good to me.

Kreijstal:
since ranger supports sixel, how do I configure it?

JosefLitos:
@Kreijstal from the looks of [the :master docs](https://github.com/ranger/ranger/blob/136416c7e2ecc27315fe2354ecadfe09202df7dd/doc/ranger.1#L380) it is available in the latest git version, but not in a release yet.

iambumblehead:
related, any plan to release next version?: https://github.com/ranger/ranger/issues/2702 _answer: no_

JosefLitos:
Also related to image previews: if `notcurses` was to be integrated, it also provides in-terminal image display and does so with a nice quality toggle highly improving speed: `ncplayer -kb half -s scale <photo>`

Issue #2476, opened 2021-10-25
Feature request: change default value for `hidden_filter`


ghost:
I've recently started using ranger, and haven't edited the default configuration yet. While running a script that backs up files, and monitoring the progress in ranger, I was surprised to find that my backup files weren't appearing in the working directory. They were listed by `ls` though. Initially, I thought this might be a bug in ranger, and I investigated that. However, I then found the `hidden_filter` option via #188, and realised that my backup files ended in `bak`, so they were excluded.

The current default setting of `hidden_filter` could be a little disconcerting for new users of ranger like myself. I recognise that a sensible default would depend on user preferences as a group, but like the commenter in #188, my idea of expected behaviour would be like the Linux `ls` command: essentially, hiding dotted files, but displaying all others.

I'm really enjoying using ranger so far, but just wanted to raise this as it was a bit of a surprise for me!

PR #2478, opened 2021-10-25
hidden filter is updated

web-d7r:
Now only files, with names started with dot are considered as hidden.
 #2476


vifon (approved):
Considering I myself forgot about this functionality, sounds like a good idea.

toonn (approved):
LGTM

Thank you for the fix! :tada:

PR #2482, opened 2021-11-04
lint: fix tests for pylint

XuehaiPan:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 20.04 LTS
- Terminal emulator and version: GNOME Terminal
- Python version: 3.9.7
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

1. apply code suggestions from `pylint`.
2. disable `consider-using-f-string` in `pylint`. (f-string is not compatible with Python 2.7 and Python 3.1~3.5)

#### MOTIVATION AND CONTEXT

Make the test pass since `pylint` upgrade. Even the tests is not pass in master version with the latest `pylint`.

#### TESTING

`make test`

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -13,7 +13,23 @@ max-branches=16
 [FORMAT]
 max-line-length = 99
 enable=no-absolute-import,old-division
-disable=cyclic-import,duplicate-code,fixme,import-outside-toplevel,locally-disabled,locally-enabled,missing-docstring,no-else-break,no-else-continue,no-else-raise,no-else-return,raise-missing-from,redefined-variable-type,stop-iteration-return,super-with-arguments,useless-object-inheritance
+disable=cyclic-import,
+        duplicate-code,
+        fixme,
+        import-outside-toplevel,
+        locally-disabled,
+        locally-enabled,
+        missing-docstring,
+        no-else-break,
+        no-else-continue,
+        no-else-raise,
+        no-else-return,
+        raise-missing-from,
+        redefined-variable-type,
+        stop-iteration-return,
+        super-with-arguments,
+        useless-object-inheritance,
+        consider-using-f-string
 
 [TYPECHECK]
 ignored-classes=ranger.core.actions.Actions
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,6 +1,8 @@
 [MASTER]
 init-hook='import sys; sys.path.append("tests/pylint")'
 load-plugins=py2_compat
+# Min Python version to use for version dependend checks.
+py-version=2.7
 
 [BASIC]
 good-names=i,j,k,n,x,y,ex,Run,_,fm,ui,fg,bg
@@ -12,7 +14,7 @@ max-branches=16
 
 [FORMAT]
 max-line-length = 99
-enable=no-absolute-import,old-division
+enable=no-absolute-import,old-division,using-f-string-in-unsupported-version
 disable=cyclic-import,
         duplicate-code,
         fixme,
--- a/.pylintrc
+++ b/.pylintrc
@@ -2,7 +2,7 @@
 init-hook='import sys; sys.path.append("tests/pylint")'
 load-plugins=py2_compat
 # Min Python version to use for version dependend checks.
-py-version=2.7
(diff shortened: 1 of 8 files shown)


toonn:
We need a Pylint checker to make sure these F-strings don't get into the code.

XuehaiPan:
> We need a Pylint checker to make sure these F-strings don't get into the code.

If the code contains f-strings, `make test` will fail with `SyntexError` on Python 2 and Python 3.5, which is a prerequisite before merging a PR.

toonn:
I'd still prefer an explicitly implemented checker. If you're not motivated to write one I'll do it but it'll have to wait until December at the earliest.

XuehaiPan:
> I'd still prefer an explicitly implemented checker.

I found the dev version of `pylint` added `using-f-string-in-unsupported-version` check in PyCQA/pylint#5139. This check will be release in 2.21.0. There will be more similar check `using-**-in-unsupported-version` in PyCQA/pylint#5134.

.pylintrc
```diff
@@ -1,6 +1,8 @@
 [MASTER]
 init-hook='import sys; sys.path.append("tests/pylint")'
 load-plugins=py2_compat
+# Min Python version to use for version dependend checks.
+py-version=2.7
```
XuehaiPan, line 5: Is it correct to set Python 2.7 here? May be 2.6 or 3.5?

.pylintrc
```diff
@@ -12,8 +14,24 @@ max-branches=16
 
 [FORMAT]
 max-line-length = 99
-enable=no-absolute-import,old-division
-disable=cyclic-import,duplicate-code,fixme,import-outside-toplevel,locally-disabled,locally-enabled,missing-docstring,no-else-break,no-else-continue,no-else-raise,no-else-return,raise-missing-from,redefined-variable-type,stop-iteration-return,super-with-arguments,useless-object-inheritance
+enable=no-absolute-import,old-division,using-f-string-in-unsupported-version
```
XuehaiPan, line 17: It is unnecessary to explicitly enable `using-f-string-in-unsupported-version` since we set `py-version` in `[MASTER]` section.

.pylintrc
toonn: 2.6 would be better for now, yeah.

toonn:
The current Py 2 compatibility checkers we need are implemented in `tests/pylint/py2_compat.py`, this is where I'm expecting a checker for this F-string syntax to be added.

XuehaiPan:
> The current Py 2 compatibility checkers we need are implemented in tests/pylint/py2_compat.py, this is where I'm expecting a checker for this F-string syntax to be added.

I tried to implement the compatibility layer to Python 2. However `astroid` in Python 2 raises `AstroidSyntaxError` because the `astroid` packages in Python 2 are not designed to check f-strings. (See the GitHub Action results)

We can use packages in 3.6 to check whether the code is compatible with Python 3.5. We cannot do that with 3.5 itself. Write another compatibility layer for `astroid` is probably out of the scope of this PR.

toonn:
Well, if we're running it with Py 2 it'll error out on the syntax anyway, right? So let's implement the check in a way that tells us the syntax isn't allowed when running with Py 3 but won't cause errors on Py 2. (Like, even check for the Python version like we do in ranger and always make it pass for Py 2.)

XuehaiPan:
> Well, if we're running it with Py 2 it'll error out on the syntax anyway, right?

yes

> So let's implement the check in a way that tells us the syntax isn't allowed when running with Py 3 but won't cause errors on Py 2. (Like, even check for the Python version like we do in ranger and always make it pass for Py 2.)

This will be covered by Py3.6 and Py3.x CI tests when pylint 2.12.0 release in the near future.

toonn:
Ok, sorry, I guess I'm dense. I misunderstood what check was going to be added in that newer Pylint version. So setting the py-version to 2.6 sounds like all we need.

XuehaiPan:
Pylint 2.12 was released yesterday. The option `py-version` is available now in tests.

PR #2484, opened 2021-11-05
[Feat/Fix] core.actions: Use ranger's Python interpreter in `scope.sh`

XuehaiPan:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 20.04 LTS
- Terminal emulator and version: GNOME Terminal
- Python version: 3.9.7
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

In this PR, the ranger's Python interpreter will be passed to `scope.sh` to ensure there is always a usable Python interpreter. This is compatible with `python` / `python2` / `python2.x` / `python3` / `python3.x` / `pypy` / `pypy3` / `pypy3.x`.

#### MOTIVATION AND CONTEXT

Python 2.7 has reached the end of its life cycle. The new releases of many mainstream OS do not have an executable named `python` (usually they are `python3`), especially in docker images.

Some file handler uses Python in `scope.sh` but they arbitrarily use the name `python` instead of checking the usability.

#### TESTING

`make test`


toonn (commented):
Thanks, this is a good oversight to point out. However, i'm not sure we need to complicate matters by passing in the interpreter from ranger. Can't we just test whether `python3` is available and set a variable accordingly?

XuehaiPan:
> However, i'm not sure we need to complicate matters by passing in the interpreter from ranger. Can't we just test whether `python3` is available and set a variable accordingly?

I have once considered checking whether `python3` is available directly, but I choose to use the ranger's interpreter and pass it to `scope.sh` for three reasons:

1. ranger's interpreter is always usable (ranger needs a Python executable to run)
2. hard code `python3` does not completely solve the problem, we need to check python / python2 / python2.x / python3 / python3.x / pypy / pypy3 / pypy3.x. What if there is a Python 4 in the future (probably not)?
3. came from 2., for performance reasons if we check all these executable names above.

toonn:
Ok, I'm convinced by the pypy argument. One caveat is that most Python programs nowadays are Python 3 only, so maybe we never want to use Python/Pypy 2 unless it's the only one available?

XuehaiPan:
I pushed a new commit to search `python` candidate in `PATH`. Either of the two commits fixes the `python` executable issue.

1. Pass ranger's Python interpreter to `scope.sh`:
    Pros: the interpreter is always useable, handles such as `python3.8`, less computation
    Cons: modifies python libraries and changes argument list of `scope.sh`

2. Check usable Python executable on runtime:
    Pros: standalone fix for `scope.sh`, does not change any functionally for python libraries and the argument list.
    Cons: duplicate computing for testing executables while previewing multiple files, cannot handle `python3.x` etc.

toonn:
Hmm, your original idea may be best after all. Means we don't depend on searching the PATH at all and maybe there's an advantage to spinning up copies of the same interpreter too.

PR #2486, opened 2021-11-07
Catch AttributeError when querying encodings

cherti:
curses can return None as a valid value, but that doesn't have any
decoding, hence the .decode fails on that call, crashing ranger in the process.
To mitigate this, the resulting AttributeError gets caught and the
codepath is skipped entirely.


vifon (dismissed):
The `None` value should be handled directly where it appears, not as an `AttributeError` caused by accessing a nonexistent method of it. This exception handler silences far too many possible errors.

toonn (commented):
I'm not such a fan of the early return. I'd prefer conditionally setting escapes. That way it's harder to accidentally add something to the window refresh at the bottom without adding it to the early return. (Also, imo, let's move fmt_tups to the else clause or drop the indirection entirely.)

cherti:
> I'm not such a fan of the early return. I'd prefer conditionally setting escapes. That way it's harder to accidentally add something to the window refresh at the bottom without adding it to the early return. (Also, imo, let's move fmt_tups to the else clause or drop the indirection entirely.)

So you say in doubt just end up with an empty `fmt_tups` and leave the early return at the bottom?

toonn:
Well, there is no early return at the bottom? Can't return later than having the function body end : )

cherti:
> Well, there is no early return at the bottom? Can't return later than having the function body end : )

Yeah, just understanding how exactly you would lay that out to make it into a solution both you and @vifon are comfortable with :)

cherti:
To structure things: In my understanding there are basically three approaches of fixing this issue:

1. Just catching any AttributeError together with UnicodeError, if anything breaks in there just skip the entire code path.
2. Check the return from curses and return early the moment we encounter an error/None, otherwise proceed
3. Check the return from curses and setup `fmt_tups` accordingly (potentially empty) and let the remaining function just silently do nothing as there is nothing to `fmt`, then return at the end

If I see everything correctly (please correct me if I'm wrong!), @vifon is in favor of 2 and @toonn is in favor of 3? Given this is not too much code, I don't see a way to combine 2 and 3, but I might not see something. In any case I'm open to proceeding with either of those, just let me know and I'll set up the PR accordingly. :)

PR #2489, opened 2021-11-11
thumbnail preview for audio

pirate486743186:
preview of the embedded images in audio files
It's a simple change in scope.sh,

#### ISSUE TYPE


toonn (approved):
LGTM

toonn:
I dropped the `exit 1` so we get fallthrough behavior in case there is no album art.

PR #2493, opened 2021-11-12
Fix some hotkeys referred in the manual

guillermogf:
I detected a few hotkeys related to previews were mistakenly referred to in the manual. I hope this commit fixes it, but please double check.


toonn (approved):
LGTM!

PR #2506, opened 2021-11-29
Load plugins fom using setup-tools entry-points

joouha:
This PR implements loading plugins from [setuptools entry-points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#advertising-behavior), and is a replacement for #2302.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: Linux 5.10.81 - NixOS
- Python version: 2.7.15 / 3.9.6
- Ranger version/commit: ranger-master v1.9.3-431-g31c5067d
- Locale: en_GB.UTF-8

#### CHECKLIST
- [x] The `CONTRIBUTING` document has been read
- [x] All changes follow the code style
- [ ] All new and existing tests pass
    - [ ] Tests have been updated

#### DESCRIPTION
Since my previous PR (#2302) I have learnt about [setuptools entry-points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#advertising-behavior), which are as close as python has to a "standard" way of implementing plugins. This approach is used by many existing packages (e.g. [pytest](https://docs.pytest.org/en/6.2.x/writing_plugins.html#making-your-plugin-installable-by-others), [sphinx](https://www.sphinx-doc.org/en/master/development/builders.html#discover-builders-by-entry-point), etc.), and I think it is a much cleaner solution.

It enables plugins to be installed as stand-alone python packages, so if a plugin is installed using `pip`, it will be loaded by `ranger` with no further steps required. All that would be required to make a package a ranger plugin is creating an entry-point in the `"ranger.plugins"` group pointing at your plugin module.

This PR adds the back-ported `importlib_metadata` package as an optional dependency for `python<3.8`, but ranger will still run without it, only entry-point plugins will not be loaded.

I've updated the man page with an example showing how to create a plugin with this new method. Briefly, if you're using `setuptools` for your package, you'd need to define an entry-point in the `ranger.plugins` group as follows:

```python
from setuptools import setup
setup(
    name="my-ranger-plugin",
    packages=["my_ranger_plugin"],
    # the following makes a plugin available to ranger
    entry_points={"ranger.plugins": ["plugin_name = myproject.pluginmodule"]},
)
```

If you're using [poetry](https://python-poetry.org/docs/pyproject/#plugins), you can add a section to your `pyproject.toml` file like this:

```toml
[tool.poetry.plugins."ranger.plugins"]
"plugin_name" = "myproject.pluginmodule"
```

It might be nice to request a new [pypi trove classifier](https://github.com/pypa/trove-classifiers) (i.e. `Framework :: ranger`) - this will facilitate plugin discovery, and make it much simpler to build a plugin manager for `ranger`.

#### MOTIVATION AND CONTEXT

The current plugin user experience is not great, as it requires a manual installation step - see my previous PR #2302 for more details on the motivation for this PR.

#### TESTING
The tests seem to be a bit broken, partially caused by the release of Python 3.10, but also because pylint suggests using f-strings everywhere.


toonn:
This sounds really awesome. I'm sorry I haven't been able to take a proper look at your original PR yet, even. Won't have much time until early next year unfortunately. Will try to at least get the Pylint fix PR in before then so you won't have to deal with Pylint nagging so much.

PR #2525, opened 2021-12-16
rc.conf: map for chmod use --

luchr:
use chmod with "--" to support filenames starting with "-"

This PR changes the maps for `chmod` calls and inserts `--` in front of the filenames. Why? In order to support filenames that start with minus "`-`". Without `--` chmod tries to interpret the filename's characters after the "`-`" as chmod flags.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version:  Fedora 35
- Terminal emulator and version: Mate Terminal 1.26.0
- Python version: 3.10.0
- Ranger version/commit:  1.9.3
- Locale: en_GB.UTF-8

#### CHECKLIST
    - [ ] Tests have been updated

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -623,17 +623,17 @@ map m<bg>   draw_bookmarks
 copymap m<bg>  um<bg> `<bg> '<bg> p`<bg> p'<bg>
 
 # Generate all the chmod bindings with some python help:
-eval for arg in "rwxXst": cmd("map +u{0} shell -f chmod u+{0} %s".format(arg))
-eval for arg in "rwxXst": cmd("map +g{0} shell -f chmod g+{0} %s".format(arg))
-eval for arg in "rwxXst": cmd("map +o{0} shell -f chmod o+{0} %s".format(arg))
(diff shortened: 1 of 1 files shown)


toonn (commented):
Unfortunately this doesn't seem portable to BSD chmod.

luchr:
I think a condition (e.g. depending on `sys.platform`) is too much overhead here (in a conf-file).

toonn:
@luchr, maybe this warrants a custom command in `commands.py`? Then we can add the complexity or even use Python's `os` to change the modes? Because I have to admit not being able to work with files that could parse as flags/options is pretty annoying.

luchr:
You mean, extending the `chmod` class  [in commands.py](https://github.com/ranger/ranger/blob/master/ranger/config/commands.py#L1178) to allow also mode values like `u+rw,o-w` (i.e. reimplementing `chmod` in python)?

toonn:
Ah, we already have a chmod command. Then maybe we can just update the bindings to use it? That way paths can't be a problem.

PR #2526, opened 2021-12-17
extend chmod command: allow mode-specs

luchr:
This is a first attempt to extend the chmod (ranger-)command (reimplementation of chmod in python) because of https://github.com/ranger/ranger/pull/2525#issuecomment-996705889

If mode-specs (e.g. `go-w,u+rw`) are understood then the mappings may be updated.

I don't know if this points in the direction toonn had in mind. Also a very limited set of mode specifications (comma separated list of `[ugoa]+[+-][rwx]+`) are included in this PR.

#### ISSUE TYPE

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -93,6 +93,7 @@ from __future__ import (absolute_import, division, print_function)
 
 from collections import deque
 import os
+import stat
 import re
 from io import open
 
@@ -1176,24 +1177,86 @@ class rename_append(Command):
 
 
 class chmod(Command):
-    """:chmod <octal number>
+    """:chmod [<octal number>|<mode-spec>]
 
-    Sets the permissions of the selection to the octal number.
+    Sets the permissions of the selection to the given mode.
 
     The octal number is between 0 and 777. The digits specify the
     permissions for the user, the group and others.
 
     A 1 permits execution, a 2 permits writing, a 4 permits reading.
     Add those numbers to combine them. So a 7 permits everything.
+
+    The <mode-spec> is a comma-separated where each entry has the
+    form [ugoa]+[+-][rwx]+
     """
 
-    def execute(self):
-        mode_str = self.rest(1)
+    def __init__(self, *args, **kwargs):
+        '''init chmod-command.'''
+        super(chmod, self).__init__(*args, **kwargs)
+        self.re_mode = None
+        self.bitmasks = None
+
+    def prepare(self):
+        '''create regex for mode-spec and bitmasks.'''
+        self.re_mode = re.compile(r'''
+            ([ugoa]+)     # (u)ser, (g)roup, (o)other, (a)ll
+            ([+-])        # add (+), remove (-)
+            ([rwx]+)      # (r)ead, (w)rite, e(x)ecute
+            $             # end
+            ''', re.X)
+        masks = {
+            'u': {'r': stat.S_IRUSR, 'w': stat.S_IWUSR, 'x': stat.S_IXUSR},
+            'g': {'r': stat.S_IRGRP, 'w': stat.S_IWGRP, 'x': stat.S_IXGRP},
+            'o': {'r': stat.S_IROTH, 'w': stat.S_IWOTH, 'x': stat.S_IXOTH}}
+        masks['a'] = {
+            'r': masks['u']['r'] | masks['g']['r'] | masks['o']['r'],
+            'w': masks['u']['w'] | masks['g']['w'] | masks['o']['w'],
+            'x': masks['u']['x'] | masks['g']['x'] | masks['o']['x']}
+        self.bitmasks = masks
+
+    def get_mask_for(self, ugoa_part, rwx_part):
+        '''return mask for "ugoa" and "rwx" string.'''
+        mask = 0
+        for part in ugoa_part:
+            for flag in rwx_part:
+                mask |= self.bitmasks[part][flag]
+        return mask
+
+    def exec_modespecs(self, mode_str):
+        '''chmod according to mode-specs.'''
+        mask_operations = []
+        for mode in mode_str.split(','):
+            match = re.match(self.re_mode, mode)  # fullmatch missing in Py2
+            if match is None:
+                self.fm.notify(
+                    "Syntax: chmod <mode spec>: cannot parse " + mode,
+                    bad=True)
+                return False
+            sub_mask = self.get_mask_for(match.group(1), match.group(3))
+            mask_operations.append(
+                (lambda found, sub=sub_mask: found | sub)
+                if match.group(2) == '+' else
+                (lambda found, sub=sub_mask: found & ~sub))
+
+        for fobj in self.fm.thistab.get_selection():
+            try:
+                mode = os.stat(fobj.path).st_mode
+                for operation in mask_operations:
+                    mode = operation(mode)
+                os.chmod(fobj.path, mode)
+            except OSError as ex:
+                self.fm.notify(ex)
+
+        return True
+
+    def exec_octal_number(self, mode_str):
+        '''chmod according to octal-number.'''
         if not mode_str:
             if self.quantifier is None:
                 self.fm.notify("Syntax: chmod <octal number> "
                                "or specify a quantifier", bad=True)
-                return
+                return False
             mode_str = str(self.quantifier)
 
         try:
@@ -1202,17 +1265,26 @@ class chmod(Command):
                 raise ValueError
         except ValueError:
             self.fm.notify("Need an octal number between 0 and 777!", bad=True)
-            return
+            return False
 
         for fobj in self.fm.thistab.get_selection():
             try:
                 os.chmod(fobj.path, mode)
             except OSError as ex:
                 self.fm.notify(ex)
+        return True
 
-        # reloading directory.  maybe its better to reload the selected
-        # files only.
-        self.fm.thisdir.content_outdated = True
+    def execute(self):
+        self.prepare()
+        mode_str = self.rest(1)
+        changes = (
+            self.exec_modespecs(mode_str)
+            if mode_str and re.search('[^0-9]', mode_str) else
+            self.exec_octal_number(mode_str))
+        if changes:
+            # reloading directory.  maybe its better to reload the selected
+            # files only.
+            self.fm.thisdir.content_outdated = True
 
 
 class bulkrename(Command):


ranger/config/commands.py
```diff
@@ -1176,24 +1177,86 @@ def execute(self):
...
 
-    def execute(self):
-        mode_str = self.rest(1)
+    def __init__(self, *args, **kwargs):
+        '''init chmod-command.'''
+        super(chmod, self).__init__(*args, **kwargs)
+        self.re_mode = None
+        self.bitmasks = None
+
+    def prepare(self):
```
toonn, line 1200: This seems like it fits more in `__init__` rather than `execute`. Whether or not it's still in its own function is up to you.

ranger/config/commands.py
```diff
@@ -1176,24 +1177,86 @@ def execute(self):
...
+            'u': {'r': stat.S_IRUSR, 'w': stat.S_IWUSR, 'x': stat.S_IXUSR},
+            'g': {'r': stat.S_IRGRP, 'w': stat.S_IWGRP, 'x': stat.S_IXGRP},
+            'o': {'r': stat.S_IROTH, 'w': stat.S_IWOTH, 'x': stat.S_IXOTH}}
+        masks['a'] = {
+            'r': masks['u']['r'] | masks['g']['r'] | masks['o']['r'],
+            'w': masks['u']['w'] | masks['g']['w'] | masks['o']['w'],
+            'x': masks['u']['x'] | masks['g']['x'] | masks['o']['x']}
+        self.bitmasks = masks
+
+    def get_mask_for(self, ugoa_part, rwx_part):
```
toonn, line 1218: Let's go with the names from the chmod grammar "who" and "perm".

ranger/config/commands.py
```diff
@@ -1176,24 +1177,86 @@ def execute(self):
...
+
+    def prepare(self):
+        '''create regex for mode-spec and bitmasks.'''
+        self.re_mode = re.compile(r'''
+            ([ugoa]+)     # (u)ser, (g)roup, (o)other, (a)ll
+            ([+-])        # add (+), remove (-)
+            ([rwx]+)      # (r)ead, (w)rite, e(x)ecute
+            $             # end
+            ''', re.X)
+        masks = {
```
toonn, line 1208: Good start and looks equivalent to the octal implementation but let's try expanding this to cover the full grammar.

ranger/config/commands.py
```diff
@@ -1176,24 +1177,86 @@ def execute(self):
...
+    def __init__(self, *args, **kwargs):
+        '''init chmod-command.'''
+        super(chmod, self).__init__(*args, **kwargs)
+        self.re_mode = None
+        self.bitmasks = None
+
+    def prepare(self):
+        '''create regex for mode-spec and bitmasks.'''
+        self.re_mode = re.compile(r'''
+            ([ugoa]+)     # (u)ser, (g)roup, (o)other, (a)ll
```
toonn, line 1203: The who part is optional.

ranger/config/commands.py
```diff
@@ -1176,24 +1177,86 @@ def execute(self):
...
+
+    def get_mask_for(self, ugoa_part, rwx_part):
+        '''return mask for "ugoa" and "rwx" string.'''
+        mask = 0
+        for part in ugoa_part:
+            for flag in rwx_part:
+                mask |= self.bitmasks[part][flag]
+        return mask
+
+    def exec_modespecs(self, mode_str):
```
toonn, line 1226: Manpage calls it "symbolic mode" let's go with that terminology.

ranger/config/commands.py
```diff
@@ -1176,24 +1177,86 @@ def execute(self):
...
+        for mode in mode_str.split(','):
+            match = re.match(self.re_mode, mode)  # fullmatch missing in Py2
+            if match is None:
+                self.fm.notify(
+                    "Syntax: chmod <mode spec>: cannot parse " + mode,
+                    bad=True)
+                return False
+            sub_mask = self.get_mask_for(match.group(1), match.group(3))
+            mask_operations.append(
+                (lambda found, sub=sub_mask: found | sub)
```
toonn, line 1238: We don't actually use the second argument so let's just close over sub_mask.

ranger/config/commands.py
```diff
@@ -1176,24 +1177,86 @@ def execute(self):
...
+                if match.group(2) == '+' else
+                (lambda found, sub=sub_mask: found & ~sub))
+
+        for fobj in self.fm.thistab.get_selection():
```
toonn, line 1242: I'd rather share this very similar behavior in the execute function. So just accumulate the bitmask, then apply it.

ranger/config/commands.py
toonn: Also, I'd rather coalesce this into a single bitmask to apply.

ranger/config/commands.py
```diff
@@ -1202,17 +1265,26 @@ def execute(self):
                 raise ValueError
         except ValueError:
             self.fm.notify("Need an octal number between 0 and 777!", bad=True)
```
toonn, line 1267: Can we expand this to cover all of the modes? Like `4000`.

toonn (commented):
Looks like a good start!

ranger/config/commands.py
toonn: We can use [this man page as a reference](https://www.man7.org/linux/man-pages/man1/chmod.1p.html) for the grammar. (Or the BSD chmod man page because that's a little clearer imo.)

luchr:
I'll try the full-grammar implementation in the next days.

ranger/config/commands.py
luchr: I moved it out of `__init__` because of ranger's behaviour:  if somebody types `:`, `c`, `h`, `m`, etc. at ranger's command line then `__init__` will be called at every keystroke (when there is only one matching command; in the `Console.on_line_change` for the "quick-Test/Call"). That's why I didn't want to stuff much initialization in `__init__` in order to not slow down this typing thing. After finding out that there is no `quick` the whole instance is thrown away.

ranger/config/commands.py
luchr: "single bitmask to apply" ... is not possible! 
Take `u-wx,u+r`  then I have for each bit *three*(!) possible actions: don't change, set or clear. There is *no* way to express this with a single bitmask that is applied. I express this with one `and` operation and an `or` operation (using lambda-Functions). 

Here is the [link to the gnu coreutils](https://fossies.org/linux/coreutils/lib/modechange.c#l_21): they are doing a similar approach because of this.

luchr:
Next round, next fun. Here is now a (re-)implementation of chmod supporting a lot of symbolic mode changes:

```
    The symbolic mode is
        symbolic mode   ::=  clause [, clause ...]
        clause          ::=  [who ...] [action ...] action
        action          ::=  op permcopy | op [perm ...]
        who             ::=  u | g | o | a
        op              ::=  = | + | -
        permcopy        ::=  u | g | o
        perm            ::=  r | w | x | s | t | X
```

Here are some examples (some of them artificial/contrived):

```
 mode   command arg                result   conditions
-----------------------------------------------------------
  0755  0744                  =>   0744  
  0700  0744                  =>   0744  
  0700  go+u                  =>   0777  
  0677  go=u                  =>   0666  
  0677  g=u                   =>   0667  
  0677  g-u                   =>   0617  
  0600  g-u                   =>   0600  
  0600  =u                    =>   0666  [for umask 0000]
  0600  =u                    =>   0644  [for umask 0022]
  0600  =u                    =>   0664  [for umask 0002]
  0621  +u                    =>   0667  [for umask 0000]
  0621  +u                    =>   0665  [for umask 0002]
  0621  +u                    =>   0665  [for umask 0022]
  0621  -u                    =>   0021  [for umask 0022]
  0321  -u                    =>   0020  [for umask 0022]
  0621  =u                    =>   0644  [for umask 0022]
  0621  u+x                   =>   0721  
  0621  u+x-w                 =>   0521  
  0621  g=u-w                 =>   0641  
  0641  g=u-w                 =>   0641  
  0651  uo=g-x                =>   0454  
  0621  g=u-w,o=              =>   0640  
  0777  =rx                   =>   0555  [for umask 0022]
  0777  =rwx                  =>   0755  [for umask 0022]
  0777  =rwx                  =>   0775  [for umask 0002]
  0777  =w=rx                 =>   0555  [for umask 0002]
  0123  u+-=r                 =>   0423  
  0123  +,-,+,+rw             =>   0767  [for umask 0002]
  0120  +,-,+,+rw             =>   0764  [for umask 0002]
  0321  u=rw,g-w,o=u          =>   0606  
  0321  u+s                   =>   4321  
  4644  =u                    =>   0664  [for umask 0002]
  4644  =u,g+s                =>   2664  [for umask 0002]
  4644  =u,g+st               =>   2664  [for umask 0002]
  4644  g=u,o+t               =>   5664  
```

I tried to satisfy (nearly) all your suggestions. Init-Code is now in the class construction, so no `prepare` any more. Coverage of `rwxstXugo`  (the last 3 as permcopy modes). Single bitmask to apply is not possible, see https://github.com/ranger/ranger/pull/2526#discussion_r777030631

ranger/config/commands.py
```diff
@@ -1176,43 +1177,203 @@ def execute(self):
...
+            (r_flag if 'r' in rwx_str else 0) |
+            (w_flag if 'w' in rwx_str else 0) |
+            (x_flag if 'x' in rwx_str else 0))
+
+    def get_mode_change_funcs(self, who, operator, perms):
+        '''returns mode-change-functions (mcf) for given action.'''
+
+        if perms == '' and operator in '+-':
+            return tuple()
+        mcfs, affected_bits = [], 0
```
toonn, line 1267: Let's use parens on the RHS. It's more unambiguously clear.

ranger/config/commands.py
```diff
@@ -1176,43 +1177,203 @@ def execute(self):
...
+        for the rwx-permissions (e.g. 'rw') input.
+        '''
+        r_flag, w_flag = self.MASKS[target]['r'], self.MASKS[target]['w']
+        x_flag = self.MASKS[target]['x']
+        return lambda rwx_str: (
+            (r_flag if 'r' in rwx_str else 0) |
+            (w_flag if 'w' in rwx_str else 0) |
+            (x_flag if 'x' in rwx_str else 0))
+
+    def get_mode_change_funcs(self, who, operator, perms):
```
toonn, line 1262: I have to admit I don't have the brainpower to properly review the logic here right now. Nothing stood out to me though.
Would you be up to writing some tests to verify the behavior?

ranger/config/commands.py
```diff
@@ -1176,43 +1177,203 @@ def execute(self):
...
+                mode = int(mode_str, 8)
+                if mode < 0  or mode > 0o7777:
+                    raise ValueError()
+                mcfs.append(lambda omode, lmode: mode)
+            except ValueError:
+                self.fm.notify(
+                    "Need an octal number between 0o0 and 0o7777!", bad=True)
+            return mcfs
+
+        set_omode = lambda omode, lmode: (lmode | (omode & ~0o7777), lmode)
```
toonn, line 1329: Isn't `omode & ~0o7777` the opposite of a tautology, i.e., always false? Since `~0o7777` is `0`?

ranger/config/commands.py
```diff
@@ -1176,43 +1177,203 @@ def execute(self):
...
 
     def execute(self):
+        '''execute chmod command.'''
         mode_str = self.rest(1)
         if not mode_str:
             if self.quantifier is None:
-                self.fm.notify("Syntax: chmod <octal number> "
-                               "or specify a quantifier", bad=True)
+                self.fm.notify(
+                    "Syntax: chmod <octal number>|<symbolic mode><octal number> "
```
toonn, line 1353: I think there's one too many `<octal number>`s here?

ranger/config/commands.py
```diff
@@ -1176,43 +1177,203 @@ def execute(self):
...
-                self.fm.notify(ex)
-
-        # reloading directory.  maybe its better to reload the selected
-        # files only.
-        self.fm.thisdir.content_outdated = True
+                omode = os.stat(fobj.path).st_mode
+                lmode = omode
+                for mask_op in mcf_list:
+                    result = mask_op(omode, lmode)
+                    if isinstance(result, tuple):
```
toonn, line 1367: Let's have the function always return a tuple, makes the logic here simpler.

ranger/config/commands.py
toonn: Also, I'm unclear about why we're changing the omode here?

ranger/config/commands.py
```diff
@@ -1176,43 +1177,203 @@ def execute(self):
...
+                        omode, lmode = result
+                    else:
+                        lmode = result
+                os.chmod(fobj.path, lmode)
+                changed = True
+            except OSError as exp:
+                self.fm.notify(exp)
+        if changed:
```
toonn, line 1375: Might as well unconditionally set `content_outdated = changed`.

toonn (commented):
Trying to catch up on the backlog, sorry for the long delays between reactions.

Looking pretty good but I do have a couple questions.

ranger/config/commands.py
```diff
@@ -1176,43 +1177,203 @@ def execute(self):
...
+    setuid/setgid/sticky bits.
 
     A 1 permits execution, a 2 permits writing, a 4 permits reading.
     Add those numbers to combine them. So a 7 permits everything.
-    """
+
+    The symbolic mode is
+        symbolic mode   ::=  clause [, clause ...]
+        clause          ::=  [who ...] [action ...] action
+        action          ::=  op permcopy | op [perm ...]
```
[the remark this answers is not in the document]
luchr: Yes single `+`, `-` and single `=` are allowed modes. You asked [for all of the modes]( https://github.com/ranger/ranger/pull/2526#discussion_r776311243). Example: `chmod = ./file` is not a no-op (and it's supported by GNU's chmod and FreeBSD's chmod). 

What do you want?

1. "all of the modes"? If yes, then we have a lot of special cases (see my answers to your other questions) and a lot of work reimplenting the wheel (see this PR).
2. your grammar (like your suggestion `op perm [perm ...]`) which is a true subset (a lot of things are not supported by this suggestion, e.g. `chmod  +-= ./file`)? If yes, then I suggest using my first attempt with a small grammar, covering a lot of the use cases, but of course not all of them.
3. something else? Then please describe.

I still vote for 3: `chmod` is not the first and won't be the last external program with the name and/or calling convention depending on the OS. If ranger supports a simple conditional statement (`if`) in the config checking for the OS, then one can use GNU's chmod calling convention and BSD's chmod calling convention, etc. And this all without the fuss reimplementing stuff that is already there.

ranger/config/commands.py
luchr: No `~0o7777` is not `0`.  Why should it be? More than 12 bits are used for the mode bits (in a typical OS). And here we have one of the special cases: the wish "all of the modes" includes changes of the modes, that depend on the current original mode "omode", e.g. a mode-spec using `X` has to test if the current file under investigation is a directory. This test is done by checking the bit `0o040000` in `omode` (see; there are more than 12 bits!). And to check this I have to pass-through the `omode` (the original mode) to the mode-chaning functions in order to implement e.g. the `X`-modespec. 

Why do I need to change the `omode` during the operations? Because "original mode" means: "mode as it was seen at the beginning of the current 'clause'", so take e.g. `chmod g=u,o=g ./file`. That has two clauses and for the second one `o=g` the `omode` is the mode after `g=u` was computed!  Hence I have to keep track of the `omode` and I have to keep the higher bits (above 12) in order for the following mode-specs (that may depend on the higher bits, see `X` example) to do the right things. Hence `omode & ~0o7777` means: keep all the higher bits.

Yes, there are a lot of special cases.  (One can have  `chmod g=u-w ./file`  and also `chmod =w=rx ./file`, etc.)

ranger/config/commands.py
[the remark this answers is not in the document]
toonn: Both 1 and 3 are acceptable. If you have a better idea for 3 I'm open to it.

ranger/config/commands.py
toonn: 👍

ranger/config/commands.py
toonn: Oh, good motivation then. I think we can still close over `sub_mask` though?

PR #2528, opened 2021-12-17
add TaskView.task_remove_all() and keybinding (dD)

### LINKED ISSUE #2527: in taskview, add functionality to delete all tasks
#### Current Behavior

When viewing tasks in taskview, I can delete one task at a type by hitting "dd".
If there are a lot of tasks, it is tedious to delete all of them individualy.

#### Expected Behavior

There should be some way to delete all tasks

#### Context

Some of the log files in /var/log on a system are large, traversing over these files generates a task "Getting preview of /var/log/foo.log" for each one. It is tedious to delete each task.

#### Possible Solutions

I've created a patch to fix this. It adds  the function task_remove_all() to the TaskView class, and sets a keybinding (dD) to call it.

#### Steps to reproduce
1. start ranger
2. go to any directory with multiple log files larger than 100MiB or so
3. traverse over each file
4. run "w" to go into taskview

njthomas:
- this patch adds the function task_remove_all() to the TaskView class
- when called, it will delete all tasks in the queue
- it adds the keybinding "dD" to the taskview interface to call the function

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Python version: 
- Ranger version/commit: 
- Locale: 

#### CHECKLIST

    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

resolves #2527

#### TESTING

I did run "make test" before and after the patch to ensure that nothing changed. It calculated the same score both times (9.37/10).

### DIFF
--- a/ranger/gui/widgets/taskview.py
+++ b/ranger/gui/widgets/taskview.py
@@ -80,6 +80,9 @@ class TaskView(Widget, Accumulator):
         if self.fm.loader.queue:
             self.fm.loader.remove(index=i)
 
+    def task_remove_all(self):
+        self.fm.loader.destroy()
+
     def task_move(self, to, i=None):  # pylint: disable=invalid-name
         if i is None:
             i = self.pointer
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -751,6 +751,7 @@ copytmap <PAGEUP>   p  b  <C-B>
 tmap J          eval -q fm.ui.taskview.task_move(-1)
 tmap K          eval -q fm.ui.taskview.task_move(0)
 tmap dd         eval -q fm.ui.taskview.task_remove()
+tmap dD         eval -q fm.ui.taskview.task_remove_all()
 tmap <pagedown> eval -q fm.ui.taskview.task_move(-1)
 tmap <pageup>   eval -q fm.ui.taskview.task_move(0)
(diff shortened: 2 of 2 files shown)


njthomas:
Just fyi, I did run "make test" before and after the patch to ensure that nothing changed. It calculated the same score both times (9.37/10).

njthomas:
resolves #2527

ranger/gui/widgets/taskview.py
```diff
@@ -80,6 +80,11 @@ def task_remove(self, i=None):
         if self.fm.loader.queue:
             self.fm.loader.remove(index=i)
 
+    def task_remove_all(self):
+        if self.fm.loader.queue:
```
chu4ng, line 84: I think you can use something like:
```
def task_remove_all(self):
    for i in self.fm.loader.queue:
        task_remove(i)
```

chu4ng (approved):
Changes is useful. I was thinking about bind this "loader.destroy" to dD, but new "task_remove_all" looks more readable. I think everything is fine.

ranger/gui/widgets/taskview.py
njthomas: So rather than mapping dD to `fm.loader.destroy()`, I kept it in the newly created `task_remove_all()` function so that the intent is clearer.

Issue #1320, opened 2018-10-01
svg re-renders preview extra times


andykais:
#### Runtime Environment
```
 $ ranger --version
ranger version: ranger-master 1.9.2
Python version: 3.7.0 (default, Jul 15 2018, 10:44:58) [GCC 8.1.1 20180531]
Locale: en_US.UTF-8
```

#### Current Behavior
when displaying the preview of an svg, ranger renders a preview multiple times. Its hard to see in this gif here, but when it attempts an svg preview, it flashes after each preview has completed rendering. I also am hitting `"w"` to show the tasks being done when an svg is attempting a preview.
![ranger](https://user-images.githubusercontent.com/6627147/46309786-7478cd00-c58b-11e8-87b9-add5d09b40fe.gif)
This does not appear to happen with any of the other image previews (`png`, `jpg`, `pdf`)

#### Expected Behavior
svg only renders a preview once.

#### Possible Solutions
I do not know the internals of ranger, but I can confirm that the issue appears to be happening before `scope.sh` is being called. I know this because I put a log in `scope.sh` to see how many times it was being called for each file preview, and for svgs it was more than once.

#### Steps to reproduce
I have uncommented the lines in `scope.sh` that allow for svg previews
![screenshot 147](https://user-images.githubusercontent.com/6627147/46309430-5363ac80-c58a-11e8-882f-653ca1446b41.png)
uncomment the relevant lines in `scope.sh` and look at an svg preview.

LexTek:
Hello, I encountered a similar problem, and fixed it by replacing the line
```sh
convert "${FILE_PATH}" "${IMAGE_CACHE_PATH}" && exit 6
```
by
```sh
convert -- "${FILE_PATH}" "${IMAGE_CACHE_PATH}" && exit 6
```
Does it works for you ?

prosoitos:
May not be related, but I could not get previews for .svg at all, but removing `+xml` in the scope.sh section for svg solved it for me.

```sh
        # SVG
	image/svg)
	    convert "${FILE_PATH}" "${IMAGE_CACHE_PATH}" && exit 6
	    exit 1;;
```

toonn:
I think you could simply do `image/svg|image/svg+xml)`.

prosoitos:
Indeed. Much less clumsy. Thank you @toonn :slightly_smiling_face:

toonn:
[...]

PR #2537, opened 2021-12-29
Add rsvg-convert for SVG previews

kankaristo:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Pop!_OS 21.04 (based on Ubuntu)
- Terminal emulator and version: Terminator 2.1.0 (zsh 5.8 + tmux 3.1c)
- Python version: 3.9.5
- Locale: en_US.UTF-8

#### CHECKLIST

    - **NOTE:** The tests in `master` don't pass for me even without these changes, but no *additional* tests fail with these changes ~~(I'm not even sure if `scope.sh` has tests)~~
    - [ ] Tests have been updated

#### DESCRIPTION

`rsvg-convert` works much more reliably for SVGs than ImageMagick's `convert`, which freezes with large SVGs (seems to convert formats first, and resize only after that), and can drop out text during the conversion. Support for SVGs isn't great in ImageMagick, since it's a raster graphics program.

`rsvg-convert` only supports conversion to PNG. The image preview works fine with PNGs, but Ranger (more specifically w3mdisplay?) assumes/requires a .jpg extension, so the file is just renamed from `xxx.jpg.png` to `xxx.jpg`. It's a bit of an ugly hack, but it works, and it's faster than a second image conversion from PNG to JPEG (and vector graphics usually look better with PNG than with JPEG, which is meant more for photos).

On Ubuntu and Debian, `rsvg-convert` is in the `librsvg2-bin` package. If I remember (and read the existing code) correctly, `scope.sh` should handle things gracefully if a command (`rsvg-convert` in this case) isn't found?

#### MOTIVATION AND CONTEXT

At least partially related issue: https://github.com/ranger/ranger/issues/1320 (see [my comment](https://github.com/ranger/ranger/issues/1320#issuecomment-733926546) in that issue)

#### TESTING

The tests in `master` don't pass for me even without these changes, but no *additional* tests fail because of these changes.

~~I'm not even sure if `scope.sh` has tests, but~~ I've been using the code in this PR for a bit over a year with no issues.

**EDIT:**  
`scope.sh` does indeed have tests, and they're passing with the latest commit (I was lazy and used GitHub's web editor, but luckily `shellcheck` caught my mistake).

`test_pylint` doesn't pass for me in `master` (even without these changes), but the tests for `scope.sh` at least pass now (it's the only file I've modified in addition to README.md).

**EDIT 2:**
The only error from `test_pylint` was that it didn't recognize `unspecified-encoding` as an error/warning to disable in `pylint`.

After updating `pylint`, I get a lot more messages, mostly `C0209: consider-using-f-string` (convention, and only available since Python 3.6) with a few instances of `R1714: consider-using-in` (refactor). If I disable those in `.pylintrc`, `test_pylint` passes, but `test_pytest` starts failing with `AttributeError: module 'pylint.testutils' has no attribute 'Message'`...

So, the tests seem to be pretty dependent on the version of `pylint` in use (although to be fair, I might have gone from a really old version to a really new one). The CI environment probably uses a working version of `pylint`, but those tests still haven't finished. But none of the failures mentioned above are from my changes, I only changed `scope.sh` (and the readme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ For enhanced file previews (with `scope.sh`):
 * `img2txt` (from `caca-utils`) for ASCII-art image previews
 * `w3mimgdisplay`, `ueberzug`, `mpv`, `iTerm2`, `kitty`, `terminology` or `urxvt` for image previews
 * `convert` (from `imagemagick`) to auto-rotate images
-* `rsvg-convert` (from `librsvg2-bin`) for SVG previews
+* `rsvg-convert` (from [`librsvg`](https://wiki.gnome.org/Projects/LibRsvg))
+  for SVG previews
 * `ffmpeg`, or `ffmpegthumbnailer` for video thumbnails
 * `highlight`, `bat` or `pygmentize` for syntax highlighting of code
 * `atool`, `bsdtar`, `unrar` and/or `7z` to preview archives
--- a/README.md
+++ b/README.md
@@ -89,7 +89,8 @@ For enhanced file previews (with `scope.sh`):
 
 * `img2txt` (from `caca-utils`) for ASCII-art image previews
 * `w3mimgdisplay`, `ueberzug`, `mpv`, `iTerm2`, `kitty`, `terminology` or `urxvt` for image previews
-* `convert` (from `imagemagick`) to auto-rotate images and for SVG previews
+* `convert` (from `imagemagick`) to auto-rotate images
+* `rsvg-convert` (from `librsvg2-bin`) for SVG previews
 * `ffmpeg`, or `ffmpegthumbnailer` for video thumbnails
 * `highlight`, `bat` or `pygmentize` for syntax highlighting of code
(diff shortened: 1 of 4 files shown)


kankaristo:
I was lazy, used GitHub's web editor, and missed the end of a case statement... Luckily, `scope.sh` *does* have tests (shellcheck), and they failed.

toonn (approved):
LGTM, I'm adding a couple doc changes.

Thank you for this contribution! 🎉 
Not having SVG preview bugs anymore would be great.

P.s: You can ignore pylint checks if you don't touch any Python files. It's a bit of a mess. PyLint changes so much that it's really hard to keep up. When we pinned the version people couldn't run checks locally because they only worked with an old version that was no longer available on PyPI. So we stopped pinning the version altogether to force ourselves to just keep up with its changes, doing all of it at once was a gargantuan task. But I have very little time and appeasing PyLint is fairly low on the priority list.

PR #2538, opened 2021-12-29
better textual preview for jupyter notebooks

ferdinandyb:
Breaking out jupyter notebooks from the json preview. The new preview uses `jupyter nbconvert` to output markdown and then `bat` to color it properly. Fallback without bat is available as is the previous json viewers in case `nbconvert` is missing.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Python version: 
- Ranger version/commit: 
- Locale: 

#### CHECKLIST

    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

Makes jupyter notebooks readable in preview (the jsons are basically useless really).

#### TESTING

I tested it on my computer and should not affect codebase.

#### IMAGES / VIDEOS

new preview:
image

original preview:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ For enhanced file previews (with `scope.sh`):
 * `mediainfo` or `exiftool` for viewing information about media files
 * `odt2txt` for OpenDocument text files (`odt`, `ods`, `odp` and `sxw`)
 * `python` or `jq` for JSON files
+* `jupyter nbconvert` for Jupyter Notebooks
 * `fontimage` for font previews
 * `openscad` for 3D model previews (`stl`, `off`, `dxf`, `scad`, `csg`)
 
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -108,7 +108,15 @@ handle_extension() {
             ;;
 
         ## JSON
-        json|ipynb)
+        json)
+            jq --color-output . "${FILE_PATH}" && exit 5
+            python -m json.tool -- "${FILE_PATH}" && exit 5
+            ;;
+
+        ## Jupyter Notebooks
+        ipynb)
+            jupyter nbconvert --to markdown "${FILE_PATH}" --stdout | env COLORTERM=8bit bat --color=always --style=plain --language=markdown && exit 5
+            jupyter nbconvert --to markdown "${FILE_PATH}" --stdout && exit 5
             jq --color-output . "${FILE_PATH}" && exit 5
             python -m json.tool -- "${FILE_PATH}" && exit 5
             ;;


ferdinandyb:
I also have an idea for getting an image preview through html, which would be slower, but is actually how jupyter would should the file, but I can't get it working, so maybe this is also a good place to ask for help with that :)

Based on [this](https://docs.jupyter.org/en/latest/reference/mimetype.html) I added a new mimetype both in `~/.mime.types` and in `data/mime.types` to make sure: `application/x-ipynb+json ipynb`.

Then I added in to the `handle_images` in `scope.sh` this:

```
        ## Jupyter Notebooks
        application/x-ipynb+json)
           jupyter nbconvert --to html "${FILE_PATH}" --stdout | wkhtmltoimage \
             -f png --width "${DEFAULT_SIZE%x*}" --crop-h "${DEFAULT_SIZE%x*}" \
             - "${IMAGE_CACHE_PATH}" && exit 6
           exit 1;;
```

The command itself works for me on the commandline (produces a png from the ipynb), but the preview doesn't show and I'm not sure how to check what is wrong.

README.md
```diff
@@ -104,6 +104,7 @@ For enhanced file previews (with `scope.sh`):
 * `mediainfo` or `exiftool` for viewing information about media files
 * `odt2txt` for OpenDocument text files (`odt`, `ods`, `odp` and `sxw`)
 * `python` or `jq` for JSON files
+* `jupyter nbconvert` for Jupyter Notebooks
```
toonn, line 107: Is nbconvert a separate command? Otherwise I think listing it as just `jupyter` is nicer, since that's what you'd want to install.

toonn (commented):
LGreatTM!

Apologies for taking so long to get around to reviewing this.

toonn:
Re the trouble you're having with the MIME match. `mime.types` is a mechanism to add MIME type matches for the `mimetypes` module which ranger uses internally. `scope.sh` is just a shell script and it uses `file(1)` to determine MIME types. I believe that's why you're having trouble getting the case to match. Check what type `file --mimetype your-notebook.ipynb` reports. If it's not a suitable MIME type you'd have to go by extension.

Rom1deTroyes:
> Check what type `file --mimetype your-notebook.ipynb` reports.

A jupyter-notebook is basically a json file, and here is what I get on a Debian/GnuLinux system :
```sh
❯ file /usr/lib/python3/dist-packages/matplotlib/backends/web_backend/nbagg_uat.ipynb
/usr/lib/python3/dist-packages/matplotlib/backends/web_backend/nbagg_uat.ipynb: ASCII text, with very long lines
❯ file --mime-type /usr/lib/python3/dist-packages/matplotlib/backends/web_backend/nbagg_uat.ipynb
/usr/lib/python3/dist-packages/matplotlib/backends/web_backend/nbagg_uat.ipynb: text/plain
```

But, I also find a file called `/usr/share/mime/application/x-ipynb+json.xml`, so there could be some hope (somewhere, if someone knows how this works :-) )

BTW, thanx for all the work done on `ranger` : it's a fabulous tool :+1:

README.md
Rom1deTroyes: It seems that 2 packages can do the same things, one need Jupyter, not the other :

```
jupyter-nbconvert/stable,stable 5.6.1-3 all
  Jupyter notebook conversion  –⋅scripts
  Depends: python3-nbconvert (= 5.6.1-3), python3:any, jupyter-core

python3-nbconvert/stable,stable 5.6.1-3 all
  Jupyter notebook conversion (Python 3)
  Depends: python3-bleach, python3-defusedxml, python3-entrypoints, python3-jinja2, python3-jupyter-core, python3-mistune, python3-nbformat, python3-pandocfilters (>= 1.4), python3-pygments, python3-testpath, python3-traitlets, python3:any
```

toonn: Ah, so nbconvert *is* something that is installed in addition to the regular Jupyter? I.e., it's not like `git log` which is a subcommand included with git?

toonn:
Ah, yeah, so `file(1)` doesn't know about the MIME type. You'll just have to rely on the extension rather than the MIME type.

README.md
ferdinandyb: It can also be installed via pip: https://nbconvert.readthedocs.io/en/latest/install.html. I think if you install jupyter it will very likely also install nbconvert, and is probably more clear than trying to guess the path to the python environment that has nbconvert installed. So IMHO calling it as jupyter notebook makes the most sense, but there might be a better solution.

toonn:
Yes, a separate PR would be good.

README.md
toonn: This is fine, since it doesn't necessarily come with Jupyter we actually want to mention the subcommand.

PR #2539, opened 2021-12-29
Escaping whitespace in arguments

### LINKED ISSUE #2470: Can't paste with `paste dest=` when `dest` has spaces
#### Runtime Environment

- Ranger version/commit: ranger-master v1.9.3-422-g1de6787c

#### Current Behavior

When I specify path that has space characters in the `paste dest=` command and execute it, Ranger says "Failed to paste. Destination is invalid". If I enclose the path with " or ' it won't work. Getting rid of space characters solves the issue.

Also, same problem occurs with bookmarked paths and with things like `%d`, `%s`.

I think the program cuts the `dest` argument off at the part where first space is.

#### Expected Behavior

Able to paste when specifying path with spaces in `paste dest=` command.

#### Steps to reproduce
1. Make a folder in home folder called "test test"
2. Make a file in home folder called "test"
3. Cut "test" file
4. Try to paste it by executing command `paste dest=test test` or `paste dest='test test'` or `paste dest="test test"`

seinmon:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Terminal emulator and version: urxvt v9.26, (+ tmux 3.2a)
- Python version: 3.10.1
- Ranger version/commit: ranger-master v1.9.3
- Locale: None.None

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

This patch adds the possibility of escaping of space character as "\ " to arguments.

#### MOTIVATION AND CONTEXT

fixes #2470 
This is useful for scenarios where the a directory name containing whitespace is passed as an argument.
 
#### TESTING

The code is tested with `make test_py` and the output shows `rated 9.85/10`.

### DIFF
--- a/ranger/api/commands.py
+++ b/ranger/api/commands.py
@@ -114,8 +114,8 @@ class Command(FileManagerAware):
         args_list = line.split()
         for index, arg in enumerate(args_list):
             if index > 0:
-                if args_list[index-1][-1] == '\\':
-                    args_list[index-1] = args_list[index-1][:-1] + ' ' + arg
+                if args_list[index - 1][-1] == '\\':
+                    args_list[index - 1] = args_list[index - 1][:-1] + ' ' + arg
                     args_list.pop(index)
 
         self.args = args_list
--- a/ranger/api/commands.py
+++ b/ranger/api/commands.py
@@ -111,7 +111,15 @@ class Command(FileManagerAware):
 
     def init_line(self, line):
         self.line = line
-        self.args = line.split()
+        args_list = line.split()
+        for index, arg in enumerate(args_list):
+            if index > 0:
+                if args_list[index-1][-1] == '\\':
+                    args_list[index-1] = args_list[index-1][:-1] + ' ' + arg
+                    args_list.pop(index)
+
+        self.args = args_list
+
         try:
             self.firstpart = line[:line.rindex(' ') + 1]
         except ValueError:


toonn:
I just pushed updates that should fix the Pylint errors. I'm not very confident this works as-is. Escaping is really hard to handle well. What we really need is to be able to quote the argument. What if the destination literally contains `\ ` for example?

seinmon:
I see! This already works with destinations that have "\" in their name, but it doesn't work if the name has "\ " in it.  (Apparently `mv` also doesn't seem to work with these kinds of names). I can also try wrapping the destination in quotation marks.

Since the codebase is rather large, are there other places that need to be changed for this patch to work as expected?

seinmon:
@toonn I tried to quote the arguments, but it becomes unnecessarily complicated. I can try to finish quoting the argument if you want, but I have to ask this first:

What if the name actually contains `"`? Wrap the argument in `'`? What if it contains `'`? Wrap it in `"`? What if it contains both? Escape them? 

I guess the chance of a name containing quotation marks are a lot higher than a name containing `\ ` (slash + space) as `\ ` is how we escape spaces in other commands like `mv`, which also doesn't handle names containing `\ ` (and this behavior probably makes escaping more obvious/familiar to the users).

Please let me know what you think.

tobiasjakobi:
How about using shlex.split()? This should handle quotes and escaped whitespaces.

seinmon:
@tobiasjakobi thank you for the great suggestion. I tried `shlex.split()`, but `ranger` tries to split commands while the user is typing. As a result, `shlex.split()` gets called with no closing quotation marks\escaped characters and crashes. (I haven't had enough time to look for a workaround.)

tobiasjakobi:
@seinmon I'm not sure what you mean by crash? I tried some input where closing quotations were missing, and I don't see any crashes of the interpreter.

seinmon:
@tobiasjakobi When I give an input like `"something` (missing closing quotation) or `some\` (missing escaped character) then the script ends with a `ValueError` exception. Isn't this the same for you?

tobiasjakobi:
Of course, but that has nothing to do with any kind of "crashing". It's just a regular Python exception (which you need to handle).

seinmon:
@tobiasjakobi Now I feel stupid! I forgot that I need to handle the exception! 🤣 It works perfectly! Thanks. I will commit changes and push them to my forked repository.

seinmon:
@toonn Can you please pull the changes again? (`Pylint` tests fail for files that I haven't modified.)

seinmon:
Up @toonn

toonn (commented):
I'm a bit fearful this is going to wreak havoc with commands like `shell` which need to pass things to the shell literally because the shell will do something that `shlex.split` tries to mimic.

PR #2542, opened 2022-01-11
Make display of directory size in linemodes less misleading

eMPee584:
.. because when the cumulative size of a directory has not yet been
determined, its size is actually the number of direct children.
So displaying byte suffixes in that case is incorrect.

### DIFF
--- a/ranger/core/linemode.py
+++ b/ranger/core/linemode.py
@@ -130,7 +130,14 @@ class SizeMtimeLinemode(LinemodeBase):
     def infostring(self, fobj, metadata):
         if fobj.stat is None:
             return '?'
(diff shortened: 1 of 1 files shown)


toonn (dismissed):
Apologies for the delay in getting to this. I'm trying to catch up again.

Hmm, it'd be more consistent with behavior elsewhere if it just showed the number of files without the suffix unless the cumulative size has been calculated.

eMPee584:
> Hmm, it'd be more consistent with behavior elsewhere if it just showed the
> number of files without the suffix unless the cumulative size has been calculated.

[...] OK, that's exactly what this patch does I think..

toonn:
Sorry for the bad review, this actually looks good to me.

PR #2543, opened 2022-01-11
runner/shell command: add -l flag to clear screen before running command

eMPee584:
This adds a flag to clear the screen before invoking a command.

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -550,6 +550,7 @@ command: (for example \f(CW\*(C`:shell \-w df\*(C'\fR)
 .Vb 3
 \& p   Redirect output to the pager
 \& s   Silent mode.  Output will be discarded.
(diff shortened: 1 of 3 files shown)


toonn (commented):
This looks okay to me. Can you describe a use case though?

eMPee584:
[…] well, use case is using the inbuilt command line often and with commands that clutter the screen.. It's much nicer to have a clean slate for every command invocation.. I actually use this as default.

eMPee584:
( Oh, when using the `-w` flag which I have just remembered is not the default.. )

toonn:
@eMPee584, so by "clearing the console" you mean the output window when using `shell -w` not the console/command line at the bottom of Ranger? Maybe we can change the documentation to say "Clear the output window before running the command"?

PR #2569, opened 2022-02-19
rifle.conf: add xviewer for image mime type

anakimluke:
- OS: Linux Mint 20.1
- ranger version: ranger 1.9.3
- python version: 3.8.10

#### CHECKLIST

    - [ ] Tests have been updated

On systems like Linux Mint 20.1 none of the alternatives for opening an image listed on `rifle.conf` are available; except for gimp. `xviewer` is available.


toonn (changes requested):
LGTM, though I think it should be a couple lines further down. How about just before `mcomix`? Matter of not changing behavior for anyone and since it comes with Mint by default it should be fairly common and thus further down.

anakimluke:
Sure. IIRC I put it there in a first place because of its similarity to Eye of Gnome. xviewer is a fork of eog.

toonn (approved):
LGTM, Better merge late than never, right? : )

Thank you for the PR. I'm hoping to get a better grip on the backlog so it won't take so long in the future.

PR #2574, opened 2022-02-21
Acknowledge and report the exit status of the rifle'd commands

vifon:
#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Currently `rifle` completely ignores the exit status of the commands it runs, even if it does wait for them (it's obviously not viable otherwise). This patch makes `rifle(1)` (the commandline application) exit with the same status the command did, and makes `rifle` the module raise `subprocess.CalledProcessError`, so it's displayed in the `ranger` status line.

#### MOTIVATION AND CONTEXT

As it stands, it was impossible to inspect whether the commands executed by `rifle` were successful or not.

#### TESTING

The errors are no longer swallowed and instead are reported back in the manner described above. If someone relied on them being swallowed, it might break their use case (but maybe for the better).

### DIFF
--- a/ranger/ext/rifle.py
+++ b/ranger/ext/rifle.py
@@ -18,6 +18,7 @@ from __future__ import (absolute_import, division, print_function)
 
 import os.path
 import re
+import shlex
 from subprocess import PIPE, CalledProcessError
 import sys
 
@@ -512,7 +513,7 @@ class Rifle(object):  # pylint: disable=too-many-instance-attributes
                     ) as process:
                         exit_code = process.wait()
                         if exit_code != 0:
-                            raise CalledProcessError(exit_code, cmd)
+                            raise CalledProcessError(exit_code, shlex.join(cmd))
             finally:
                 self.hook_after_executing(command, self._mimetype, self._app_flags)
 
--- a/ranger/ext/rifle.py
+++ b/ranger/ext/rifle.py
@@ -18,7 +18,7 @@ from __future__ import (absolute_import, division, print_function)
 
 import os.path
 import re
-from subprocess import PIPE
+from subprocess import PIPE, CalledProcessError
 import sys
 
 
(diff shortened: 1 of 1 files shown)


toonn (commented):
LGTM

Ranger still swallows these unless run with `--debug`, right?

vifon:
It swallows them loudly: with the red text in the notification area.

toonn (approved):
Tested and working.

Maybe we can render the list as a single string? If it's not too much effort. I think that'd be slightly easier for debugging because you could then simply copy the string to a terminal and go from there.

toonn (changes requested):
How about that improved error message? ; )

vifon:
Sorry for the delay. I still remember about it, I just need to get around to this.

ranger/ext/rifle.py
```diff
@@ -18,7 +18,7 @@
 
 import os.path
 import re
-from subprocess import PIPE
+from subprocess import PIPE, CalledProcessError
```
toonn, line 21:
```suggestion
import shlex
from subprocess import PIPE, CalledProcessError
```

ranger/ext/rifle.py
```diff
@@ -510,7 +510,9 @@ def execute(self, files,  # noqa: E501 pylint: disable=too-many-branches,too-man
                     with Popen23(
                         cmd, env=self.hook_environment(os.environ)
                     ) as process:
-                        process.wait()
+                        exit_code = process.wait()
+                        if exit_code != 0:
+                            raise CalledProcessError(exit_code, cmd)
```
toonn, line 515:
```suggestion
                            raise CalledProcessError(exit_code, shlex.join(cmd))
```

toonn (commented):
I think this is all we'd need? Feel free to either batch and apply the suggestions or to simply confirm and I'll do it locally : )

toonn (approved):
LGTM

PR #2586, opened 2022-03-12
rifle_sxiv: more correct fallback and add more img support.

N-R-K:
#### ISSUE TYPE
- Bug fix

#### CHECKLIST

      (Ran shellcheck and it doesn't report any problems)
    - [ ] Tests have been updated

### add some more image extensions

Imlib2 v1.8.0 adds loaders for svg, heif and avif.
ico support seems to have been there for a while.
add all these to known image extensions.

### properly fallback on non-valid extensions

originally we checked the extension to make sure the argument is an
image to avoid edge cases like this:

	./a        # is a dir
	./a.jpg

where running `rifle_sxiv a` would open up `a.jpg` instead of `a` the
directory.

however the is_img function was not correct as it would fail on upper
case extensions, such as `a.JPG`.

instead of doing extension check in two places, now we just have a
is_img_extension() function which checks for extension
case-insensitively via grep.

in short:

* this checks the extension case-insensitively
* reduces some code duplication
* sxiv will print an error msg for invalid files instead of the script
silently exiting

this is probably a tad bit slower, but IMO worth the trade.

### DIFF
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -33,7 +33,7 @@ tmp="$TMPDIR/sxiv_rifle_$$"
 
 listfiles () {
     find -L "///${1%/*}" \( ! -path "///${1%/*}" -prune \) -type f -print |
-      grep -iE '\.(jpe?g|png|gif|webp|tiff|bmp)$' |
+      grep -iE '\.(jpe?g|png|gif|svg|webp|tiff|heif|avif|ico|bmp)$' |
       sort | tee "$tmp"
 }
 
--- a/examples/rifle_sxiv.sh
+++ b/examples/rifle_sxiv.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Compatible with ranger 1.6.0 through 1.7.*
+# Compatible with ranger 1.6.0 through 1.9.*
 #
 # This script searches image files in a directory, opens them all with sxiv and
 # sets the first argument to the first image displayed by sxiv.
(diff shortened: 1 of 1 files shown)


examples/rifle_sxiv.sh
```diff
@@ -31,27 +31,27 @@
...
+    find -L "///${1%/*}" \( ! -path "///${1%/*}" -prune \) -type f -print |
+      is_img_extension | sort | tee "$tmp"
 }
 
 open_img () {
-    is_img "$1" || exit 1
-    trap 'rm -f $tmp' EXIT
-    count="$(listfiles "$1" | grep -nF "$1")"
+    # only go through listfiles() if the file has a valid img extension
+    if echo "$1" | is_img_extension >/dev/null 2>&1; then
```
toonn, line 45: Previously the behavior was that if we decided something's not an image, we exit. This changes that to rely on `count` not being set, right? That seems a bit too implicit to me and `rifle` relies on exit codes to disambiguate what to do with the output. Are you sure we're still getting the exit code `rifle` expects and an explicit `exit` wouldn't be more understandable?

toonn (commented):
Sorry for taking so long to get around to this, there's a long backlog and very little time to go through it : s

wantom85:
oh dont worry...i fixed it....thanks for the help

examples/rifle_sxiv.sh
N-R-K: > Previously the behavior was that if we decided something's not an image, we exit.

That's a behavior this PR is trying to correct. Sxiv is capable of opening directories, but the script would reject it due to extension check.

> `rifle` relies on exit codes to disambiguate what to do with the output. Are you sure we're still getting the exit code `rifle` expects and an explicit `exit` wouldn't be more understandable?

If `sxiv` fails to open the argument, it will exit with 1. I use this script as a standalone wrapper but I don't use ranger (and thus not rifle either).

However, if the concern is around exit code, the only difference this PR makes is that:

0. Previously if the extension wasn't "valid" we rejected right away and exited with `1`. This meant that certain valid inputs (image without extension/directory etc) would get rejected due to false positive on the extension check.
0. Now we let `sxiv` try to open it, if it doesn't succeed it will exit with `1`.

```console
# valid directory containing images

# before
$ rifle_sxiv ~/pic # rifle script rejects the input
$ echo $?
1

# after
$ rifle_sxiv ~/pic # sxiv opens the directory

# invalid input

# before
$ rifle_sxiv /tmp
$ echo $?
1

# after
$ rifle_sxiv /tmp
sxiv: no more files to display, aborting
$ echo $?
1
```

In other words, the exit code will still be `1` upon failure. Difference being that we let `sxiv` try to open it first instead of deciding to exit ourselves (which leads to some false positives).

toonn (dismissed):
LGTM, thanks for the contribution, and the patience!

N-R-K:
> LGTM, thanks for the contribution, and the patience!

Glad to help!

---

And for anyone else who uses sxiv_rifle as a standalone wrapper - there's a modified version of this script for nsxiv [available here](https://codeberg.org/nsxiv/nsxiv-extra/src/branch/master/scripts/nsxiv-rifle) which also contains some extra features like `file://` and `trash://` URI support.

I see that there's a lot of review pressure on ranger. And since `sxiv_rifle` is just an example, I won't be creating more review pressure trying to port those features over here :)

So just making this information public so anyone interested in this can take a look.

PR #2632, opened 2022-05-26
Fix width calculation when line numbers are enabled

toonn:
This fixes the width of files being off whenever a line number with an extra digit appears,
 shifting the width one item too early, shifting relative line numbers more than necessary and a trailing space on the file size.

@NoSuck, I created this PR because you did not seem interested in continuing discussion/development on PR #2346 after @markus-bauer's comments. Thank you for bringing the issue to my attention! A review of this PR would be really welcome.
 
 @markus-bauer, thank you for digging into this issue.  I'd appreciate a review. I considered your suggestion to replace `line_numbers` with `linum_text_len` in the key but it seemed like improperly overloading the meaning of that variable and I don't really see any harm in making this key a little bigger.

In the last commit I did add a fix related to the width to be used when relative line numbers are enabled, which is necessarily nearly always smaller. This has introduced enough logic in setting `linum_text_len` that we could just as easily set it to zero and have that be synonymous with disabling line numbers. If there's a reason for doing so. I don't think making the key one item bigger is much of a problem but maybe I'm wrong?

### DIFF
--- a/ranger/gui/widgets/browsercolumn.py
+++ b/ranger/gui/widgets/browsercolumn.py
@@ -275,7 +275,7 @@ class BrowserColumn(Pager):  # pylint: disable=too-many-instance-attributes
 
         # Set the size of the linum text field to the number of digits in the
         # visible files in directory.
-        linum_text_len = len(str(self.scroll_begin + self.hei))
+        linum_text_len = len(str(self.scroll_begin + self.hei - 1))
         linum_format = "{0:>" + str(linum_text_len) + "}"
 
         selected_i = self._get_index_of_selected_file()
--- a/ranger/gui/widgets/browsercolumn.py
+++ b/ranger/gui/widgets/browsercolumn.py
@@ -321,7 +321,7 @@ class BrowserColumn(Pager):  # pylint: disable=too-many-instance-attributes
                    drawn.path in copied, tagged_marker, drawn.infostring,
                    drawn.vcsstatus, drawn.vcsremotestatus, self.target.has_vcschild,
                    self.fm.do_cut, current_linemode.name, metakey, active_pane,
-                   self.settings.line_numbers, linum_text_len)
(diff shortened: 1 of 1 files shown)


markus-bauer:
I don't have much time for this right now, but here are two quick comments:
1) The calculation has to take the setting `one_indexed` into account.

Example:
Let's say self.hei=40.

scroll_begin = 0
0 indexed: line numbers 0 -- 39
1 indexed: line numbers 1 -- 40
self.scroll_begin + self.hei - 1 = 39
self.scroll_begin + self.hei = 40

scroll_begin = 60
0 indexed: line numbers 60 -- 99
1 indexed: line numbers 61 -- 100
self.scroll_begin + self.hei - 1 = 99
self.scroll_begin + self.hei = 100

2) You can also take into account how many files are actually shown.
There might be fewer files than self.hei:
```python
number_of_lines = min(self.hei, len(self.target))
scroll_end = scroll_begin + number_of_lines
```

toonn:
@markus-bauer, thanks for pointing out those two edge cases! Off-by-one errors are the bane of my existence o.o

zaeph:
Tested everything I could think off, and LGTM!

Reviewing now.  With `relative`, I find the visual shift distracting when you switch focus from the 99th to the 100th file (or bigger jumps with `J`/`K`), but Emacs does the same, so I’m going to assume this is standard.  I guess it’s preferable to computing the max-digits from the total number of files in the cwd.  If we really wanted to address it, we could make the minimum be 3 or 4 to reduce visual disturbance in most usecases.

toonn:
I'm gonna go ahead and merge this.

@NoSuck, that's a reasonable feature request. Feel free to submit an issue asking for it or a PR implementing it. (Do keep in mind the slow turnaround time though, I cannot promise getting around to it in a timely fashion.)

PR #2635, opened 2022-06-06
Fix "Loader work process failed" error when copying dangling symlinks

eMPee584:
#### ISSUE TYPE
- Bug fix

#### CHECKLIST

#### DESCRIPTION
When copying folders that contain dangling symlinks, ranger throws an error message, although the copying itself seems to succeed. This change fixes it. Tried to fix the pylint complaints, not sure about the other tests.

### DIFF
--- a/ranger/ext/shutil_generatorized.py
+++ b/ranger/ext/shutil_generatorized.py
@@ -49,9 +49,9 @@ else:
             pass
 
         # follow symlinks (aka don't not follow symlinks)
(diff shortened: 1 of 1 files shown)


eMPee584:
( To explain: the added `os.path.exists(src)` condition is only `true` when the symlink points to a valid target, so dangling symlinks will not be dereferenced. )

toonn (approved):
LGTM

I floated the exists check outwards because we don't want to follow_symlinks even if both src and dst are symlinks.

toonn:
Thank you for the fixes!

PR #2648, opened 2022-06-13
browsercolumn: Only add the separator when needed

jmcantrell:
When predisplay_right is empty (which is the case when vcs_aware is
off), the separator should be omitted. This fixes the error introduced
by my last commit.


toonn (approved):
LGTM, and this time I did do the barest amount of testing : )

toonn:
Thanks for reporting on this again!

jmcantrell:
Happy to. I love this tool. :D

PR #2654, opened 2022-06-15
ui: Guard terminal capabalities

### LINKED ISSUE #2587: ranger with `update_title` set crashes in terminal emulators that don't support/advertise statusline capabilities
#### Runtime Environment

```
$ ranger --version
ranger version: ranger 1.9.3
Python version: 3.10.2 (main, Jan 15 2022, 19:56:27) [GCC 11.1.0]
Locale: en_US.UTF-8
```

- Operating system and version: Up-to-date archlinux
- Terminal emulator and version: `foot version: 1.11.0 +pgo +ime +graphemes -assertions`

#### Current Behavior

If using a terminal emulator that does not support updating the window title (no statusline capabilities in terminfo), and `rc.conf` contains `set update_title true`, ranger does not start (crash).

#### Expected Behavior

Ranger should not crash but print a warning and not update the window title.

#### Context

AIUI whether a terminal emulator supports updating the window title or not is determined by the respective terminfo file. On archlinux and voidlinux, the regular `foot` terminfo is provided by ncurses (v6.3), but [currently lacks statusline capabilities](https://codeberg.org/dnkl/foot/issues/872) (this will be fixed in the next ncurses release according to the linked issue). This is how I found the issue.

#### Possible Solutions

On archlinux and voidlinux, install `foot-terminfo`, which installs foot's own terminfo as `foot-extra`. Then set `term=foot-extra` in `foot.ini` (or run foot with `--term=foot-extra`). Then foot advertises statusline capabilities correctly and ranger updates the window title, and does not crash.

#### Steps to reproduce

1. Use a terminal emulator that does not support updating the window title (no statusline capabilities in terminfo).
2. `set update_title true` in `rc.conf`
3. start ranger

#### Traceback

```
ranger version: ranger 1.9.3
Python version: 3.10.2 (main, Jan 15 2022, 19:56:27) [GCC 11.1.0]
Locale: en_US.UTF-8
Current file: '/home/camoz/bin'

  File "/usr/lib/python3.10/site-packages/ranger/core/main.py", line 201, in main
    fm.loop()
  File "/usr/lib/python3.10/site-packages/ranger/core/fm.py", line 377, in loop
    ui.redraw()
  File "/usr/lib/python3.10/site-packages/ranger/gui/ui.py", line 333, in redraw
    self.draw()
  File "/usr/lib/python3.10/site-packages/ranger/gui/ui.py", line 375, in draw
    curses.tigetstr('tsl').decode('latin-1'),
AttributeError: 'NoneType' object has no attribute 'decode'


```

toonn:
We call `decode` on the result of calls to `tigetstr` but that returns
`None` if the current terminfo does not support the capability we ask
for.

Fixes #2587

### DIFF
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -381,11 +381,14 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
             try:
                 fixed_cwd = cwd.encode('utf-8', 'surrogateescape'). \
                     decode('utf-8', 'replace')
-                escapes = [
-                    curses.tigetstr('tsl').decode('latin-1'),
-                    ESCAPE_ICON_TITLE
-                ]
-                bel = curses.tigetstr('fsl').decode('latin-1')
+                titlecap = curses.tigetstr('tsl')
+                escapes = (
+                    [titlecap.decode("latin-1")]
+                    if titlecap is not None
+                    else [] + [ESCAPE_ICON_TITLE]
+                )
+                belcap = curses.tigetstr('fsl')
+                bel = belcap.decode('latin-1') if belcap is not None else ""
                 fmt_tups = [(e, fixed_cwd, bel) for e in escapes]
             except UnicodeError:
                 pass


jmcantrell:
I can confirm that this PR does not crash ranger on foot terminal with the default term setting.

PR #2686, opened 2022-08-05
Fix flattened names

### LINKED ISSUE #2625: Vcs doesn't work correctly with flat.
#### Runtime Environment
ranger version: ranger-master v1.9.3-492-g391f061c

#### Current Behavior
When using flat, the vcs backend doesn't show the status of files correctly.

`container/directory.py` is not correct.

#### Steps to reproduce
- To create the image, I added an empty line to directory.py.
- run ranger with --clean
- activate vcs (I only tested git)
```
set vcs_aware true
set vcs_backend_git enabled
```
- `:flat -1`

---
By the way: 
Before trying to fix this (or any flat related issues), someone should have a look at flat again (https://github.com/ranger/ranger/issues/1085).

toonn:
I believe I've managed to figure out what was wrong with flattening.

Apologies for the low-effort PR description but I've run out of time and will be away until the 16th.

Any and all reviews and testing are more than welcome as this is a rather finicky problem domain :heart: 

I'm hoping, though not hopeful that this could also fix #2625.

Highlighting the important parts of the commit messages:

We can't set a `Directory` with multiple `basename_is_rel_to` values and it can appear simultaneously in both flattened and non-flattened directories, or differently flattened directories, so it can in fact be simultaneously rendered relative to multiple directories.

The `relative_path` property of an `FSObject` is now always semantically meaningful, relative to the current working directory. This does mean `relative_path` is no longer usable as a way to get an `FSObject`'s "display name."

The flat level of a directory changes the display names and therefore we cannot reuse cached `display_data` when the flat level changes.

To get a proper `infostring` we provide the column's target as a base to calculate relative paths to.

Changes to `Directory` mean the `relative_path` attribute is no longer a (bad) proxy for the way the object is displayed in the browser columns.

This means line modes can no longer use it as such. Using the path relative to the target of the current column, or the basename otherwise seems like an appropriate alternative.

While working on issues related to flattening I noticed unfortunate interplay between being in a hidden directory and the hidden filter. If we are in a hidden directory there's no point in hiding that directory from the parent directory. It causes confusion because the directory we are in does not appear in the parent and instead a different child of the parent is highlighted, leading us to potentially believe we are in a different directory.

I've also stopped filtering away hidden files and directories if they are highlighted, which makes moving left (to the parent) and then right (into the highlighted child) again a more idempotent operation.

~~EDIT: Need to fix up the various linting errors still.~~
EDIT: I changed `examples/plugin_linemode.py` because PyLint complains about the abstract base class and not reimplementing `filetitle` with the proper arguments. Ideally we wouldn't force any plugin filemodes to adopt the default parameter value and maybe not even the parameter at all.

### DIFF
--- a/examples/plugin_linemode.py
+++ b/examples/plugin_linemode.py
@@ -17,7 +17,7 @@ from ranger.core.linemode import LinemodeBase
 class MyLinemode(LinemodeBase):
     name = "rot13"
 
-    def filetitle(self, fobj, metadata):
+    def filetitle(self, fobj, metadata, relative_base=None):
         return codecs.encode(fobj.relative_path, "rot_13")
 
     def infostring(self, fobj, metadata):
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -140,16 +140,15 @@ class cd(Command):
         if self.arg(1) == '-r':
             self.shift()
             destination = os.path.realpath(self.rest(1))
-            if os.path.isfile(destination):
-                self.fm.select_file(destination)
-                return
         else:
             destination = self.rest(1)
 
         if not destination:
             destination = '~'
 
-        if destination == '-':
+        if os.path.isfile(destination):
+            self.fm.select_file(destination)
+        elif destination == '-':
             self.fm.enter_bookmark('`')
         else:
             self.fm.cd(destination)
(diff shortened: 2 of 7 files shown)


zaeph:
Ran out of time for testing this morning, but you’ve got two bugs:
1) History is FUBAR (probably a mixture of greedy `history_action` additions):

https://user-images.githubusercontent.com/12202828/202646226-0b877347-75cc-47d0-a547-93b37c26b4b0.mp4

2) Unpredictable ordering because of some init condition:

https://user-images.githubusercontent.com/12202828/202646192-e9da814d-89e8-473a-ae67-232fef16e624.mp4

Issue #2690, opened 2022-08-14
Transparent palette images issue PIL warning and render with a green background in kitty


gandalf3:
#### Runtime Environment

- Operating system and version: Arch
- Terminal emulator and version: kitty 0.25.2
- Python version: Python 3.10.5
- Ranger version/commit: fcf44f6f12506fa0096cb6468de6a2706bbfaf51
- Locale: en_US.UTF-8

#### Current Behavior
When attempting to preview .png files with palette color type and with transparent
I get this warning: ` /usr/lib/python3.10/site-packages/PIL/Image.py:992: UserWarning: Palette images with Transparency expressed in bytes should be converted to RGBA images` and the image is displayed with a green background.

#### Expected Behavior
Image renders with a transparent background.

#### Steps to reproduce
1. Open an image such as this one, which is a png in palette color mode with transparency:
<img src="https://user-images.githubusercontent.com/11167504/184516839-2ab6fa6d-da81-44e8-a7ed-f461f237a4fd.png" width="150">
3. Preview it in ranger in kitty (I wasn't able to test any of the other image preview methods, w3m and urxvt methods didn't work for me and I didn't try terminology).

PR #2691, opened 2022-08-14
Fix image display for palette images with transparency

gandalf3:
Even if image is not in RGB or RGBA format, it could still contain
transparency information.

I used the "transparency" key in the image info dict, see documentation here: https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html
Basically this key only exists for some formats in some modes, when the image contains transparency. I didn't test any formats besides a rudimentary test of palette mode .png; this change may also affect .eps, .gif, .gd, .xpm.

#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version:  Arch
- Terminal emulator and version: kitty 0.25.2
- Python version: 3.10.5
- Ranger version/commit: fcf44f6f12506fa0096cb6468de6a2706bbfaf51
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

Fixes: https://github.com/ranger/ranger/issues/2690

#### TESTING

I ran the tests and got a 9.99/10 reported failure rate; but AFAICT there aren't any tests which cover this change.

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -644,7 +644,7 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
                                  self.backend.LANCZOS)
 
         if image.mode not in ('RGB', 'RGBA'):
-            image = image.convert('RGB')
+            image = image.convert('RGBA' if image.info.get("transparency", False) else 'RGB')
         # start_x += ((box[0] - image.width) // 2) // self.pix_row
         # start_y += ((box[1] - image.height) // 2) // self.pix_col
         if self.stream:
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -644,7 +644,9 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
                                  self.backend.LANCZOS)
 
         if image.mode not in ('RGB', 'RGBA'):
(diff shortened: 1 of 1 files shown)


toonn:
Apologies for how long it's taken to get around to taking a look.

This seems like a pretty reasonable fix. I just pushed a commit which might simplify it further though.

The only case where my code will differ is if `info["transparency"] is None`, in which case my code converts to RGBA while the original converts to RGB. Can that key can ever map to `None` while a conversion to RGB is preferable?

Another simplification would be always converting to RGBA, if we can test a good amount of varied images in various formats and measure the difference in time and size of conversion to RGB/RGBA.

Let me know what you think?

gandalf3:
No worries! Maintaining an insanely popular project has got to be really hard, yall are amazing! 

I actually found both style of checks within pillow itself, and while skimming that documentation I don't see anything about `None` being special. The no transparency case seems to be consistently represented by the absence of the `transparency` attribute in the image `info`, at least as described in the documentation.

So TL;DR your change looks good to me!

toonn (commented):
LGTM

Thank you! Pretty good first contribution! 🎉 

I will try not to have it take so long to review and merge next time though.

PR #2696, opened 2022-08-28
using bash variable expansion  to get the lower case file extension i…

raphaelhenrique:
…nstead of tr command, for better performance

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: XUbuntu 20.04.4
- Terminal emulator and version: urxvt 9.22
- Python version: 3.8.10
- Locale: pt_BR.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

using variable expansion to get  file extension in lower case.
FILE_EXTENSION_LOWER="${FILE_EXTENSTION,,}"
instead of using "printf" and "tr" commands, for better performance

#### MOTIVATION AND CONTEXT

using pipe and "tr" command creates an unecessary new subshell. Plus, "tr" command is slower than variable expansion

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -36,7 +36,7 @@ IMAGE_CACHE_PATH="${4}"  # Full path that should be used to cache image preview
 PV_IMAGE_ENABLED="${5}"  # 'True' if image previews are enabled, 'False' otherwise.
 
 FILE_EXTENSION="${FILE_PATH##*.}"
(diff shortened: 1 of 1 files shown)


toonn:
Hi, thanks for the contribution and apologies for not getting around to it sooner.

How much of a difference does this really make performance wise? Attempts have been made in the past to make `scope.sh` POSIX sh compatible, it's not quite there yet but dropping all the "local" keywords and `[[ ]]` would reduce it to a handful of bashisms. So IMO we can only accept this if there's a big performance improvement to be had.

Note that alternative scope implementations that optimize for performance do already exist. I believe there's one implemented in Go.

PR #2703, opened 2022-09-17
update config/commands.py

### LINKED ISSUE #2575: Pressing tab on `scout` does nothing when there's no input.
#### Current Behavior
Pressing the Tab key on a `scout` command followed by some input cycles through the matching results. But it does nothing when there's no input. This can lead to inconsistent and confusing behavior for the user, especially when using find or travel.

#### Expected Behavior
Pressing tab should be consistent and cycle through the entries, even without input.

#### Possible Solutions
I think it happens at this check of `not pattern`:
https://github.com/ranger/ranger/blob/7a5f820e760c7ce2f93e2a5f0e91677b46252ea9/ranger/config/commands.py#L1706-L1707

Removing this check makes tab work:

```python
        # if not pattern or not cwd.files:
        if not cwd.files:
            return 0
        if pattern == '.':
            return 0
        if pattern == '..':
            return 1
```
Please try it out and see if the ux is better this way.

A working solution for the case of an empty pattern has to move the cursor and return a value that doesn't negatively interfere with existing code. AFAIK, `_count` is used in 3 places:
- In `scout.execute()` and `scout.quick()`, to determine if there's only one match, and if auto_open should trigger. This is checked with `_count() == 1`.
- In `scout.tab()`, to tab through the entries. There's no check for the returned value of `_count()`.

At the moment, an empty pattern returns 0. After the change it would return 2, 1, or 0. Only returning 1 would interfere with existing code. One example is this:
- enter a directory with only one entry
- evoke travel
- the entry will be opened (because _count returns 1)

One solution is this: You could still intercept this case at the top of the function, like before, but additionally move the pointer by one unit, and return a custom return value:

```python
    def _count(self, move=False, offset=0):
        ...
        if not cwd.files:
            return 0
        if pattern == '.':
            return 0
        if pattern == '..':
            return 1

        if not pattern:
            if move and len(cwd.files) > 1:
                cwd.move(to=(cwd.pointer + offset) % len(cwd.files))

            # This could return anything, as long as it doesn't interfere negatively with existing code:
            return -1
        ...

```
Another solution is to let the movement be handled in the loop like before, but do the returning of the correct value at the end.

Disclaimer: There might be other side effects I didn't consider.

---
### Update: Full working example
```diff
@@ -1699,17 +1699,25 @@ class scout(Command):
         return self._regex
 
     def _count(self, move=False, offset=0):
-        count = 0
         cwd = self.fm.thisdir
         pattern = self.pattern
 
-        if not pattern or not cwd.files:
+        if not cwd.files:
             return 0
         if pattern == '.':
             return 0
         if pattern == '..':
             return 1
 
+        files_len = len(cwd.files)
+
+        if not pattern:
+            if move and files_len > 1:
+                cwd.move(to=(cwd.pointer + offset) % files_len)
+            return -1
+
+        count = 0
+
         deq = deque(cwd.files)
         deq.rotate(-cwd.pointer - offset)
         i = offset
@@ -1718,7 +1726,7 @@ class scout(Command):
             if regex.search(fsobj.relative_path):
                 count += 1
                 if move and count == 1:
-                    cwd.move(to=(cwd.pointer + i) % len(cwd.files))
+                    cwd.move(to=(cwd.pointer + i) % files_len)
                     self.fm.thisfile = cwd.pointed_obj
             if count > 1:
                 return count
```

nadapez:
Fix  #2575 
Method scout._count updated.
Now moves the current pointer when pressing tab, even when there is no pattern.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1703,7 +1703,11 @@ class scout(Command):
         cwd = self.fm.thisdir
         pattern = self.pattern
 
-        if not pattern or not cwd.files:
+        if not cwd.files:
+            return 0
+        if not pattern:
+            cwd.move(to=(cwd.pointer + offset) % len(cwd.files))
+            self.fm.thisfile = cwd.pointed_obj
             return 0
         if pattern == '.':
             return 0


toonn (commented):
We should check the `move` flag.

Do I understand correctly that pressing enter on `:travel ` would *not* make this move into a highlighted directory? I think that's desirable behavior, start `:travel `, change your mind and tab to a directory instead, press enter.

PR #2707, opened 2022-09-24
[Feature] Add navigation scope

### LINKED ISSUE #2693: [FR] A way to prevent `l`/`<right>` from opening files, but still be able open directories
I find myself often opening files by accident because I pressed `<right>` one too many times (either because I was pressing it repeatedly or because I was confused about which column I'm in). This is inconvenient, since it often opens a GUI program or has my browser try to "download" a random file.

I'd like some solution to this. The best one I can imagine is some way to prevent `l/<right>/<mouse button>` from opening files, but have that open directories only (and, perhaps, have a different configurable action, e.g. open a pager). To open files, I'd use `<enter>` or perhaps ` <double-click>`.

I asked on IRC, and according to `zaeph`, there's currently no way to do this. They pointed to the following snippet as a place to look at for implementing it.

https://github.com/ranger/ranger/blob/fcf44f6f12506fa0096cb6468de6a2706bbfaf51/ranger/core/actions.py#L496

Thanks for making ranger!

Gigitsu:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version:  macOS 12.6
- Terminal emulator and version: iTerm2 3.4.16
- Python version: 3.10.6 (main, Aug 30 2022, 05:12:36)
- Ranger version/commit: fcf44f6f12506fa0096cb6468de6a2706bbfaf51
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Added `--scope` arguments to limit ranger browsing only inside the provided scope directory

#### MOTIVATION AND CONTEXT
Ranger can browse the entire filesystem. If used to navigate inside a project (e.g. as vim file explorer) is useful to limit ranger to browse only in a specific directory.

Fix #2693

### DIFF
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -44,12 +44,13 @@ class FM(Actions,  # pylint: disable=too-many-instance-attributes
     _visual_pos_start = None
     _visual_move_cycles = None
 
-    def __init__(self, ui=None, bookmarks=None, tags=None, paths=None):
+    def __init__(self, ui=None, bookmarks=None, tags=None, paths=None, root=None):
         """Initialize FM."""
         Actions.__init__(self)
         SignalDispatcher.__init__(self)
         self.ui = ui if ui is not None else UI()
         self.start_paths = paths if paths is not None else ['.']
+        self.root_dir = root if root is not None else '/'
         self.directories = {}
         self.bookmarks = bookmarks
         self.current_tab = 1
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -6,6 +6,7 @@
 from __future__ import (absolute_import, division, print_function)
 
 import atexit
+from email.policy import default
 import locale
 import os.path
 import shutil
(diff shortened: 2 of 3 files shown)


toonn (commented):
I appreciate the idea and I don't see any obvious flaws. But this sounds like a feature ripe for exploitation. I'm not sure we should try implementing something that sounds like `chroot`, we're bound to do it badly. What about `:shell` for example? That won't be limited by this at all.

Gigitsu:
Hi @toonn , thank you.

It's perfectly fine to change from `--scope` to `--roof` or `--top` (I personally prefer `--top`).

And I agree, this is absolutely not a security/sandboxing feature, I added this feature to help integrate ranger into other tools (ex. use it as a vim file navigator). 

It's not clear what you mean with:
`What about :shell for example?`

toonn:
I meant that commands run through `:shell` would have easy access past the top-dir.

If you rename the option and add some very explicit documentation this can move forward.

PR #2710, opened 2022-09-27
[Feature] Display count of files to be deleted

cibinmathew:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Python version: 
- Ranger version/commit: 
- Locale: 

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

while deletion, additionally display the total count of files/directories getting deleted in the prompt
image

#### MOTIVATION AND CONTEXT
Currently if there are large number of files, it helps the user to know how many files may get deleted

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -765,7 +765,7 @@ class delete(Command):
         confirm = self.fm.settings.confirm_on_delete
         if confirm != 'never' and (confirm != 'multiple' or many_files):
             self.fm.ui.console.ask(
-                "Confirm deletion of: %s (y/N)" % ', '.join(files),
+                "Confirm deletion of (total %s): %s (y/N)" % (len(files), ', '.join(files)),
                 partial(self._question_callback, files),
                 ('n', 'N', 'y', 'Y'),
             )
@@ -831,7 +831,7 @@ class trash(Command):
 
         if confirm != 'never' and (confirm != 'multiple' or many_files):
             self.fm.ui.console.ask(
-                "Confirm deletion of: %s (y/N)" % ', '.join(file_names),
+                "Confirm deletion of (total %s): %s (y/N)" % (len(files), ', '.join(files)),
                 partial(self._question_callback, files),
                 ('n', 'N', 'y', 'Y'),
             )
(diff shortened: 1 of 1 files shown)


ranger/config/commands.py
```diff
@@ -757,7 +757,7 @@ def is_directory_with_files(path):
         confirm = self.fm.settings.confirm_on_delete
         if confirm != 'never' and (confirm != 'multiple' or many_files):
             self.fm.ui.console.ask(
-                "Confirm deletion of: %s (y/N)" % ', '.join(files),
+                "Confirm deletion of (total %s): %s (y/N)" % (len(files), ', '.join(files)),
```
toonn, line 760:
```suggestion
                "Confirm deletion of {0} files: {1} (y/N)".format(len(files), ', '.join(files)),
```

ranger/config/commands.py
```diff
@@ -819,7 +819,7 @@ def is_directory_with_files(path):
         confirm = self.fm.settings.confirm_on_delete
         if confirm != 'never' and (confirm != 'multiple' or many_files):
             self.fm.ui.console.ask(
-                "Confirm deletion of: %s (y/N)" % ', '.join(file_names),
+                "Confirm deletion of (total %s): %s (y/N)" % (len(files), ', '.join(files)),
```
toonn, line 822:
```suggestion
                "Confirm deletion of {0} files: {1} (y/N)".format(len(files), ', '.join(files)),
```

toonn (commented):
This is a great idea!

One thing though, I believe `self.files` can include directories so maybe we need to recursively count the files or alternatively list counts for files and directories.

ranger/config/commands.py
hut: This makes `file_names` an unused variable, right? should either use it or remove it.

hut:
> This is a great idea!
> 
> One thing though, I believe `self.files` can include directories so maybe we need to recursively count the files or alternatively list counts for files and directories.

Recursing into directories sounds potentially extremely slow. It's bad enough when we do it for copy/pasting to calculate the progress bar. But here I'd rather avoid it.

Really important feature though!

hut (approved):

Issue #2717, opened 2022-10-14
Improving image preview performance


Microwave-WYB:
#### Runtime Environment

- Operating system and version: Manjaro Linux
- Terminal emulator and version: kitty 0.26.3
- Python version: 3.10.7
- Ranger version/commit: ranger-master
- Locale: en_US.UTF-8

#### Current Behavior
Image preview is slowing down file navigation. You have to wait until the preview is loaded before moving to the next file

#### Expected Behavior
Image preview should not slow down file navigation

#### Context

When browsing a lot of image files, especially when these files are large, it is very hard to navigate smoothly between files because you are always waiting for this huge preview to be loaded.

#### Possible Solutions
1. Interrupt image preview with file navigation
2. Generate smaller preview for large files to improve performance
3. Cache more image previews in the directory

#### Steps to reproduce
1. Open a folder containing large images
2. Navigate between files with `j` and `k`

Microwave-WYB:
I actually tried to implement a multi threaded image preview so that preview will not stop you from moving, but I'm encountering weird bugs when too many threads are spawned in a short time. Another possible thing we can do is to preprocess down-scaled image thumbnails in the background and store in a cache, instead of down-scaling the image each time the user is trying to access the image.

Friptick:
Any comments on this, @toonn ? It seems a clone of #1808, which you closed as unfixable.

I am using the Kitty implementation and, as @PoisonFrog says:
>Unfortunately it is just better to use ueberzug (ueberzugpp) as the image preview method. The kitty implementation in ranger is slow and crashes when you use commands.

The `ueberzug` method leaves ghost windows everywhere and is basically unusable, but it _is_ much faster.

I also use `feh` and that is also much faster at going thru a directory, even in full-screen mode. It does not seem to be caching the images (or at least I have not located the cache).

On a [related issue](#2834), you pointed out that:
>Another factor that'll slow down things like network filesystems is ranger's tendency to do a lot of stat calls.

Absolutely not an expert, but does this not all suggest that a major part of the sluggishness might come from those calls? I guess they are being used to populate the status bar, which suggests a quick fix - **a setting to disable the status bar**! In my personal workflow I hardly every look at the status bar anyway, would be annoying if that is what is slowing things down.

Thoughts?

NB. Thanks for all your work.

toonn:
@Friptick, no, if the stat calls were to blame then the Ueberzug method wouldn't be fast either.

[...]

PR #2718, opened 2022-10-15
Improve performance when browsing large image files with preview

Microwave-WYB:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Manjaro Linux (stable)
- Terminal emulator and version: kitty 0.26.3
- Python version: 3.10.7
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

In `gui/ui.py`, i created a new function called `image_thread_target`. As the name suggests, this is the target function of a image preview thread. The function does the same thing as the original `draw_images` function, except some checks before drawing the images. 

The original `draw_images` now calls `finalize` first, then spawn a image preview thread, targeting to function `image_thread_target`. 

#### MOTIVATION AND CONTEXT

When browsing large image files with preview on, the user must wait until the preview is fully rendered before navigating to the next file, which results in bad, laggy browsing experiment. This patch is intended to fix this issue #2717 opened by myself. 

In this patch to `gui/ui.py`, `UI.draw_image` no longer blocks user input. Instead, it spawns a thread to draw the preview image. The newly spawned thread will sleep for `0.05s` first before doing anything. If any other thread is spawned within this `0.05s`, the older thread will not draw preview, but return immediately, as the user has moved to a different file within a very short interval. This also prevents too many threads to be spawned within a short interval. 

I encountered weird previewing bugs when moving too fast between files. This seems to be addressed by running `self.finalize()` before spawning a new thread. I have no idea how that works but the bug disappeared after this fix. 

#### TESTING

I did `make test` and got this result, which is the same as the current `master` branch.
```
Your code has been rated at 9.91/10
```
I had a smooth experience in a lot of manual testing on folders containing large image files (>9M), and did not encounter any unexpected behaviors. 

This change should not affect any other areas of the codebase.

#### IMAGES / VIDEOS

https://user-images.githubusercontent.com/68882116/195982273-14ad08b3-8d4f-433e-93fd-4ab1c938b242.mp4


Microwave-WYB:
Just found a bug, when you navigate in the image folder and go back after the preview is drawn, files under the selected folder is missing.

Microwave-WYB:
Found a bug here, in terminals where image preview is not supported, it breaks everything. I need to add check to make sure the terminal is kitty

Microwave-WYB:
The improvements are now restricted to `kitty` terminal with `kitty` image preview mode.

toonn:
This is absolutely cool! Apologies for taking so long to get around to responding.

About the timeout, what happens when multiple threads are spawned outside of the timeout and then complete out of order? Is whatever thread happened to finish last going to determine the preview that is actually shown?
Maybe we need something like a thread pool instead? Where we can preempt the oldest thread if we need a new one?

I hope you're still interested in working on this, it's something we've been meaning to do since forever.

Issue #2521, opened 2021-12-15
AttributeError: 'NoneType' object has no attribute 'pointer'


gncs:
#### Runtime Environment

- Operating system and version: Fedora 35
- Terminal emulator and version: GNOME Terminal 3.42.1 using VTE 0.66.1 +BIDI +GNUTLS +ICU +SYSTEMD

#### Current Behavior

Ranger suddenly crashed. Probably related: there was a remote directory mounted with SSHFS and that mount point broke before.

#### Traceback

```
ranger version: ranger 1.9.3
Python version: 3.10.0 (default, Oct  4 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
Locale: en_US.UTF-8

Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/ranger/core/main.py", line 201, in main
    fm.loop()
  File "/usr/lib/python3.10/site-packages/ranger/core/fm.py", line 377, in loop
    ui.redraw()
  File "/usr/lib/python3.10/site-packages/ranger/gui/ui.py", line 334, in redraw
    self.finalize()
  File "/usr/lib/python3.10/site-packages/ranger/gui/ui.py", line 391, in finalize
    DisplayableContainer.finalize(self)
  File "/usr/lib/python3.10/site-packages/ranger/gui/displayable.py", line 268, in finalize
    displayable.finalize()
  File "/usr/lib/python3.10/site-packages/ranger/gui/widgets/view_base.py", line 59, in finalize
    col_y = self.main_column.y + self.main_column.target.pointer \
AttributeError: 'NoneType' object has no attribute 'pointer'

ranger crashed. Please report this traceback at:
https://github.com/ranger/ranger/issues
```

toonn:
Could you try running ranger from the master branch and see if it reproduces?

MasterTonLab:
I have the same issue on my Fedora 35.
OS: Linux fedora35comp 5.15.14-200.fc35.x86_64 1 SMP Tue Jan 11 16:49:27 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
ranger version: ranger 1.9.3
Python version: 3.10.1 (main, Dec  9 2021, 00:00:00) [GCC 11.2.1 20211203 (Red Hat 11.2.1-7)]
Locale: en_US.UTF-8

~/.config/ranger/rc.conf
set save_tabs_on_exit true

[...]

PR #2724, opened 2022-10-25
Prevent duplicate tab from being created on startup, manage tabs with "dead" path

awerebea:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: Linux Mint 21 (Vanessa)
- Terminal emulator and version: kitty 0.26.4, tmux next-3.4, zsh 5.8.1
- Python version: 3.10.6
- Ranger version/commit: ranger-master v1.9.3-527-g10d5b8f0
- Locale: en_US.UTF-8

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
With the `save_tabs_on_exit=true` option, every time Ranger is started without explicitly specifying which tabs to open, a new tab is first created for the current working directory, and then the previously saved tabs from the last session are added. So if the current directory was already in the list from the previous session, then it was saved twice on the next exit.
Try opening multiple tabs in Ranger, and then opening and closing Ranger from the same path multiple times in a row. You will see that every time you run it, a new tab will be created with the same path. Since it doesn't make much sense to keep multiple tabs with the same path, I suggest removing duplicates from `fm.start_paths` at startup.

UPD: Added filter for non-existent paths in the list of saved tabs.

UPD 2023.03.13:
Added logic to silently skip tabs with "dead" paths when switching them, or ask for confirmation to delete that tab according to the new `confirm_to_close_dead_tab` setting.
Added control of mouse events on tabs with "dead" paths.

#### MOTIVATION AND CONTEXT
Get rid of annoying duplicate tabs when you actively switch between terminal and Ranger with the `save_tabs_on_exit` option enabled.
Also, with this change, Ranger automatically removes non-existent paths from the fm.start_paths list at startup (optional, according to the new `filter_dead_tabs_on_startup` setting).

#### TESTING
This is a minor local change that does not affect other areas of the codebase in any way.
I did `make test` and all tests passed!


eMPee584 (dismissed):
Tested this and it works well. Also preserves order of tabs by using a dict instead of `set()`.
Fixes #2521.

ranger/core/main.py
```diff
@@ -161,6 +161,7 @@ def main(
                     with open(tabs_datapath, 'r', encoding="utf-8") as fobj:
                         tabs_saved = fobj.read().partition('\0\0')
                         fm.start_paths += tabs_saved[0].split('\0')
+                        fm.start_paths = list(filter(os.path.isdir, dict.fromkeys(fm.start_paths)))
```
toonn, line 164: I don't think dictionaries actually guarantee any ordering in the keys? As eMPee584 points out maintaining the ordering is desirable.

Also, can we maybe notify the user of any paths that are filtered out for no longer existing? It's possible a user forgot about mounting a partition and doesn't necessarily want to lose the saved tab. So being informed of tabs being removed and for which paths would be nice.

toonn (changes requested):
Great idea and gracefully handling missing paths is long overdue.

ranger/core/main.py
toonn: Actually, maybe we can even keep the tabs for non-existent directories. If we can keep Ranger from crashing when switching to tabs for which the path doesn't exist. That way you could keep tabs open dedicated to network volumes or removable drives.

awerebea: > maintaining the ordering is desirable

Maintain tab order while filtering out duplicates and dead entries

Keep only the first occurrence of a tab with a duplicate path, if a tab with the current launch path already exists, switch to that tab after launch instead of creating a new one at the top of the list.
34e813e8243bb8f7c42bf32799b7fb13ecd9a10e

awerebea: Regarding keeping tabs with non-existent paths (unmounted network volumes etc.), what should we do with them when switching tabs? Just skip it and go directly to the next one? Ok, but what about deleted dirs, which never get back again? How should we manage them? Of course, we can switch to tab and if its path doesn't exist go to some default one (user's $HOME for example), but what is the point if we destroy all these tabs with unmounted volumes this way?

There are many possibilities to revert to paths that didn't exist at startup if they become available again (built-in bookmarks, [fzf-marks](https://github.com/laggardkernel/ranger-fzf-marks), [fasd](https://github.com/gotbletu/shownotes/blob/master/ranger_fasd_fzf.md), [cd .. keymaps](https://github.com/ranger/ranger/blob/master/ranger/config/rc.conf?plain=1#L450-L466) for favorite directories, maybe something else...).
I mean, if we want to keep things simple, just clearing tabs with "dead" paths on startup is a good enough solution, IMO.

awerebea:
Added logic to silently skip tabs with "dead" paths when switching them, or ask for confirmation to delete that tab according to the new `confirm_to_close_dead_tab` setting.

Cleaning up "dead" tabs on startup is now optional due to the new `filter_dead_tabs_on_startup` setting.

Mouse events on tabs with "dead" paths are also managed.

ranger/core/main.py
toonn: Ah, interesting. I missed the change to dict, used to be ordered by the hashes of the objects if I'm not mistaken. We do support older versions than Python 3.6 but let's not have that hold us back in this instance and just add a note.

My desired behavior re tabs with unavailable paths is they simply display the usual "empty" or whatever placeholder Ranger always uses and not interfere with moving across tabs or moving tabs.
Testing on master puts dead tabs in the first still existing parent AFAICS, so we could change this to reflect the original path and show "empty" in the parents if they don't exist. But since it doesn't cause crashes it's OK if you'd rather leave this unaltered.

Removing duplicate paths isn't actually a good idea IMO, now that I've thought about it more. It can be totally valid for someone to always want multiple tabs that are in the same directory. I think the main issue is we always create a new tab when starting ranger? We could suppress doing so if the path is already open in a tab and move to that tab instead. We do need to consider multiple paths passed on the command line. If I pass multiple paths I do expect them to open in the order I pass them. If you already have them open in saved tabs they might be in a different order, we could look for the greatest overlap and then add missing tabs on either end but that seems fairly complicated.

Actually I noticed that Ranger seems to only restore tabs when no paths are passed on the command line so that may not be a problem. I'm not sure that's desirable behavior though, should we restore tabs even when paths are passed on the command line?

To avoid the duplicate directories, maybe we simply want to move to the PWD if it is already an open tab and no path is passed?

Don't feel a need to implement code whenever I say something BTW, let's discuss what we really want first to avoid extra work on your part.

awerebea: > let's discuss what we really want first

Hello! Yes, sure! I completely agree, and let's discuss. I just got some free time this weekend, and I'm enjoying improving my favorite console FM a bit. 😄

> Removing duplicate paths isn't actually a good idea IMO

It doesn't make any sense to me personally, but ok, I agree that duplicating tabs at startup can be useful to someone (although the content of these tabs will be exactly the same, I mean sorting, cursor position ... everything). If you really need a copy of the tab, you can just press Ctrl+n as soon as you switch to that tab and that's it. But again, I'm totally OK with keeping duplicates on startup, but maybe we can do that optional? Personally, I'm much more annoyed by switching and deleting duplicate tabs from a previous session than making a copy of a tab when I need it. For some, it might be the other way around. So maybe the user has to decide on this part of the startup approach?

> I think the main issue is we always create a new tab when starting ranger? We could suppress doing so if the path is already open in a tab and move to that tab instead.

In fact, this part is already implemented, so regardless of the optional deduplication at startup, we can already switch to the PWD tab if it already exists, instead of creating a new one (or switch to the first instance of existing tabs if we keep duplicates and have this tab more than once at startup).

> should we restore tabs even when paths are passed on the command line?

I don't think so, currently, we ignore everything else if we specify certain paths that we want to have in tabs on startup and I think that makes sense. However, it's up to you to decide, because it's absolutely not my use case. I never start Ranger with any arguments, but always just start it from PWD, get my saved tabs, and jump to where I need to.

> Testing on master puts dead tabs in the first still existing parent AFAICS, so we could change this to reflect the original path and show "empty" in the parents if they don't exist.

Maybe I didn't understand you completely, but it makes no sense for me to let the user switch to a tab with a "temporarily" unavailable path and show him "empty" or any other placeholder instead of the content of the desired dir. Switching to it and immediately jumping from there to the closest existing parent or any other path makes no sense as well, IMO. Because if we decided to keep tabs with unavailable paths to use them when these paths become available again, we don't really want to lose these tabs this way. So I still think that the best approach is to silently skip these tabs until they are "resurrected", or (optionally) ask the user whether to delete such a tab (if its path is never expected to be available again, for example).

awerebea: P.S.:
It just occurred to me: imagine that the user in charge of the annual financial report for a very important audit, on the eve of a major transaction for the company that depends on this report, switches to the network drive tab where this report is stored, which suddenly turned out to be unmounted from the file system and sees there `empty `. Who will take responsibility for his heart attack? :laughing:

toonn: Current behavior doesn't need that much changing I think. You can see what I'd expect on a dead tab by navigating somewhere in Ranger and then deleting a parent directory in another shell. Then run `:reset` in Ranger. It should show `empty` for the parent and `not accessible` for the current directory. The only part that could bear improvement IMO is that `:cd ..` or `h` will change to the first accessible parent instead of a single step of the path. But that change can be left for another PR.

The setting isn't good enough because the "safe" option, i.e., not simply deleting the tabs, *requires* attentive interaction every time you move over the tab, and you need to answer correctly every single time. It's not hard to accidentally hit `y` when you really meant `n`. Brainfarts happen.

That scenario is actually a vote in favor of what I'm suggesting. I'd rather realize the official copy isn't accessible right now, rather than accidentally skip to the next tab that has something potentially similar open and risk not realizing I'm not actually working on the important document.

If you're not interested in implementing this, that's absolutely fine and you can cut the PR down to just avoiding duplicate creation.

awerebea: I have made several attempts to implement the described approach, and the best I have come up with is the following behavior: 

Suppose we have some directory open in a tab:
image
image

Now we remove this directory from another tab and return to this tab:
image
image
We have received the desired `not accessible` message.

Here's what it looks like with a longer nested directory path:
image
image
image
image

If we create this directory again, the tab will come to life again:
image
image

So far everything looks good and I was almost happy with this "fix", but then I noticed some shortcomings.

1. If we launch Ranger in a state where the open tab path is "dead" and switch to that tab, we see only empty space with no columns with the message "not available".
image
2. If we try to create any file or directory in this "dead" tab, it will be created in the path of the previous "live" tab.
image
image

I tried to solve these problems many times over several days without success.
It seems my approach is not the correct way how we should achieve this.
I've pushed this WIP commit, in case you're interested in taking a look and suggesting how to fix it.
Otherwise, I will cut down this commit and leave only getting rid of the duplicate existing tab at startup.

awerebea: > Would you be interested in looking into this further?

Well, yes and no. I have already made several attempts to somehow improve what you saw here. But no luck, unfortunately. And I think I don't have enough free time to continue like this (reverse coding blindly). As I said before, I'm not a very experienced Python developer and not entirely familiar with the entire Ranger codebase, and I'm stuck. Therefore, I'm afraid that if no one supports me in this matter, I will not continue this fix, sorry.
If anyone else is interested in this later and decides to continue from now on, I can fork and keep these changes in another branch of my fork of this repo or even create another issue with that branch. But cut off everything but duplicate tabs on startup fix in current branch, so you can easily merge it.

toonn: Totally understandable. No need to keep this in a fork though. I intend to merge it regardless.

ranger/gui/widgets/titlebar.py
```diff
@@ -101,28 +101,40 @@ def _get_left_part(self, bar):
...
-            if path.is_link:
-                clr = 'link'
-            else:
-                clr = 'directory'
-
-            bidi_basename = self.bidi_transpose(path.basename)
-            bar.add(bidi_basename, clr, directory=path)
+        if self.fm.thisdir:
+            pathway = self.fm.thistab.pathway
+            if self.settings.tilde_in_titlebar \
```
toonn, line 106: I assume you've tried reducing the duplication here?

toonn (commented):
Looks like this is in pretty good shape to me. I took the liberty of pushing a minor refactor because you seem a little fatigued on this PR.

If you could squash all the reverted commits out of the history and rephrase the "WIP" commit message I think this is ready.

ranger/gui/widgets/titlebar.py
awerebea: Actually no, this change is related to the control of switching to "dead" tabs, more specifically, to avoid errors of calling a non-existent `pathway` attribute for a tab with `thisdir` of type `None` instead of the `Directory` class.

toonn: Yes, I understand but small parts of the code are duplicated here so I was wondering whether you'd attempted to reduce that?

awerebea: > If you could squash all the reverted commits out of the history and rephrase the "WIP" commit message I think this is ready.

Done!
I rebased the branch by squashing all commits related to closing dead tabs with a confirmation when switching between them since this logic was later discarded.
And improved the commit message for the initial version of dead-path tab switch control.

BTW, thanks for your refactoring. :wink:

awerebea: Another way to deduplicate code (left part of screen - "before", right - "after"):

Formally, we get rid of duplicating pieces of code, but we have to check the condition of the external if operator three times, which was previously checked only once.
In my opinion, this construction reads worse, so I would prefer not to push this commit.

toonn: Up to you, feel free to use your best judgement. Do you still consider this a draft? The changes look fine to me for the most part.

awerebea: By draft, I mean that there are still unresolved drawbacks and limitations in the dead-path tab switching control implementation, which I've described here and in the related commit message.  But as we discussed earlier, it seems that we can merge it in the current state and come back to improve the current implementation in the future.

toonn (approved):
Merging as discussed.

I did notice some rendering issues on the tab numbers when closing tabs. Not sure if that's a regression introduced here?

awerebea:
> I did notice some rendering issues on the tab numbers when closing tabs. Not sure if that's a regression introduced here?

What kind of issues? I didn't notice any on my end.

toonn:
I was on a tab with a long filename (I think that might be relevant) and some of the tab numbers stuck around after closing the tab, without a space between that number and the one for the next tab. I couldn't really reproduce it immediately though.

awerebea:
I see. Well, I'm pretty sure this issue is not related to this PR. At least it's hard for me to imagine which particular change here could cause the described behavior.

PR #2726, opened 2022-10-27
Introduce setting to hide infostring in parent directories column.

zebolino:
Added a setting to hide the infostring in the left column showing parent directories.

I am using https://github.com/alexanderjeurissen/ranger_devicons and _default_linemode devicons-sizemtime_. But I wanted to omit _sizemtime_ only in the left column.

#### ISSUE TYPE
- Improvement

#### RUNTIME ENVIRONMENT
- Operating system and version: Manjaro Linux 22.0.0
- Terminal emulator and version: st 0.8.2, tmux 3.3_a-2, zsh 5.9-1
- Python version: 3.10.7
- Ranger version/commit: ranger-master v1.9.3-526-g7171fe43
- Locale: de_DE.UTF-8

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
Just added a flag to hide the infostring in parent directories column (level < 0).

#### MOTIVATION AND CONTEXT
When the width of the left column is too small to show names and sizemtime I wanted to hide sizemtime (infostring).

#### TESTING
make test runs fine

#### ADDITIONAL INFORMATION
When I executed _make man_ there have been a lot of changes in _doc/ranger.1_ which would revert most changes of commit 7171fe43b7f3568fc8fcfc085077ae615010dc1a. So I left these changes out of this pull request.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -29,6 +29,9 @@ set viewmode miller
 # How many columns are there, and what are their relative widths?
 set column_ratios 1,3,4
 
+# Hide infostring in the parent directories column?
+set hide_infostring_in_parent_directories false
+
 # Which files should be hidden? (regular expression)
 set hidden_filter ^\.|\.(?:pyc|pyo|bak|swp)$|^lost\+found$|^__(py)?cache__$
 
--- a/doc/ranger.pod
+++ b/doc/ranger.pod
@@ -1081,6 +1081,10 @@ this pattern will hide all files that start with a dot or end with a tilde.
 
  set hidden_filter ^\.|~$
 
+=item hide_infostring_in_parent_directories [bool]
+
+Hide infostring in the parent directories column?
+
 =item hint_collapse_threshold [int]
 
 The key hint lists up to this size have their sublists expanded.
--- a/ranger/container/settings.py
+++ b/ranger/container/settings.py
@@ -46,6 +46,7 @@ ALLOWED_SETTINGS = {
     'freeze_files': bool,
     'global_inode_type_filter': str,
     'hidden_filter': str,
+    'hide_infostring_in_parent_directories': bool,
     'hint_collapse_threshold': int,
     'hostname_in_titlebar': bool,
     'size_in_bytes': bool,
--- a/ranger/gui/widgets/browsercolumn.py
+++ b/ranger/gui/widgets/browsercolumn.py
@@ -388,19 +388,20 @@ class BrowserColumn(Pager):  # pylint: disable=too-many-instance-attributes
             # info string
             infostring = []
             infostringlen = 0
-            try:
-                infostringdata = current_linemode.infostring(drawn, metadata)
-                if infostringdata:
-                    infostring.append([" " + infostringdata,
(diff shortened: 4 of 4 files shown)


ranger/config/rc.conf
```diff
@@ -29,6 +29,9 @@ set viewmode miller
 # How many columns are there, and what are their relative widths?
 set column_ratios 1,3,4
 
+# Hide infostring in the parent directories column?
+set hide_infostring_in_parent_directories false
```
toonn, line 33: I don't like the double negative. Let's flip this to show and default to true.

toonn (changes requested):
I think it could be cool to be able to specify linemodes for different column levels but this seems like a pretty useful first step.

ranger/config/rc.conf
zebolino: Just flipped the hide to show.

toonn:
Testing this out it doesn't actually change any behavior for me. Could you describe a more detailed scenario where the difference is observable? How wide should the parent column be exactly, for example?

zebolino:
> Testing this out it doesn't actually change any behavior for me. Could you describe a more detailed scenario where the difference is observable? How wide should the parent column be exactly, for example?


See these screenshots of my setup:
 
- Setting: `set show_infostring_in_parent_directories false`

screenshot-22Jul24-039116419

- Setting: `set show_infostring_in_parent_directories true`

screenshot-22Jul24-245002866

My column ratios are:
`set column_ratios 2,5,7`

toonn:
Finally getting back to reviewing PRs. It's still a bit wonky for me. If I change the linemode in a parent directory and then set it to hide the infostring in parent directories, it doesn't take effect. If I change the parent directory to a different linemode it does.

I also realize that I shouldn't have asked to switch the hide_* for show_* because bool settings that are not explicitly set default to `False`. So we have to go with hide_* to preserve the behavior for users who still have a local `rc.conf` that will not include the new setting.

zebolino:
> Finally getting back to reviewing PRs. It's still a bit wonky for me. If I change the linemode in a parent directory and then set it to hide the infostring in parent directories, it doesn't take effect. If I change the parent directory to a different linemode it does.

Can you explain that more? Did not get the problem.
 
> I also realize that I shouldn't have asked to switch the hide_* for show_* because bool settings that are not explicitly set default to `False`. So we have to go with hide_* to preserve the behavior for users who still have a local `rc.conf` that will not include the new setting.

I changed that back to `hide_infostring_in_parent_directories`

toonn:
Another shot at it: In a directory change the linemode to whatever, go into a child directory and set it to hide infostring in parent dirs, it doesn't take effect on the parent dir until you change its linemode. Ranger's probably caching the infostring and not being instructed to recompute it.

eMPee584:
Oh the caching and linemode implementation in general surely have got their sharp edges, just getting deep into it 😅🩸
@toonn does a `R`/`reload_cwd` do anything?

toonn:
`reload_cwd` doesn't seem to affect it but `^R` does. Though both of them interact badly with infostrings on plain files seemingly.

PR #2736, opened 2022-11-12
Don't strip commands of trailing spaces

Gelox:
It is nice be able to have trailing spaces in commands because it reduces the amount of keystrokes the user needs to use. For example if a user wants to create a shortcut for `touch` they'd like to have a trailing space which allows them to immediatly start typing the name of the file they are creating.

No longer remove trailing whitespaces from commands in rc.conf

#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Removed a single whitespace character from the strip function when reading commands from `rc.conf`.

#### MOTIVATION AND CONTEXT

Removing the trailing whitespaces from commands makes users have to press space after using a custom mapped command.
For example if you map "to" to "console touch " you will end up with "console touch" which will force you to press space.

I don't know of any relevant issues, this fix is so small that I figured it would be easier to do it myself than open an issue.

#### TESTING

All of them

It does not

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -375,7 +375,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
         # pylint: disable=unspecified-encoding
         with open(filename, 'r', encoding="utf-8") as fobj:
             for line in fobj:
-                line = line.strip(" \r\n")
+                line = line.strip("\r\n")
                 if line.startswith("#") or not line.strip():
                     continue
                 try:


Gelox:
I'm not sure why pylint is failing, it does not fail locally. I'm assuming it is not due to the single character change that I made and that it would be feature creep for this PR to try to fix this. If this is a wrong assumption then let me know.

toonn:
This was intentionally added in #720. A problem with trailing whitespace is that some editors/git configurations strip them. So we need some non-whitespace symbol for trailing whitespace anyway. I think if we added the option to quote strings so they will include the space without the final character on the line being actual whitespace, this could be accepted.

PR #2739, opened 2022-11-17
fm: Add nsxiv to sxiv workaround

### LINKED ISSUE #2671: Add nsxiv to sxiv_workaround_hook(...)
Sxiv now is unmaintained and [nsxiv](https://github.com/nsxiv/nsxiv) came as a replacement of it. It fixes some bugs (for example some gifs don't play in sxiv for me) and «adding simple, sensible features».

But ranger doesn't open all files in the current directory with this fork and [nsxiv-rifle](https://codeberg.org/nsxiv/nsxiv-extra/src/branch/master/scripts/nsxiv-rifle) doesn't keeping the order of files the same as in ranger.

I don't know about ranger's implementation details but seems like you just need to add `nsxiv` to [this](https://github.com/ranger/ranger/blob/391f061cb0b0cfa8266c0651f2a6d948f22e01dd/ranger/core/fm.py#L154) line and exactly the same condition block for it as for sxiv. nsxiv accepts exactly the same arguments.

```
re.match(r'^(feh|sxiv|nsxiv|imv|pqiv) ', command):

...

    if command[0:6] == 'nsxiv ':
        number = images.index(self.thisfile.relative_path) + 1
        new_command = command.replace("nsxiv ", "nsxiv -n %d " % number, 1)
```

As a temporary solution (for those who came across to this feature request) one can create a symlink to nsxiv:

`ln -s /usr/bin/nsxiv /usr/bin/sxiv`

and everything will work.

toonn:
Fixes #2671

### DIFF
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -161,7 +161,7 @@ class FM(Actions,  # pylint: disable=too-many-instance-attributes
                         "$@" in command:
                     new_command = None
 
-                    if command[0:5] == 'sxiv ' or command[0:6] == 'nsxiv':
+                    if command[0:5] == 'sxiv ' or command[0:6] == 'nsxiv ':
                         number = images.index(self.thisfile.relative_path) + 1
                         new_command = command.replace("sxiv ", "sxiv -n %d " % number, 1)
 
--- a/ranger/core/fm.py
+++ b/ranger/core/fm.py
@@ -144,14 +144,15 @@ class FM(Actions,  # pylint: disable=too-many-instance-attributes
         # The requirements to use it are:
         # 1. set open_all_images to true
         # 2. ensure no files are marked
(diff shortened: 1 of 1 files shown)


ranger/core/fm.py
```diff
@@ -160,7 +161,7 @@ def sxiv_workaround_hook(command):
                         "$@" in command:
                     new_command = None
 
-                    if command[0:5] == 'sxiv ':
+                    if command[0:5] == 'sxiv ' or command[0:6] == 'nsxiv':
```
gardockt, line 164: This PR fixes the issue for me - just a small typo needs to be fixed
```suggestion
                    if command[0:5] == 'sxiv ' or command[0:6] == 'nsxiv ':
```

toonn: Thanks!

PR #2750, opened 2022-11-23
img_display: Restore cleanup of Kitty tempfiles

### LINKED ISSUE #2699: kitty image preview temporary files not automatically removed after viewing on kitty 0.26.0+
#### Runtime Environment

- Terminal emulator and version: kitty 0.26.2
- Python version: 3.9.12 (05fbe3aa5b0845e6c37239768aa455451aa5faba, Aug 10 2022, 21:31:33) [PyPy 7.3.9 with GCC 12.1.1 20220730]
- Locale: en_AU.UTF-8

#### Current Behavior
Temporary files generated for image previews in kitty are never deleted.

#### Expected Behavior
Temporary files generated for image previews in kitty should be automatically deleted after being read by the terminal.

#### Context

After browsing some amount of images in ranger my `/tmp` folder fills up, most notably causing image previews to stop working. This can temporarily be resolved by deleting the generated files.
kitty 0.26.0 ([changelog](https://sw.kovidgoyal.net/kitty/changelog)) introduces a new restriction on deletion of temporary files, which is the cause of the issue:
> - Graphics protocol: Only delete temp files if they have the string `tty-graphics-protocol` in their file paths. This prevents deletion of arbitrary files in `/tmp`.

#### Possible Solutions
The source of the issue is this line:
https://github.com/ranger/ranger/blob/fcf44f6f12506fa0096cb6468de6a2706bbfaf51/ranger/ext/img_display.py#L669
I suggest changing `prefix='ranger_thumb_'` to `prefix='tty-graphics-protocol-'` or similar.

#### Steps to reproduce
1. Open ranger in kitty 0.26.0 or newer
2. set `preview_files=true`, `preview_images=true` and `preview_images_method=kitty`
3. Browse through images (/anything that has an image preview?)
4. Run out of space in `/tmp`

toonn:
Kitty version 0.26.0 stopped removing temporary files used in the Kitty Image Protocol unless their full path contains the string "tty-graphics-protocol," so we need to adapt to the new behavior.

By using a subdirectory of the directory used for temporary files with the appropriate name, Kitty should again delete the temporary PNGs we generate. An added benefit of using a subdirectory is avoiding cluttering up the directory for temporary files. This subdirectory is not cleaned up by Ranger and will remain as an empty directory (if previews work correctly) until removed by other means.

Fixes #2699

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -27,7 +27,7 @@ from collections import defaultdict
 import termios
 from contextlib import contextmanager
 import codecs
-from tempfile import NamedTemporaryFile
+from tempfile import gettempdir, NamedTemporaryFile
 
 from ranger import PY3
 from ranger.core.shared import FileManagerAware, SettingsAware
@@ -607,6 +607,17 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
         # if resp.find(b'OK') != -1:
         if b'OK' in resp:
             self.stream = False
+            self.tempFileDir = os.path.join(
+                gettempdir(), "tty-graphics-protocol"
+            )
+            try:
+                os.mkdir(self.tempFileDir)
+            except OSError:
+                raise ImgDisplayUnsupportedException(
+                    "Could not create temporary directory for previews : {d}".format(
+                        d=tempFileDir
+                    )
+                )
         elif b'EBADF' in resp:
             self.stream = True
         else:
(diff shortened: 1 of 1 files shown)


EvanCarroll:
Can we get review on this patch? What's holding it up? If this fixes the problem it'd be great. Does it need to be rebased. The failing CI stuff can't possibly because of the patch.

toonn:
@socksthefennec I've finally been able to incorporate your review, thank you so much!
Another review/testing would be welcome, also directed to @EvanCarroll.

socksthefennec (approved):
Tested with python 3.10.9 and seems to work as intended.

Issue #2007, opened 2020-06-21
Ranger won't launch


mtwb47:
Receiving the following error when launching Ranger. 

```
ranger version: ranger 1.9.3
Python version: 3.8.3 (default, May 17 2020, 18:15:42) [GCC 10.1.0]
Locale: en_US.UTF-8

Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/ranger/core/main.py", line 171, in main
    fm.initialize()
  File "/usr/lib/python3.8/site-packages/ranger/core/fm.py", line 127, in initialize
    self.bookmarks.load()
  File "/usr/lib/python3.8/site-packages/ranger/container/bookmarks.py", line 47, in load
    new_dict = self._load_dict()
  File "/usr/lib/python3.8/site-packages/ranger/container/bookmarks.py", line 237, in _load_dict
    dct[key] = self.bookmarktype(value)
  File "/usr/lib/python3.8/site-packages/ranger/container/directory.py", line 150, in __init__
    assert not os.path.isfile(path), "No directory given!"
AssertionError: No directory given!

ranger crashed. Please report this traceback at:
https://github.com/ranger/ranger/issues
```
This started happening without any updates to the system. It worked one time and then it did this the next time. 
OS: Arcolinux
Linux Kernel : 5.7.4
zsh 4.8
bash 5.0.17

mtwb47:
launching ranger with -c will allow it to launch. Don't know what that means though. I'm assuming something in rc.conf is what is causing this? 

--I'm going to go change it so the icons don't show up, I'm betting that's the issue. font awesome is always screwing things up

mtwb47:
Okay, even getting rid of the ranger config files and having ranger rebuild them with the defaults won't let ranger load without the -c flag. So it wasn't the icon or anything I added to the config.

toonn:
Looks like you have a non-existing dir bookmarked. Could be the `` ` `` bookmark for the last-visited directory. Probably easiest to remove the bookmarks file which should be in `${XDG_DATA_HOME}/ranger/bookmarks`.

mtwb47:
Yup that worked. I'm guessing this is less a bug, and more my fingers hit a key somewhere along the line without me knowing it.

toonn:
It's possible but we should just warn about the invalid bookmarks rather than crash.

hulmakerik:
I have exactly this issue too. The problem started after running the polybar. ranger -c works

[...]

PR #2754, opened 2022-11-29
Tolerate missing paths

toonn:
Bookmarks can sometimes point to invalid paths, be they on removable media or since replaced by files rather than directories. Ranger rather inelegantly crashes when this happens without notifying users about the reason at all. (Unless Python's run with the optimize flag in which case Ranger happily creates a `Directory` object that actually represents a file until such time when shenanigans happen.)

This PR makes instantiating a `Directory` object with a file path an exception, preventing shenanigans. It reports bookmarked paths that are invalid. And it tries to retain invalid bookmarks across Ranger invocations so the user can deal with them at their leisure.

### DIFF
--- a/ranger/container/bookmarks.py
+++ b/ranger/container/bookmarks.py
@@ -242,6 +242,7 @@ class Bookmarks(FileManagerAware):
                             try:
                                 dct[key] = self.bookmarktype(value)
                             except ValueError:
+                                dct[key] = str(value)
                                 self.fm.notify(
                                     "Bookmark {0} points to file '{1}', instead of a directory".format(
                                         key, value
--- a/ranger/container/bookmarks.py
+++ b/ranger/container/bookmarks.py
@@ -239,7 +239,14 @@ class Bookmarks(FileManagerAware):
                     if self.load_pattern.match(line):
                         key, value = line[0], line[2:-1]
                         if key in ALLOWED_KEYS:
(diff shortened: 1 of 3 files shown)


qrvd:
I only just saw this after working on a bugfix. Should we consider issue #2007 solved? :)

toonn:
@qrvd, thanks for finding this and bringing it up. This never got merged because I don't like doing so without any review or testing by others. In looking at the changes I do believe we should notify of the exception whenever a bookmark is not an existing directory, rather than only if it is an existing file. Your review would be appreciated!

Your approach also has merit though, especially because it allows rechecking whenever the bookmark is attempted to be accessed. Let me know whether you feel strongly one way or the other.

In a perfect world I think we'd allow file bookmarks maybe. Navigating to the bookmark would then involve changing to the directory it is in and then selecting it. I don't see a good reason to disallow this, though the practical implementation of selecting a file programmatically is stupidly hard in the current code.

hut:
I just added this bookmark in `~/.config/ranger/bookmarks`:

```
x:/tfsdkfjl/asdjfklsd
```

and started ranger, without any crash. Even `ranger --debug` doesn't produce any.

Am I missing something or is this fixed by now?

hut:
I'll move this out of the 1.9.5 milestone because I can't reproduce it.

PR #2788, opened 2023-02-15
Update actions.py

soraxas:
I was getting `None` type has no attribute `.encode`, because the function `read_text_file` can return `None`. 

See here:

https://github.com/ranger/ranger/blob/07e1ca3eb59714ec105d6a365cba711cfb90f079/ranger/core/actions.py#L1217-L1220

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1146,7 +1146,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 data['foundpreview'] = False
             elif rcode == 2:
                 text = self.read_text_file(path, 1024 * 32)
-                if not isinstance(text, str):
+                if text is not None and not isinstance(text, str):
                     # Convert 'unicode' to 'str' in Python 2
                     text = text.encode('utf-8')
                 data[(-1, -1)] = text


eMPee584:
You were getting that when you did _what_? A bit more context, please..

soraxas:
Before my changes (see file changed tab), it is written as
```python
text = self.read_text_file(path, 1024 * 32)
if not isinstance(text, str):
    ...
```
..., but `self.read_text_file` can return `None`.

So I'm adding a check for None. My apology, I thought a check for None is self-explanatory.

eMPee584:
Thanks for answering, but you didn't answer the actual question.. The code change is clear, but please describe a test case.

soraxas:
> Thanks for answering, but you didn't answer the actual question.. The code change is clear, but please describe a test case.

What? Hey dude, look at line 1220 that I've quoted. It literally reutns `None` bro, I don't know what would makes it makes more clear than what value it's returning.

eMPee584:
That's the code. What user interaction triggers the problem?

soraxas:
If it's a possible return value by-design then you should always check against it, no?

eMPee584:
Please understand that I am not an official ranger maintainer, so I do not know where this code is triggered from and was not questioning the validity of your fix, I just wanted to understand the context to properly test it. It is good practice to provide a minimal test case with a bug fix. You said: "I was getting `None` type has no attribute `.encode`", so why do you not want to share under which circumstances this error condition might occur?

toonn (approved):
I agree with @eMPee584 that knowing when this happened would be helpful but the change does look pretty simple and reasonable in this case.

Please take into consideration that any explanation, especially if it's a way to reproduce an issue, helps us to gain confidence in a code change, no matter how simple and is therefore really appreciated.

PR #2797, opened 2023-02-28
Remove the legacy filename filters, replace them with a compatibility layer utilizing the filter stack

vifon:
I've removed the original filtering mechanism ranger used to use, and replaced it with a thin compatibility layer that utilizes the newer filter stack mechanism. While I was at it, I also added live previews for the `:filter_stack add name` command just like the ones we have in `:filter`.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: NixOS 22.11
- Terminal emulator and version: Alacritty 0.11.0
- Python version: 3.10.9
- Ranger version/commit: N/A
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Assuming no explicit filter stack manipulation occurs, the semantics should be the same as before. If filter stack is being actively used, the compatibility layer either edits the topmost filter (if it's a name-based one), or adds a new element to the filter stack (all the other cases). Both such a new filter stack node and the legacy filtering mechanism are logically AND-ed with the rest of the filter stack, so the semantics are still mostly preserved.

The "inode type filters" are left intact for now. Considering they are both more complex (there are both global and local inode type filters) and less commonly used, it might be reasonable to just leave them be, at least for a time being.

It might be desirable to replace the internal "temporary filter" mechanism with a fully fledged "temporary filter stack node" that would be processed by the unified filter stack system. Possibly with a similar compatibility layer.

Some definitions got moved to `filter_stack.py` to prevent circular imports between it and `directory.py`. I believe these definitions should end up there anyway if we continue to unify the filtering system.

#### MOTIVATION AND CONTEXT

Currently ranger has 3-5 (depending on who you ask and how you count) independent filtering systems. They work, but it's confusing both for the users and for the maintainers. This is a start.

#### TESTING

I've manually tested `.n`, `zf` and the other filter stack keys to check if they still work correctly.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1815,6 +1815,18 @@ class filter_stack(Command):
             )
             return
 
+        # Cleanup.
+        self.cancel()
+
+    def quick(self):
+        if self.rest(1).startswith("add name "):
+            self.fm.thisdir.temporary_filter = re.compile(self.rest(3))
+            self.fm.thisdir.refilter()
+
+        return False
+
+    def cancel(self):
+        self.fm.thisdir.temporary_filter = None
         self.fm.thisdir.refilter()
 
 
--- a/ranger/container/directory.py
+++ b/ranger/container/directory.py
@@ -11,8 +11,8 @@ import re
 from collections import deque
 from time import time
 
-import ranger.core.filter_stack as filter_stack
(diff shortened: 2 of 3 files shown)


toonn (approved):
LGTM

Issue #2491, opened 2021-11-12
File sizes displayed in Ranger differ from file sizes displayed in `ls -l`


max-cx:
#### Runtime Environment

- Operating system and version: RHEL 8
- Terminal emulator and version: XTerm 331 (this issue is not related to the terminal!)
- Python version: Python 3.6.8
- Ranger version/commit: ranger 1.9.3
- Locale:  en_US.UTF-8
- 
#### Current Behavior
The rounded file sizes in Ranger differ from the unrounded file sizes in `ls -l`.
In Ranger:
![image](https://user-images.githubusercontent.com/78564304/141448384-8e14a50b-2043-4cfd-87cd-6b67269734bf.png)
In `ls -l`:
![image](https://user-images.githubusercontent.com/78564304/141448624-bb8cb7d0-2fe2-49ec-aa49-828138797b9f.png)

3.194 in `ls -l` ≠ 3.12 in `ranger`
3.185 in `ls -l` ≠ 3.11 in `ranger`

#### Expected Behavior
The rounded file size data in Ranger must match the unrounded file size data in `ls -l`.

#### Context

I see different file size data in `ls -l` and `ranger`.
It goes without saying that as the Ranger user, I trust the file size info in `ls -l` more than in Ranger, which also means that I now do not trust the file size info displayed in Ranger.

#### Possible Solutions
You're taking your file size data from a different place or using a different method or settings than `ls -l` does.
Please align file size info in `ranger` with `ls -l`.
Take the same unrounded file size number as `ls -l`.
There's probably an option for `ls` that does file size rounding, you can align it with that too.

#### Steps to reproduce
1. Open a directory with files in Ranger.
2. Open the same directory in `ls -l`.
3. Compare the rounded file sizes in Ranger to the unrounded file sizes in `ls -l`.

#### Traceback

```

```

toonn:
Nice catch, not sure what sort of weird behavior we have that rounds an 8 downwards.

max-cx:
@toonn I doubt this is a rounding issue, because also 3.194 is rounded to 3.2, not 3.12.

My bet is that the size-counting method Ranger using counts some extra bytes toward the file size.

seinmon:
@max-cx On my machine, the difference is due to the roundings of ls.

[...]

PR #2802, opened 2023-03-05
Changed the default unit prefixes for file sizes from binary to decimal

sz-hankus:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: macOS 12.6 (Darwin 21.6.0)
- Terminal emulator and version: iTerm2 3.4.19
- Python version: 3.11.2
- Ranger version/commit: ranger-master (`git describe` doesn't work)
- Locale: en_US.UTF-8

#### CHECKLIST
    - [x] Tests have been updated

#### DESCRIPTION
Ranger currently uses binary unit prefixes. In short, it means that there are 1024 bytes in a "kilobyte", 1024 "kilobytes" in a "megabyte" etc. According to the [IEC standard](https://en.wikipedia.org/wiki/ISO/IEC_80000#Part_13:_Information_science_and_technology), these units should actually be referred to as kibibytes, and mebibytes. The [wikipedia page](https://en.wikipedia.org/wiki/Binary_prefix) is very informative, so if you want to dig a little deeper, it's a great place to start. This pull request was inspired by issue [#2491](https://github.com/ranger/ranger/issues/2491).

The `human_readable` function converts the number of bytes to an appropriate string as follows:
https://github.com/ranger/ranger/blob/ccaf898a923b91c8546ec7e95a656bf0dfb45efc/ranger/ext/human_readable.py#L29-L56

As you can see, ranger uses the power-of-2 (binary) units for determining the appropriate prefix. From what I've read, nowadays most programs and systems use power-of-10 (decimal) prefixes, so I guess that ranger could follow suit.

We could add a global setting (e.g. `binary_size_prefix` that defaults to False) which would be read by the `human_readable` function each time it's called. This way, the user has control over which type of prefixes are used.  What I propose would look something like this:
```python
def human_readable(byte_count, separator=' '):
    # handle automatically_count_files false
    if byte_count is None:
        return ''
    if byte_count <= 0:
        return '0'
    if SettingsAware.settings.size_in_bytes:
        return format(byte_count, 'n')  # 'n' = locale-aware separator.

    binary = bool(SettingsAware.settings.binary_size_prefix)
    if binary:
        prefixes = ('B', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi')
        unit = 1024
    else:
        prefixes = ('B', 'k', 'M', 'G', 'T', 'P')
        unit = 1000

    ind = 0
    while byte_count >= unit:
        byte_count /= unit
        ind += 1

    format_str = '%.3g%s%s' if byte_count < 1000 else '%.4g%s%s'  # e.g. 1023 requires '%.4g'
    return format_str % (byte_count, separator, prefixes[ind]) if ind < len(prefixes) else '>9000'
```

There is one caveat - the comment above the current solution mentioning the performance. We could, of course, write an additional if statement and check for the appropriate prefix the same way it's done now, but with powers of 10 instead. However, what I've written is not that far off (in terms of performance) from the original, and looks cleaner. I tested it on 1 million random integers (ranging from 1 to 10^12 - a terabyte) and the current version took about 0.28s, while mine took about 0.35s - approximately 25% decrease in speed. I dug a little deeper and found the [previous version](https://github.com/ranger/ranger/blob/871c502d58055c7611f0763eaa71a7fafad7efdc/ranger/ext/human_readable.py) (what the author of the comment mentioned) and it took about 0.56s. Here's a [quick-and-dirty script](https://gist.github.com/sz-hankus/008b0fcf4b1f536249c995f39af37405), that I've written to test the performance. Let me know what you think :)

#### MOTIVATION AND CONTEXT
I think it would be great if ranger followed the [IEC standard](https://en.wikipedia.org/wiki/ISO/IEC_80000#Part_13:_Information_science_and_technology). It would provide greater consistency and eliminate confusion (which is what caused issue [#2491](https://github.com/ranger/ranger/issues/2491)).

#### TESTING
Because the default behavior is different than before, I had to change the tests to accommodate the new defaults. Previously, there were 3 test-cases. I've added the same 3 test-cases, but with `binary_unit_prefix` set to False, which changes the output.


sz-hankus:
Having read [1891](https://github.com/ranger/ranger/pull/1891), it seems like the issue was using `%` formatting instead of the `.format` method. It slows thing down a bit, but if you're fine with it, then I can use `.format` as well. I can run some tests to see how my implementation performs with `.format` if you'd like.

What my pull request was targeting was the several if statements, which would have to be doubled if we decided to give the user an option to display either binary or decimal prefixes. I chose a different approach, so I guess the if statements would have to go.

ranger/ext/human_readable.py
```diff
@@ -8,52 +8,49 @@
...
-        return '%.3g%sT' % ((byte / 2**40), separator)
-    if byte < 2**50:
-        return '%.4g%sT' % ((byte / 2**40), separator)
-    if byte < 2**50 * 999:
-        return '%.3g%sP' % ((byte / 2**50), separator)
-    if byte < 2**60:
-        return '%.4g%sP' % ((byte / 2**50), separator)
-    return '>9000'
+    # If you attempt to shorten this code, take performance into consideration.
+    binary = bool(SettingsAware.settings.binary_size_prefix)
```
toonn, line 39: Can we maybe turn this into a default argument to the function? That way it'd be easier to override if need be.

toonn (commented):
Some performance tests would be highly appreciated. As long as we can stay within the same order of magnitude it's probably fine but it would be nice if we don't have to drop the previous work by the wayside.

ranger/ext/human_readable.py
sz-hankus: I tried it, and it threw an error `type object 'SettingsAware' has no attribute 'settings'`. That's probably due to the fact that when the method is defined and the default arguments are evaluated, `SettingsAware` has not yet had its method `.settings_set` called (so it doesn't contain the `.settings` attribute). I'm not an expert in the codebase, so there might be something different going on. I'm not sure.

toonn: Ah, darn, then I'd say to go ahead with `None` as the default argument and checking with `is None`.

sz-hankus: @toonn I pushed the changes, but forgot about the "line too long" pylint error. Maybe I should just turn this long conditional expression into a proper if statement?
```python
binary = use_binary if use_binary is not None else bool(SettingsAware.settings.binary_size_prefix)
```
Or maybe just omit the `bool()` call?

toonn: I believe we can drop the `bool()`. If you declare the setting as bool Ranger should take care of converting regardless of case and I believe yes/no should also work.

And I think flipping the conditional is also slightly clearer.

`binary = SettingsAware.settings.binary_size_prefix if use_binary is None else use_binary`

sz-hankus:
@toonn Sorry for the delay but I finally got around to testing the performance. I updated the [script](https://gist.github.com/sz-hankus/c8c2ef960f62ea958db687c91fe42375) to run each version of the function for 1 million randomly generated integers ranging from 1 to 10^12 (a Terabyte) 10 times. I then collected the results and printed out the mean and standard deviation.

The results look as follows:
```python
curr:   avg=0.2553052960050991  std=0.0010652014755020084
this:   avg=0.3783249541942496  std=0.0015415990498268597
other:  avg=0.4652608875039732  std=0.002089931178004608
```
`curr` being the version in the current codebase, `this` - the version in this pull request and `other` - the version from 
[#1891](https://github.com/ranger/ranger/pull/1891)

The results for `this` were pretty much exactly the same when using the argument `use_binary=True` or `use_binary=False` or not specifying `use_binary` at all (For that, I created a mock `SettingsAware` object)

toonn (approved):
LGTM, let's merge this in favor of #1891

PR #2805, opened 2023-03-09
Unstuck tasks

toonn:
Building on work by @kevinhwang91 in #2197 to fix blocking on previews.
Adding improvements to error reporting and clearing the throbber when the task queue is emptied.

Testing would be appreciated!

### DIFF
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -229,7 +229,7 @@ class CommandLoader(  # pylint: disable=too-many-instance-attributes
                 selectlist.append(fd_out)
             if not self.silent:
                 selectlist.append(fd_err)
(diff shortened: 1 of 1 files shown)


huyz:
@toonn can you give more details about what symptoms this PR is fixing?  Every time I try to quit ranger it says I have preview tasks still running. I always have to `:qa!` or go into the task view to kill the preview tasks.  And I'm not sure if this PR addresses my issue.

toonn:
I don't think this addresses that particular issue. This is about clearing the queue so the status bar no longer indicates tasks in progress if they aren't.

toonn (commented):
Mostly cosmetic improvements so I'm self-merging this.

PR #2809, opened 2023-03-19
Fix the ABC checks on Python 3

vifon:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: N/A
- Terminal emulator and version: N/A
- Python version: Python 3.10.9 and Python 2.7.18.6
- Ranger version/commit: N/A
- Locale: N/A

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION, MOTIVATION AND CONTEXT

The previous solution used the ABCMeta metaclass directly as the ABC non-meta class didn't exist in Python 2.

Python 2 and 3 have different syntax for metaclasses.  The one used by Python 2 doesn't work on Python 3 but is not an error either.  The one used by Python 3 is an error in Python 2.  We used to use the Python 2 syntax while acknowledging it merely doesn't do anything on Python 3.

The regular class ABC allows to avoid using the metaclass syntax at all, but it's only available in Python 3.  This solution implements it on our side (using the Python 2 metaclass syntax) if it's not available, but otherwise uses the standard ABC class with regular inheritance.

#### TESTING

Manual ranger startup on both Python 2 and Python 3. Then I've removed once mandatory method that `abc` should catch check to see if ranger will refuse to start. The checks are now working on both Python 2 and Python 3.

### DIFF
--- a/ranger/core/linemode.py
+++ b/ranger/core/linemode.py
@@ -11,10 +11,17 @@ from datetime import datetime
 from ranger.ext.human_readable import human_readable, human_readable_time
 from ranger.ext import spawn
 
+try:
+    from abc import ABC
+except ImportError:
+    class ABC(object):
+        __metaclass__ = ABCMeta
+
+
 DEFAULT_LINEMODE = "filename"
 
 
-class LinemodeBase(object):
+class LinemodeBase(ABC):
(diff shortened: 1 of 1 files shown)


toonn (approved):
LGTM

PR #2816, opened 2023-03-29
get_executables: Use platform instead of reading /proc

### LINKED ISSUE #2815: Recent in_wsl check tries to read a path in `/proc` which only exists in Linux-like environments
```
ranger version: ranger-master v1.9.3-613-g1a70e7f4
Python version: 3.11.1 (main, Feb 11 2023, 14:35:02) [Clang 11.1.0 ]
Locale: en_US.UTF-8

  File "/Users/toonn/src/ranger/ranger/core/main.py", line 174, in main
    fm.initialize()
  File "/Users/toonn/src/ranger/ranger/core/fm.py", line 133, in initialize
    self.ui.initialize()
  File "/Users/toonn/src/ranger/ranger/gui/ui.py", line 137, in initialize
    self.handle_multiplexer()
  File "/Users/toonn/src/ranger/ranger/gui/ui.py", line 508, in handle_multiplexer
    elif _in_screen():
         ^^^^^^^^^^^^
  File "/Users/toonn/src/ranger/ranger/gui/ui.py", line 60, in _in_screen
    and 'screen' in get_executables())
                    ^^^^^^^^^^^^^^^^^
  File "/Users/toonn/src/ranger/ranger/ext/get_executables.py", line 20, in get_executables
    _cached_executables = get_executables_uncached()
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/toonn/src/ranger/ranger/ext/get_executables.py", line 43, in get_executables_uncached
    in_wsl = _in_wsl()
             ^^^^^^^^^
  File "/Users/toonn/src/ranger/ranger/ext/get_executables.py", line 26, in _in_wsl
    with open('/proc/sys/kernel/osrelease', encoding="utf-8") as file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/proc/sys/kernel/osrelease'
```

toonn:
Not every system has a `/proc` file system. We could catch the error but it is simpler to check `platform.release()`. This string contains `WSL2` for WSL version 2 but the only common substring for version 1 is lower case `microsoft`. On Windows this string should simply contain a major version number and not include that substring.

Fixes #2815

### DIFF
--- a/ranger/ext/get_executables.py
+++ b/ranger/ext/get_executables.py
@@ -3,9 +3,10 @@
 
 from __future__ import (absolute_import, division, print_function)
 
-from stat import S_IXOTH, S_IFREG
 from os import listdir, environ, stat
+import platform
 import shlex
+from stat import S_IXOTH, S_IFREG
 
 from ranger.ext.iter_tools import unique
 
@@ -23,8 +24,9 @@ def get_executables():
 
 def _in_wsl():
     # Check if the current environment is Microsoft WSL instead of native Linux
-    with open('/proc/sys/kernel/osrelease', encoding="utf-8") as file:
-        return 'microsoft' in file.read().lower()
+    # WSL 2 has `WSL2` in the release string but WSL 1 does not, both contain
+    # `microsoft`, lower case.
+    return 'microsoft' in platform.release()
 
 
 def get_executables_uncached(*paths):


toonn:
@awerebea, since you contributed this code, could you test to verify this fix still works for WSL?

Issue #2511, opened 2021-11-30
Btrfs copy deduplication


manujchandra:
Hi,

I am using Fedora 35 XFCE edition. One thing I have noticed:

I copied a 27 GB folder in Thunar in the same drive, the GUI file manager of XFCE. It took 4 seconds to copy.

If I run `sudo btrfs fi du ~`

It shows 27 GB is shared. In other words, only metadata was copied, not the actual data.

But if I use yy and pp on the same folder in Ranger, it takes a few minutes to copy to the same drive. It means ranger is duplicating the files. I can see the free space also decreases.

I found [here](https://unix.stackexchange.com/questions/249772/how-to-duplicate-a-file-without-copying-its-data-with-btrfs)

that we can use `cp --reflink=auto` to copy with deduplication.

Is it possible to integrate this behavior of Thunar in Ranger where it copies with de-duplication?

Thanks! 🙏🏽

toonn:
The easiest way is probably to simply create a key binding, `map pd shell cp --reflink=auto %c %d` would pass the contents of the copy buffer as a list of paths to the cp command and `%d` is the current directory in ranger.

manujchandra:
This thought crossed my mind. But it wont appear in the task list as per my knowledge.

If someone knows where the code of copy and paste commands are in Ranger, then maybe I can look into how to modify that to support reflinks. It must be using some internal command to copy files.

toonn:
It's either in `actions.py` or `commands.py`.

manujchandra:
I am leaving these links here just for future reference:

https://stackoverflow.com/questions/65492317/copy-file-in-python-with-copy-on-write-cow

https://bugs.python.org/issue37157

I will investigate them later.

manujchandra:
Leaving this here for future reference:

https://pypi.org/project/reflink/

```
from reflink import reflink

# Reflink copy 'large_file.img' to 'copy_of_file.img'
reflink("large_file.img", "copy_of_file.img")
```

ayouknowwho:
I was having a look at this, sorry I'm very new to python and this project so I'm just exploring. I made some notes on this issue:

ranger/core/actions.py contains the actions 'cut', 'copy', 'paste'
'copy' writes the selected filenames into self.copy_buffer
'cut' simply calls copy and sets self.do_cut = True
'paste' checks the destination is valid then calls CopyLoader from ranger/core/loader.py

CopyLoader initialises and then calls its 'generator' method
[...]

PR #2832, opened 2023-05-03
Use os.copy_file_range where available to enable copy-on-write

ayouknowwho:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
In a VM, python 3.9.2
Locale en_GB.UTF-8
Debian Bullseye
Termit (KDE)
Latest ranger version from git with below commits

#### CHECKLIST
    - [X] Tests have been updated

#### DESCRIPTION
The implementation of file copying was not compatible with copy-on-write functionality. By using os.copy_file_range which is available in Python 3.8+ we can enable this functionality.
A new configuration setting which defaults to False enables the functionality when set to True. A check that the configuration setting is True AND the PY3 is True is performed before diverting flow to os.copy_file_range rather than using python file buffers.
CopyLoader was made SettingsAware so that the setting information can be passed down through function call parameters to copyfileobj in shutil_generatorized

#### MOTIVATION AND CONTEXT
issue https://github.com/ranger/ranger/issues/2511

#### TESTING
make test passes all tests
with the setting off a 500MB file transfer takes several seconds
with the setting on a 500MB file transfer takes less than a second

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -133,7 +133,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.9.3" "2023-05-01" "ranger manual"
+.TH RANGER 1 "ranger-1.9.3" "2023-04-25" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
 .\" way too many mistakes in technical documents.
 .if n .ad l
@@ -1011,9 +1011,6 @@ on top of those for \fIdraw_borders\fR:
 .IX Item "draw_progress_bar_in_status_bar [bool]"
 Draw a progress bar in the status bar which displays the average state of all
 currently running tasks which support progress bars?
-.IP "enable_copy_on_write [bool]" 4
-.IX Item "enable_copy_on_write [bool]"
-Enable Copy-on-Write functionality (if supported by os.copy_file_range)
 .IP "filter_dead_tabs_on_startup [bool]" 4
 .IX Item "filter_dead_tabs_on_startup [bool]"
 Remove tabs with unavailable paths on startup. Setting this to false (the
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -133,7 +133,7 @@
 .\" ========================================================================
 .\"
 .IX Title "RANGER 1"
-.TH RANGER 1 "ranger-1.9.3" "2023-04-25" "ranger manual"
+.TH RANGER 1 "ranger-1.9.3" "2023-05-01" "ranger manual"
 .\" For nroff, turn off justification.  Always turn off hyphenation; it makes
 .\" way too many mistakes in technical documents.
 .if n .ad l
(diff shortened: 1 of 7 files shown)


ayouknowwho:
I have realised there is an additional bug that will need fixing, os.copy_file_range only works if both files are on the same filesystem, which we can't guarantee, so I do still need to enable it to fallback to the original method.

toonn:
@ayouknowwho, can you switch the PR to "draft" until you've addressed that issue?

ayouknowwho:
Should be able to get this done on Monday

ayouknowwho:
I have now updated to revert to the old method in any cases where copy_file_range raises an error. I have tested on several different file system combinations successfully.

toonn (commented):
I haven't found a way to check whether a file is indeed a reflinked copy of another. How can we make sure that's what's happening?

I'm somewhat coming around on making this default behavior, FYI. However, I think we'd need changes to how settings work to be able to do that.

I also realized we already have the paste symlink and hardlink commands and key mappings, so I think we should first add a key map for pasting with reflinking and then afterwards we can add a setting like implemented here. You can either open a new PR adding the binding and rebase this one afterwards or change this one. Or we can discuss further if you disagree, I'm clearly open to changing my mind on this >.<

ayouknowwho:
There's a couple of methods that could be used to check whether something was reflinked indirectly:
A: Copying a large file on a small disk without enough space for the hard copy and see if it completes
B: Check disk usage before and after copy, ensure they are the same (works best with big file on a small disk)
I'm not sure if there's a better way, but there could well be.

There are advantages to using copy_file_range where available outside of the copy-on-write functionality, as it is all done in kernel, which supposedly offers performance benefits, and I quite like how the code works now trying this method first and falling back to the old method when we get errors.

I'm not sure about allowing the user to purposefully choose between reflinking and hard copying data, above simply enabling or disabling the functionality, what do you think the use case is for that choice being useful file-by-file? Surely you'd always want to reflink when possible and hard copy if reflinking is not available? I'm interested to hear your thoughts.

If you want to make it default behaviour I guess you can either have a disable_copy_on_write setting instead or drop the setting altogether (keeping the PY3 check I think so we aren't trying to call a function that doesn't exist)

ayouknowwho:
I merged in the latest changes so the pull request is still up to date. There didn't seem to be any conflicts. Have you had any more thoughts on how you'd like to proceed?

toonn:
You've convinced me that this should just be the default behavior. If it's good enough for `cp(1)` it should be good enough for us.

iambumblehead:
cc @ayouknowwho

ayouknowwho:
Sorry for the delayed response, I've been away on holiday, I get back later this week so I'll review properly then

ayouknowwho:
Yes this seems like a nice simple solution, it doesn't disadvantage users in old environments heavily, since as soon as a single block fails it uses the old method from then on.

ayouknowwho:
As far as I was aware it was ready to be merged.

toonn:
I still haven't found a good way to verify the files created are really reflinks. But I'm taking @ayouknowwho's last comments as approval of my changes so I'm going ahead and merging. Release the Kraken^Wcopy-on-write behavior!

Issue #2846, opened 2023-05-20
Ranger can't preview images in kitty with termux even though it now seems possible.


RexAckermann:
#### Runtime Environment

- Operating system and version: Fedora 38
- Terminal emulator and version: kitty 0.28.1
- Python version: Python 3.11.3
- Ranger version/commit: ranger 1.9.3
- Locale: 

#### Current Behavior
kitty with tmux can't preview images
#### Expected Behavior
Should be able to now because it is now possible to preview images with termux in kitty.

#### Context

Can't get image preview in kitty tmux.
#### Possible Solutions
Not exactly sure.But reference to the fact that tmux on kitty can now preview images: [Reference](https://github.com/gokcehan/lf/issues/1143)
``` 
I tried the kitty +kitten icat <path-to-image>
```
And it seems to work fine.
#### Steps to reproduce
1.open kitty 
2. initiate tmux
3. try to preview image file

#### Traceback

```

```

toonn:
You're not running the Nightly build of Kitty that's mentioned in the linked issue.

RexAckermann:
Thanks for responding.
So, Can it be fixed ? Just commenting it out didn't solve the issue ?

otakutyrant:
@toonn Looking forward to your fix. You have my appreciation from the bottom of my heart.

toonn:
I'm not working on this, Ethsan is in #2856.

kohane27:
I've been waiting for tmux+kitty+ranger+wayland support for a long time. Then I just realized [ueberzugpp](https://github.com/jstkdng/ueberzugpp) supports tmux on sway. It just works out of the box: https://github.com/ranger/ranger/wiki/Image-Previews#with-ueberzug

hut:
@otakutyrant It should be fixed now thanks to https://github.com/ranger/ranger/pull/3086 . Can you check on the master branch?

otakutyrant:
@hut It is fixed. Just note that PILLOW and `tmux set -g allow-passthrough on` are required. Thank you.

Issue #1603, opened 2019-06-22
Freeze when executing shell command while Kitty preview visible


wilecoyote2015:
#### Runtime Environment

- Operating system and version: Arch Linux
- ranger version: ranger-master 1.9.2
- Python version: 3.7.3 (default, Mar 26 2019, 21:43:19) [GCC 8.2.1 20181127]
- Locale: en_GB.UTF-8

#### Current Behavior
In Kitty terminal with preview method kitty, when an image File is selected so that a preview is shown, pressing `CTRL-s` and entering a command (for example `mkdir test`), the command is executed, but ranger freezes and needs to be restarted.

#### Expected Behavior
Ranger should continue to work after command execution.

#### Steps to reproduce
1.  Set preview method to kitty
2. Navigate to an image file so that the preview is shown
3. Press `CTRL-s` and enter `mkdir test`

#### Traceback
I cannot start ranger in debug mode (bug report will follow soon if not already present)

toonn:
That's a terminal feature though, it's called flow control. `C-s` freezes output to the terminal and `C-q` resumes it.

wilecoyote2015:
Ah, I see; thank you!
Thus, pressing `C-q` after executing a shell command should cancel the alleged freeze? Unfortunately, this I cannot reproduce this.

toonn:
Hmm, then it sounds like a bug. However, I can't reproduce it. What did you map `<C-s>` to? It's not bound in the default config.

toonn:
Ah, actually. I think you don't have `<C-s>` bound at all. I got it to freeze up by typing `mkdir test<CR>`. Which is basically equivalent to `r test<CR>` I think.

i007c:
same issue

i007c:
any shell command while previewing an image will freeze the ranger
even going to the man page will freeze it

i007c:
@toonn yep its been fix. tnx <3

PR #2856, opened 2023-06-01
Fix image display with tmux + kitty

Ethsan:
Fix image display with tmux + kitty

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch Linux 6.3.3
- Terminal emulator and version: kitty 0.28.1
- ranger version: ranger-master v1.9.3-678-g1a007158
- Python version: 3.11.3 (main, Apr  5 2023, 15:52:25) [GCC 12.2.1 20230201]
- Locale: fr_FR.UTF-8

#### CHECKLIST

#### DESCRIPTION
Fix escape sequence for tmux passthrough
Add the possibility to render images wiith placehoder ( necessary with tmux )

#### MOTIVATION AND CONTEXT
Could not view image with tmux+kitty but work in icat https://github.com/ranger/ranger/issues/2846

#### TESTING
Tested with 
- kitty without tmux
- kitty with tmux and allow-passthrough not set
- kitty with tmux and allow-passthrough set

i ran ```make test```

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -22,7 +22,7 @@ import warnings
 import json
 import mmap
 import threading
-from subprocess import Popen, PIPE, check_call, CalledProcessError
+from subprocess import Popen, PIPE, check_call, check_output, CalledProcessError
 from collections import defaultdict, namedtuple
 
 import termios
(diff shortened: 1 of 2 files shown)


ranger/ext/img_display.py
```diff
@@ -685,6 +693,8 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
         https://github.com/kovidgoyal/kitty/blob/master/graphics-protocol.asciidoc"""
     protocol_start = b'\x1b_G'
     protocol_end = b'\x1b\\'
+    answer_end = protocol_end
+    use_placeholder = False
```
toonn, line 697: It's not clear to me what the placeholder is about, is this for testing?

ranger/ext/img_display.py
```diff
@@ -43,6 +43,12 @@
...
...00002CF1", "\U00002DE0", "\U00002DE1", "\U00002DE2", "\U00002DE3", "\U00002DE4", "\U00002DE5", "\U00002DE6", "\U00002DE7", "\U00002DE8", "\U00002DE9", "\U00002DEA", "\U00002DEB", "\U00002DEC", "\U00002DED", "\U00002DEE", "\U00002DEF", "\U00002DF0", "\U00002DF1", "\U00002DF2", "\U00002DF3", "\U00002DF4", "\U00002DF5", "\U00002DF6", "\U00002DF7", "\U00002DF8", "\U00002DF9", "\U00002DFA", "\U00002DFB", "\U00002DFC", "\U00002DFD", "\U00002DFE", "\U00002DFF", "\U0000A66F", "\U0000A67C", "\U0000A67D", "\U0000A6F0", "\U0000A6F1", "\U0000A8E0", "\U0000A8E1", "\U0000A8E2", "\U0000A8E3", "\U0000A8E4", "\U0000A8E5", "\U0000A8E6", "\U0000A8E7", "\U0000A8E8", "\U0000A8E9", "\U0000A8EA", "\U0000A8EB", "\U0000A8EC", "\U0000A8ED", "\U0000A8EE", "\U0000A8EF", "\U0000A8F0", "\U0000A8F1", "\U0000AAB0", "\U0000AAB2", "\U0000AAB3", "\U0000AAB7", "\U0000AAB8", "\U0000AABE", "\U0000AABF", "\U0000AAC1", "\U0000FE20", "\U0000FE21", "\U0000FE22", "\U0000FE23", "\U0000FE24", "\U0000FE25", "\U0000FE26", "\U00010A0F", "\U00010A38", "\U0001D185", "\U0001D186", "\U0001D187", "\U0001D188", "\U0001D189", "\U0001D1AA", "\U0001D1AB", "\U0001D1AC", "\U0001D1AD", "\U0001D242", "\U0001D243", "\U0001D244"]  # noqa: E501
```
toonn, line 50: Do we really need this? I'm worried it'll be hard to keep up to date going forward.

toonn (commented):
I still need to test this but if it fixes image previews in Tmux in Kitty that's awesome!

ranger/ext/img_display.py
Ethsan: I agree that awful and i hate it but there is no workaround ( hopefully it will probably not change in the future). Best i can do is put in comment the bash command to generate the list.

ranger/ext/img_display.py
Ethsan: https://sw.kovidgoyal.net/kitty/graphics-protocol/#unicode-placeholders

> You can also use a special Unicode character U+10EEEE as a placeholder for an image. This approach is less flexible, but it allows using images inside any host application that supports Unicode and foreground colors (tmux, vim, weechat, etc.) and has a way to pass escape codes through to the underlying terminal.

Because tmux doesn't understand the protocol, the image will not properly render. For example in the case where the user swap two tmux pane, the image will be at the same place.

ranger/ext/img_display.py
toonn: Maybe we can simply run the command that provides these when the method is initialized?

Ethsan: I feel like it's a bit silly to parse a [file](https://unicode.org/Public/UNIDATA/UnicodeData.txt) of 1,82Mo each time the method is initialized, simply to get a list of only 256 elements. Moreover kitty will probably never change the character chosen due to the risk of breaking a lot of implementations.

toonn: OK, fine. Can you add some more explanation as to what the diacritics are for to the comment?

BigBrotherJu:
@Ethsan Can I patch ranger on my own using this PR? Would things work as expected? Can't wait for this PR to go into the next release.

Ethsan:
Feel free, remember to allow passthrough in tmux and it should work like a charm. I only tested it on my machine, if there is a problem don't hesitate to tell me.

thenbe:
This issue was happening to me as well, even though I don't use tmux. While using kitty to preview images, ranger would freeze after opening any image file. This PR fixed it for me. Thank you.

For anyone wanting to patch this on nix, you need to add a couple of extra deps. As an overlay:

```nix
(self: super: {
  # Use ranger PR, fixes freeze after opening image in kitty: https://github.com/ranger/ranger/pull/2856
  ranger = super.ranger.overrideAttrs (old: {
    version = "1.9.3";
    src = super.fetchFromGitHub {
      owner = "Ethsan";
      repo = "ranger";
      rev = "71a06f28551611d192d3e644d95ad04023e10801";
      sha256 = "sha256-Yjdn1oE5VtJMGnmQ2VC764UXKm1PrkIPXXQ8MzQ8u1U=";
    };
    propagatedBuildInputs = old.propagatedBuildInputs ++ (with super.python3Packages; [ astroid pylint ]);
  });
})
```

Ethsan:
I think I see why it work but I am not sure. Do you use a terminal multiplexer ? What is your TERM variable ?

thenbe:
Whatever is set by kitty by default, I haven't changed it.

```
$ echo $TERM
xterm-kitty
```

Ethsan:
Lmao I don’t know why it work but good for you.

thenbe:
@Ethsan I think I was affected by this bug: https://github.com/ranger/ranger/issues/1603#issuecomment-1538288679. Originally, I was on [latest stable release](https://github.com/ranger/ranger/tree/v1.9.3). I only got the fix for my freezing issue when I patched ranger to point to your PR, since the fix for my issue has only been [merged to master but not released yet](https://github.com/ranger/ranger/issues/1603#issuecomment-1538288679). Still, thank you for contributing.

kohane27:
Not sure if this is the appropriate thread but hopefully it's still useful to some.

I've been waiting for tmux+kitty+ranger+wayland support for a long time. Then I just realized [ueberzugpp](https://github.com/jstkdng/ueberzugpp) supports tmux on sway. It just works out of the box: https://github.com/ranger/ranger/wiki/Image-Previews#with-ueberzug

foobarops (approved):
- it works
- mergeable with current master

Ethsan:
New PR: #3086

PR #2860, opened 2023-06-09
added check for torrent part files

ask1234560:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Arch linux
- Terminal emulator and version: Alacritty 0.12.1 (5fdfd47f)
- Python version: Python 3.11.3
- Ranger version/commit: 1a007158
- Locale: en_US.UTF-8

#### CHECKLIST

#### DESCRIPTION
To skip generating thumbnail for a torrent video part file.

#### MOTIVATION AND CONTEXT
For a torrent video part file ffmpeg give error for video part file but it takes some time to give the error. 
The change skips generating thumbnail for part file.

#### TESTING
workflow tests and tested if the change works


ranger/data/scope.sh
```diff
@@ -172,6 +172,8 @@ handle_image() {
 
         ## Video
         # video/*)
+        #     # check if file is incomplete torrent video file
+        #     [ "$FILE_EXTENSION_LOWER" = "part" ] && exit 1
```
toonn, line 176: Maybe we should `break` instead? That way we still get the `file(1)` generated preview identifying the file.

toonn (commented):
Looks like a nice improvement!

ranger/data/scope.sh
ask1234560: Hi, thanks for reviewing. 

Like this `[ "$FILE_EXTENSION_LOWER" = "part" ] && break`  ?

toonn: I think that should work, yeah.

ask1234560: But it is not working for me, still an image is getting previewed instead of going to [here](https://github.com/ranger/ranger/blob/1a007158cef1279df68d0e8567b126def6b68046/ranger/data/scope.sh#L452) .

ranger/data/scope.sh
```diff
@@ -468,7 +468,8 @@ handle_fallback() {
 
 MIMETYPE="$( file --dereference --brief --mime-type -- "${FILE_PATH}" )"
 if [[ "${PV_IMAGE_ENABLED}" == 'True' ]]; then
-    handle_image "${MIMETYPE}"
+    # don't generate thumbnail for incomplete torrent video file
+    [[ "$MIMETYPE" =~ video/* && "$FILE_EXTENSION_LOWER" = "part" ]] || handle_image "${MIMETYPE}"
```
ask1234560, line 472: I added the check here, now it's showing text as preview.

toonn: I don't like this, it's not in the correct spot. Things related to a case should exclusively go inside the case.

Let's look into why the break wasn't working in the case instead.
The reason you were still seeing a preview is probably that the preview was cached?

ranger/data/scope.sh
```diff
@@ -172,6 +172,8 @@ handle_image() {
 
         ## Video
         # video/*)
+        #     # don't generate thumbnail for incomplete torrent video file
+        #     [[ "$FILE_EXTENSION_LOWER" = "part" ]] && return
```
ask1234560, line 176: for `break` i am getting `break: only meaningful in a for, while, or until loop` but `return` works

markus-bauer:
Not all incomplete video files are broken. I don't think you should disable all video preview just based on the .part extension.
Most files I tried generate a valid preview a preview using ffmpegthumnailer.
I don't use torrent though. But partial download by firefox or youtube-dl work.

markus-bauer:
Youtube-dl files can still be broken, it's just that not *all* .part files are broken.

markus-bauer:
There might be a bigger issue with previews.  And I suspect the issue is with ranger, not with scope.
Here are some quick observations. Please validate these yourself. **I have no motivation to investigate this further.**

1) There's a difference between using ffmpeg and ffmpegthumnbailer. The later will produce an output image even if it fails (at least the version I have installed).  It returns an error, but there will now be an empty image in the ranger cache folder.
2) If you enable ffmpegthumnbailer in scope.sh, and try to image-preview a broken video with ranger, the task will hang. But if you use scope.sh directly, it returns immediately.
```
~/.config/ranger/scope.sh "./broken.mp4.part" 100 100 "./out.jpg" "True"; echo $?
```
3) If you only enable ffmpeg in scope (and disable ffmpegthumnbailer), ranger doesn't hang.
4) The hanging is probably related to the CommandLoader. Just for testing, I made these changes in get_preview, and it didn't hang anymore. This is not a solution, but perhaps a way to narrow down the issue.
```diff
diff --git a/ranger/core/actions.py b/ranger/core/actions.py
index e1fd2afe..9b08bf17 100644
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1183,7 +1183,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             args=[self.settings.preview_script, path, str(width), str(height),
                   cacheimg, str(self.settings.preview_images)],
             read=True,
-            silent=True,
+            silent=False,
             descr="Getting preview of %s" % path,
         )
         loadable.signal_bind('after', on_after)
```

---
A note about testing scope generated image previews: If you cancel the preview in ranger and try again, it will not hang, because it now uses the (empty) cacheimg. You first have to delete that.

ask1234560:
The thing i noticed was suppose if you have a .part file and trying to preview. But multiple ffmpegthumbnailers are getting called for previews, some times it causes spikes in cpu and ranger gets hang. I think the reason behind is the .part file is getting downloaded.

ranger/data/scope.sh
toonn: Shell's case statement uses the rather less clear `;;&` for what I meant with `break`. The points raised by markus-bauer are interesting though, let's see if we can get a better solution first?

ask1234560: Yeah, i agree with it. Can we make ranger to invoke only one scope.sh at a time. That way we don't need this pull request.

PR #2886, opened 2023-08-22
report file byte size in IIP

### LINKED ISSUE #2883: wrong size parameter for iterm image protocol
The iterm image protocol gets the wrong size value here:
https://github.com/ranger/ranger/blob/1a007158cef1279df68d0e8567b126def6b68046/ranger/ext/img_display.py#L367-L369

Per protocol spec the size should denote the file's byte size, not the base64-encoded size. Currently this breaks with vscode's image support, which does a sanity check against the size parameter to spot data errors.

jerch:
#### ISSUE TYPE

- Bug fix: Fixes #2883.

#### RUNTIME ENVIRONMENT

- Terminal emulator and version: vscode 1.80+ integrated terminal
- Python version: 3.10

#### CHECKLIST

#### DESCRIPTION

With `preview_images_method iterm2` being set, vscode terminal does not show images. This is due to ranger wrongly sending the size of the base64 encoded content, instead of the byte size of the original file/data.

#### TESTING

- `set preview_images_method iterm2` in conf file
- open integrated terminal in vscode
- run ranger
- navigate to folder with images
- images now show up

#### Additional note
 Vscode terminal will still refuse to show an image that creates a rather big sequence (>20MB, again a security/sanity check). Here it might make sense to do an interim resizing with PIL if available. (not part of this PR)

### DIFF
--- a/ranger/ext/img_display.py
+++ b/ranger/ext/img_display.py
@@ -357,7 +357,7 @@ class ITerm2ImageDisplayer(ImageDisplayer, FileManagerAware):
             return ""
         image_width = self._fit_width(
             image_width, image_height, max_cols, max_rows)
-        content = self._encode_image_content(path)
+        content, byte_size = self._encode_image_content(path)
         display_protocol = "\033"
         close_protocol = "\a"
         if os.environ["TERM"].startswith(("screen", "tmux")):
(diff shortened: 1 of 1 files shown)


toonn:
Thanks for the fix! :tada:

PR #2900, opened 2023-09-14
use urllib for URIs

gigamicro:
fixes percent encoding being broken (eg when steam opens ranger as default file manager)
may break paths that are actually named in percent encoding (that are not themselves percent encoded)
i imagine there's a better way to do this but this seems to work so far

#### ISSUE TYPE
- Bug fix
- Breaking changes
(i feel like depending on definition it could be any or all of these)

#### RUNTIME ENVIRONMENT
- Operating system and version: Linux Lite 5.4
- Terminal emulator and version: alacritty 0.12.2
- Python version: Python 3.8.10
- Ranger version/commit: 136416c7e2ecc27315fe2354ecadfe09202df7dd
- Locale: en_US.UTF-8

#### CHECKLIST

- [x] All changes follow the code style **[REQUIRED]**  (at least as far as i can tell after reading it, the import from within a function seems odd but matches the rest of the file afaik)
    - [ ] Tests have been updated

#### DESCRIPTION

uses urllib.parse.urlparse and urllib.parse.unquote to, respectively, parse the path from a URI and decode URI percent encoding
potential better way would be to only unquote() the path if it is actually a file:// uri but i literally do not know python

#### MOTIVATION AND CONTEXT

my previous fix was to make the exo preferred filemanager a shell script that removed the file:// and replaced %20's with spaces before passing any path to ranger (the current apt version of ranger has a bug where it removes the leading slash of file uris, but that was fixed by cloning the repo)

#### TESTING
"What tests have been run?"
make test_py; make test
"How does the changes affect other areas of the codebase?"
shouldn't, apart from changing the argument parsing

### DIFF
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -255,9 +255,8 @@ https://github.com/ranger/ranger/issues
 
 def get_paths(args):
     if args.paths:
-        prefix = 'file://'
-        prefix_length = len(prefix)
-        paths = [path[prefix_length:] if path.startswith(prefix) else path for path in args.paths]
+        from urllib.parse import urlparse, unquote
+        paths = [unquote(urlparse(path).path) for path in args.paths]
     else:
         start_directory = os.environ.get('PWD')
         is_valid_start_directory = start_directory and os.path.exists(start_directory)


hut:
Thanks for your work, @gigamicro. As you said, your code did not cover all the cases. I tried fixing this in https://github.com/ranger/ranger/commit/bd9b37faee9748798f3970468bd0d1dedbf0e99f . Hope this works for you.

PR #2902, opened 2023-09-17
Restore previous behaviour of move_parent to ignore files (fixes #2325)

### LINKED ISSUE #2325: [Bug Report] move_parent should ignore the item if it is not a directory
#### Runtime Environment

- Operating system and version: Manjaro-Linux
- Terminal emulator and version: Alacritty/Kitty
- Python version: v3.9.4
- Ranger version/commit: 741929906f, manually installed 
- Locale: en_US.UTF-8

#### Current Behavior
3 columns, when I use the `move_parent -1` or `move_parent 1` binding to move the parent's focus,  it works OK if the parent is a directory, but if it is not a directory, the left column will become the middle column which destroys the whole work flow.

#### Expected Behavior
The `move_parent` command should ignore the item when it is not a directory, not handle it or just skip the file and move focus to the next directory parent.

#### Steps to reproduce
1. 
2. 
3. 

#### Traceback

```

```

eMPee584:
Well, from git history it is clear that `enter_dir()` was expanded to handle files gracefully by selecting them, which is a nice feature. Even with that, the old behaviour of `move_parent()` (ignoring files) can easily be regained by adding the check for fs object type therein, as in this patch.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -569,7 +569,9 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             if parent.pointer + n < 0:
                 n = 0 - parent.pointer
             try:
-                self.thistab.enter_dir(parent.files[parent.pointer + n])
+                fobj = parent.files[parent.pointer + n]
+                if fobj.is_directory:
+                    self.thistab.enter_dir(fobj)
             except IndexError:
                 pass
 


hut (approved):

hut:
Thank you @eMPee584!

While it would probably be nicer if `:move_parent` skips over files instead of doing nothing in case of a file, I'm guessing that most people use the `sort_directories_first` option, which makes this a non-issue.

PR #2904, opened 2023-09-29
fix: preview for `tsx`, `html`, `lua`

Zeioth:
The fix is very simple: Instead of trusting the file mimetype, which `file` very often fail to determine accurately, we pygmentize the file based on its file extension.

screenshot_2023-09-30_00-14-11_998031005

From: https://www.reddit.com/r/ranger/comments/16s46s4 and many other similar threads trying to achieve this.

### DIFF
--- a/ranger/data/scope.sh
+++ b/ranger/data/scope.sh
@@ -50,6 +50,28 @@ OPENSCAD_COLORSCHEME="${RNGR_OPENSCAD_COLORSCHEME:-Tomorrow Night}"
 SQLITE_TABLE_LIMIT=20  # Display only the top <limit> tables in database, set to 0 for no exhaustive preview (only the sqlite_master table is displayed).
 SQLITE_ROW_LIMIT=5     # Display only the first and the last (<limit> - 1) records in each table, set to 0 for no limits.
 
+pygmentize() {
+    ## Syntax highlight
+    if [[ "$( stat --printf='%s' -- "${FILE_PATH}" )" -gt "${HIGHLIGHT_SIZE_MAX}" ]]; then
+        exit 2
+    fi
+    if [[ "$( tput colors )" -ge 256 ]]; then
+        local pygmentize_format='terminal256'
+        local highlight_format='xterm256'
+    else
+        local pygmentize_format='terminal'
+        local highlight_format='ansi'
+    fi
+    env HIGHLIGHT_OPTIONS="${HIGHLIGHT_OPTIONS}" highlight \
+        --out-format="${highlight_format}" \
+        --force -- "${FILE_PATH}" && exit 5
+    env COLORTERM=8bit bat --color=always --style="${BAT_STYLE}" \
+        -- "${FILE_PATH}" && exit 5
+    pygmentize -f "${pygmentize_format}" -O "style=${PYGMENTIZE_STYLE}"\
+        -- "${FILE_PATH}" && exit 5
+    exit 2
+}
+
 handle_extension() {
     case "${FILE_EXTENSION_LOWER}" in
         ## Archive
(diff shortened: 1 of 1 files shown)


Zeioth:
The reason for also adding `lua`, is `file` fails to determine its filetype when mixed with vimscript variables, as in neovim config files.

Zeioth:
`js` and `ts` added to the fix. I've tested 22+ different programming languages and the preview seem to work fine on all of them now.

ranger/data/scope.sh
```diff
@@ -50,6 +50,30 @@ OPENSCAD_COLORSCHEME="${RNGR_OPENSCAD_COLORSCHEME:-Tomorrow Night}"
 SQLITE_TABLE_LIMIT=20  # Display only the top <limit> tables in database, set to 0 for no exhaustive preview (only the sqlite_master table is displayed).
 SQLITE_ROW_LIMIT=5     # Display only the first and the last (<limit> - 1) records in each table, set to 0 for no limits.
 
+# HELPERS
+# =============================================================================
+pygmentize() {
```
klevvit, line 55: Okay this part is just the lines moved from `text/* | */xml`, former line 417.

Is it okay to name a function the same as `pygmentize` utility?

ranger/data/scope.sh
```diff
@@ -461,17 +462,15 @@ handle_mime() {
...
+# =============================================================================
 MIMETYPE="$( file --dereference --brief --mime-type -- "${FILE_PATH}" )"
 if [[ "${PV_IMAGE_ENABLED}" == 'True' ]]; then
     handle_image "${MIMETYPE}"
 fi
 handle_extension
 handle_mime "${MIMETYPE}"
-handle_fallback
+# No handler for the current format or mimetype? Preview without color.
+cat -- "${FILE_PATH}"
```
klevvit, line 474: Are you sure this is a good idea? I have doubts about scrolling through large binary files: `cat` will take a ton of resources for actually nothing

ranger/data/scope.sh
```diff
@@ -93,23 +117,13 @@ handle_extension() {
...
-
         ## XLSX
         xlsx)
             ## Preview as csv conversion
             ## Uses: https://github.com/dilshod/xlsx2csv
             xlsx2csv -- "${FILE_PATH}" && exit 5
             exit 1;;
 
-        ## HTML
-        htm|html|xhtml)
```
klevvit, line 105: Here we remove support of html command-line browsers & replace it with a preview of html file, not html page.
I think this is wrong, and we should leave this part as is, maybe add optional setting to view html as text, not as page

klevvit (changes requested):
I really like the idea of better rendering, but I consider we shouldn't remove or change significantly the options some users may actively use. Additional modifications, e.g. replacement of `handle_fallback()` with `cat` or switch to html as text not as page may be useful for some users, but it has huge influence on ranger functionality, so it's better to make it as an easy-to-switch option

ranger/data/scope.sh
Zeioth: I think our priority should be addressing the case of use that is gonna affect 99% of users, which is file preview. But for the border case you mention ranger already has `preview_max_size`

ranger/data/scope.sh
Zeioth: Again, you are proposing removing a case of use that affect 99% of users, to ship `html command-line browser support`, which is a 1% of users case of use. And which people interested easily enable anyway editing their `scope.sh`.

ranger/data/scope.sh
Zeioth: It implements the utility pygmentize. That's why the name is pygmentize.

Yes it is the same code. We just made it a function so it can be re-used in the code.

PR #2939, opened 2024-01-22
handle middle mouse click and fix pylint tests

m4ttm:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: Debian 12
- Terminal emulator and version: Kitty 0.26.5 & tmux 3.3a_3
- Python version: Python 3.9.18
- Ranger version/commit: 136416c
- Locale: en_GB.utf8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Use middle mouse click to close and fix Pylint dependency problem

#### MOTIVATION AND CONTEXT

Not able to customize the middle mouse button because ranger doesn't recognise it

Allows seamless use of the mouse, preventing the need to touch the keyboard to close

n/a

#### TESTING

make test

n/a

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 flake8
 pylint
 pytest
+astroid>2.5,<3.0
--- a/ranger/gui/mouse_event.py
+++ b/ranger/gui/mouse_event.py
@@ -13,6 +13,7 @@ class MouseEvent(object):
         curses.BUTTON2_PRESSED,
         curses.BUTTON3_PRESSED,
         curses.BUTTON4_PRESSED,
+        curses.BUTTON5_PRESSED,
     ]
     CTRL_SCROLLWHEEL_MULTIPLIER = 5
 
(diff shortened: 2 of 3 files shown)


Rom1deTroyes:
Handling the middle mouse button is surely a good behavior, but having this to close the application seems unexpected.

I expect middle mouse button to paste my previously mouse selection if we are in an input, or behave like most file browser.

For reference, as Ubuntu is a popular GNU/Linux distribution, maybe we can take this as a base for reflexion : https://help.ubuntu.com/stable/ubuntu-help/mouse-middleclick.html.en

m4ttm:
I agree it's non standard but I wanted to use the mouse to cd to a directory, I can start ranger from my tmux right click menu and have it configured to the chosen directory on exit. My neovim right click menu has everything I need to jump through codebases so this smoothes out everything and keeps me away from the keyboard until it's time to type again. Maybe more standard behaviour would be preferable for other users?

m4ttm:
Maybe the best thing to do would be to make the mouse buttons user configurable with more standard behaviour as default

toonn:
I agree that we can't make this default behavior. Configurable mouse buttons sounds awesome.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 flake8
 pylint
 pytest
+astroid>2.5,<3.0
```
eMPee584, line 4: :thinking: … why did this slip in?

Issue #1173, opened 2018-05-14
Ranger performance issue with huge files number


nmiculinic:
#### Runtime Environment
```
ranger version: ranger-master 1.9.1
Python version: 3.6.5 (default, Apr 14 2018, 13:17:30) [GCC 7.3.1 20180406]
Locale: en_US.UTF-8
```

#### Current Behavior

Upon encountering a large number of files, ranger visually blocks/hangs and it's impossible to stop the operation

#### Expected Behavior

It should be as opportunistic as possible -- First, it should never hang and be unable to switch tabs/kill navigation to back dir. 

#### Context
I have some folders which have around 100k files in them, and this is where ranger keeps halting. The reason for having 100k files is domain specific unfortunatelly, and there's not much I can do about it. 

#### Possible Solutions

Move the folder listing/related operations off the main thread and into the background. I'm not sure 

#### Steps to reproduce
1. Create folder with a lot of files (( e.g. 100k )) 
2. ranger <folder> 
3. Tab switching shall work badly, ranger shall hang/block, other UI issues

nmiculinic:
Hmmm, how do you handle ranger in cases where files are on slow storage (e.g. network)? 

> proper multithreading is a difficult problem and a constant maintenance burden.

True, it's harder than a single thread. This can be overcome with the proper design though, where concurrency is obvious. 

I'm glancing over ranger source:
```python
            while True:
                loader.work()
                if loader.has_work():
                    throbber(loader.status)
                else:
                    throbber(remove=True)

                ui.redraw()

                ui.set_load_mode(not loader.paused and loader.has_work())

                ui.draw_images()

                ui.handle_input()
```

so I guess this is the main loop? and I assume loader, among other things, is responsible for loading folder context? And I assume on each folder change, synchronously reads the whole folder and caches is somewhere?

One design idea would be using https://docs.python.org/3/library/concurrent.futures.html --> running folder scanning in executor, and when drawing on the screen. If folder changes, the future can be canceled and forgotten, since it's no longer needed. It makes writing concurrent code easier to reason about. 

There's some python2.7 equivalent too: 
http://elliothallmark.com/2016/12/23/requests-with-concurrent-futures-in-python-2-7/

toonn:
We still try to support python 2.6. The ideal solution would stream the necessary information on demand so you only have to load the files you can see, then potentially keep loading the others in the background but cap it somewhere because loading 100k files doesn't really seem sensible to me : )

[...]

PR #2943, opened 2024-01-28
basename_natural: Improve performance

mingmingrr:
#### ISSUE TYPE

- Breaking changes

#### RUNTIME ENVIRONMENT

- Operating system and version: NixOS 24.05
- Terminal emulator and version: WezTerm 20230712-072601-f4abf8fd
- Python version: 2.6.9 2.7.10 3.11.6
- Ranger version/commit: 136416c7e2ecc27315fe2354ecadfe09202df7dd
- Locale: en_US.UTF-8

#### CHECKLIST

    - [x] Tests have been updated

#### DESCRIPTION

Various optimizations to `basename_natural` functions.

**Breaking change:** The regex was from `\D` to `\D+` so that all non-digits are grouped together and less elements are generated in `basename_list`, this results in the sorting order being changed so that numbers come before non-numbers after non-number prefixes. For example:

```
Original:
"a" < "a b" < "a#" < "a1" < "ab"
After changes:
"a" < "a1" < "a b" < "a#" < "ab"
```

Tests have been updated to reflect this. If this change is undesirable then I do have a version that keeps the original behavior, however it's slightly slower and much more complex.

#### MOTIVATION AND CONTEXT

Helps with #1173, but doesn't completely address it.

In certain scenarios with file sorting method left as the default natural sort, a large chunk of time used in viewing large directories is spent on sorting. Running `ranger --profile` on a directory with 76k files on an NVMe drive shows 16 seconds out of the 24 seconds startup time is spent generating the natural sort keys.

#### TESTING

Ran tests with `make test_py` on Python 3.11.

Results of profiling with `ranger --profile` on my `/nix/store` directory, which has 76k files:

```
Original:
       1    0.002   24.282  fm.py:388(loop)
   76079   14.642   16.219  fsobject.py:167(basename_natural_lower)
After changes:
       1    0.002   10.809  fm.py:388(loop)
   76079    1.826    2.966  fsobject.py:167(basename_natural_lower)
```

Testing only `basename_natural` with `timeit` shows an improvement from 15.0s to 1.3s on the same directory. (Alternate version that keeps the sorting behavior runs in 2.2s).

### DIFF
--- a/ranger/container/fsobject.py
+++ b/ranger/container/fsobject.py
@@ -43,7 +43,7 @@ BAD_INFO = '?'
 
 _UNSAFE_CHARS = '\n' + ''.join(map(chr, range(32))) + ''.join(map(chr, range(128, 256)))
 _SAFE_STRING_TABLE = maketrans(_UNSAFE_CHARS, '?' * len(_UNSAFE_CHARS))
-_EXTRACT_NUMBER_RE = re.compile(r'(\d+|\D)')
+_EXTRACT_NUMBER_RE = re.compile(r'\d+|\D+')
 
 
 def safe_path(path):
@@ -157,21 +157,21 @@ class FileSystemObject(  # pylint: disable=too-many-instance-attributes,too-many
     @lazy_property
     def basename_natural(self):
         basename_list = []
-        for string in _EXTRACT_NUMBER_RE.split(self.relative_path):
-            try:
-                basename_list += [('0', int(string))]
-            except ValueError:
-                basename_list += [(string, 0)]
+        for string in _EXTRACT_NUMBER_RE.findall(self.relative_path):
+            if string[0].isdigit():
+                basename_list.append(('0', int(string)))
+            else:
+                basename_list.append((string,))
         return basename_list
 
     @lazy_property
     def basename_natural_lower(self):
         basename_list = []
-        for string in _EXTRACT_NUMBER_RE.split(self.relative_path_lower):
-            try:
-                basename_list += [('0', int(string))]
-            except ValueError:
-                basename_list += [(string, 0)]
+        for string in _EXTRACT_NUMBER_RE.findall(self.relative_path_lower):
+            if string[0].isdigit():
+                basename_list.append(('0', int(string)))
+            else:
+                basename_list.append((string,))
         return basename_list
 
     @lazy_property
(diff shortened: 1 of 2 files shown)


toonn:
This looks like a pretty serious improvement.

I am a bit wary of affecting the collation order in somewhat unexpected ways though. My understanding of the current implementation is that it's faithful to [the definition Wikipedia gives](https://en.wikipedia.org/wiki/Natural_sort_order):

> In computing, natural sort order (or natural sorting) is the ordering of strings in alphabetical order, except that multi-digit numbers are treated atomically, i.e., as if they were a single character.

We put numbers into a tuple with `'0'` as the first element, so any other character will compare to the number as if it's a single character. The only case where we do numeric comparison is if both tuples contain a number and therefore have the same first element.

However, further reading of some links on that article, like [Dave Koelle's blog](https://web.archive.org/web/20210803201519/http://www.davekoelle.com/alphanum.html) seem to show an algorithm that's more in line with what you have here. Though [Martin Pool makes a good point here](https://sourcefrog.net/projects/natsort/) about leading zeroes. `1.002 < 1.01` at least to our  "natural" thinking if we interpret them as fractional numbers but both the current implementation and the new one would compare `[('0', 1), ('.', 0), ('0', 2)] > [('0', 1), ('.', 0), ('0', 1)]`.

I'm not sure about fixing this though because what about scientific notation or dates, dates have a natural sort order but they come in many formats. And there's other cases where natural sorting runs afoul of expectations, like hashes, I don't actually want coincidental stretches of digits affecting their sort order if I'm being honest.

After all of this I figured we should just defer to the wisdom of our elders and go with what `ls` does. GNU ls has the `-v` flag, which is equivalent to `--sort=version` and intends to sort version numbers naturally. However, that seems to lead to yet another order? `a   a1   ab  'a 2'  'a b'  'a #1'   a#`

In summary, I'm open to this change but want to check in with the community before changing the behavior. It wouldn't be hard to introduce a setting to flip between the old and new orders or introduce a new "naturalish" sort order.

mingmingrr:
I made some additional changes that should keep the previous sort behavior, at the cost of a ~25% performance penalty (instead of ~70% penalty previously):
```
Original:
55140  4.119  4.634  fsobject.py:167(basename_natural_lower)
Sort behavior changed:
55140  0.998  1.310  fsobject.py:167(basename_natural_lower)
Sort behavior unchanged:
55140  1.238  1.664  fsobject.py:168(basename_natural_lower)
```

PR #2950, opened 2024-02-18
Load plugins from system confdir (/etc/ranger/plugins)

amozeo:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
- Operating system and version:  NixOS 23.11
- Terminal emulator and version: kitty 0.31.0
- Python version: 3.11.6 (main, Oct  2 2023, 13:45:54) [GCC 12.3.0]
- Ranger version/commit: ranger-master v1.9.3-720-gcb9ab8df
Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
This patch makes ranger load plugins from /etc/ranger/plugins

#### MOTIVATION AND CONTEXT
I wanted to setup ranger plugin system-wide using nixos system configuration, however I noticed that ranger does not load plugins placed in /etc/ranger/plugins
Since ranger already loads configuration from /etc/ranger/rc.conf, I excpected it would load plugins from the same system confdir too. This change fixes that.

#### TESTING
All test have ran successfully,
I believe this change doesn't affect other areas of the codebase

### DIFF
--- a/ranger/core/main.py
+++ b/ranger/core/main.py
@@ -438,49 +438,53 @@ def load_settings(  # pylint: disable=too-many-locals,too-many-branches,too-many
         load_custom_commands(system_comm_path, custom_comm_path)
 
         # XXX Load plugins (experimental)
-        plugindir = fm.confpath('plugins')
-        try:
-            plugin_files = os.listdir(plugindir)
-        except OSError:
-            LOG.debug('Unable to access plugin directory: %s', plugindir)
-        else:
-            plugins = []
-            for path in plugin_files:
-                if not path.startswith('_'):
-                    if path.endswith('.py'):
-                        # remove trailing '.py'
-                        plugins.append(path[:-3])
-                    elif os.path.isdir(os.path.join(plugindir, path)):
-                        plugins.append(path)
+        def try_load_plugins_from(plugindir):
+            try:
+                plugin_files = os.listdir(plugindir)
+            except OSError:
+                LOG.debug('Unable to access plugin directory: %s', plugindir)
+            else:
+                plugins = []
+                for path in plugin_files:
+                    if not path.startswith('_'):
+                        if path.endswith('.py'):
+                            # remove trailing '.py'
+                            plugins.append(path[:-3])
+                        elif os.path.isdir(os.path.join(plugindir, path)):
+                            plugins.append(path)
 
+                ranger.fm = fm
+                for plugin in sorted(plugins):
+                    try:
+                        try:
+                            # importlib does not exist before python2.7.  It's
+                            # required for loading commands from plugins, so you
+                            # can't use that feature in python2.6.
+                            import importlib
+                        except ImportError:
+                            module = __import__('plugins', fromlist=[plugin])
+                        else:
+                            module = importlib.import_module('plugins.' + plugin)
+                            fm.commands.load_commands_from_module(module)
+                        LOG.debug("Loaded plugin '%s'", plugin)
+                    except Exception as ex:  # pylint: disable=broad-except
+                        ex_msg = "Error while loading plugin '{0}'".format(plugin)
+                        LOG.error(ex_msg)
+                        LOG.exception(ex)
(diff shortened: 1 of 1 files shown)


ranger/core/main.py
```diff
@@ -444,49 +444,53 @@ def load_custom_commands(*paths):
...
+                        LOG.debug("Loaded plugin '%s'", plugin)
+                    except Exception as ex:  # pylint: disable=broad-except
+                        ex_msg = "Error while loading plugin '{0}'".format(plugin)
+                        LOG.error(ex_msg)
+                        LOG.exception(ex)
+                        fm.notify(ex_msg, bad=True)
+                ranger.fm = None
 
+        if os.path.exists(fm.confpath('plugins')):
             if not os.path.exists(fm.confpath('plugins', '__init__.py')):
```
toonn, line 484: This should happen for the system wide plugins directory too, no?

ranger/core/main.py
```diff
@@ -444,49 +444,53 @@ def load_custom_commands(*paths):
...
+                        else:
+                            module = importlib.import_module('plugins.' + plugin)
+                            fm.commands.load_commands_from_module(module)
+                        LOG.debug("Loaded plugin '%s'", plugin)
+                    except Exception as ex:  # pylint: disable=broad-except
+                        ex_msg = "Error while loading plugin '{0}'".format(plugin)
+                        LOG.error(ex_msg)
+                        LOG.exception(ex)
+                        fm.notify(ex_msg, bad=True)
+                ranger.fm = None
```
toonn, line 481: Why are we setting `ranger.fm` and then setting it back to `None`?

ranger/core/main.py
```diff
@@ -444,49 +444,53 @@ def load_custom_commands(*paths):
...
-                        module = importlib.import_module('plugins.' + plugin)
-                        fm.commands.load_commands_from_module(module)
-                    LOG.debug("Loaded plugin '%s'", plugin)
-                except Exception as ex:  # pylint: disable=broad-except
-                    ex_msg = "Error while loading plugin '{0}'".format(plugin)
-                    LOG.error(ex_msg)
-                    LOG.exception(ex)
-                    fm.notify(ex_msg, bad=True)
-            ranger.fm = None
+        try_load_plugins_from(fm.confpath('plugins'))
```
toonn, line 492: Should probably have the user plugins override the system plugins so I think we should flip the order

toonn (commented):
I'm OK with the change in principle.

ranger/core/main.py
amozeo: This existed before my patch, see changes without trailing whitespace

ranger/core/main.py
amozeo: since `/etc/ranger/plugins` is a system configuration directory, ranger should not try to modify it, especially when ranger is ran as a regular user. it can at least warn about it being missing.

toonn: While reasonable I assume it needs to exist to allow Ranger to import plugins from the directory?

ranger/core/main.py
toonn: Not a good enough reason to have it be present : )

Any idea why it's there?

ranger/core/main.py
amozeo: Given that I use this patch for over a year, no
see PEP 420, python versions older than 3.3 have needed it. all of those python versions are [end of life](https://devguide.python.org/versions/)

ranger/core/main.py
amozeo: I'd say so plugins can have access to `fm` object when plugins are being loaded. [github search](https://github.com/search?q=%22ranger.fm%22+language%3APython+-repo%3Aranger%2Franger&type=code) shows only 1 use of that

ranger/core/main.py
toonn: OK, so this does break Python 2.7 compatibility then?

PR #2964, opened 2024-04-06
Disable terminal focus reporting

### LINKED ISSUE #2864: Receiving Keyboard signals when switching focus in window managers
Whenever I open `ranger` inside my terminal and when I  switch focus to/from ranger it receives signals. I have tried different window managers and with different terminals but it is still the case.

The problem doesn't occur when I open ranger as separate process with `kitty -e ranger`  but only when I do `ranger` inside terminal. 

#### Runtime Environment
- Operating system and version:  Arch linux 6.1.34-1-lts
- Terminal emulator and version: kitty 0.28.1 and alacritty 0.12.1 (5fdfd47f)
- Python version: 3.11.3 (main, Jun  5 2023, 09:32:32) [GCC 13.1.1 20230429]
- Locale: en_US.UTF-8

#### Current Behavior
Receives signals when switching focus.

#### Expected Behavior
Should not receive signals.

#### Steps to reproduce
1. Open ranger inside terminal 
2. Switch focus to and from ranger to other windows 
3. ranger is now in rename mode and is receiving signals when switching focus

output2

krobelus:
Terminals can be asked to report focus events with

        print("\x1b[?1004h")

For some reason, our use of curses sends a sequence like that (not
exactly that one, but it contains "1004", you can observe it by
running ranger inside "script").

Then the terminal will send ^[[I and ^[[O for "focus in" and "focus
out" events respectively.  This is quite annoying. The proper way
to deal with these is to map them ourselves in all modes (to a nop,
or optionally add some UI feedback).

I do not know how to do that, so turn it off altogether with the
corresponding "l" sequence.

Fixes #2864

### DIFF
--- a/ranger/gui/ui.py
+++ b/ranger/gui/ui.py
@@ -130,6 +130,7 @@ class UI(  # pylint: disable=too-many-instance-attributes,too-many-public-method
             self.win.addstr("loading...")
             self.win.refresh()
             self._draw_title = curses.tigetflag('hs')  # has_status_line
+        print("\x1b[?1004l")
 
         self.update_size()
         self.is_on = True


abrac (approved):
All good. Can't see any issues that could arise from this fix.

krobelus:
ranger 1.9.4 no longer seems to send the "enable focus reporting sequence" (the one with 1004h).
Probably something in curses changed.
So this is no longer necessary

toonn:
@krobelus, did the original issue happen on 1.9.3 or on master? Because 1.9.4 has very little difference to 1.9.3.

krobelus:
I don't know, it might depend on ncurses etc.
Again, it's easy to test with `script`

toonn:
I'm not sure what `script` is. I assume you mean calling Ranger in a shell script?

krobelus:
> I'm not sure what `script` is. I assume you mean calling Ranger in a shell script?

https://manpage.me/?q=script
also provided by util-linux and others

PR #2978, opened 2024-05-24
Add option to configure word separators while editing in the console.

jojhelfer:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Ubuntu 22.04.4 LTS
- Terminal emulator and version: Gnome Terminal Version 3.44.0 for GNOME 42
- Python version: 3.10.12
- Ranger version/commit: ranger-master 38bb8901004b75a407ffee4b9e176bc0a436cb15
- Locale: en_us

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

This adds a configuration option `console_word_separators` to allow characters other than spaces to be considered as word separators when using (by default) `<a-b>` and `<a-f>` to move between words in the console.

A typical value for this setting might be `"_-."`, but I left the default value as `""`, so as not to break compatibility.

I also would have preferred if this option has *all* characters which are separators (rather than all characters besides space), but that would require the default value to be `" "`, which seems to be impossible with the current config parser.

#### MOTIVATION AND CONTEXT

Often filenames have hyphens and underscores rather than spaces as word separators, and it is nice to be able to move between words in a filename.

#### TESTING

I was not quite able to run the tests, because running pylint in the root directory (i.e., with the project's `.pylintrc`) gives me the error described [here](https://github.com/pylint-dev/pylint/issues/5723) (this is the case even before making the changes in this commit).

When I remove `python3` from `load-plugins` in `.pylintrc`, the tests run and give `Your code has been rated at 10.00/10` but then also `make: *** [Makefile:98: test_pylint] Error 8`.


DanilaBerezin:
Unrelated question: judging by your commit history, it looks like you had an issue running `make` and `make test` due to [this] issue, but then it seems you somehow fixed it. How were you able to?

jojhelfer:
> Unrelated question: judging by your post's edit history, it looks like you had an issue running `make` and `make test` due to [this](https://github.com/pylint-dev/pylint/issues/5723) issue, but then it seems you somehow fixed it. How did you fix that issue? Running into the exact same thing.

Ah, good question. That was also to do with the pylint version; see [here](https://pylint.readthedocs.io/en/stable/whatsnew/2/2.13/full.html#what-s-new-in-pylint-2-13-0).
I had been using the pylint package on Ubuntu 22.04, which is pylint version 2.12, but apparently the issue involving `W1601` was resolved only in version 2.13.

So, when I installed pylint with `pipx`, giving me the latest version 3.2.2, it resolved that issue, but then it created the above-mentioned incompatibility with ranger's tests, for which I need version <3.

PR #2983, opened 2024-06-10
fix(actions.py): misplaced name and value in error message

belkka:
Problem: the error message
```
"Invalid value `%s' for option `%s'!" % (name, value)
```
displays option name after "value" and option value after "option", e.g.
```
:set use_preview_script xxx
Invalid value `use_preview_script' for option `xxx'!
```
Apparently, it's supposed to say
```
Invalid value `xxx' for option `use_preview_script'!
```

Solution: swap arguments (name, value) after % string interpolation operator

Note: I believe using python f-strings [rather than % operator] is a better approach. In particular, f-strings could reduce this type of mistakes:
```
f"Invalid value `{value}' for option `{name}'!"
```
Also, see "Note" at https://docs.python.org/3.12/library/stdtypes.html#printf-style-string-formatting
Unfortunately, f-strings are only added in python 3.6, but ranger seems to be supporting plenty of older python versions including python 2 (?) :/

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -134,7 +134,7 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
             return value
         if list in types:
             return value.split(',')
-        raise ValueError("Invalid value `%s' for option `%s'!" % (name, value))
+        raise ValueError("Invalid value `%s' for option `%s'!" % (value, name))
 
     def toggle_visual_mode(self, reverse=False, narg=None):
         """:toggle_visual_mode


belkka:
The linter issue seems to be valid, but irrelevant to these changes (a different function many lines below)

```
ranger/core/actions.py:822:47: E0606: Possibly using variable 'fnc' before assignment (possibly-used-before-assignment)
```

At first glance, it's not immediately obvious to me how to properly fix the error (there is indeed a logical error). Should I resolve it?

Note, tools like https://github.com/udifuchs/pylint-silent or https://github.com/mbarkhau/pylint-ignore might be used to prevent legacy issues from blocking new changes in the same file (found in this discussion https://github.com/pylint-dev/pylint/issues/5403, never used these personally)

belkka:
@toonn could you advice what should I do? I don't mind helping to resolve linter error(s) in the file, as long as my changes will be eventually reviewed.

Sorry for bothering, I've selected you randomly from https://github.com/orgs/ranger/people (you are number 3), because current CI/CD configuration might affect arbitrary other contributions in a similar way.

image

hut:
Thank you for your fix!

Sorry for taking so long to merge this, and thank you for reaching out on IRC to remind us. All those issues/PRs are just overwhelming.

toonn:
> All those issues/PRs are just overwhelming.

Yep, only just got here. Happy to see I'm no longer needed : )

Thank you for the fix! :heart:

PR #3009, opened 2024-09-12
Update vim_file_chooser.vim to work in Neovim and GUI

Konfekt:
This [adapts](https://gist.github.com/Konfekt/0181e371b985a4b9918242ba87f7882e) to https://github.com/ranger/ranger/blob/bd9b37faee9748798f3970468bd0d1dedbf0e99f/examples/vim_file_chooser.vim#L3 so that it should work in Neovim and GUI without xterm. It bails out to other file pickers in case ranger happens not to be installed.

### DIFF
--- a/examples/vim_file_chooser.vim
+++ b/examples/vim_file_chooser.vim
@@ -1,40 +1,79 @@
 " Compatible with ranger 1.4.2 through 1.7.*
 "
-" Add ranger as a file chooser in vim
+" Add ranger as a file chooser in vim.
+"
+" You need to either
+"
+" - copy the content of this file to your ~/.vimrc resp. ~/.config/nvim/init.vim,
+" - or put it into ~/.vim/plugin respectively ~/.config/nvim/plugin,
+" - or source this file directly:
+"
+"     let s:fp = "/path/to/vim_file_chooser.vim"
+"     if filereadable(s:fp) | exec "source" s:fp | endif
+"     unlet s:fp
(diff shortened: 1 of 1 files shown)


toonn:
@Konfekt, I see that you've started FilePicker.vim since submitting this. Maybe add a comment pointing to your plugin? Might also be a nice addition to our wiki.

Konfekt:
I added https://github.com/ranger/ranger/wiki/Open-Ranger-from-Vim to the Wiki; where should the comment pointing to https://github.com/Konfekt/filepicker.vim be made? I proposed this PR since I think this little plug-in, though more general, achieves in particular what the current official plug-in does (if I didn't overlook something).

toonn:
@Konfekt, I meant adding a comment to the vim_file_chooser.vim example. But maybe a pointer on the wiki is better than an example altogether, we don't really pay much attention to the examples so they can bitrot.

Konfekt:
I am a bit confused; this replaces vim_file_chooser.vim

toonn:
Yes, but maybe removing it from the examples and adding a clearer pointer to the upstream plugin is better. It's better for users to use the upstream rather than an included example that might be outdated.

toonn (commented):
Echoing a deprecation message would be nice. That way current users won't suddenly have a broken setup without at least a pointer to how to fix it.

toonn (approved):
LGTM

PR #3050, opened 2025-02-04
Added ability to use environmental variables in preview_script option

Wolodia-M:
`os.path.expandvars()` is now used to expand environmental variables for configuration key `preview_script`.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Terminal emulator and version: kitty 0.39.1
- Python version: 3.13.1 (main, Dec  4 2024, 18:05:56) [GCC 14.2.1 20240910]
- Ranger version/commit: ranger-master c646588
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

I use different directory for configuration files and I have environmental variable that points to this directory. So I want to simply use this environmental variable in ranger's configuration file to specify path to preview script.


toonn (changes requested):
I think this is reasonable. Even though it is technically backwards incompatible, "$HOME" is, in principle, a legal name for a directory.

Let's swap the order though, do expanduser last, avoids an unlikely edge case but also works more like users would expect with putting `~/some/path` in a variable.

Wolodia-M:
Swapped the order of function calls.

toonn (approved):
LGTM

Thanks for the contribution! :tada:

PR #3051, opened 2025-02-05
Add GNOME papers to document viewers

septatrix:
Add the new GNOME Papers PDF viewer to the list of applications for lookup by rifle


toonn (approved):
LGTM and I think it's fair to assume that Papers will be slightly less common than Evince so the location of the rules seems good.

PR #3055, opened 2025-02-11
add: support for tilde in bookmarks

### LINKED ISSUE #3032: [FR]: pls support environment variables like `~` or `$HOME` in `bookmarks`
#### Steps to reproduce
1. put sth like `d:~/Documents` in `~/.local/share/ranger/bookmarks`
2. open ranger
3. the recent path gets prepended to `~/Documents`
4. bookmarks file is broken/can't be used

prashantrahul141:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version:  popOS 22.04
- Terminal emulator and version:  alacritty 0.14.0-dev (fd1a3cc7)
- Python version: 3.10.12 (main, Nov  6 2024, 20:22:13) [GCC 11.4.0]
- Ranger version/commit:  ranger-master
- Locale: en_US.UTF-8

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
While loading saved bookmarks, we check if the path starts with `~` if it does then we replace the first occurrence of `~` with the current home directory.

#### MOTIVATION AND CONTEXT
This fixes #3032 

#### TESTING
Manual tests were conducted with various edge cases to ensure the changes doesn't break anything.

### DIFF
--- a/ranger/container/bookmarks.py
+++ b/ranger/container/bookmarks.py
@@ -9,6 +9,7 @@ import os
 from io import open
 
 from ranger import PY3
+from ranger.container import fsobject
 from ranger.core.shared import FileManagerAware
 
 ALLOWED_KEYS = string.ascii_letters + string.digits + "`'"
@@ -182,6 +183,8 @@ class Bookmarks(FileManagerAware):
                 for key, value in self.dct.items():
                     if key in ALLOWED_KEYS \
                             and key not in self.nonpersistent_bookmarks:
+                        if isinstance(value, fsobject.FileSystemObject):
+                            value = value.original_path
                         key_value = "{0}:{1}\n".format(key, value)
                         if not PY3 and isinstance(key_value, str):
                             key_value = key_value.decode("utf-8")
--- a/ranger/container/fsobject.py
+++ b/ranger/container/fsobject.py
@@ -6,7 +6,7 @@ from __future__ import (absolute_import, division, print_function)
 import re
 from grp import getgrgid
 from os import lstat, stat
-from os.path import abspath, basename, dirname, realpath, relpath, splitext
+from os.path import abspath, basename, dirname, realpath, relpath, splitext, expanduser
 from pwd import getpwuid
 from time import time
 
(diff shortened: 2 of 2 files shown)


toonn (changes requested):
Use `os.path.expanduser` instead.

prashantrahul141:
I had previously noticed that the bookmarks file gets overwritten with an absolute path after opening Ranger, even if you have something like `~/Documents` in it. Initially, I thought this wasn't really a problem. However, I later realized that this would completely undermine the option of syncing bookmarks if you have a different username, since the absolute path includes the username as well.

So now `FileSystemObject` also keeps a reference to the original path it was given before expanding it to the absolute path. When saving bookmarks, it ensures that the value is of type `FileSystemObject` and use `original_path`, otherwise the old behaviour.

ranger/container/bookmarks.py
```diff
@@ -183,7 +182,7 @@ def save(self):
                 for key, value in self.dct.items():
                     if key in ALLOWED_KEYS \
                             and key not in self.nonpersistent_bookmarks:
-                        key_value = "{0}:{1}\n".format(key, value)
+                        key_value = "{0}:{1}\n".format(key, value.original_path)
```
toonn: If a bookmark is updated this value is updated too?

toonn (commented):
Most of it looks OK.

ranger/container/bookmarks.py
prashantrahul141: https://github.com/user-attachments/assets/f71179e5-186a-4aaa-8698-46c79d4fbffc



is this what you mean?

EmmAanue:
EXe

EmmAanue (approved):

ranger/container/bookmarks.py
toonn: Yep, perfect.

toonn (dismissed):
LGTM

toonn:
Can you clean up the commit history a bit and rebase?

prashantrahul141:
:sweat_smile: my bad

toonn (approved):
LGTM

For future reference, multiple commits in a PR are fine, it's just things that amount to reverts and fixups that are better cleaned up before merging.

PR #3063, opened 2025-02-27
default colorscheme: change border / separator lines from white to gray

### LINKED ISSUE #3016: [Question]
Is there any way to set the border characters and color?
I would like to have rounded borders with gray dim but I cannot find a place where I can set it
╭╮╯╰─│

eMPee584:
Fixes #3016

Actually does look smoother to me..

### DIFF
--- a/ranger/colorschemes/default.py
+++ b/ranger/colorschemes/default.py
@@ -28,6 +28,7 @@ class Default(ColorScheme):
             if context.empty or context.error:
                 bg = red
             if context.border:
+                attr |= dim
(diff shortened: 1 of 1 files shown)


toonn (changes requested):
LGTM

Can you drop the Fixes #3016 from the commit message? It only addresses part of the issue and I think the other PR addresses the more significant part of it.

eMPee584:
> Can you drop the Fixes #3016 from the commit message?

Sure, if that's the blocker.. 😅

PR #3064, opened 2025-02-27
miller / multipane view: add settings.draw_borders_rounded

### LINKED ISSUE #3016: [Question]

eMPee584:
Fixes #3016

How to fix the code duplication? 🤔

### DIFF
--- a/doc/ranger.pod
+++ b/doc/ranger.pod
@@ -1075,6 +1075,10 @@ on top of those for I<draw_borders>:
 
  active-pane    draw an outline around the active pane only
 
+=item draw_borders_rounded [bool]
+
+Use the unicode characters "╭ ╰ ╮ ╯" to draw borders with rounded corners.
+
 =item draw_progress_bar_in_status_bar [bool]
 
 Draw a progress bar in the status bar which displays the average state of all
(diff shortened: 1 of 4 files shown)


toonn (commented):
I'm not sure about this change. Isn't the point of the ACS_* characters that terminal emulators are allowed to draw actual lines rather than characters? So using normal characters for the corners might end up with a weird, continuous line, bit of whitespace around a corner character that doesn't necessarily match the line thickness, etc., no?

You could refactor it into a separate module if you want.

Another option for the setting is we could add a list as the value of draw_borders, boolean values would have the same behavior as now but passing a list would have Ranger use those characters for the corners (probably best to include all the possible line drawing characters, vertical, horizontal, crossing, various T crossings).

toonn:
@eMPee584, what do you think of the suggestion of accepting a list of border-drawing characters?

eMPee584:
[…] I've included this in the patch series I posted and removed the duplication by moving `_draw_border_rectangle()` to `view_base`. Reading your suggestions again @toonn, I'm all in  favor of generalizing it to other border character sets (like double line or bold) but (in favor of 80/20 efficiency) would consider that a rather rare use case and choose to postpone implementation of these until anyone actually needs / asks for it. Would you agree that's a sensible approach? 😅

PR #3068, opened 2025-03-05
7-zip now has an official Linux version (7zz)

adam8157:
p7zip (`7zr`, `7za`, `7z`) is unmaintained.

ref: https://news.ycombinator.com/item?id=31846317


doc/ranger.1
toonn, line 1: Don't regenerate the man pages. They usually need to be regenerated after a merge anyway.

toonn (approved):
LGTM

PR #3075, opened 2025-03-25
add :narrow!

### LINKED ISSUE #3074: need an easy way to disable :narrow
The `:narrow` command which only shows marked files is great. However, once it is enabled, there is **no easy way to disable it.**

The kludgy way I do it is to run `:filter` with a pattern that has zero matches, run `:narrow` on this empty set (which disables it), then reset the filter.

There should be an easy way to do this. Perhaps if `:narrow` is run when it is already enabled, that should disable it. Looking at the code, this should not be too hard to do.

njthomas:
Adds command `:narrow!`, which will disable `:narrow` if it is enabled.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: FreeBSD 14.2-p2 and macOS 14.7.4 (23H420).
- Terminal emulator and version: iTerm2 Build 3.5.11
- Python version: Python 3.11.11 (FreeBSD) and Python 3.13.2 (macOS Homebrew)
- Ranger version/commit: `master/a19517`
- Locale: `en_US.UTF-8`

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
This patch adds a new command `:narrow!` which disables `:narrow`. If narrow is not enabled, it has no effect.

#### MOTIVATION AND CONTEXT
There is no easy way to disable the narrow feature without unselecting files. This command will allow the user to do that.

I initially thought to add a toggle to `:narrow` itself, which would obviate the need for an extra command, however this would prevent the user from unselecting files within a *narrowed context* and run narrow again to show the new subset of narrowed files. Adding this extra command allows the user to exit the narrowed environment entirely.

Fixes #3074.

#### TESTING
`make test` has been run successfully on FreeBSD and macOS.

Apart from adding a single class (`narrow_bang`) to `config/commands.py`, there are no changes to other areas of the codebase.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1753,6 +1753,21 @@ class narrow(Command):
         self.fm.thisdir.refilter()
 
 
+class narrow_bang(Command):
+    """
+    :narrow!
+
+    Disable narrowing.
+    """
+    name = 'narrow!'
+    allow_abbrev = False
+
+    def execute(self):
+
+        self.fm.thisdir.narrow_filter = None
+        self.fm.thisdir.refilter()
+
+
 class filter_inode_type(Command):
     """
     :filter_inode_type [dfl]


toonn (commented):
The change is well motivated but I think we can avoid adding the new command.
As an aside, the naming preference would be `unnarrow`, our commands ending with `!` imply a more forceful form of the command.

I think we can add a default argument to the `narrow` command that defaults to `True`, then we can simply add `alias unnarrow narrow False` to the default `rc.conf`.

njthomas:
> I think we can add a default argument to the `narrow` command that defaults to `True`, then we can simply add `alias unnarrow narrow False` to the default `rc.conf`.

Ah, makes sense. I'll throw away this patch and create a new one to that end.

njthomas:
Will redo this.

PR #3084, opened 2025-04-25
add :unnarrow to disable :narrow mode

njthomas:
Add `:unnarrow` functionality to disable `:narrow` mode even if files are selected.

Implemented using a default argument for :narrow that defaults to True.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
- Operating system and version: FreeBSD 14.2-p3 and macOS Sonoma 14.7.4 (23H420)
- Terminal emulator and version: iTerm2 Build 3.15.3
- Python version: Python 3.11.11 (FreeBSD) and Python 3.13.3 (macOS Homebrew)
- Ranger version/commit: `master/b31db0` 
- Locale:  `en_US.UTF-8`

#### CHECKLIST
    - [ ] Tests have been updated

#### DESCRIPTION
This patch adds a new command alias `:unnarrow` which disables `:narrow` functionality. If narrow is not enabled, it has no effect.

#### MOTIVATION AND CONTEXT
There is no easy way to disable narrow without unselecting files. This command will allow the user to do that.

#### TESTING
`make test` has been run successfully on FreeBSD and macOS.

Apart from adding a single alias (`:unnarrow`) to `config/rc.conf` and adding a default argument to `class narrow` in `config/commands.py`, there are no changes to other areas of the codebase.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1743,13 +1743,23 @@ class narrow(Command):
 
     Show only the files selected right now. If no files are selected,
     disable narrowing.
+
+    To disable narrow mode when selection is still active, use the
+    :unnarrow command.
     """
+
     def execute(self):
-        if self.fm.thisdir.marked_items:
+
+        # by default, calling :narrow enables narrow mode
+        # uness it is called with :narrow False
+        narrow_mode = self.arg(1).lower() != "false"
+
+        if narrow_mode and self.fm.thisdir.marked_items:
             selection = [f.basename for f in self.fm.thistab.get_selection()]
             self.fm.thisdir.narrow_filter = selection
         else:
             self.fm.thisdir.narrow_filter = None
+
         self.fm.thisdir.refilter()
 
 
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1749,9 +1749,8 @@ class narrow(Command):
     """
 
     def execute(self):
-
-        # by default, calling :narrow enables narrow mode
-        # uness it is called with :narrow False
+        # Calling :narrow enables narrow mode unless it is called with a
+        # case-insensitive `false` argument
         narrow_mode = self.arg(1).lower() != "false"
 
         if narrow_mode and self.fm.thisdir.marked_items:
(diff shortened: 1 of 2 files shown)


ranger/config/commands.py
```diff
@@ -1743,13 +1743,26 @@ class narrow(Command):
...
+
     def execute(self):
-        if self.fm.thisdir.marked_items:
+
+        # by default, calling :narrow enables narrow mode
+        self.narrow_mode = True
+
+        # turn off narrow mode if called with ":narrow False"
+        if self.arg(1) and self.arg(1) == "False":
+            self.narrow_mode = False
```
toonn, line 1758: I think we can simplify this to `narrow_mode = self.arg(1).lower() != "false"`. Can be an if statement but I prefer not having to keep track of changing state.

Doesn't need to be an instance variable if it's only used in this method.

toonn: The case lowering is because we tend to accept bools regardless of case.

ranger/config/commands.py
```diff
@@ -1743,13 +1743,26 @@ class narrow(Command):
 
     Show only the files selected right now. If no files are selected,
     disable narrowing.
+
+    To disable files when selection is active, use the :unnarrow
```
toonn, line 1747: I think you meant "filtering" or something rather than "files."

toonn (commented):
I like this approach.

ranger/config/commands.py
njthomas: Gah. Malapropism on my part. I reworked the docs in my new patch.

ranger/config/commands.py
njthomas: Great idea, I redid the patch. It makes the code much simpler.

toonn:
This looks ready to merge. I'm holding off until I adapt CI to the newer lint warnings, hopefully tomorrow. Feel free to ping me if I don't get around to it within a week or so.

toonn (approved):
LGTM
I'm gonna drop a blank line and reword the comment and then merge.

Issue #3072, opened 2025-03-14
Make use of unicode placeholders for KIP in programs like Tmux


toonn:
> Came across [this discussion](https://github.com/ghostty-org/ghostty/discussions/2496#discussioncomment-11727878) which states that image previews inside tmux work. Tested it out and it does.
> 
> 
> 
> Edit: I tried out the timg -pK command mentioned in that thread which worked properly, but ranger seems to have issues inside tmux. Exact message was: `terminal did not respond to kitty graphics query; disabling` 

 _Originally posted by @dinesh-58 in [d3e51ae](https://github.com/ranger/ranger/commit/d3e51aefce1e8caec7a0c011c8121d55d6307caa#r153648580)_

hut:
Fixed in #3086

Issue #2414, opened 2021-08-11
Ranger Freezes when trying to display an image (Using Kitty with Tmux)


riceset:
#### Runtime Environment

- Operating system and version: MacOS Big Sur 11.5.1
- Terminal emulator and version: Kitty 0.22.2
- Python version: 3.6.12
- Ranger version/commit: 1.9.3
- Locale: en_US.UTF-8

#### Current Behavior
The application completely freezes when trying to display an image.

#### Possible Solutions
It probably has to do with Tmux because it works normally without it.

#### Steps to reproduce
1. Open kitty
2. Open Tmux
3. Open Ranger
4. Try viewing an image

#### `rc.config` contents
```
set preview_images true
set preview_images_method kitty
```

toonn:
Could you try reproducing this on master? It's probably still an issue but that gives us a recent point of reference.

MattiaSilvestri:
Are there any updates on this? I have the same issue.

toonn:
Can you reproduce it on the master branch?

MattiaSilvestri:
Yes, just tried and I get the same behaviour (ranger version 1.9.3).

toonn:
Thank you. #2846 seems related.

hut:
It should be fixed now. Can you please check again on the master branch @MattiaSilvestri ?

PR #3086, opened 2025-04-27
use tmux passthrough for kitty graphics protocol

l4zygreed:
Updated Ethan PR for current master

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

- Operating system and version: Macos sequoia 15.4
- Terminal emulator and version: Ghostty 1.1.3
- Python version: 3.13.3
- Ranger version/commit: ranger-master v1.9.3-794-gcf51b12f
- Locale: en_US.UTF-8

#### CHECKLIST

#### DESCRIPTION

Updated PR #2856 to be compatible with current master branch.

#### MOTIVATION AND CONTEXT
Want to finally see images inside tmux on my mac and outside of iterm/kitty terminal 😆 
Found old Ethan PR and decided to resolve conflicts and update it for current master 

#### TESTING

- `make test` executed successfully 
- kitty without tmux
- kitty with tmux and allow-passthrough not set
- kitty with tmux and allow-passthrough set
- ghostty without tmux
- ghostty with tmux and allow-passthrough not set
- ghostty with tmux and allow-passthrough set

#### IMAGES / VIDEOS

https://github.com/user-attachments/assets/a550437e-80e7-48b0-ade5-fe03c7ff761d


ranger/ext/img_display.py
```diff
@@ -805,6 +844,8 @@ def draw(self, path, start_x, start_y, width, height):
         # finish initialization if it is the first call
         if self.needs_late_init:
             self._late_init()
+        if self.use_placeholder:
+            cmds.update({'U': 1})
```
Ethsan, line 848:
```suggestion
            # U: 1=enable virtual placement using unicode placeholder
            cmds.update({'U': 1})
```
The past me didn't seems to like comment

ranger/ext/img_display.py
```diff
@@ -738,16 +759,17 @@ def _late_init(self):
             # read response(s); DA1 response should always be last
             resp = b''
             #          (DA1 resp start   )     (DA1 resp end     )
-            while not ((b'\x1b[?' in resp) and (resp[-1:] == b'c')):
+            # while not ((b'\x1b[?' in resp) and (resp[-2:] == self.answer_end)):
```
Ethsan, line 762: remove the comment

ranger/ext/img_display.py
```diff
@@ -694,6 +706,10 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
         https://github.com/kovidgoyal/kitty/blob/master/graphics-protocol.asciidoc"""
```
Ethsan, line 706: link seem broken, we can maybe update it in a separate commit ? With like https://sw.kovidgoyal.net/kitty/graphics-protocol/#display-images-on-screen ?

Ethsan (commented):
Hey, thanks for the keeping the pr alive and the update, it's cleaner now!

Should we separate the commit in two ? Something like  :
- support for virtual placement using unicode placeholder
- support for tmux allow-passthrough mode

Moreover some minor changes below.

ranger/ext/img_display.py
```diff
@@ -43,6 +44,16 @@
...
...00002CF1", "\U00002DE0", "\U00002DE1", "\U00002DE2", "\U00002DE3", "\U00002DE4", "\U00002DE5", "\U00002DE6", "\U00002DE7", "\U00002DE8", "\U00002DE9", "\U00002DEA", "\U00002DEB", "\U00002DEC", "\U00002DED", "\U00002DEE", "\U00002DEF", "\U00002DF0", "\U00002DF1", "\U00002DF2", "\U00002DF3", "\U00002DF4", "\U00002DF5", "\U00002DF6", "\U00002DF7", "\U00002DF8", "\U00002DF9", "\U00002DFA", "\U00002DFB", "\U00002DFC", "\U00002DFD", "\U00002DFE", "\U00002DFF", "\U0000A66F", "\U0000A67C", "\U0000A67D", "\U0000A6F0", "\U0000A6F1", "\U0000A8E0", "\U0000A8E1", "\U0000A8E2", "\U0000A8E3", "\U0000A8E4", "\U0000A8E5", "\U0000A8E6", "\U0000A8E7", "\U0000A8E8", "\U0000A8E9", "\U0000A8EA", "\U0000A8EB", "\U0000A8EC", "\U0000A8ED", "\U0000A8EE", "\U0000A8EF", "\U0000A8F0", "\U0000A8F1", "\U0000AAB0", "\U0000AAB2", "\U0000AAB3", "\U0000AAB7", "\U0000AAB8", "\U0000AABE", "\U0000AABF", "\U0000AAC1", "\U0000FE20", "\U0000FE21", "\U0000FE22", "\U0000FE23", "\U0000FE24", "\U0000FE25", "\U0000FE26", "\U00010A0F", "\U00010A38", "\U0001D185", "\U0001D186", "\U0001D187", "\U0001D188", "\U0001D189", "\U0001D1AA", "\U0001D1AB", "\U0001D1AC", "\U0001D1AD", "\U0001D242", "\U0001D243", "\U0001D244"]  # noqa: E501
```
Ethsan, line 55: I am still ashamed of this lines, I am open to better idea. Maybe putting the list inside a separate python file ? or json file ?

Ethsan: Hi,

> I think parsing json each time is not a good idea. It’s not a sensative/dynamic data or config to put it into json or another format which will require additional resources and time for parsing.

Agreed

> 
> It’ll probably will be instantaneous on most modern systems, but anyway, why do more when we can do less.
> 
> About moving to another python file, idk seems fine to me. I think this is for maintainers to decide.

@toonn, what do you prefer ?

Ethsan:
Resolve #2414, Resolve #2846, Resolve #3072

ranger/ext/img_display.py
Ethsan: I don't think the unicode code points will change, since it would break all current implementations of the kitty graphical protocol.

l4zygreed: For me having it there is totally fine. And creating a separate file for a single list sounds meh… 😐

And also don’t think that they'll change unicode points after 2+ years after they added 🤔 

So I guess we can mark this as resolved?

toonn: I'd still prefer an upstream source or at least a copy of an upstream source so that it's easy to verify it's still in sync down the line.

Any reason not to, past "It's every so slightly easier not to read bytes from a file?"

Ethsan:
Hi,
@toonn do you prefer something like that ? or should we revert ?

ranger/ext/img_display.py
```diff
@@ -898,6 +934,57 @@ def _format_cmd_str(self, cmd, payload=None, max_slice_len=2048):
...
 
+    # Lazily load and parse diacritics use to index row and column from
+    # rowcolumn-diacritics.txt
+    def _get_diacritics(self):
+        if (self.diacritics is not None):
+            return self.diacritics
+
+        try:
+            with open(
+                self.fm.relpath("rowcolumn-diacritics.txt"),
```
l4zygreed, line 945:
```suggestion
                self.fm.relpath(“data”, “rowcolumn-diacritics.txt"),
```
@Ethsan you placed file in data folder, so the path is incorrect and throws error.
also why file named txt while it’s treated like csv? 🤔

ranger/data/rowcolumn-diacritics.txt
toonn, line 1: I was kinda hoping to read this from wherever Kitty or Ghostty install it on the system but I guess this makes it compatible with other terminals too.

ranger/ext/img_display.py
```diff
@@ -898,6 +934,57 @@ def _format_cmd_str(self, cmd, payload=None, max_slice_len=2048):
...
+                newline="",
+                encoding="utf-8",
+            ) as file:
+                reader = csv.reader(file, delimiter=";")
```
toonn, line 949: Is `csv.reader` more performant than rolling our own? It's just some hex digits from the starts of lines, right? I'd expect parsing the entire lines to take a little bit longer.

ranger/ext/img_display.py
```diff
@@ -792,6 +806,23 @@ def _late_init(self):
...
+        # check if tmux allows passthrough
+        # use tmux escape sequence if it's enabled
+        # throw error if it's disabled
+        try:
+            tmux_allow_passthrough = check_output(
+                ['tmux', 'show', '-Apv', 'allow-passthrough']).strip()
+        except CalledProcessError:
+            tmux_allow_passthrough = b'off'
+        if tmux_allow_passthrough == b'off':
+            raise ImageDisplayError("allow-passthrough no set in Tmux!")
```
toonn, line 819:
```suggestion
            raise ImageDisplayError("allow-passthrough is not set in Tmux!")
```

ranger/ext/img_display.py
```diff
@@ -898,6 +934,57 @@ def _format_cmd_str(self, cmd, payload=None, max_slice_len=2048):
...
+            ) as file:
+                reader = csv.reader(file, delimiter=";")
+                # extract first column containing hex code and filter out commented line
+                hex_list = [
+                    row[0] for row in reader if not row[0].lstrip().startswith("#")
+                ]
+                # convert hex code to int and then to unicode char
+                self.diacritics = [chr(int(hex, 16)) for hex in hex_list]
+        except Exception as err:
+            raise ImageDisplayError("Error while loading diacritics" + str(err))
```
toonn, line 957:
```suggestion
            raise ImageDisplayError("Error while loading diacritics: " + str(err))
```

ranger/ext/img_display.py
```diff
@@ -691,9 +695,14 @@ class KittyImageDisplayer(ImageDisplayer, FileManagerAware):
                  |  base64 encoded payload
         key: value pairs as parameters
     For more info please head over to :
-        https://github.com/kovidgoyal/kitty/blob/master/graphics-protocol.asciidoc"""
+        https://sw.kovidgoyal.net/kitty/graphics-protocol/#display-images-on-screen"""
     protocol_start = b'\x1b_G'
     protocol_end = b'\x1b\\'
+    answer_start = protocol_start
+    answer_end = protocol_end
```
toonn, line 702: This part is honestly a bit confusing. Why are we initializing two sets of these? Do we need `answer_start` after changing `protocol_start` while also not being able to use it instead of `protocol_start` before changing it?

ranger/ext/img_display.py
```diff
@@ -720,6 +729,11 @@ def __init__(self):
         self.temp_file_dir = None  # Only used when streaming is not an option
 
     def _late_init(self):
+        # check if we inside tmux
+        # if true then we use the tmux escape sequence
```
toonn, line 733:
```suggestion
        # Use Tmux escape sequence if appropriate.
```

ranger/ext/img_display.py
```diff
@@ -792,6 +806,23 @@ def _late_init(self):
         self.pix_row, self.pix_col = x_px_tot // n_rows, y_px_tot // n_cols
         self.needs_late_init = False
 
+    def _handle_init_tmux(self):
+        # check if tmux allows passthrough
+        # use tmux escape sequence if it's enabled
+        # throw error if it's disabled
```
toonn, line 812:
```suggestion
        # Check whether allow-passthrough is enabled
```

ranger/ext/img_display.py
```diff
@@ -857,9 +891,11 @@ def draw(self, path, start_x, start_y, width, height):
         with temporarily_moved_cursor(int(start_y), int(start_x)):
             for cmd_str in self._format_cmd_str(cmds, payload=payload):
                 self.stdbout.write(cmd_str)
+            if self.use_placeholder:
+                self._print_placeholder(self.image_id, start_x, start_y, width, height)
         # catch kitty answer before the escape codes corrupt the console
         resp = b''
-        while resp[-2:] != self.protocol_end:
+        while resp[-2:] != self.answer_end:
```
toonn, line 898: Why not use `endswith` here too?

toonn (commented):
FWIW in the future I'd prefer if people tried to PR to the PR branch of the original PR author, possibly linking to the downstream PR in our PR comments. It's inconvenient to keep track of things if people keep opening new PRs that are really the same thing. Rebasing on master is usually quite trivial so you can rest assured that that is *not* what holds back a PR from getting merged.

In this case Ethsan seems perfectly happy with the situation but it can come across as trying to take credit for someone else's work. All of this is easily avoided by opening the PR against their PR branch.

ranger/ext/img_display.py
Ethsan: Hi,
> @Ethsan you placed file in data folder, so the path is incorrect and throws error.

Yeah sorry about that, I move it will trying ot fix linting issues and forgot to update path.

> also why file named txt while it’s treated like csv? 🤔

I have simply reused the name given upstream [rowcolumn-diacritics.txt](https://github.com/kovidgoyal/kitty/blob/master/gen/rowcolumn-diacritics.txt)

ranger/ext/img_display.py
l4zygreed: replaced

ranger/ext/img_display.py
l4zygreed: Because we need to send messages with escape sequence, but receive without it? Code doesn't have single "send code" function, where we can alter protocol, so it's easier to alter variables. Maybe we can refactor code and unify code sending? 🤔

ranger/ext/img_display.py
l4zygreed: I thinks it's better for enduser if we don't do any parsing at all, doen't it? 🤔
I've replaced csv with python file and added instruction from which file (url) it's maken from, just in case. I think take first column from file and edit it into a python list is not a big deal for any other dev. (But I believe it will not be altered, at least for nearest few years.)

l4zygreed:
Sorry, been busy. (War and busy work - great combination, lol 🤣) 
Thought Ethsan would add the fixes. But I can work on PR a little bit now.

Ethsan:
I wanted too but I was busy and then completely forgot the PR. Sorry about that

ranger/ext/img_display.py
```diff
@@ -3,6 +3,7 @@
 # Author: Emanuel Guevel, 2013
 # Author: Delisa Mason, 2015
 
+# pylint: disable=too-many-lines
```
Ethsan, line 6: I was thinking about may be splitting up the file like ext/vcs. But it's probably outside the scope of this PR

ranger/ext/img_display.py
Ethsan: Agreed, that probably the best solution. And that what seems to do kitten icat too (but in go).

ranger/ext/img_display.py
```diff
@@ -805,6 +813,23 @@ def _late_init(self):
         self.pix_row, self.pix_col = x_px_tot // n_rows, y_px_tot // n_cols
         self.needs_late_init = False
 
+    def _handle_init_tmux(self):
+        # check if tmux allows passthrough
+        # use tmux escape sequence if it's enabled
+        # throw error if it's disabled
+        try:
+            tmux_allow_passthrough = check_output(
+                ['tmux', 'show', '-Apv', 'allow-passthrough']).strip()
```
eMPee584, line 811: `tmux  show` is the shorthand for `tmux show-options`, I think it's better to use the long form because the meaning is more obvious..

l4zygreed:
@toonn Hello, can you review new commits? 🙃

l4zygreed:
@toonn 👋🏻😪

l4zygreed:
@toonn Notice me, senpai 😢

l4zygreed:
@toonn 
a-cartoon-of-a-woman-sitting-on-a-bed-with-the-words-mum-mum-mum-below-her

l4zygreed:
@toonn

l4zygreed:
@toonn check the commits, senpai

ranger/ext/img_display.py
```diff
@@ -719,7 +727,55 @@ def __init__(self):
...
 
+    def _wrap_bite_string(self, msg):
+        # Wrap bite string into protocol.
+        # With tmux passthrough if inside tmux.
+        if self.is_tmux:
+            return b'%s%s%s%s%s' % (
+                self.tmux_protocol_start,
+                self.protocol_start,
+                msg,
+                self.tmux_protocol_end,
```
toonn: I *think* I see what you're doing with `tmux_protocol_{start,end}` but that seems confusing honestly, because of the interleaving.

And couldn't the `msg` contain `\x1b`? I think being explicit about doubling the `\x1b`s would be more straightforward code.

ranger/ext/img_display.py
```diff
@@ -719,7 +727,55 @@ def __init__(self):
         self.pix_row, self.pix_col = (0, 0)
         self.temp_file_dir = None  # Only used when streaming is not an option
 
+    def _wrap_bite_string(self, msg):
+        # Wrap bite string into protocol.
+        # With tmux passthrough if inside tmux.
+        if self.is_tmux:
+            return b'%s%s%s%s%s' % (
```
toonn: Let's use `"".format()` syntax, `%` is really old.

toonn (commented):
Some minor things I wanted to ask, let me know what you think.

ranger/ext/img_display.py
l4zygreed: Bytes doesn't have `format` attribute, so this syntax AFAIK is the only formatting syntax for them.

Workarounds are:
* possible double conversion - format as string and then `encode` into bytes
* do something like `b''.join(protocal_start, msg, protocol_end)`

ranger/ext/img_display.py
l4zygreed: messages containing `\x1b` doesn't use wrap function.

but updated anyway, now should be less confusing.

ranger/ext/img_display.py
```diff
@@ -727,8 +727,8 @@ def __init__(self):
         self.pix_row, self.pix_col = (0, 0)
         self.temp_file_dir = None  # Only used when streaming is not an option
 
-    def _wrap_bite_string(self, msg):
-        # Wrap bite string into protocol.
+    def _wrap_bytes_string(self, msg):
```
Jagholin: I think it's byte_string :)

l4zygreed: another_one_bites_the_dust 🤔

ranger/ext/img_display.py
toonn: My bad, %-formatting is fine then.

ranger/ext/img_display.py
toonn: > messages containing `\x1b` doesn't use wrap function.

`wrapped` includes `protocol_{start,end}` so there's at least those `b'\x1b'`s, no?

l4zygreed: in new code, yes. Messages which require wrapping are wrapped in single x1b's (protocol_start/end), if inside tmux - additionally wrapped into tmux_protocol_start/end with original x1b's doubled.

toonn (approved):
I do think it's clearer this way. Looks ready now. Thank you for the contribution, previews working inside Tmux is awesome! :rocket: :tada:

PR #3107, opened 2025-06-02
fixed bug with command info staying

dgshanee:
The screen is now cleared at the appropriate time. The command info is cleared when the cmd line is cleared.


toonn:
@ask1234560, could you test this PR? It should be as simple as fetching dgshanee's branch and running the `ranger.py` file.

ananthu-sk:
@toonn tested, it's working. Great work @dgshanee :)

toonn:
Thanks for the fix!

PR #3109, opened 2025-06-12
img_display: Avoid unicode escape sequences for Ueberzug input

toonn:
In [ueber-devel/ueberzug#26](https://github.com/ueber-devel/ueberzug/pull/26) Ueberzug stopped accepting `\` escape characters in input file names.

We should simply pass along the original filename rather than replace non-ASCII characters with `\uXXXX` escape sequences.


vejkse (approved):
This is what I have been using for some time, and it fixes the problem with ueberzug and non-ASCII characters, including with a non-UTF-8 locale.

Issue #2984, opened 2024-06-12
Preview archives only if smaller than a specified size


flyingrhinonz:
ranger --version
ranger version: ranger 1.9.3
Python version: 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Locale: en_NZ.UTF-8

#### Current Behavior

Slow to preview large archives.

#### Expected Behavior

Request config option to exclude this behavior if archive is bigger than specified size.

#### Context

Very slow if a multi Gb archive is selected

PR #3114, opened 2025-06-25
Preview archives only if smaller than a specified size #2984

initumX:
#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
ranger version: ranger-master 
Python version: 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0]
Locale: en_US.UTF-8
Debian 12 "Bookworm"

#### CHECKLIST

- [X ] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [ X] All changes follow the code style **[REQUIRED]**
- [ X] All new and existing tests pass **[REQUIRED]**
- [ X] Changes require config files to be updated
    - [ X] Config files have been updated
- [X ] Changes require documentation to be updated
    - [ ] Tests have been updated

#### DESCRIPTION

- added new setting `max_archive_preview_size` on rc.conf, it looks like this
`set max_archive_preview_size 20M`
- added this setting into `ALLOWED_SETTINGS` inside `container/settings.py`file like this:
`'max_archive_preview_size': str,`
- added `parse_human_readable` function into `ext/human_readable.py` file (used for parsing human readable option 20M and convert it into bytes)
- added `is_archive_preview_allowed` function into `container/file.py` file. This function gets `max_archive_preview_size` from settings, compare its value with current file size and returns True or False
- modified `has_preview` function from the same file. This is modified part:
```
if self.fm.settings.preview_script and \
                self.fm.settings.use_preview_script:
    if self.container:
        return self.is_archive_preview_allowed()
    return True
  ```
this `self.is_archive_preview_allowed(self.path)` returns true if `max_archive_preview_size` is blank or equal to zero or if its value is greater or equal to file size. So, nothing is broken. All these modifications affects only on "container" files and only if their size is greater than `max_archive_preview_size` (and only when preview script is used). In all other cases everything should work as worked before.

UPD:
Modified `mkdir` from `commands.py`. Now it can create multiple folders at once

#### MOTIVATION AND CONTEXT
#2984

### DIFF
--- a/ranger/colorschemes/solarized.py
+++ b/ranger/colorschemes/solarized.py
@@ -93,7 +93,7 @@ class Solarized(ColorScheme):
         elif context.in_titlebar:
             attr |= bold
             if context.hostname:
-                fg = 16 if context.bad else 255
+                fg = default if context.bad else 255
                 if context.bad:
                     bg = 166
             elif context.directory:


toonn:
Please split the mkdir changes into a separate PR.

initumX:
I've deleted mkdir changings from this PR. Idk how to add it from new PR, it pushes everything here

initumX:
lol, It appeared here again. I have no idea how does github work. This is just 30 lines of code, doing very simple job. No mistakes there. But one issue is there. This line `self.fm.notify("Created directory: {0}".format(full_path))` shows only the last one created folder. If it's a problem, you can create a list `created_dirs = []` and append a dirname there right after makedirs(full_path), like this 
`created_dirs.append(dirname)` and then use it in fm.notify `self.fm.notify("Created directories: {0}".format(created_dirs))`. This way you can use mkdir like this:

`:mkdir folder1/folder2 yetonefolder`
`mkdir folder3 folder4`
`mkdir folder5`

PR #3127, opened 2025-08-03
Various patches for review [DRAFT]

eMPee584:
This is the first iteration of the consolidation of my very messy ranger tree 🎄😆
As school starts again tomorrow, I just wanted to get this out the door.. Most
patches should be in a good state and are thoroughly tested, some have room for
improvement. Let me know! I learned a lot about object oriented programming with
python, and discovered some new tools for development (after spending too much
time adding and removing `self.fm.notify` statements..):

- https://github.com/inducer/pudb - Full-screen console debugger for Python
- https://github.com/kylepollina/objexplore - A terminal UI to inspect and explore Python objects
- https://github.com/namhyung/uftrace - Function graph tracer for C/C++/Rust/Python
- https://github.com/gaogaotiantian/viztracer - trace and visualize python code execution

Oh and I still have to do the `make test` thing, admittedly. This is the first iteration..

### Changelog (prefilled from list of commits)

- **Add tag filter to filter_stack**
- **Make tag filter more explicit about behaviour when no tag is specified**
- **Make format spec in added tag filter explicit (E4220 implicit-format-spec)**
- **runner/shell command: add -l flag to clear screen before running command**
- **Fix opening tabs on files (regression from fe7c3b2 remove-duplicate-tabs-on-startup)**
- **Show statusline indicator when show_hidden is true**
- **Change show_hidden status indicator to colored asterisk (*)**
- **handle middle mouse click**
- **tab cursor positions: don't crash if a parent folder is not readable**
- **default colorscheme: change border / separator lines from white to gray**
- **miller / multipane view: add settings.draw_borders_rounded**
- **miller / multipane view: rename variable & reduce indention**
- **default colorscheme: change color of inactive pane contents to dim gray**
- **browsercolumn widget: EMPTY instead of empty**
- **browsercolumn widget: don't enter dir repeatedly**
- **widgets.view_miller: add preview_min_width setting as collapse trigger**
- **bookmarks: support Directory and str bookmark types throughout**
- **Switch default bookmark type from directory objects to string list**
- **fm: mark directory as unloaded on garbage collection**
- **fm: also free unfiltered list of files on garbage_collection**
- **ui: correctly inherit global tmux options (requires tmux >3.0)**
- **ext.rifle: fix mysterious off-by-one errors in open_with**
- **ext.human_readable: explicitly show Byte unit for size 0, too**
- **core.actions: close hints menu before get_cumulative_size is invoked**
- **core.linemode: add MimeTypeLinemode**
- **core.main: set process title to 'ranger'**
- **core.runner: less 🧟 zombies!**
- **core.runner: clear zombies**
- **Add a fudge_symlink_stat setting to allow influencing sorting / display**
- **Document unhandled miller levels (for more than three columns) with FIXMEs**
- **gui.ui: make suspend() more robust against flukes**
- **gui.widgets: move _draw_border_rectangle() to view_base and share**
- **browsercolumn widget: supposedly fixes some scrolling issue [DRAFT]**
- **gui.widgets.statusbar: redraw on show_hidden change**
- **gui.widgets.taskview: remove a finalize() that does nothing but crash**
- **gui.widgets.taskview: these two addstr() calls are actually addnstr()**
- **gui.widgets.view_miller: keep preview collapse status updated**
- **gui.widgets.statusbar: show FROZEN marker even if directory not loaded**
- **Add an idle_auto_freeze setting to freeze ranger after # seconds**
- **container.directory: we want the _current_ settings, yes really, thanks**
- **widgets.statusbar: add a space left of the show_hidden marker**
- **gui.widgets.view_multipane: no more phantom columns 👻**
- **gui.widgets.view_multipane: one main column only**
- **gui.widgets.view_multipane: load unloaded panes on poke**
- **gui.widgets.view_base: don't draw border when height is too low**
- **gui.widgets.view_multipane: drop explicit redraws, already handled**
- **config.commands: fix the default_linemode command to update display**
- **gui.widgets.view_multipane: fix layouting so borders don't cover tags**
- **gui.widgets.view_multipane: implement a click handler so mouse works**
- **rc.conf: give hash filter mapping current file as default parameter**
- **These have a request_redraw() function so make use of that**
- **core.tab.enter_dir(): actually return False for failure modes**
- **gui.widgets.statusbar: indicate the lack of all that could be shown**
- **Add a linemode_shellcommand plugin**


eMPee584:
ah damn.. tried to rename the branch but github doesn't seem to allow that ..

PR #3131, opened 2025-08-17
Plugin to replace draw_possible_programs with fzf (almost working)

eMPee584:
This worked sufficiently well (apart from not handling resizes while fzf
is shown) when I patched the actions module directly, but as a plugin,
the override function is never called..👽

Any ideas?

### DIFF
--- a/examples/plugin_fzf_open_with.py
+++ b/examples/plugin_fzf_open_with.py
@@ -1,0 +1,63 @@
+# This plugin replaces the open_with menu with fzf.
+from __future__ import (absolute_import, division, print_function)
+
+import logging
+import subprocess
+
+import ranger.core.actions
+
+LOG = logging.getLogger(__name__)
+
+# Save original function
+DRAW_INFO_OLD = ranger.core.actions.Actions.draw_possible_programs
+HOOK_INIT_OLD = ranger.api.hook_init
+
+# Define a new info drawing function
+@classmethod
+def draw_possible_programs_override(self):
+    import pudb
+    pu.db
+
+    DRAW_INFO_OLD(self)
+    program_info = self.ui.browser.draw_info
+    self.ui.browser.draw_info = []
+
+    win_hei = self.ui.browser.hei + 2
+    win_wid = self.ui.browser.wid
+    fzf_menu_hei = min(win_hei, len(programs) + 3)
+
+    self.ui.win.move(win_hei - fzf_menu_hei, 0)
+    self.ui.win.refresh()
+
+    label = ' open %s with '
+    label_pos = 3
+    path_str = target.relative_path
+    max_len = win_wid - (len(label) + label_pos)
+    if len(path_str) > max_len:
+        path_str = path_str[:max_len - 1] + self.ui.browser.ellipsis[self.settings.unicode_ellipsis]
+
+    process = subprocess.Popen(['fzf', '--exact', '--accept-nth=1', '--height=%d' % fzf_menu_hei,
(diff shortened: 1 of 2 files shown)


eMPee584:
ranger-testing_2025-08-17_21-37-20

vejkse:
> This worked sufficiently well (apart from not handling resizes while fzf is shown) when I patched the actions module directly, but as a plugin, the override function is never called..👽

In the `load_settings` function: https://github.com/ranger/ranger/blob/968b7067d97bcaf3d23479780efdecd203c3e779/ranger/core/main.py#L387 the `Actions` are loaded by `fm.commands.load_commands_from_object` *before* plugins are loaded.  Depending on how exactly `load_commands_from_object` works, the commands generated by it might be stuck with the original version of `draw_possible_programs`.

If that is the case, you might have to make it into a `Command` rather than an `Action`, if it is possible.

---

Apart from the resizing issue, I noticed another problem: after cancelling or selecting an action, once we come back to ranger, we get the original `:open_with ` prompt.

eMPee584:
Ah cool @vejkse thanks for the cues, going to try to apply and see if I can make it work! 😎
> Apart from the resizing issue, I noticed another problem: after cancelling or selecting an action, once we come back to ranger, we get the original `:open_with ` prompt.

Ah yeah good catch, I had `map r  draw_possible_programs` only in my personal config to override `map r  chain draw_possible_programs; console open_with%space`, that either needs to be set from the plugin or in the default config .. thanks for testing again! Will have another go at it soon..ish 😅💦

vejkse:
I tried making your `Action` into a `Command` and it’s mostly a question of replacing `self.ui` by `self.fm.ui`.

Here’s what I’m using now, with some personal changes in the fzf interface.  I also directly use `rifle.list_commands` and `rifle.execute`, which may be less stable in the long run than what you do.  There is more flickering than with the default `chain draw_possible_programs; console open_with%space`; it may be unavoidable.
```python
import subprocess


class fzf_open_with(Command):
    """
    :fzf_open_with

    Select with fzf which rifle command to open the selection with.
    """

    def execute(self):
        targets = [target.path for target in self.fm.thistab.get_selection()]
        commands = list(self.fm.rifle.list_commands(targets))

        win_hei = self.fm.ui.browser.hei + 2
        fzf_menu_hei = min(win_hei, len(commands) + 1)
        self.fm.ui.win.move(win_hei - fzf_menu_hei, 0)
        self.fm.ui.win.refresh()

        formatted_commands = [
            "{}\t{}\t{}".format(label or "", id, com)
            for id, com, label, _flags in commands
        ]
        process = subprocess.Popen(
            [
                "fzf",
                "--height={}".format(fzf_menu_hei),
                "--layout=reverse-list",
                "--pointer=",
                "--marker=",
                "--no-info",
                "--prompt=Open with … ",
                "--tiebreak=index",
            ],
            text=True,
            stdin=subprocess.PIPE,
            stdout=subprocess.PIPE,
        )
        stdout, _ = process.communicate(input="\n".join(formatted_commands))

        self.fm.ui.suspend()
        self.fm.ui.initialize()
        if stdout:
            _, id, _ = stdout.split("\t", maxsplit=2)
            self.fm.rifle.execute(targets, number=int(id))
```

eMPee584:
Now finally got to this 😄

> I tried making your `Action` into a `Command` and it’s mostly a question of replacing `self.ui` by `self.fm.ui`.

Ah cool that is a much saner way and much less convoluted code, thanks! going to poke it a bit...

toonn:
As a plugin it'd be best self published and then we can just link to it from the wiki.

Issue #1925, opened 2020-04-19
infinite recourse on copying folder into sub-folder


sadid:
#### Runtime Environment
```
ranger version: ranger 1.9.3
Python version: 3.8.2 (default, Apr  8 2020, 14:31:25) [GCC 9.3.0]
Locale: en_US.UTF-8

```
- Operating system and version:  Archlinux (5.4.32-1-lts)
- Terminal emulator and version: urxvt
- Python version: 3.8.2

#### Current Behavior
1. Create folder/sub-folder `parent/child`
2. Copy parent and paste it into child

The result will be infinite nested sub-folders `parent/child/parent/child/parent/child/...`

#### Expected Behavior
just like cp it should create one level of nest aka `parent/child/parent/child` in the folder structure.

#### Possible Solutions

1. Using standard `cp` (at least in Linux/Mac) might be helpful. However, I'm not aware of special decision on using implement copy mechanism
2. ranger --debug crash (while ranger don't). It can be detected at least.

#### Steps to reproduce
1. Create folder/sub-folder `parent/child`
2. Copy parent and paste it into child

#### Extra Note

Trash-Cli also remove the file but trash-empty will give error as:
```
RecursionError: maximum recursion depth exceeded while calling a Python object
```

However, `rm -rf` can fix the problem.

#### Trackback
This happens only on `ranger --debug` and not regular renger:
```
ranger version: ranger 1.9.3
Python version: 3.8.2 (default, Apr  8 2020, 14:31:25) [GCC 9.3.0]
Locale: en_US.UTF-8

Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/ranger/core/loader.py", line 423, in work
    next(item.load_generator)
  File "/usr/lib/python3.8/site-packages/ranger/core/loader.py", line 127, in generate
    for n in shutil_g.copytree(
  File "/usr/lib/python3.8/site-packages/ranger/ext/shutil_generatorized.py", line 242, in copytree
    raise Error(errors)
[...]

PR #3147, opened 2025-10-18
shutil_g.copytree: fix infinite recursion bug (#1925)

qrvd:
#### ISSUE TYPE

x Bug fix

#### RUNTIME ENVIRONMENT

- Python version: 3.11.2
- Ranger version/commit: master 6d5e0d8dc4cc5ddf53211e19f48c5b5e9ee47b18
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Record the destination directory in a set of "ignored trees" and pass this variable down to recursive instances of `copytree()`.

Remove the `ignore` keyword argument mostly because Pylint complains that the function is too big otherwise. Fortunately it was never used anyway.

#### MOTIVATION AND CONTEXT

When copying a folder into its subtree, `copytree()` will
recursively copy the newly created directory, and enter an infinite recursive descent, because it doesn't know which of the files it already created.

This is somewhat of an edge case and some file managers outright refuse to handle it. Even the version of `cp` running on my machine won't do this. But it seems like a simple bug for us to fix and it'd give the user more freedom.

#1925

#### TESTING

Create a directory, and create another directory inside of it.

Optionally create additional files and add them to either directory.

Now copy the first directory (`:copy`), navigate into the second directoy, and run the `:paste` command. All the contents of the source directory will be in the second directory as a user would expect.

Changes to `shutil_generatorized` affect the `CopyLoader` class, which is its only user. The specific code paths involved would only affect copying directories using `:copy` followed by `:paste`.

### DIFF
--- a/ranger/ext/shutil_generatorized.py
+++ b/ranger/ext/shutil_generatorized.py
@@ -204,13 +204,12 @@ def copytree(  # pylint: disable=too-many-locals,too-many-branches
     it is false, the contents of the files pointed to by symbolic
     links are copied.
     """
-    realsrc = os.path.abspath(os.path.expanduser(src))
+    realsrc = os.path.realpath(os.path.expanduser(src))
     if _ignored_trees is None:
         _ignored_trees = set()
-        realdst = os.path.abspath(os.path.expanduser(dst))
+        realdst = os.path.realpath(os.path.expanduser(dst))
         realdstdir = os.path.dirname(realdst)
-        if _destinsrc(realsrc, realdstdir) \
-                and not (symlinks and os.path.islink(realdstdir)):
+        if _destinsrc(realsrc, realdstdir):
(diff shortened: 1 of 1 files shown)


toonn (changes requested):
I'm not a fan of dropping functionality for the sake of appeasing the linter. I'm also not sure that some plugin doesn't depend on this so I'd prefer to leave it in. It also seems like we could use `ignore` to tackle this problem, no?

How does this handle a directory name that occurs multiple times throughout the directory trees involved?

What does upstream `copytree` do in this case, create a recursively nested structure like we do or do they handle it better?

hut:
Love this. This issue occurred to me so many times, but strangely I never actually considered fixing it. Thank you.

Can't review it atm, but hope to get to this soon.

Issue #3087, opened 2025-04-28
trash-put is equivalent to delete (i.e. command 'rm') when the number (before '| trash-put ***') is selected.


webridging:
There is a issue that bothers me. The numbers,  '0 | ***' and the like,  are discontinuous.
 
![Image](https://github.com/user-attachments/assets/08488ccc-cbd5-4048-ac4a-c4933145b459)

![Image](https://github.com/user-attachments/assets/9a5c1acb-bbf0-4552-8a6f-326a6bd8c2e1)
This is equivalent to "rm"

![Image](https://github.com/user-attachments/assets/d18369ef-f285-4857-a049-f4c5206f40fa)
![Image](https://github.com/user-attachments/assets/91f17ee4-346c-4e3f-aa4e-0776a637cc3a)
![Image](https://github.com/user-attachments/assets/84ba1143-3a4f-4aa9-a9b9-9836324ec8c6)
These are equivalent to "trash-put"

![Image](https://github.com/user-attachments/assets/33cac17a-4abc-4a2a-9723-e974f3134a9f)

toonn:
This is a known issue caused by an attempt to fix misnumbering IIRC.

qrvd:
Command `:open_with` should probably not allow users to pass a command number that is not listed, so that for example `rm` isn't executed by accident. I'm working on a patch to ensure this so that it's safer to use.

qrvd:
#1756

hut:
Where does ranger get the idea from to run a `rm`? Do you have it in your rifle.conf? @webridging

webridging:
> Where does ranger get the idea from to run a `rm`? Do you have it in your rifle.conf? [@webridging](https://github.com/webridging)

I checked my rifle.conf, and there is no rm entry configured there.

<img width="852" height="759" alt="Image" src="https://github.com/user-attachments/assets/64f76ac2-3ca7-4971-902d-1ad807e51fda" />

hut:
Ok, thank you.

@toonn @qrvd does anyone of you understand where the deletion came from?

Issue #1756, opened 2019-11-16
draw_possible_programs sometimes skips menu entries


toonn:
`draw_possible_programs` (bound to `r`) sometimes skips entries in the menu.
Example:
```
0 | ${VISUAL:-$EDITOR} -- "$@"
1 | "$PAGER" -- "$@"
3 | mkdir -p -- ${XDG_DATA_DIR:-$HOME/.ranger}/ranger-trash; mv -- "$@" ${XDG_DA
```

This can be confusing, especially because that entry is *not* disabled for `open_with`.

qrvd:
This was difficult to see but it seems that we only exclude the "ask" option when generating the list of programs for display:

https://github.com/ranger/ranger/blob/6d5e0d8dc4cc5ddf53211e19f48c5b5e9ee47b18/ranger/core/actions.py#L942-L949

but not when we select the program to execute, which is inconsistent, because Rifle will re-create its own list of commands internally in our call to`Rifle.execute()` and then use our chosen label or number to find the command it thinks we chose:

https://github.com/ranger/ranger/blob/6d5e0d8dc4cc5ddf53211e19f48c5b5e9ee47b18/ranger/ext/rifle.py#L421-L431

So the two generated lists will have differently numbered commands. Rifle ultimately picks the command from a list that includes `ask` but the user chooses a number from a list that excludes it.

vifon's hotfix #1711 fixes the bug *visually* so that the command numbers match. But besides causing gaps in the list users see, users can still select the command numbers that aren't shown, which will execute whatever Rifle internally thinks is at that position, causing issues like #3087.

qrvd:
I think we'd really want the two lists to always be exactly the same: users should always get what they chose. I think the best way to ensure this, with minimal maintenance, is by storing the list we show the user and having Rifle choose the command from that list later, with the same numbering. I wrote a patch that does this, about to submit a PR: https://github.com/qrvd/ranger/commit/1851c1c3375c644f4ba9048bbbc9bb730b5ff0a6

qrvd:
I just pushed another possible fix for comparison https://github.com/qrvd/ranger/commit/737197abb7a07a2a49df3a34daed7d3c6dd5cbf3

toonn:
The problem I expect with 1851c1c is having multiple matches with the same label. I think the only way to reliably select in that case is by number, no?

737197a adds skip_ask to the corresponding execute call, right?

[...]

PR #3150, opened 2025-10-22
Fix command numbering bugs (#1756, #3087)

qrvd:
Don't skip command numbers in `draw_possible_programs`. Before:
image
After:
image

And don't allow users to execute program numbers that aren't shown in the list. `:open_with 4` used to be allowed in the "Before" version, with unpredictable results.

#### ISSUE TYPE

x Bug fix

#### RUNTIME ENVIRONMENT

- Python version: 3.11.2
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

- New field `FM.displayed_commands` is used to track the list of commands displayed to the user in `draw_possible_programs`.
- `Actions.execute_file` passes the command list to `Rifle.execute`
- `Rifle.execute` gets a new keyword argument `command_list` which it may use instead of recreating the list of programs internally
- Ranger's `:open_with` command will not allow for selecting commands by number when the list of programs is not visible to the user. So if `:open_with (number)` is executed without a preceding call to `draw_possible_programs`, it will fail. This should increase safety but may be overkill. The current version of the patch breaks the use case where a user gave a command a specific number in `rifle.conf` and just wants to execute it by that number.
- Rifle, when run standalone, will store the generated list of commands so that it prints the same commands shown to the user as it will select from, which makes it more consistent.

#### MOTIVATION AND CONTEXT

Explained the bug in issue comments: https://github.com/ranger/ranger/issues/1756#issuecomment-3419132737

Some relevant issues: #1756, #3087

#### TESTING

The list of programs to be executed now doesn't skip any numbers, and Ranger will now refuse to execute a command number that isn't on the list.

**How do the changes affect Rifle?**
The newly introduced keyword arguments to `Rifle.execute()` have default values that correspond to the old behavior, so Rifle should continue working the same.
When run standalone, Rifle now stores its generated list of commands, which makes it more consistent.

**How do the changes affect Ranger?**
The changes to `Actions.execute_file()` should not break any code that calls to it since previously.

### DIFF
--- a/ranger/ext/rifle.py
+++ b/ranger/ext/rifle.py
@@ -627,12 +627,14 @@ def main():  # pylint: disable=too-many-locals
         rifle = Rifle(conf_path)
         rifle.reload_config()
         # print(rifle.list_commands(sys.argv[1:]))
+        commands = list(rifle.list_commands(positional))
         if options.l:
-            for count, cmd, label, flags in rifle.list_commands(positional):
+            for count, cmd, label, flags in commands:
                 print("%d:%s:%s:%s" % (count, label or '', flags, cmd))
         else:
             try:
-                result = rifle.execute(positional, number=number, label=label, flags=options.f)
+                result = rifle.execute(positional, number=number, label=label,
+                                       flags=options.f, command_list=commands)
             except CalledProcessError as ex:
                 exit_code = ex.returncode
             else:


eMPee584:
.. did you see commit 1ac12db97f22e04f1d3a08fbe79e20c0ec277755 which fixed it for me?
**Please** test against this.. Not sure the code needs to become more complicated to solve the problem! And no time to dive back into it right now..

qrvd:
Hi @eMPee584. Last week I saw your commit linked in issue #1756 and was happy to immediately try your fix, just like you I like simple elegant solutions.

However I kept reexperiencing the issue. So I started over and really racked my brain trying to debug the root cause of the problem and this is what I came up with. I understand we're all busy so whenever you have the time maybe you could let me know if I misunderstood the problem? [(link)](https://github.com/ranger/ranger/issues/1756#issuecomment-3419132737)

eMPee584:
Hah, that's interesting, can you describe how to recreate the issue with 1ac12db97f22e04f1d3a08fbe79e20c0ec277755 applied on top of master? then I'll have another look at it too..
Sorry I was overreacting a bit, got scared that you were working on some of the same issues i already broke my head about to fix.. 😅💦 it really is a tricky problem, lol, and I'm definitely not claiming I understood it fully. It's good to hear that you did indeed test my patch first before coming up with this (and the other version), so that's somewhat of a relief 👌

(What's more, I started reading the recently merged #2237 and it almost seemed like AI commits to me always with a perfect explanation and redoing code sections over and over.. But I didn't dive into it as CTRL+Z is just not part of my workflow at all (I use ephemeral ranger panes in tmux.. a _lot_ of them) so it just gave me a strange feeling admittedly :man_shrugging: .. But hey, obviously that PR is properly reviewed code that was in the making for years, so I knew that to be kinda implausible.)

qrvd:
No problem I totally understand your reaction. I'm sorry too, I probably should have communicated with you since you were already trying to fix the same problem. We're all just trying to improve Ranger

> Hah, that's interesting, can you describe how to recreate the issue with https://github.com/ranger/ranger/commit/1ac12db97f22e04f1d3a08fbe79e20c0ec277755 applied on top of master? then I'll have another look at it too..

I tested using [this `rifle.conf` file](https://github.com/ranger/ranger/blob/7ad50fadca91646d78a8d6cb3108a29b256df2b8/ranger/config/rifle.conf). Apparently when I select a directory and I hit `r` to bring up the list, I see a list of programs that skips a number. In my tests I also reproduced the issue when I select a file with some unusual extension such as `.pyc`. So with 1ac12db whether it's skipped depends on which programs are in the list and results may be different depending on your `rifle.conf`.

> But hey, obviously that PR is properly reviewed code that was in the making for years, so I knew that to be kinda implausible.

I don't use AI, this is just how I write and code. As you can see I started with a flawed fix and eventually worked through the problems with toonn's guidance. Some of my longer explanations in that PR took me hours of searching and testing because I really wanted everything to be correct. I'm flattered that you find them perfect, that pull request was really hard work.

PR #3151, opened 2025-10-22
Fix UI freeze when initiating a copy/move of a big file tree

qrvd:
Before moving or copying files, CopyLoader traverses the full directory trees two times to calculate the total file size. This traversal is now only done once and generatorized to prevent Ranger from freezing when there is a large number of files to move. File moves that require just a rename will also yield now, again to avoid a UI hiccup when there are many items to move or we're on a slow/networked file system.

#### ISSUE TYPE

x Bug fix

#### RUNTIME ENVIRONMENT

- Python version: 3.11.2
- Locale: en_US.UTF-8

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

- `shutil_generatorized.move()`: yield after renaming
- `CopyLoader._calculate_size()`: yield after calculating the size of a file or listing a directory
- `CopyLoader.generate()`: calculate the file size by traversing the tree only once. Calculate `self.percent` in terms of bytes since this is what `shutil_generatorized` gives us

#### MOTIVATION AND CONTEXT

Traversing the full tree of files to be moved can take a lot of time, rendering Ranger completely unresponsive. Such work should be generatorized.

We also fix the percentage calculation and no longer traverse the tree twice, see commit message: https://github.com/ranger/ranger/pull/3151/commits/3e39b6738fab0c761a427f5e586c4f151b46ac05

`shutil_generatorized.move()` will yield periodically when moving files, except when a file can be renamed to its destination. Instead it should probably always yield since renaming files is an I/O operation with performance impact depending on how slow the filesystem is or the number of files being renamed.

#### TESTING

Tested by moving directories containing 100K+ files in total, from one folder to another. After the changes, the freezes no longer occur.

Changes should improve the performance of the `:paste` command. No adverse changes to the rest of the codebase since only `CopyLoader` uses `shutil_generatorized`.

### DIFF
--- a/ranger/core/loader.py
+++ b/ranger/core/loader.py
@@ -4,7 +4,6 @@
 from __future__ import (absolute_import, division, print_function)
 
 import errno
-import math
 import os.path
 import select
 from collections import deque
@@ -94,12 +93,13 @@ class CopyLoader(Loadable, FileManagerAware):  # pylint: disable=too-many-instan
                     pass
         yield size
 
-    def generate(self):
+    def generate(self):  # pylint: disable=too-many-branches
         if not self.copy_buffer:
             return
 
         from ranger.ext import shutil_generatorized as shutil_g
         # TODO: Don't calculate size when renaming (needs detection)
+        size = 0
         for size in self._calculate_size():
             yield
         size = max(1, size)


qrvd:
https://github.com/ranger/ranger/blob/21e3c19c020ed399fa5b2f6c7667e7186806f908/ranger/core/loader.py#L103-L107

Ranger does a lot of I/O here to pre-calculate the total size of files moved, just for the progress bar. Is it really worth it or shouldn't we remove it / make it a setting? On my system it makes a huge difference to just skip this step and go ahead with the move/copy.

PR #3159, opened 2025-10-31
Fix PyPy3 doctest in signals.py

hut:
Fixes the pypy3 doctest in signals.py and adds this warning:

> WARNING: Signal.signal_bind(..., weak=True) is buggy on PyPy!
> 
> As of 2025-10-31, PyPy has delayed garbage collection, which results in
> signal handlers being run on objects that should have been deleted already.
> DO NOT put side effects in signal handler functions that might cause changes on
> the file system for this reason.

### DIFF
--- a/ranger/ext/signals.py
+++ b/ranger/ext/signals.py
@@ -17,6 +17,13 @@ deleted when trying to call them (in signal_emit), but if they are never
 called, they accumulate and should be manually deleted with
 signal_garbage_collect().
 
+WARNING: Signal.signal_bind(..., weak=True) is buggy on PyPy!
+
+As of 2025-10-31, PyPy has delayed garbage collection, which results in
+signal handlers being run on objects that should have been deleted already.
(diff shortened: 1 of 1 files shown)


hut:
The pypy2 test still fails but that's probably not worth fixing and should be removed.

fwam:
As we talked on the IRC channel on 30.10.2025
> 18:39 <@vifon> My stance would be: don't break the code that works in 2.7 for the sake of it, but if dropping it from the CI would makes things vastly easier, it might be worth it without sacrificing much.

fwam:
`pylint` still has has some issues. (ignoring the `invalid-name` issues. i'd add it to `.pylintrc` ignored warnings.)
```
ranger/core/actions.py:822:47: E0606: Possibly using variable 'fnc' before assignment (possibly-used-before-assignment)
ranger/core/actions.py:1064:23: W4902: Using deprecated method open() (deprecated-method)
```

PR #3162, opened 2025-11-02
feat(runner): add support for non-sudo in 'r' flag

fwam:
#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
Add support for `doas` and `run0` in `-r` flag (no support for `f` flag yet.)

#### MOTIVATION AND CONTEXT
Hard-coded `sudo` might cause problems on systems where `sudo` isn't installed.

### DIFF
--- a/ranger/core/runner.py
+++ b/ranger/core/runner.py
@@ -226,13 +226,26 @@ class Runner(object):  # pylint: disable=too-few-public-methods
                 wait_for_enter = True
         if 'r' in context.flags:
             # TODO: make 'r' flag work with pipes
-            if 'sudo' not in get_executables():
-                return self._log("Can not run with 'r' flag, sudo is not installed!")
+            # XXX: is there any way to make it better?
+            root_commands = ['sudo', 'doas', 'run0']
+            root_command = None
+            for tested_command in root_commands:
+                if tested_command not in get_executables():
+                    continue
+                else:
+                    root_command = tested_command
+                    break
             f_flag = ('f' in context.flags)
-            if isinstance(action, str):
-                action = 'sudo ' + (f_flag and '-b ' or '') + action
-            else:
-                action = ['sudo'] + (f_flag and ['-b'] or []) + action
+            # TODO: 'f' flag for doas/run0
+            match(root_command):
+                case 'sudo':
+                    action = 'sudo ' + ('-b ' if f_flag else '') + action
+                case 'doas':
+                    action = 'doas ' + action
+                case 'run0':
+                    action = 'run0 ' + action
+                case _:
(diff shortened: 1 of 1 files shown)


ranger/core/runner.py
```diff
@@ -226,13 +226,26 @@ def __call__(
...
         if 'r' in context.flags:
             # TODO: make 'r' flag work with pipes
-            if 'sudo' not in get_executables():
-                return self._log("Can not run with 'r' flag, sudo is not installed!")
+            # XXX: is there any way to make it better?
+            root_commands = ['sudo', 'doas', 'run0']
+            root_command = None
+            for tested_command in root_commands:
+                if tested_command not in get_executables():
+                    continue
```
belkka:
```python
for x in seq:
   if not condition:
      continue
   else:
      # do smth
```
Could it be simplified, avoiding one of `if` branches and `continue`?

```python
for x in seq:
   if condition:
      # do smth
```
---
BTW, an alternative way to reach the same result in a single assignment, instead of nested `for` + `if`:

```python
root_command = next(
   (cmd in root_commands if cmd in get_executables()),
   None,
)
```
generator expression is used to filter out elements, and
built-in [`next()`](https://docs.python.org/2.7/library/functions.html#next) is used to select first element of sequence, or `None` if empty

ranger/core/runner.py
```diff
@@ -226,13 +226,26 @@ def __call__(
...
+                else:
+                    root_command = tested_command
+                    break
             f_flag = ('f' in context.flags)
-            if isinstance(action, str):
-                action = 'sudo ' + (f_flag and '-b ' or '') + action
-            else:
-                action = ['sudo'] + (f_flag and ['-b'] or []) + action
+            # TODO: 'f' flag for doas/run0
+            match(root_command):
```
belkka: Note: `match` statements are added in Python 3.10

AFAIK (based on setup.py content), ranger is supporting Python 3.1 and even 2.7 :disappointed:

toonn:
@belkka, if you want to pick up this PR, feel free. Support for run0 and doas is desirable.

belkka:
Noted.

Could you comment on whether `match:` syntax is actually allowed in the project, and whether the `if` brach in the existing code is unreachable? In case anyone picks PR up.

To be fair, i wasn't interesred in the feature specifically. Rather wanted to help dealing with an [overwhelming][1] number of PRs, for the project I like.

[1]: https://github.com/ranger/ranger/pull/2983#issuecomment-3447927002

Aug 15, 2026 10:29:56 toonn ***@***.***>:

PR #3169, opened 2025-11-05
Revert "rifle: rename global variable to fix lint error"

toonn:
This reverts commit dafecb97712e7a06eb9b9d1d9f140f87cdebafc9.

This is currently the only error I get locally running `make test`. Maybe this is due to a version difference? Creating a PR to check whether CI is of a different mind.


toonn:
Yep, CI is running more recent versions.

PR #3170, opened 2025-11-06
Revert "Remove pypy2.7 test from the workflow"

### LINKED ISSUE #3168: Reinstate pypy2 CI checker
Apparently github actions doesn't support python 2.x syntax checking and the pypy2 CI test (removed in #3160) was our only one that checks for python2 syntax errors.

Unless there's a better way to check for python 2.x syntax, let's add pypy2 CI back in.

toonn:
This reverts commit c6238a4265dba9e8bdd972eabd10a864fe16386c.

Closes #3168

We could expand the task to *Fixing* the test if you want an open issue for it.

Enabling CI makes it more visible that we want this fixed and easier for people to check out what the problems are in the logs rather than having to set up a local Pypy 2.7 and such.

IMO there's no point delaying re-enabling this.

Second commit message provides some context around dropping the linting from Pypy:

Pypy 2.7 is stuck with older versions of Flake8 and Pylint. Old Pylint
    would be attractive to run the Python 2 compatibility checks that have
    since been removed but we can't run all the lints because old and new
    Pylint can disagree, which would make it impossible to satisfy both CI
    checks.

    Pypy 3 seems to lag behind CPython at least on GHActions, this leads to
    version mismatches, at the time of this commit Flake8 is at version
    7.4.0 for CPython and 7.3.0 for Pypy 3.10 for example. This too can make
    it impossible to satisfy both CI checks unless we constrain the versions
    of the linters.

    We unconstrained the linters intentionally to keep up with changes
    gradually rather than having to do sweeping refactors every time we
    decide to bump the linter versions.

    Functional tests are more important than linting so we'll have to rely
    on those to catch syntactical regressions. This isn't ideal because our
    tests don't have great coverage but it's the most practical way forward.

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -719,8 +719,10 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
         if func is not None:
             self.settings['sort'] = str(func)
 
-    def mark_files(  # pylint: disable=redefined-builtin,too-many-arguments
-        self, *, all=False, toggle=False, val=None, movedown=None, narg=None
+    def mark_files(
+        # pylint: disable=redefined-builtin,too-many-arguments
+        # pylint: disable=too-many-positional-arguments
+        self, all=False, toggle=False, val=None, movedown=None, narg=None
     ):
         """A wrapper for the directory.mark_xyz functions.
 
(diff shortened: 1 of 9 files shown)


toonn:
The Pypy 2 tests already caught the first regression! I had introduced a Python 3-ism in the form of using anonymous `*` to force the use of keyword arguments.

If we merge this PR we can make the Pypy CI a required check on PRs.

hut (dismissed):

Issue #2937, opened 2024-01-14
SIXEL preview failed


freeconst:
After recent ArchLinux updates, SIXEL preview show error.
However command `convert <image> -geometry ${GEOMETRY:=800x480} sixel:-` works correctly.

#### Runtime Environment

- Operating system and version: Arch Linux (rolling)
- Terminal emulator and version: Foot version: 1.16.2 +pgo +ime +graphemes -assertions
- Python version: 3.11.6 (main, Nov 14 2023, 09:36:21) [GCC 13.2.1 20230801]
- Ranger version/commit: ranger-master (latest git)
- Locale: ru_RU.UTF-8

#### Current Behavior
Ranger show error on SIXEL preview.

#### Expected Behavior
Ranger show image in preview pane.

#### Steps to reproduce
1. Install ArchLinux and update it;
2. Install latest Ranger (git version);
3. Enable SIXEL preview in rc.conf;
4. Try to view the image.

![image_2024-01-14_19-50-28](https://github.com/ranger/ranger/assets/5403268/c7dd994e-ce60-4623-9a7f-13ac3df118c7)
![image_2024-01-14_19-54-00](https://github.com/ranger/ranger/assets/5403268/542bacda-8dbe-43f1-9286-b8a70c0ae0a7)
![image_2024-01-14_20-00-19](https://github.com/ranger/ranger/assets/5403268/d775ff55-9066-405f-99dc-865a41a55914)

toonn:
@freeconst, is that really the command getting run? Should it not be `magick <image_path>[0] -geometry 800x480> -dither FloydSteinberg sixel:-`?
Do you get an error with this version of the command?

@pablo-888, you could run that as well.

pvl1:
on my system, the function `get_font_dimensions()` always returns `0, 0`, and there for runs magick with geometry 0x0

toonn:
@pvl1 that sounds like a separate issue actually. I assume the preview doesn't technically fail in your case, but rather a 0x0 preview shows nothing?

pvl1:
@toonn I agree
I noticed, sixel works fine on alacritty and kitty but not terminator
the geometry on terminator remains 0x0. will dig more on this

toonn:
@acarpenter628, digging into the `get_font_dimensions` function might turn up something. Maybe at least enough to ask on the Konsole bug tracker.

PR #3185, opened 2026-01-30
Fix sixel previews in Windows Terminal and/or over telnet

MadReasonable:
When determining sixel preview panel size, fall back to XTWINOPS if ioctl fails to get terminal dimensions.

#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT
- Operating system and version: Debian GNU/Linux 13.3 (trixie)
- Terminal emulator and version: Windows Terminal v1.23.20211.0
- Python version: 3.13.5
- Locale: en-US

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION
This fix adds a fallback in `get_font_dimensions()` for `get_terminal_size()`, which fails for terminals that do not properly support ioctl TIOCGWINSZ calls, like Windows Terminal.

- In `get_font_dimensions()`
  - If `get_terminal_size()` returns 0px for width or for height
    - Call `get_xtwinops_cell_size()` to query the terminal directly via XTWINOPS control sequence `CSI 16 t`
    - Parse response in the format `ESC [ 6 ; <cell_height_px> ; <cell_width_px> t`

#### MOTIVATION AND CONTEXT
Windows Terminal has had stable sixel support since v1.22 released on Feb. 2025, but sixel previews in ranger render a 1x1 pixel dot.  This fix enables proper sixel preview rendering in Windows Terminal builds that support sixel ( >= v1.22).

No open issues, just this comment:
https://github.com/ranger/ranger/issues/2937#issuecomment-2449877470

#### TESTING
- pytests run in IDE
- make test run from command line
- Terminals
  - Windows Terminal v1.23
  - Konsole v25.04.2
  - tmux v3.5a
  - telnet v2.6

The function `get_font_dimensions()` is also called for iTerm2 previews.  iTerm2 behavior should be unaffected since any terminal that supports TIOCGWINSZ will never execute the fallback code.

#### IMAGES / VIDEOS
*Before*
Screenshot 2026-01-30 130149

*After*
Screenshot 2026-01-30 130450


MadReasonable:
Two points I'd like to discuss:

1. Windows Terminal always returns **10x20** pixels for cell dimensions but it seems to also handle scaling so this solution still works, though images are often slightly distorted because there is no way to know the true aspect ratio of a character cell.  Maybe there are other terminals without *TIOCGWINSZ* support that can take better advantage of this fix.

3. I included a hard-coded fallback of **5x7** pixels that I'm happy to replace with original behavior, **0x0** pixels.  It just seemed better to render something assuming minimal readable font sizes instead of generating a single pixel as happens now.

ranger/ext/img_display.py
```diff
@@ -84,13 +94,66 @@ def get_terminal_size():
...
+
 def get_font_dimensions():
     """
     Get the height and width of a character displayed in the terminal in
     pixels.
     """
     rows, cols, xpixels, ypixels = get_terminal_size()
-    return (xpixels // cols), (ypixels // rows)
+    if xpixels > 0 or ypixels > 0:
+        return (xpixels // cols), (ypixels // rows)
```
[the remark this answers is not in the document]
MadReasonable: I've pushed a new commit to this MR with this change.  This does mean more terminals will fallback to my new code.  I was originally trying to keep impact as small as possible.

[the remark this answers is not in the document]
j4james: > What's the policy for resolving conversations?

I don't see a resolve button, so I think only you can resolve it.

j4james (approved):
Looks good to me. Thanks for fixing this.

ranger/ext/img_display.py
```diff
@@ -84,13 +94,66 @@ def get_terminal_size():
...
     Get the height and width of a character displayed in the terminal in
     pixels.
     """
     rows, cols, xpixels, ypixels = get_terminal_size()
-    return (xpixels // cols), (ypixels // rows)
+    if xpixels > 0 and ypixels > 0:
+        return (xpixels // cols), (ypixels // rows)
+
+    global _terminal_rows, _terminal_cols  # pylint: disable=global-statement,invalid-name
+    global _xtwinops_cell_size             # pylint: disable=global-statement,invalid-name
```
toonn, line 148: I don't love doing this with globals. I think memoizing `get_xtwinops_cell_size` and passing in the rows and cols is a better solution.

ranger/ext/img_display.py
```diff
@@ -2,6 +2,9 @@
 # License: GNU GPL version 3, see the file "AUTHORS" for details.
 # Author: Emanuel Guevel, 2013
 # Author: Delisa Mason, 2015
+# Author: Lloyd Park, 2026
+
+# pylint: disable=invalid-name,too-many-lines
```
toonn, line 7: I prefer local disabling of lints.

toonn (changes requested):
I don't love relying on a timeout but I suppose there's no alternative.

Do want to not rely on globals though.

ranger/ext/img_display.py
MadReasonable: I've removed `disable=invalid-name`.  I don't have a low impact way of removing `too-many-lines`.  I am happy to move XTWINOPS functionality into it's own ext file.  I already have a branch where I've started this work.  Or we could just leave `too-many-lines`

ranger/ext/img_display.py
MadReasonable: I'm updating `get_xtwinops_cell_size(...)` to use an updated version of `cached_function(...)`.  I've added support for a maxsize parameter, and for custom keys using a subset of passed arguments.

ranger/ext/cached_function.py
```diff
@@ -3,17 +3,92 @@
...
+            if maxsize is not None:
+                order.append(key)
+                while len(order) > maxsize:
+                    del cache[order.pop(0)]
             return value
-    inner_cached_function._cache = cache  # pylint: disable=protected-access
-    return inner_cached_function
+
+        # pylint: disable=protected-access
+        wrapper._cache = cache  # type: ignore[reportFunctionMemberAccess]
```
toonn, line 89: What does this achieve? `_cache` doesn't seem to be accessed anywhere.

ranger/ext/img_display.py
```diff
@@ -84,13 +95,58 @@ def get_terminal_size():
     return struct.unpack("HHHH", fretint)
 
 
+@cached_function(maxsize=1, key_args=['rows', 'cols'])
+def get_xtwinops_cell_size(rows, cols, timeout=0.05):
+    # pylint: disable=unused-argument
+    # rows and cols are only used as cache key, so we can ignore them
+    """
+    Query terminal for character cell size using XTWINOPS sequence `CSI 16 t`.
+    Returns (cell_width_px, cell_height_px) or None. Memoized on (rows, cols)
```
toonn, line 104: Should we consider changes in font size? That'd have greater `cell_*_px` for the same `(rows, cols)`, no?

toonn (commented):
Just some questions on my part.

ranger/ext/cached_function.py
MadReasonable: This was in the previous version in master and I didn't want to change the surface area of this function any more than I had to.  It doesn't look like anything is/was referencing this so I'm happy to remove it if you'd like.
https://github.com/ranger/ranger/blob/682a45d5a971cedc581da7896cccb8144c67d6a0/ranger/ext/cached_function.py#L18

ranger/ext/img_display.py
MadReasonable: I don't believe there is a reliable way to query for font or font-size in a reliable way across terminals, or even get them at all in most terminals.  I don't think these edge cases are realistically solvable.

When I was testing this I could find no situation where a change in font-size didn't also change the number of rows or columns.  Changes in the font-family itself could cause the image to get stretched. which is admittedly not great.  My thinking is at least you can see something at all when using Windows terminal or Telnet, which wasn't possible at before.

ranger/ext/cached_function.py
toonn: Yeah, let's remove it : )

cgranier:
Independent confirmation of the bug this fixes, plus measurements that may help with the two open questions above.

I hit this today on a fresh master install and traced it to the same line before finding this PR.

### Environment

  - ranger master @ 3187384e4078d91848bd22ea0dc8eb5fe2866f59 (pip install from git)
  - Ubuntu 24.04.4 LTS, Python 3.12.3, ImageMagick 6.9.12
  - Windows Terminal Preview 1.25.1912.0, over Windows OpenSSH
  - preview_images_method sixel, imagemagick installed, scope.sh in place

### The failure

get_font_dimensions() reads TIOCGWINSZ and on this transport the pixel fields are zero, because ConPTY sizes consoles in character cells only — the Windows SSH client has no pixel values to put in the pty-req / window-change messages:

TIOCGWINSZ on the SSH tty: rows=61 cols=270 xpixel=0 ypixel=0
    -> get_font_dimensions() = (0, 0)

That propagates to -geometry 0x0>, and the important part is that ImageMagick does not fail:

`$ convert "shot.png[0]" -geometry "0x0>" -dither FloydSteinberg sixel:- | wc -c  35`

A valid 35-byte sixel with a 1×1 raster (\x1bP0;0;0q"1;1;1;1...). So there is no CalledProcessError, no ImageDisplayError, nothing in the log — just an invisible dot. That silence is what makes it expensive to diagnose.

Same setup, inside tmux, works

tmux does exactly what this PR proposes — queries the terminal with CSI 16t and populates the pane pty's pixel fields — so ranger's unmodified ioctl read succeeds there:

|       context    | TIOCGWINSZ pixels  | get_font_dimensions()  |  ImageMagick output  |
| --- | --- | --- | --- |
| direct SSH (ConPTY) | 0 × 0  | (0, 0)   | 35 bytes, 1×1 raster |
| inside tmux 3.7b  | 2700 × 1200   | (10, 20)  | 823 KiB, correct |


In other words the CSI 16t value this PR would fetch is already known-good on this terminal: tmux derives 10x20 from it and previews render correctly at 1000×558 px. The only difference is who asks.

Re: point 1, aspect ratio

Windows Terminal answers CSI 16t with ESC[6;20;10t here, i.e. 10×20 px. At default font settings that appears to be the real cell size rather than a placeholder: 270 cols × 10 px = 2700 and 60 rows × 20 px = 1200 match the pane geometry tmux reports, and images came out correctly proportioned. I have not tested non-default font sizes or display scaling.

Re: point 3, the 5×7 fallback

The current 0x0 behaviour is worse than any non-zero fallback specifically because it fails silently — ImageMagick returns success and a well-formed 1×1 image, so nothing surfaces to the user. A fallback that renders something visibly wrong is strictly easier to debug than one that renders something invisible. If a hard-coded value is contentious, even keeping 0x0 but raising ImageDisplayError when the computed font size is zero would remove the silent-failure mode.

Note this also affects preview_images_method iterm2, which shares get_font_dimensions().

I have not applied or tested the patch itself — this is confirmation of the underlying bug and of the CSI 16t value on this terminal, not a review.

PR #3196, opened 2026-02-25
shutil_generatorized: Return to raise StopIteration

toonn:
When refactoring `copyfile` in 4a5c9fd1f2cac3a40bb75b9e7140025f359c42fe, I forgot to return after `copyfileobj_range` finishes successfully. This means `copyfileobj` is called right after and copies the entire file **again**, this time without potentially making a COW copy!

@qrvd, I noticed this when reviewing #3151, so double thanks : ) A review'd be welcome.

### DIFF
--- a/ranger/ext/shutil_generatorized.py
+++ b/ranger/ext/shutil_generatorized.py
@@ -154,6 +154,7 @@ def copyfile(src, dst):
             try:
                 for done in copyfileobj_range(fsrc, fdst):
                     yield done
+                return
             except OSError:
                 # Return to start of files first, then use old method
                 fsrc.seek(0, 0)


vejkse:
> This means `copyfileobj` is called right after and copies the entire file **again**, this time without potentially making a COW copy!

As far as I understand, since the `seek(0, 0)` commands are skipped if everything went fine with `copyfileobj_range`, we are already at the end of the files when trying to run `copyfileobj` and nothing happens.

I rechecked whether it really does COW by checking the size on disk on btrfs filesystems of the original and the copy together and the total size was just the size of one file, as expected.

But adding `return` would still skip a bit of unnecessary work.

hut (approved):
Looks good to me.

PR #3205, opened 2026-03-24
Enhance rifle.conf

followtherabbit99:
Added support for Markdown files and updated archive handling and disables automatic script execution

#### ISSUE TYPE

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

adds markdown and archive handling with ouch, disabled the automatic script execution.

#### MOTIVATION AND CONTEXT

general improvements and safety
While the script rules are overruled by the editor options, having these lines seems dangerous since a change in the editor section would lead to automatic execution of scripts. This way, anyone who actually wants automatic script execution can still uncomment these lines (changing rifle.conf was already required to use these lines)

improved file opener support

#### TESTING

manual tests

they dont

### DIFF
--- a/ranger/config/rifle.conf
+++ b/ranger/config/rifle.conf
@@ -82,7 +82,7 @@ ext x?html?, has lynx,              terminal = lynx -- "$@"
 ext x?html?, has w3m,               terminal = w3m "$@"
 
 # Markdown
-ext md, has glow, terminal = glow -t -- "$@"
+ext md, has glow, flag t = glow -t -- "$@"
 
 #-------------------------------------------
 # Misc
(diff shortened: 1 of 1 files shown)


hut:
Thanks for the patch. Could you please fix a few things:

1. you remove mpv and then add it again in 2 commits. Could you squash the commits so that the mpv line remains untouched?
2. can you add a "has glow" condition to the glow line?
3. swayimg is duplicated, can you remove one of the lines?
4. the comment "dangerous!" is a bit too informal. Can you change it to something like: "Potentially harmful. Enable these as needed."

@toonn : would you agree that the automated script execution lines should be disabled by default? (like `ext py  = python -- "$1"`) It's been like this since forever, but on second thought, it does seem too careless to me as well.

toonn:
I disagree about it being dangerous and it never having been a problem backs me up on that. Commenting out those rules means that users can't select the option from the ask menu that pops up with `r`.

BTW, @hut, might be worth checking out [this page on how to request changes](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request), that way each change requested can be marked as addressed and discussed in its own thread.

PR #3209, opened 2026-03-28
Fix: loading of cached directories

### LINKED ISSUE #3081: Folder shows up as empty in ranger after moving, even though it isn't.
I was transferring some data from my internal laptop SSD to an external SSD. When going over the files, I noticed that ranger showed one of the folders was empty. I copied the "empty" folder back to the laptop, and it wasn't empty anymore. 
I copied it back to the external SSD with another name, it wasn't empty. I renamed it to the original name, and it was empty again. 
I deleted it and made a new folder with `mkdir`, and it was still empty (even after adding some files).
I created a file with `touch`, and that worked without a problem; I then deleted that file and made a new folder with `mkdir`, and it was again stuck on empty.

I checked with `ls` and with PCManFM that the folder was not really empty, and ranger just showed it that way.

I tried to close ranger while the "empty" folder was selected and it told me I couldn't, since a task was being performed. I selected a different folder and it had no problem quitting. After restarting, the problem was no longer present and the folder was displayed normally even after deleting it and copying it over again.

If you need any more info or if you have ideas how to recreate it and need me for that, please tell me. I really would like to include steps to recreate this, but I haven't managed to. I still wanted to inform you that this happened in case you have an idea why it did.

yinghao-w:
Inaccessible directories when deleted and recreated under looser permissions were still set to not runnable, correctly reload them as runnable. Fixes #3081, specifically https://github.com/ranger/ranger/issues/3081#issuecomment-2797950360.

#### ISSUE TYPE
- Bug fix

#### CHECKLIST
- [X] All changes follow the code style
- [X] All new and existing tests pass

### DIFF
--- a/ranger/container/fsobject.py
+++ b/ranger/container/fsobject.py
@@ -315,6 +315,9 @@ class FileSystemObject(  # pylint: disable=too-many-instance-attributes,too-many
         # Set some attributes
 
         self.accessible = bool(new_stat)
+        if self.is_directory:
+            self.runnable = self.accessible
+
         mode = new_stat.st_mode if new_stat else 0
 
         fmt = mode & 0o170000


hut (approved):
Looks good to me.

`self.runnable` is, for directories, normally defined through a `lazy_property` in `directory.py`, but that's not flexible enough to cover reloads, so I'd agree that it has to go in `fsobject.load()`.

toonn:
Does #3219 fix this issue actually? It prevents the Directory object from being reused after deletion.

yinghao-w:
> Does #3219 fix this issue actually? It prevents the Directory object from being reused after deletion.

Checked out the PR, sadly not. This issue is when the directory is deleted from outside of ranger, so the code in the PR doesn't actually run.

toonn:
Do you have steps to reproduce the issue then? When I `rmdir` a directory with `x` permissions and `mkdir` the same directory, Ranger reloads the directory just fine and I can then access it.

yinghao-w:
> Do you have steps to reproduce the issue then? When I `rmdir` a directory with `x` permissions and `mkdir` the same directory, Ranger reloads the directory just fine and I can then access it.

I just did exactly as in https://github.com/ranger/ranger/issues/3081#issuecomment-2797950360, and got precisely the same result. Could you describe the exact steps you took?

toonn (commented):
It feels wrong to have directory-specific code in fsobject. Also, maybe we should call `runnable__reset()` instead? (Probably for all the `lazy_property`s.)

PR #3214, opened 2026-04-16
Enhance security of rifle file parsing

followtherabbit99:
#### ISSUE TYPE

- Bug fix

#### RUNTIME ENVIRONMENT
any

#### CHECKLIST

    - [ ] Tests have been updated

#### DESCRIPTION

Fixes get_executables to require a real executable bit instead of accidentally matching any regular file, and replaces the broken sudo -E su root -mc prefix with a plain sudo /bin/sh -c for the r (run-as-root) flag.

Improves the terminal-fallback path parsing by remembering the raw action string instead of trying to re-parse the already-built shell command with fragile split(';')/split('--') calls, which mangled filenames containing those characters.

#### MOTIVATION AND CONTEXT

rifle is used as a file opener and thereby runs on potentially untrusted data. This patch ensures that filenames cannot trick the file opener 

#### TESTING

Manual tests

it doesnt

### DIFF
--- a/ranger/ext/get_executables.py
+++ b/ranger/ext/get_executables.py
@@ -6,7 +6,7 @@ from __future__ import (absolute_import, division, print_function)
 from os import listdir, environ, stat
 import platform
 import shlex
-from stat import S_IXOTH, S_IFREG
+from stat import S_IXOTH, S_IXUSR, S_IXGRP, S_IFREG
 
 from ranger.ext.iter_tools import unique
 
@@ -56,7 +56,8 @@ def get_executables_uncached(*paths):
                 filestat = stat(abspath)
             except OSError:
                 continue
-            if filestat.st_mode & (S_IXOTH | S_IFREG):
+            if filestat.st_mode & S_IFREG and \
+                    filestat.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH):
                 executables.add(item)
     return executables
 
--- a/ranger/ext/rifle.py
+++ b/ranger/ext/rifle.py
@@ -50,7 +50,7 @@ except ImportError:
         else:
             paths = ['/usr/bin', '/bin']
 
-        from stat import S_IXOTH, S_IFREG
+        from stat import S_IXOTH, S_IXUSR, S_IXGRP, S_IFREG
         paths_seen = set()
         cached_executables = set()
         for path in paths:
@@ -67,7 +67,8 @@ except ImportError:
                     filestat = os.stat(abspath)
                 except OSError:
                     continue
-                if filestat.st_mode & (S_IXOTH | S_IFREG):
+                if filestat.st_mode & S_IFREG and \
+                        filestat.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH):
                     cached_executables.add(item)
         return cached_executables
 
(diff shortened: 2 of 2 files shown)


ranger/ext/get_executables.py
```diff
@@ -56,7 +56,8 @@ def get_executables_uncached(*paths):
                 filestat = stat(abspath)
             except OSError:
                 continue
-            if filestat.st_mode & (S_IXOTH | S_IFREG):
+            if filestat.st_mode & S_IFREG and \
+                    filestat.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH):
```
toonn, line 60: I see what you're trying to do but ORing the executable bits and then ANDing with `st_mode` means you're checking for files being executable by *all* of user, group and other.

Checking user and group executability is also not enough, you'd need to check the current user's is said user or in the group.

OTOH, shells don't care about executability and will just emit an error if you try to execute something that's not executable.

My shell's actually pretty strict wrt to executable permission, if I'm the owner, it has to have u+x, if I'm not the own but in the group it has to have g+x, neither of the above then o+x is enough.

ranger/ext/rifle.py
```diff
@@ -455,7 +459,7 @@ def execute(
             self.hook_before_executing(command, self._mimetype, self._app_flags)
             try:
                 if 'r' in flags:
-                    prefix = ['sudo', '-E', 'su', 'root', '-mc']
+                    prefix = ['sudo', '/bin/sh', '-c']
```
toonn, line 462: This doesn't preserve the user's environment depending on their sudo configuration and it also doesn't behave like an interactive shell. I'm not sure what problem it's trying to address?

ranger/ext/rifle.py
```diff
@@ -518,8 +522,9 @@ def execute(
                     if term in ['tilda', 'pantheon-terminal', 'terminology',
                                 'termite']:
 
-                        target = command.split(';')[0].split('--')[1].strip()
-                        app = command.split(';')[1].split('--')[0].strip()
+                        target = "' '".join(f.replace("'", "'\\''")
+                                            for f in files if "\x00" not in f)
```
toonn, line 526: We should probably use shlex instead to get proper shell escaping?

ranger/ext/rifle.py
```diff
@@ -435,6 +438,7 @@ def execute(
         else:
             if label and label in get_executables():
                 cmd = '%s "$@"' % label
+                raw_action = cmd
```
toonn, line 441: I don't see the point of `raw_action`, `cmd` is just a command to be executed files are not passed as part of that but rather through the `files` argument.

ranger/ext/rifle.py
```diff
@@ -490,7 +494,7 @@ def execute(
                                          "change fallbacks in rifle.conf.")
                         self._mimetype = 'ranger/x-terminal-emulator'
                         self.execute(
-                            files=[command.split(';')[1].split('--')[0].strip()]
```
toonn, line 493: I agree that reparsing this is bad. Splitting on the last `;` wouldn't be as bad but better if we don't need to do it at all.

I think this is trying to get the command out of `cmd` though, no? `split(';')[1]` is (intended to be) the part after the `;` of `set -- %s; %s` and `split('--')[0]` on that is trying to get whatever goes before `--` of the second `%s` which is `cmd`?

toonn (commented):
There's potential here but give me some proper test scenarios that demonstrate how it's an improvement.

PR #3215, opened 2026-04-16
Fix hashing for duplicate, unique and hash filter stack filters

toonn:
Somewhere along the line the hash we use started requiring strings be encoded explicitly.

Fixing that issue revealed some more bugs to me, which I try to fix here.


hut (approved):

hut:
@toonn Looks good to me. However, I never use this feature and I didn't notice any bugs either, so I leave it to your discretion to merge this.

hut:
damn, we have a merge conflict here. for reference: it came from merging https://github.com/ranger/ranger/pull/2130

PR #3217, opened 2026-04-21
actions: prevent spawning of multiple preview tasks per file

### LINKED ISSUE #1320: svg re-renders preview extra times

### LINKED ISSUE #2113: ranger becomes a fork bomb when previewing svg
#### Runtime Environment

- Operating system and version:  Arch Linux current
- Terminal emulator and version: gnome-terminal 3.36.2-1 + tmux 3.1_b-1
- Python version: 3.8.5-2
- Ranger version/commit: 240d8a0ea1b82690372ba067c24a3f581a29c782
- Locale: zh_CN.UTF-8
- imagemagick-fftw 7.0.10.31-1 and inkscape 1.0.1-1

#### Current Behavior

Lots of child `identify` and `inkscape` processes

#### Expected Behavior

Not too many to bring the system down.

#### Steps to reproduce
1. in `~/.config/ranger/rc.conf` put `set preview_images true`
1. protect yourself with e.g. systemd + cgroups. If you have systemd + cgroups v2, you can run `systemd-run --user --scope -p CPUQuota=10% -p TasksMax=100 -p MemoryMax=1G -u ranger ./ranger.py`
2. navigate to a (large) svg, e.g. one generated with `systemd-analyse plot`
3. see quite a few child processes being created (in htop or `systemctl --user status ranger.scope`). If not too many, try `Ctrl-C` one or more times. Memory is using up, CPU time is using up.
4. to cleanup, run `systemctl --user kill ranger.scope` and `systemctl --user kill -s KILL ranger.scope` several times.

eMPee584:
.. by not clearing preview state too zealously.

Fixes #1320 (?)
Fixes #2113 (?)

When accidentally enabling previews for virtual files like
`/var/lib/lxcfs/proc/diskstats`, I stumbled over this. The ever-changing
timestamp had ranger create vast numbers of preview tasks.. After some tracing
with pudb, I found this code for removing stale previews was also removing the
loading state marker, causing preview tasks to be rererecreated. Chaning to not
clear the dict entry when a preview tasks is still pending solves that.
Previews will still update on next refresh.

I could not quite reproduce the issues with SVG files again even without the
patch, but most likely this fixes those as well..

### DIFF
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1035,7 +1035,8 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
     # --------------------------
     def update_preview(self, path):
         try:
-            del self.previews[path]
+            if not self.previews[path]['loading']:
+                del self.previews[path]
             self.signal_emit('preview.cleared', path=path)
         except KeyError:
             return False
--- a/ranger/core/actions.py
+++ b/ranger/core/actions.py
@@ -1037,11 +1037,10 @@ class Actions(  # pylint: disable=too-many-instance-attributes,too-many-public-m
         try:
             if not self.previews[path]['loading']:
                 del self.previews[path]
-            self.signal_emit('preview.cleared', path=path)
+                self.signal_emit('preview.cleared', path=path)
+                self.ui.need_redraw = True
         except KeyError:
-            return False
-        self.ui.need_redraw = True
-        return True
+            pass
 
     @staticmethod
     def sha512_encode(path, inode=None):


ranger/core/actions.py
```diff
@@ -1035,7 +1035,8 @@ def scroll_preview(self, lines, narg=None):
     # --------------------------
     def update_preview(self, path):
         try:
-            del self.previews[path]
+            if not self.previews[path]['loading']:
+                del self.previews[path]
             self.signal_emit('preview.cleared', path=path)
```
toonn: This should be guarded too, no? We shouldn't emit the signal unless we do the `del`.

toonn (commented):
I think I follow your reasoning. The main problem is that we're not killing the preview process along with the clearing of `previews` so we end up respawning it over and over because we lose the bookkeeping information that the processes still exist. Maybe the correct solution would rather be to kill the process but this is reasonable as well, previews that don't really exist yet need not be cleared.

ranger/core/actions.py
eMPee584: Yeah makes sense.. same would apply to setting the redraw marker, right? No change no redraw? Oh and the return value is never used in the lone single call of this function..
Sometimes looking at these old blame logs, I wonder whether whatever @hut was smoking was still ok 😂

eMPee584: Oh actually 1687e0f41f172cef6923f75f4c6f1038f19671dc was @nfnty :sweat_smile: 
Make it make sense to me :smirk:

toonn (commented):
Why are we dropping the return altogether? Did you check that nothing relies on that API?

hut:
https://en.wiktionary.org/wiki/Hyrum%27s_law

"given a sufficiently popular system, any observable behavior — even if it is not guaranteed by contract — will be depended upon by some user."

PR #3219, opened 2026-04-25
actions: Drop Directory object

### LINKED ISSUE #1613: Content of directory does not update properly when renaming directories to names of deleted directories
#### Runtime Environment

- Operating system and version: 5.1.15-arch1-1-ARCH
- Terminal emulator and version: urxvt 9.22-9
- Python version: 3.7.3-2
- Ranger version/commit: 1.9.2-1
- Locale: LC_COLLATE=C

#### Current Behavior

When a directory gets renamed with a name matching the name of a recently deleted directory (recently = within the session), instead of keeping its content and simply be renamed, it *becomes* the old directory (so the content turns into the content of the old and deleted directory which had that name).
 
This is very dangerous because files can easily be lost (if the user does not realize what happened and ends the session before reverting the name).

#### Expected Behavior

Being able to rename a directory, even with a name formerly used, without the content being affected.

#### Steps to reproduce

1. Create 2 directories somewhere: test and test_
2. Add a file in test_
3. Delete test
4. Rename test_ into test
The file has disappeared because test_ renamed into test is being confused with the *old* test directory. Rename test into test_ and the file re-appears. This happens until ranger is closed and re-opened.

toonn:
Additionally erase its signal handlers. Since those were registered as weak references, this should also clear the associated bound signals.

Fixes #1613

### DIFF
--- a/ranger/container/directory.py
+++ b/ranger/container/directory.py
@@ -203,6 +203,9 @@ class Directory(  # pylint: disable=too-many-instance-attributes,too-many-public
             return Vcs(self)
         return None
 
+    def clear_signal_handlers(self):
+        self._signal_functions = None
+
     def signal_function_factory(self, function):
         def signal_function():
             self.load_if_outdated()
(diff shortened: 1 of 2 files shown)


toonn:
@qrvd or @vejkse, a review would be appreciated :heart:

hut (approved):

hut:
I'm a bit confused why the signal cleaning is necessary, since it is weak references and if the directory is deleted, they should be deleted too. But maybe the deletion is delayed or there's some residual references.

Either way, this works and lgtm. Good job <3

PR #3221, opened 2026-04-30
fix: ValueError introduced by b4618e0

### LINKED ISSUE #535: Filenames containing '²' cause range to show empty folders
As said in the title. Any directory which contains another directory or file with "²" or any other superscript character (³, etc) in their filename causes Ranger to not display anything inside it, and to return the error: 
`Invalid literal for int() with base 10: '²'`

To fix it I have to open a terminal and rename the file with mv.

### LINKED ISSUE #3220: Regression: ValueError introduced by b4618e0
The commit b4618e0 removes the logic used to resolve the issue https://github.com/ranger/ranger/issues/535

aquanjsw:
b4618e0 removes the logic used to resolve #535

Fixes: #3220

- add test filenames


vejkse:
It may not be the best way to fix this, since according to this comment by @mingmingrr: https://github.com/ranger/ranger/pull/2943#discussion_r2463678998, replacing `try`/`else` by `if`/`then` improved the performance of `basename_natural` the most.

The original fix for https://github.com/ranger/ranger/issues/535 did not actually use a `try`/`else` but an `if`/`then`: https://github.com/ranger/ranger/commit/899e615b63a1a68fbc9bcc389eef18432c74ca9e.  This was changed later in https://github.com/ranger/ranger/commit/009973f7f4b650aa20109e67edae673faf4f25f4.

Going back to it may fix it while keeping the performance gain?  There is a profiling script in @mingmingrr comment.

vejkse:
I tested a list of about 240,000 file names with an adaptation of @mingmingrr script:
[basename_natural_profile_exponents_fix.py](https://github.com/user-attachments/files/27755374/basename_natural_profile_exponents_fix.py)
and I got somewhat surprising results:
```
bn_fixed_with_try_except_findall: 2.106677005000165
bn_fixed_with_try_except_split: 1.0328283599992574
bn_fixed_with_list_findall: 0.8192235639999126
bn_fixed_with_list_split: 0.82025006099866
```
`findall` was supposed to be faster than `split`, but with the other optimizations in place it is not.

And if I run it on full paths of the same files (which can happen in ranger with `:flat`), I get:
```
bn_fixed_with_try_except_findall: 4.18750348300091
bn_fixed_with_try_except_split: 1.6021870210006455
bn_fixed_with_list_findall: 1.6844126010000764
bn_fixed_with_list_split: 1.6882480260010198
```
In this case, because it uses `split`, your original fix with `try`/`except` is a bit faster than the `if`/`else` version with a list of digits.

aquanjsw:
So maybe we can just keep the list+findall method for now?

mingmingrr:
Maybe add a `"hello²"`/`"hello².txt"` to `‎tests/ranger/container/test_fsobject.py`?

aquanjsw:
Test filenames added.

mingmingrr:
Using re.split with the current regex produces lists of empty strings. You'll have to wrap the entire regex in a group to preserve the matches. Here are the times with fixed re.split:
 
```
bn_fixed_with_try_except_findall: 0.5587730840197764
  bn_fixed_with_try_except_split: 1.0031945340160746
      bn_fixed_with_list_findall: 0.20884695299901068
        bn_fixed_with_list_split: 0.22023708000779152
```

mingmingrr:
To be clear, re.findall works fine (and in fact performs better) without being wrapped in a capture group. I was trying to explain the performance difference between try/except + split and findall was actually the result of a bug.

aquanjsw:
> To be clear, re.findall works fine (and in fact performs better) without being wrapped in a capture group. I was trying to explain the performance difference between try/except + split and findall was actually the result of a bug.

emm... My fault, not reading your comment carefully, reverted.

hut (approved):

hut:
Thank you for catching this and fixing this!

PR #3223, opened 2026-05-06
Add size filter to filter_stack

eMPee584:
(crafted in vim assisted by 🤖 Codeium plugin)

### DIFF
--- a/ranger/core/filter_stack.py
+++ b/ranger/core/filter_stack.py
@@ -91,6 +91,43 @@ class MimeFilter(BaseFilter, FileManagerAware):
         return "<Filter: mimetype =~ /{pat}/>".format(pat=self.regex.pattern)
 
 
+@stack_filter("size")
+class SizeFilter(BaseFilter, FileManagerAware):
+    # size argument is a string of the form [+-<>=]?[0-9]+[KMG]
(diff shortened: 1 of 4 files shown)


ranger/core/filter_stack.py
```diff
@@ -91,6 +91,43 @@ def __str__(self):
...
+            self.operator = "="
+        size = size.replace(" ", "")
+        if size[-1].lower() == "k":
+            self.size = int(size[:-1]) * 2 ** 10
+        elif size[-1].lower() == "m":
+            self.size = int(size[:-1]) * 2 ** 20
+        elif size[-1].lower() == "g":
+            self.size = int(size[:-1]) * 2 ** 30
+        else:
+            self.size = int(size)
```
toonn, line 117: This parsing feels extremely fragile. IMO, drop the `+/-`, not intuitive as a prefix. Include allowing `<=` and `>=`. I also feel like we need to do better than hardcoding `k/m/g`, support all the SI prefixes, including "Ki/Mi/etc."

I know that you could do ranges with multiple filters but I feel like it might be such a common use case that we should either support a syntax for it `5K-10G` or allow multiple arguments and add a filter to the stack for each of them.

PR #3224, opened 2026-05-06
Add shell command filter to filter_stack

eMPee584:
Works nicely with grep, comments welcome.

I was about to take out the timeout again but wanted to get someone else's
perspective on it. A `-t` parameter would be nice, but replicating
`ranger.api.commands.parse_flags()` seems over the top for this, so I tend
towards removing the timeout again. Similar thing about the caching, it does
work but cached entries are bound to the specific filter instance and never
cleared, so not sure about that either.

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -742,6 +742,9 @@ Apply a new mimetype filter.
 .IP .n 14
 .IX Item ".n"
 Apply a new filename filter.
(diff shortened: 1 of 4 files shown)


toonn (commented):
Another good idea for a filter! Would be great if we could get a couple people testing this.

PR #3144, opened 2025-10-16
shutil_generatorized: always move() symbolic links as files (#2881)

qrvd:
#### ISSUE TYPE

x Bug fix

#### RUNTIME ENVIRONMENT

- Operating system and version: 
- Terminal emulator and version: Tilix 1.9.5
- Python version: 3.11.2
- Ranger version/commit: master 6d5e0d8dc4cc5ddf53211e19f48c5b5e9ee47b18
- Locale: en_US.UTF-8

#### CHECKLIST

- [x] The `CONTRIBUTING` document has been read **[REQUIRED]**
- [x] All changes follow the code style **[REQUIRED]**
- [ ] All new and existing tests pass **[REQUIRED]**
- [ ] Changes require config files to be updated
    - [ ] Config files have been updated
- [ ] Changes require documentation to be updated
    - [ ] Documentation has been updated
- [ ] Changes require tests to be updated
    - [ ] Tests have been updated

#### DESCRIPTION

Please read my comment on issue https://github.com/ranger/ranger/issues/2881#issuecomment-3412954268

#### MOTIVATION AND CONTEXT

https://github.com/ranger/ranger/issues/2881

#### TESTING

Tested moving a symbolic link pointing to a directory, between filesystems as described by @vejkse in #2881. After this commit the issue no longer occurs.

I grepped over the codebase and it looks like `shutil_generatorize.move()` is only used by `CopyLoader` in `ranger/core/loader.py`. `CopyLoader` is only used by the `:paste` command (`Actions.paste`). So this change should not affect
anything else.

PR #3226, opened 2026-05-11
shutil_generatorized: skip _samefile check for symlinks in move() (#3145)

### LINKED ISSUE #3145: Can't move a symlink to a folder into the folder it points to
#### Runtime Environment
* Terminal emulator and version: Tilix 1.9.5
* Python version: 3.11.2
* Ranger version/commit: master 6d5e0d8dc4cc5ddf53211e19f48c5b5e9ee47b18
* Locale: en_US.UTF-8

#### Current Behavior
Moving a symlink into the folder that is the destination of the symlink causes an error.

#### Expected Behavior
The symlink should be moved into the folder.

#### Steps to reproduce

1. Create a directory `A`.
2. Create a symlink `B` pointing to `A`, for example by running `ln -s A B`
3. In ranger, navigate to symlink `B` and `:cut` it (by pressing `dd`)
5. Navigate into directory `A`
6. Execute `:paste` (`pp`).

This shows the error message:
```
Loader work process failed: moving: /.../B (0) (Percent: 0)
```

Symbolic link `B` is still where it was before we attempted the move.

If we instead try to copy `B` into `A`, by executing `:copy` (`yy`) instead of `:cut` in step (3) above, then Ranger performs as expected, and I see a copy of `B` created in dierctory `A`.

kite-sense:
#### ISSUE TYPE
Bug fix

#### RUNTIME ENVIRONMENT
- OS: Linux 6.8.0
- Python: 3.12
- Ranger: master (51e19b8c)

#### CHECKLIST
- [x] The CONTRIBUTING document has been read
- [x] All changes follow the code style
- [x] All new and existing tests pass
- [x] Tests have been updated

#### DESCRIPTION
Add `and not os.path.islink(src)` guard to the `_samefile` check in `shutil_generatorized.move()`, preventing `IsADirectoryError` when moving a symlink into the directory it points to.

#### MOTIVATION AND CONTEXT
Fixes #3145

The `_samefile()` function uses `os.stat()` which follows symlinks, so `_samefile(symlink_B, dir_A)` returns True when B → A. The subsequent `os.rename(symlink, dir)` fails with `IsADirectoryError`.

This matches the upstream CPython fix: python/cpython#21759.

Maintainer toonn approved the fix: "I agree with all of this, including the proposed fix."

Complements PR #3144 which fixed the `except OSError` fallback path in the same function.

#### TESTING
- Added 8 test cases in `tests/ranger/ext/test_shutil_generatorized.py`:
  - Exact bug reproduction (symlink into target directory)
  - Symlink into symlink-to-directory
  - Regression: regular file, directory, file-to-file moves
  - Edge cases: broken symlinks, symlink-to-file, unrelated directory
- All 23 tests pass (full suite)
- `flake8` passes on changed files

### DIFF
--- a/ranger/ext/shutil_generatorized.py
+++ b/ranger/ext/shutil_generatorized.py
@@ -299,7 +299,7 @@ def move(src, dst, overwrite=False, make_safe_path=get_safe_path):
     """
     real_dst = dst
     if os.path.isdir(dst):
-        if _samefile(src, dst):
+        if _samefile(src, dst) and not os.path.islink(src):
             # We might be on a case insensitive filesystem,
             # perform the rename anyway.
             os.rename(src, dst)
--- a/tests/ranger/ext/test_shutil_generatorized.py
+++ b/tests/ranger/ext/test_shutil_generatorized.py
@@ -1,0 +1,146 @@
+from __future__ import (absolute_import, division, print_function)
+
+import os
+
+from ranger.ext.shutil_generatorized import move
+
+
+def consume(generator):
+    for _ in generator:
+        pass
+
+
+# === Core regression: exact scenario from issue #3145 ===
+
+
+def test_move_symlink_into_target_directory(tmpdir):
+    """Regression test for #3145.
+
+    Setup:  directory A, symlink B -> A
+    Action: move B into A
+    Before fix: IsADirectoryError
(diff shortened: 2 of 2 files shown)


toonn (approved):
LGTM

Thanks for getting this across the finish line : )

PR #3227, opened 2026-05-11
fix(rename): cursor not moving after rename under symlink

### LINKED ISSUE #3175: When renaming under a symlink, the cursor doesn’t move
### Ranger version

Latest commit (08913377).  **This issue was already present in 1.9.3**.

### Steps to reproduce

1. Run `ranger --clean` and go to a temporary directory.
2. Create in it a subdirectory `a` and a symlink `b` pointing to `a`.
3. `:cd a`, `:touch 1`, `:touch 2`.
4. On `1`, run `:rename 3`.  `1` is renamed to `3` and the cursor moves to `3`, as expected.
5. `:cd ../b`.
6. On `2`, run `:rename 4`.  `2` is renamed to `4` but the cursor is now on `3`, not `4`.

### Workaround

Adding the following file in one’s `ranger/plugin` directory fixes this issue:
```python
import os

class rename(Command):
    """:rename <newname>

    Changes the name of the currently highlighted file to <newname>
    """

    def execute(self):
        from ranger.container.file import File
        from os import access

        new_name = self.rest(1)

        if not new_name:
            return self.fm.notify('Syntax: rename <newname>', bad=True)

        if new_name == self.fm.thisfile.relative_path:
            return None

        if access(new_name, os.F_OK):
            return self.fm.notify("Can't rename: file already exists!", bad=True)

        if self.fm.rename(self.fm.thisfile, new_name):
            file_new = File(os.path.abspath(os.path.join(self.fm.thisdir.path, new_name)))
            self.fm.bookmarks.update_path(self.fm.thisfile.path, file_new)
            self.fm.tags.update_path(self.fm.thisfile.path, file_new.path)
            self.fm.thisdir.pointed_obj = file_new
            self.fm.thisfile = file_new

        return None

    def tab(self, tabnum):
        return self._tab_directory_content()
```
The only difference with the default `rename(Command)` is the line `file_new = …` which is originally just `file_new = File(new_name)`.  This was inspired by the case of `touch`/`mkdir`: https://github.com/ranger/ranger/pull/2038.  I’m a bit worried that this might have undesired side effects.

kite-sense:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
- Operating system and version: Linux 6.8
- Python version: 3.12
- Ranger version/commit: master (08913377+)

#### CHECKLIST
    - [x] Tests have been updated

#### DESCRIPTION

In the `rename` command (`ranger/config/commands.py`), `File(new_name)` resolves the path via `os.getcwd()`, which returns the **physical** (symlink-resolved) path. However, `Directory`'s internal file list is built using `self.path`, the **logical** path that preserves symlinks. When these two paths diverge, `Directory.sync_index` cannot match the `pointed_obj` by path, and the cursor falls back to a default position instead of following the renamed file.

**Fix:** Construct the absolute destination path using `self.fm.thisdir.path`, consistent with how `touch`/`mkdir` already handle this (PR #2038):

```python
dest = abspath(join(self.fm.thisdir.path, expanduser(new_name)))
# ...
file_new = File(dest, path_is_abs=True)
```

Changes:
- `ranger/config/commands.py` — 4 lines changed in `rename.execute()`
- `tests/ranger/config/test_commands_rename.py` — 9 new test cases

#### MOTIVATION AND CONTEXT

Closes #3175

When navigating into a directory via a symlink (e.g., `cd b` where `b -> a`), renaming a file succeeds but the cursor does not move to the renamed file. This was already present in ranger 1.9.3.

#### TESTING

All tests pass: `make test_py` — pylint 10.00/10, flake8 clean, doctests pass, 24/24 pytest green.

New test cases in `tests/ranger/config/test_commands_rename.py`:

| Test | Coverage |
|---|---|
| `test_rename_normal_directory` | Baseline: rename in regular dir |
| `test_rename_under_symlink` | **Core bug**: CWD is physical path, File path must be logical |
| `test_rename_same_name` | No-op when renaming to same name |
| `test_rename_existing_name` | Error when target exists |
| `test_rename_no_argument` | Error on missing argument |
| `test_rename_symlink_path_matches_directory_list` | Path scheme matches `Directory.load_bit_by_bit` |
| `test_rename_into_subdir_under_symlink` | Rename into subdirectory under symlink |
| `test_rename_absolute_destination` | Absolute path destination |
| `test_issue_3175_exact_reproduction` | Exact issue reproduction steps |

The fix only affects path construction in the `rename` command. No other commands, config files, or public APIs are touched.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1120,6 +1120,7 @@ class rename(Command):
 
     def execute(self):
         from ranger.container.file import File
+        from os.path import join, expanduser, abspath
         from os import access
 
         new_name = self.rest(1)
@@ -1130,11 +1131,13 @@ class rename(Command):
         if new_name == self.fm.thisfile.relative_path:
             return None
 
-        if access(new_name, os.F_OK):
+        dest = abspath(join(self.fm.thisdir.path, expanduser(new_name)))
+
+        if access(dest, os.F_OK):
             return self.fm.notify("Can't rename: file already exists!", bad=True)
 
-        if self.fm.rename(self.fm.thisfile, new_name):
-            file_new = File(new_name)
+        if self.fm.rename(self.fm.thisfile, dest):
+            file_new = File(dest, path_is_abs=True)
             self.fm.bookmarks.update_path(self.fm.thisfile.path, file_new)
             self.fm.tags.update_path(self.fm.thisfile.path, file_new.path)
             self.fm.thisdir.pointed_obj = file_new
--- a/tests/ranger/config/test_commands_rename.py
+++ b/tests/ranger/config/test_commands_rename.py
@@ -1,0 +1,378 @@
+"""Tests for the rename command — regression for issue #3175.
+
+When renaming a file in a directory accessed via a symlink, the cursor
+must move to the renamed file.  The bug was caused by File(new_name)
+resolving the path via os.getcwd() (physical path) while the directory's
+internal file list uses the logical path (preserving symlinks).
+"""
+
+from __future__ import absolute_import, division, print_function
+
+import os
+import sys
+
+from ranger.container.file import File
+from ranger.core.shared import FileManagerAware
+
+# Make ranger importable from the source tree
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "..", ".."))
+
+
+# ---------------------------------------------------------------------------
+# Helpers: lightweight FM mock
+# ---------------------------------------------------------------------------
+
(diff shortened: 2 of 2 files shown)


toonn (dismissed):
LGTM

Appreciate the added tests.

toonn (commented):
Actually, after rereading #2038, I'm now not confident this is a correct fix. vejkse did some great digging in that thread, if you can show that all the cases they mention either don't apply or work out correctly this is ready to merge.

PR #3237, opened 2026-05-30
tags: Snapshot items before mutating in update_path (fixes #3176)

### LINKED ISSUE #3176: When renaming, tags are not updated
### Ranger version

Latest commit (08913377).  **This issue was already present in 1.9.3** (unless this is due to a somewhat recent behaviour change of Python).

### Steps to reproduce

(This cannot be tested with `ranger --clean`, as far as I can tell, but I tested it with my configuration stashed away.)

1. `:touch abcd`.
2. Tag the file: `:tag_add`.  The file is shown to be tagged.
3. Rename the file: `:rename efgh`.
4. The file is renamed, and shown to be tagged.  An error appears in red: `RuntimeError: dictionary keys changed during iteration`.
5. Reset: `:reset` or quit and restart ranger.  It appears the file `efgh` is not tagged (but `abcd` is still recorded as tagged in `share/ranger/tagged`.

### Comment

This is an old issue, I just thought until now this was a limitation of ranger that it doesn’t update tags when renaming, that it had to be done by hand.

But when investigating the issue I created just before this one, I discovered that tags were supposed to be updated.

I may be missing something, since apparently nobody raised it until now.

### Partial solution

The dictionary keys error suggests a solution: replace the line
https://github.com/ranger/ranger/blob/08913377c968d39f11fa2d546aa8d53a99bb5e98/ranger/container/tags.py#L119
with
```python
        for path, tag in list(self.tags.items()):
```

This eliminates the error and fixes the issue when no symlinks are involved.

### Steps to reproduce the issue with the partial solution

1. Create a directory `ab` and a symlink `ef` to `ab`.
2. In `ab`: create a file `cd` and tag it.  It is shown to be tagged.
3. In `ef` (where `cd` is not shown to be tagged), rename `cd` to `gh`.
4. Go back to `ab`: `gh` is not shown to be tagged (indeed, in `share/ranger/tagged`, it is `ab/cd` that is still recorded as tagged).

(This residual issue looks related to the discussion in https://github.com/ranger/ranger/pull/1068.)

mokashang:
#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
- Operating system and version: macOS 15 (also reproduces on Linux)
- Terminal emulator and version: any
- Python version: 3.13 (and any Python 3.6+)
- Ranger version/commit: master at 057c8799
- Locale: en_US.UTF-8

#### CHECKLIST
    - [x] Tests have been updated

#### DESCRIPTION
`Tags.update_path` walked `self.tags.items()` while deleting the old key and inserting the renamed one inside the same loop. On Python 3 the dict iterator raises `RuntimeError: dictionary keys changed during iteration` as soon as the first key is rewritten, which means the function exits early — before `dump()` — so the tag is silently lost from the on-disk tag file even though the in-memory rename appears to succeed in the UI.

Snapshotting the items with `list(self.tags.items())` before the loop lets the body mutate the dict safely while still doing only one disk write at the end.

A new `tests/ranger/container/test_tags.py` covers:

- A single tagged file being renamed (the original `RuntimeError` repro) — asserts the new path is both in memory and persisted to the tag file.
- A directory-prefix rename with multiple tagged children, mixing the default `*` tag and custom tag characters (`a:`, `b:`).
- The no-op path: when nothing matches, the tag file's mtime is left untouched so we don't churn disk.
- The separator boundary check: `/tmp/foo` must not match `/tmp/foobar`.

#### MOTIVATION AND CONTEXT
Fixes #3176. Quoting the reporter's steps:

> 1. `:touch abcd`.
> 2. Tag the file: `:tag_add`.  The file is shown to be tagged.
> 3. Rename the file: `:rename efgh`.
> 4. The file is renamed, and shown to be tagged.  An error appears in red: `RuntimeError: dictionary keys changed during iteration`.
> 5. Reset: `:reset` or quit and restart ranger.  It appears the file `efgh` is not tagged (but `abcd` is still recorded as tagged in `share/ranger/tagged`).

`update_path` is called from the `rename` command in `ranger/config/commands.py` right after `fm.rename(...)`, so any tagged path goes through this code path.

The reporter also identified the exact line and proposed wrapping the iteration in `list(...)`; this PR does that and pins the behavior with tests so the regression cannot silently come back.

#### TESTING
- `make test_pytest` — 19 passed (4 new tests in `test_tags.py`, all existing tests still green).
- `make test_flake8` — clean on the two touched files.
- `make test_pylint` — `tags.py` and `test_tags.py` both rated 10.00/10.
- Reverting just the one-line fix while keeping the tests confirms the two interesting tests fail (`RuntimeError` on the single-file rename, partial mutation on the directory-prefix rename), so the regression coverage is real.

#### IMAGES / VIDEOS
N/A — pure behavior fix.

### DIFF
--- a/ranger/container/tags.py
+++ b/ranger/container/tags.py
@@ -116,7 +116,13 @@ class Tags(FileManagerAware):
     def update_path(self, path_old, path_new):
         self.sync()
         changed = False
-        for path, tag in self.tags.items():
+        # Iterate over a snapshot of the items because the loop body mutates
+        # ``self.tags`` (deletes the old key, inserts the new one).  Iterating
+        # ``self.tags.items()`` directly raises
+        # ``RuntimeError: dictionary keys changed during iteration`` on
+        # Python 3, which would abort the update before ``dump()`` runs and
+        # leave the renamed path silently un-tagged on disk.
+        for path, tag in list(self.tags.items()):
             pnew = None
             if path == path_old:
                 pnew = path_new
--- a/tests/ranger/container/test_tags.py
+++ b/tests/ranger/container/test_tags.py
@@ -1,0 +1,99 @@
+from __future__ import (absolute_import, division, print_function)
+
+import os
+import pytest
+
+from ranger.container.tags import Tags
+from ranger.core.shared import FileManagerAware
+
+
+class _StubFM(object):  # pylint: disable=too-few-public-methods
+    """Minimal ``fm`` substitute for ``Tags`` -- only ``notify`` is exercised."""
+
+    def notify(self, *args, **kwargs):
+        pass
+
+
+@pytest.fixture(autouse=True)
+def _fm_stub():
+    # ``Tags`` inherits from ``FileManagerAware`` and calls ``self.fm.notify``
+    # from its IO helpers; install a no-op stub for the duration of each test.
+    previous_fm = getattr(FileManagerAware, "fm", None)
+    FileManagerAware.fm_set(_StubFM())
+    yield
+    if previous_fm is None:
+        try:
+            del FileManagerAware.fm
+        except AttributeError:
+            pass
+    else:
+        FileManagerAware.fm_set(previous_fm)
+
+
+def _write_tagfile(tmpdir, lines):
+    tagfile = tmpdir.join("tagged")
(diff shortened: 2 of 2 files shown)


toonn (approved):
LGTM

Having some more tests is definitely welcome.

PR #3243, opened 2026-06-17
Default to case insensitive filter_stack name filter

### LINKED ISSUE #3242: feature request: case-insensitive name filtering in filter stack
When using the filter stack, name filtering is case-sensitive by default.

I can use the `(?i)pattern` pragma to get around this, but I think non-case-sensitive filtering should be the default. (For *most* use cases, I don't think case-sensitive searches are that common (or ergonomic).)

Would it be possible to add a filter stack name filter that does case-insensitive searches?

Alternatively, you may be able to add a dot (`.N`?) command that automatically prepends `(?i)` to the search pattern input when users start typing.

vifon:
Resolves #3242

This only changes the default prompt as defined in the shipped rc.conf. As demonstrated by the issue reporter, the underlying mechanism already supports passing arbitrary arguments supported by the `re` module thanks to its inline flags. I'd rather avoid inventing our own syntax for that instead.

If the user want a case insensitive search, explicitly removing the inline flag is still possible. The flag getting auto-inserted might be confusing but at the same time it improves the discoverability of this feature as a whole, so YMMV.

### DIFF
--- a/ranger/config/rc.conf
+++ b/ranger/config/rc.conf
@@ -631,7 +631,7 @@ map .d filter_stack add type d
 map .f filter_stack add type f
 map .l filter_stack add type l
 map .m console filter_stack add mime%space
-map .n console filter_stack add name%space
+map .n console filter_stack add name (?i)
 map .# console filter_stack add hash%space
 map ." filter_stack add duplicate
 map .' filter_stack add unique


hut:
Thanks for the PR <3

I'd merge it, but I'd like to at least mention first that it does seem like a bad idea to expose regular expression flags to the user in such a manner.

I use REs myself quite frequently, and yet I'm not sure that I'd understand what's going on if a key binding opens the console with a "(?i)" in the terminal. It would appear like a bug to me.

It's an elegant solution, yes, but is it really a good idea?

Are there any alternative approaches?

toonn (approved):
I'm in favor of this. Maybe slap a comment right above the keybind that explains what the hell `(?i)` means, point to the Python re documentation.

PR #3247, opened 2026-07-05
Add Pix: support for opening images with pix

blackarez:
Line to allow opening images with the Pix application. Pix is ​​an image organizer and viewer that is automatically installed on Linux Mint, and it will continue to work even if Xviewer is uninstalled, allowing you to open images with Ranger on Linux Mint.


blackarez:
view Pix image.
imagen

hut (approved):

hut:
Thank you!

Issue #3246, opened 2026-06-27
Trash fail


ParsaShemirani:
Happens when I select a file and type :trash, followed by y to confirm.

parsahome@Parsas-MacBook-Pro ~ % ranger
ranger version: ranger 1.9.4
Python version: 3.14.5 (main, May 10 2026, 10:21:34) [Clang 21.0.0 (clang-2100.0.123.102)]
Locale: en_US.UTF-8
Current file: '/Users/parsahome/Desktop/jamietownjam/first_gold_drive_extended_ctl.txt'

Traceback (most recent call last):
  File "/opt/homebrew/Cellar/ranger/1.9.4/libexec/lib/python3.14/site-packages/ranger/core/main.py", line 201, in main
    fm.loop()
    ~~~~~~~^^
  File "/opt/homebrew/Cellar/ranger/1.9.4/libexec/lib/python3.14/site-packages/ranger/core/fm.py", line 383, in loop
    ui.handle_input()
    ~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/ranger/1.9.4/libexec/lib/python3.14/site-packages/ranger/gui/ui.py", line 266, in handle_input
    self.handle_key(key)
    ~~~~~~~~~~~~~~~^^^^^
  File "/opt/homebrew/Cellar/ranger/1.9.4/libexec/lib/python3.14/site-packages/ranger/gui/ui.py", line 196, in handle_key
    elif not DisplayableContainer.press(self, key):
             ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/opt/homebrew/Cellar/ranger/1.9.4/libexec/lib/python3.14/site-packages/ranger/gui/displayable.py", line 275, in press
    focused_obj.press(key)
    ~~~~~~~~~~~~~~~~~^^^^^
  File "/opt/homebrew/Cellar/ranger/1.9.4/libexec/lib/python3.14/site-packages/ranger/gui/widgets/console.py", line 188, in press
    self.type_key(key)
    ~~~~~~~~~~~~~^^^^^
  File "/opt/homebrew/Cellar/ranger/1.9.4/libexec/lib/python3.14/site-packages/ranger/gui/widgets/console.py", line 214, in type_key
    self._answer_question(answer)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
[...]

PR #3248, opened 2026-07-06
Add cross-platform trash support via ranger.ext.trash

web-d7r:
- New module handles macOS (Finder/AppleScript, trash utility) and Linux (trash-put, gio, local fallback)
- Normalize file objects to paths before trashing
- Update trash label rules to rifle.conf
#3246

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT

Debian GNU/Linux 13 (trixie) x86_64
Python 3.13.5
Ranger-master

- Python version: 
- Ranger version/commit: 
- Locale: 

#### CHECKLIST

- [ ] The `CONTRIBUTING` document has been read **[REQUIRED]**
    - [ ] Tests have been updated

#### MOTIVATION AND CONTEXT

Because ranger can't move files into trash in MacOs

#3246


hut:
Hey, thanks for taking the time to implement this :)

From what I see, your implementation is not ideal though.

You saw that there's already a mechanism to choose different trashing tools depending on the system configuration, at the very end of rifle.conf:

```
label trash = mkdir -p -- "${XDG_DATA_HOME:-$HOME/.local/share}/ranger/trash"; mv -- "$@" "${XDG_DATA_HOME:-$HOME/.local/share}/ranger/trash"
```

This uses trash-put if available, and otherwise has its own basic trash implementation.

You extended this with some new options, "trash" and "gio". However, by deleting the line `self.fm.execute_file(files, label='trash')` in commands.py (which uses rifle.conf to "open" the selected file with the program marked with the "trash" label), you bypass this mechanism completely. Then you reimplement this bypassed trash handling logic in a new python module, leaving the rifle.conf trash handlers as "dead code"

I think a better approach would be to add a new line in rifle.conf that triggers your newly added `osascript` trash method if MacOS is running, instead of reimplementing the command selection logic in ranger/ext/trash.py.

This mainly leaves the question of how to check whether the computer is running MacOS inside rifle.conf. I don't know how to best do this best, but one suggestion would be `has osascript`:

```
label trash, has osascript = osascript [...your implementation here...]
```

What do you think? Did I miss something?

P.S. you should be aware that some of your commits have commit messages that don't describe what's actually happening in the commit. Like https://github.com/ranger/ranger/pull/3248/changes/68e9b42d529772b3fcdb0884257e94023c6afb24 which is described as "add import" but it actually removes a comment.

web-d7r:
I've deleted trash as extension and used riffle. You are right, I think `label trash, has osascript = osascript -e 'on run argv' -e 'repeat with f in argv' -e 'tell application "Finder" to delete POSIX file f' -e 'end repeat' -e 'end run' "$@"` should work. I've done a lot of extra work which wasn't really required. You just need to add a line from above into riffle.conf. But I don't have mac, so someone should check whether this line works on mac or not. I've confused with all these githubing

hut:
Thanks! Let's see if we can find a mac user to test this.

I would prefer to squash the commits, to avoid going back-and-forth with the changes. But I can do that for you.

One question though, can you explain this to me?

```
        if isinstance(files, str):
            files = [self.fm.thisfile]
        elif not isinstance(files, (list, tuple)):
            files = [files]
```

When would `files` ever be a str? Or when would it ever be a `File` object that's not in a list/tuple?

Is this maybe related to https://github.com/ranger/ranger/issues/3246 ?  But there, the problem is that it's a list of str, like:

```
files = ["a", "b"]
```

whereas your code covers cases like

```
files = "a"
files = File("a")
```

web-d7r:
You're right, I was targeting the wrong case. #3246 problem was already solved somewhere in master branch. We need update only riffle.conf, which we discussed before

hut:
Is the "trash" utility from trash-cli?

I'm wondering why there's an extra line that calls `trash` when there's already a line for `trash-put`, which is also from trash-cli.

Is that an accidental duplicate?

ranger/config/commands.py
```diff
@@ -851,6 +851,7 @@ def _trash_files_catch_arg_list_error(self, files):
         Executes the fm.execute_file method but catches the OSError ("Argument list too long")
         that occurs when moving too many files to trash (and would otherwise crash ranger).
         """
+
```
toonn, line 854:
```suggestion
       """
```

According to PEP 257 doc-strings shouldn't be followed by a blank line.

ranger/config/rifle.conf
```diff
@@ -310,6 +310,9 @@ label pager,  !mime ^text, !ext xml|json|ya?ml|toml|conf|cfg|ini|env|csv|tex|py|
 # Execute a file as program/script.
 mime application/x-executable = "$1"
 
-# Move the file to trash using trash-cli.
+# Move the file to trash
 label trash, has trash-put = trash-put -- "$@"
-label trash = mkdir -p -- "${XDG_DATA_HOME:-$HOME/.local/share}/ranger/trash"; mv -- "$@" "${XDG_DATA_HOME:-$HOME/.local/share}/ranger/trash"
+label trash, has trash = trash "$@"
+label trash, has gio = gio trash "$@"
```
toonn, line 316:
```suggestion
label trash, has gio = gio trash "$@"
```

This is the same command as `trash-put`, the latter is preferred because it's more specific and the existing line also includes `--` making it safer. Does `gio trash` support the convention that `--` signals to stop parsing for flags and options? If so the same `--` should be added before `"$@"`.

toonn (changes requested):
It's a good change to make but needs some finishing touches.

ranger/config/commands.py
```diff
@@ -851,6 +851,7 @@ def _trash_files_catch_arg_list_error(self, files):
...
         try:
```
toonn, line 855:
```suggestion
       """
try:
```

According to PEP 257 doc-strings shouldn't be followed by a blank line.

Had to try again, Github's suggestions really don't like suggesting to remove a single blank line.

hut:
The osascript line causes this error on a tester's MacOS machine:

```
63:82: execution error: Finder got an error: Can’t get POSIX file "/Users/xyz/Downloads/abc.pdf". (-1728)
```

I tried prefixing it with `file://` but that didn't help.

Issue #2041, opened 2020-07-08
Renumber tabs (like tmux config 'renumber-windows')


scottkosty:
Enhancement: if I open three tabs and then close the first, the remaining two tabs are numbered "2" and "3". I would like for them to be numbered as "1" and "2", so that I can switch back and forth between them with alt+1 and alt+2.

tmux has a config "renumber-windows" that does this. Could the same be implemented for ranger?

toonn:
This sounds like a sensible feature. It wouldn't be hard to implement either, just a matter of adding a function in `ranger/core/actions.py` that iterates over the tabs dictionary and renumbers.

Ablesius:
I'd like to take a look into this. I'm wondering why Tab in tab.py doesn't have a `name` field @toonn @vifon. Would it make sense to add it?

toonn:
What would be the use? "Names" are currently recorded in the dictionary of tabs.

roverway:
I need this too

toonn:
Sounds like a sensible name, yes. I'd implement this as a command first. That would already allow adding the renumbering to any command that closes tabs. Then as a second feature a setting that allows configuring this without having to change all the mappings that close tabs.

PR #3253, opened 2026-07-22
Renumber tabs (feature, #2041)

JRPdata:
Manual command renumber_tabs (skips setting), and auto renumber tabs on tab close conditional on rc.conf setting.

#### ISSUE TYPE

#### RUNTIME ENVIRONMENT
Ubuntu 24.04.4 LTS
GNOME Terminal 3.52.0 using VTE 0.76.0 +BIDI +GNUTLS +ICU +SYSTEMD
ranger version: ranger-master v1.9.3-947-g26a25793
Python version: 3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:50:58) [GCC 12.3.0]
Locale: en_US.UTF-8

#### CHECKLIST
- [X] The code was not generated by an LLM **[REQUIRED]**
    - [ ] Tests have been updated

#### DESCRIPTION
Added renumber_tabs_on_tab_close setting:
    - default is False
    - added it also to relevant docs Added renumber_tabs() to actions.py (helper):
    - forces renumbering of tabs irrespective of setting Added tabsrenumber() to actions.py (helper handler):
    - tab_close() in actions.py (checking the config)
    - renumber_tabs() in actions.py (calling it with force=True) Added tabs_renumber() to actions.py (actual execution of renumbering the tabs).
    - enumerates tabs starting from 1
    - only actually changes the dict and (conditionally) emits a layout_change if the numbering changes
    - conditionally emits the layout_change if emit=True (emit=False when called from tabs_close) Modified tabs_close():
    - Added call to tabsrenumber(), making sure emit=False, so 'layoutchange' is only emitted once

#### MOTIVATION AND CONTEXT
Tab numbers become irregular when closing tabs, and opening new tabs opens them in irregular places as a result.
Discussed already and addresses issue #2041:
[Renumber tabs (like tmux config 'renumber-windows')](https://github.com/ranger/ranger/issues/2041)

#### TESTING
Tested implemented features in multiplane and miller (automatic and manual command).
Tested it loaded with rc.conf variable set to true.
Appears to work fine.
Shouldn't affect anything else in the code base other than tab_close (as intended).

### DIFF
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -1081,6 +1081,9 @@ the content.
 .IX Item "relative_current_zero [bool]"
 When line_numbers is set to relative, show 0 on the current line if
 true or show the absolute number of the current line when false.
+.IP "renumber_tabs_on_tab_close [bool]" 4
+.IX Item "renumber_tabs_on_tab_close [bool]"
+After closing a tab the tabs will automatically renumber if true.
 .IP "save_backtick_bookmark [bool]" 4
 .IX Item "save_backtick_bookmark [bool]"
 Save the \f(CW\*(C`\`\*(C'\fR bookmark to disk.  This bookmark is used to switch to the last
--- a/doc/ranger.1
+++ b/doc/ranger.1
@@ -1082,6 +1082,9 @@ the content.
 .IX Item "relative_current_zero [bool]"
 When line_numbers is set to relative, show 0 on the current line if
 true or show the absolute number of the current line when false.
+.IP "renumber_tabs_on_tab_close [bool]" 4
+.IX Item "renumber_tabs_on_tab_close [bool]"
+After closing a tab the tabs will automatically renumber if true.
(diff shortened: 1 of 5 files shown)


hut:
Thanks!

I noticed that the method names in actions.py are very similar. Could you rename them to clarify their differences?

Also, looks like 2 of them are helper functions. I think it would make sense to prefix the names with a "_" so they are "private" methods and not exposed as ranger console commands.

JRPdata:
@hut,  I've amended the commit (and commit message) with the helper function names refactored as follows (sorry for all the pushes):

tabs_renumber() -> _do_tabs_renumber()
("do" since it does the actual renumbering logic)

tabsrenumber() -> _try_tabs_renumber_handler()
("try" since it is the handler for both tab_close (which can be conditional on the setting) and the command)

hut (approved):

Issue #1388, opened 2018-11-25
ranger doesnt save tabs with $SiGTERM


rtng1:
#### Runtime Environment

- Operating system and version: Debian GNU/Linux 9.6 (stretch) x86_64 
- Terminal emulator and version: terminator 1.90
- Python version: Python 2.7.13
- Ranger version/commit: ranger-master 1.9.2
- Locale: /home/rtng/local/bin/ranger

#### Current Behavior 
ranger doesnt save tabs with $SiGTERM

#### Expected Behavior
ranger saves tabs with $SiGTERM

#### Context

using ranger in a multi-tabbed layout of terminator, ranger does not save it's own tabs, when just quit a terminator-session without closing ranger with 'Q' or ':quit'
with: kill -TERM, kill -HUP or kill -9 had all the same conseqence: no tabs saved .  

#### Steps to reproduce
1. 
2. 
3. 

#### Traceback

```

```

j-lakeman:
Reproduced with

#### Runtime Environment
- Operating system and version: Debian Sid
- Terminal emulator and version: alacritty 0.13.2
- Python version: 3.11.9
- Ranger version/commit: ranger 1.9.3
- Locale: en_AU.UTF-8

Any updates on this? This is quite major if working with tabs IMHO.

PR #3254, opened 2026-07-22
Save tabs on exit on sigterm/sighup (bugfix, #1388)

JRPdata:
Trap sigterm/sighup to save the tabs.

#### ISSUE TYPE
- Bug fix

#### RUNTIME ENVIRONMENT
Ubuntu 24.04.4 LTS
GNOME Terminal 3.52.0 using VTE 0.76.0 +BIDI +GNUTLS +ICU +SYSTEMD
ranger version: ranger-master v1.9.3-947-g26a25793
Python version: 3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:50:58) [GCC 12.3.0]
Locale: en_US.UTF-8

#### CHECKLIST
- [X] The code was not generated by an LLM **[REQUIRED]**
    - [ ] Tests have been updated

#### DESCRIPTION
Traps sigterm and sighup.
Previously the tabs were only saved on a graceful exit using F10 or :quit.

The respective if block that handled the tab saving, which was only reached in those circumstances, has been now moved into save_tabs_on_exit(), and the previous if block replaced with a call to that function.
The signal handlers for sigterm and sighup now call save_tabs_on_exit() and then exit.

Note: since the sigterm sighup handlers can be called twice, an extra variable (self._already_saved_tabs_on_exit) was needed to keep track to make sure we don't push onto the stack the session's tabs twice when the program exits.

#### MOTIVATION AND CONTEXT
Losing tabs by accidentally hitting alt-f4 is a bug. No more lost tabs! (Except for power outages and computer crashes).
Hitting Alt-F4 or killing the program will now save the tabs before exiting.
Addresses bug already discussed in issue #1388;
[ranger doesnt save tabs with SiGTERM #1388](https://github.com/ranger/ranger/issues/1388)

Note: A more comprehensive solution was considered, such as periodically saving the tabs, but this idea was abandoned since it seems to go against ranger's design and is also out of scope of the issue. This could be addressed in the future with a different issue.

#### TESTING
Tested by monitoring the tab file size, and verifying the tabs changed during a session were present on reload of the program.
Monitored file size (ls -al) of ~/.local/share/ranger/tabs during testing to make sure we didn't push extra copies of the session's tabs.

Testing Process:
1) Check 'tabs' file size.
2) Loaded ranger.
3) Changed directory of last tab to something memorable.
4) Hit TAB key so so first tab is active (to make sure it saves the changes to the last tab)
5) In this step I varied the program exit conditions:
    a) F10
    b) :exit
    c) ALT-F4
    d) kill <PID>
6) Load ranger again
7) Verify the last tab's directory is where I last left it after step 3.
8) Check 'tabs' file size again. Verify it is within the same order of magnitude.
9) Repeat All above steps, while varying step 5 conditions.

Seems to work ok.

As far as I know these changes shouldn't affect anything else.


ranger/core/fm.py
```diff
@@ -233,6 +234,20 @@ def sigtstp_handler(signum, frame):
...
+            self.save_tabs_on_exit()
+            sys.exit(0)
+
+        signal.signal(signal.SIGTERM, sigterm_handler)
+
+        def sighup_handler(signum, frame):
+            # pylint: disable=unused-argument
+            self.save_tabs_on_exit()
+            sys.exit(0)
+
```
toonn: This is the exact same code as `sigterm_handler`, just bind the same function to both signals.

ranger/core/fm.py
```diff
@@ -312,6 +327,16 @@ def destroy(self):
                 if debug:
                     raise
 
+    def save_tabs_on_exit(self):
+        if self._already_saved_tabs_on_exit:
```
toonn: IMO, drop this, add the inverse condition to the other conditions and push the setting of `_already_saved...` into the if-statement.

toonn (changes requested):
I could see a case made for wanting to quit without saving with F10 or whatever but I don't feel strongly about it so I'm open to this.

JRPdata:
@toonn Thanks; your suggested changes are done.

PR #3257, opened 2026-08-05
Fix bulkrename tag handling on Python 3

chengzhipengczp:
## Summary

Fix bulkrename tag migration issues on Python 3.

## Changes

- Decode temporary script content before comparison so unchanged scripts are detected correctly.
- Fix bare `fm.notify()` call to use the command's file manager instance.
- Add regression tests for both cases.

## Testing

Passed:

- pytest tests/ranger/config/test_commands.py -q
- flake8 ranger/config/commands.py tests/ranger/config/test_commands.py
- pylint ranger/config/commands.py

Full pytest suite could not run on Windows due to existing Unix-only test dependencies.

### DIFF
--- a/ranger/config/commands.py
+++ b/ranger/config/commands.py
@@ -1300,7 +1300,12 @@ class bulkrename(Command):
             # script was modified by the user
             self.fm.execute_file([File(cmdfile.name)], app='editor')
             cmdfile.seek(0)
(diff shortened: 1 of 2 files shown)


chengzhipengczp:
Thanks for reviewing.

CI is now passing. I kept the changes limited to the Python 3 bulkrename tag handling fixes and added regression tests for both cases.

PR #3264, opened 2026-08-10
Warn about `:console [...] %s [...]` commands

### LINKED ISSUE #3261: Show a warning on incorrect macro usage in `:console` command
Ranger should display a warning if the user has a key binding that uses `:console` followed by `%s`, because it will resolve the macro at the wrong/unexpected time: during the call of `:console`, not during the call of the command whichever is following console.

The warning should explain the proper way to do this, which is using `%%s` instead of `%s`.

See https://github.com/ranger/ranger/issues/1734#issuecomment-5208042939

Example for correct command (from ranger's rc.conf):

`map @ console -p6 shell  %%s`

Example for a dangerous command:

`map X console shell rm -rf %s`

If the files in `%s` contain spaces or other special characters, it will mess with the command. For example, if `%s` is a directory called 'My Documents', typing the X key would make it expand to:

```
:shell rm -rf My Documents
```

which would delete 2 directories/files called `My` and `Documents`, not the directory/file `My Documents`.

hut:
The issue is described in https://github.com/ranger/ranger/issues/3261. This commit also fixes https://github.com/ranger/ranger/issues/3261.

Solution:

1. Add a warning when using commands matching the regex `\bconsole .*[^%]%s\b` with `fm.notify(..., bad=True)`
2. Add the macro `%nowarn` to silence the warning, e.g. in case of false positive warning

### DIFF
--- a/doc/ranger.pod
+++ b/doc/ranger.pod
@@ -472,6 +472,9 @@ Note that macros are expanded twice when using chain. For example, to insert
 a space character in a chained command, you would write %%space:
  chain command1; command2%%space
 
+The macro %nowarn anywhere in the command will silence warnings about dangerous
+commands and key bindings.
+
 =head2 BOOKMARKS
 
 Type B<m<keyE<gt>> to bookmark the current directory. You can re-enter this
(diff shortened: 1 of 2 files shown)


hut:
Closing, in favor of @vifon's suggestion to use a secondary macro for %s that does not warn, such as %sNoWarn, instead of "abusing" the macro system for such a hack. (or another, better solution)

